PR #5, opened 2019-02-04
Minor bugfix in s1_min_channels

ErikHogenbirk:
The number of channels requirement needed a >=, not a >. If this fix is implemented, if the setting for s1_min_channels is, say, 3, then an S1 with exactly 3 channels will pass the cut (which I think it should).


JelleAalbers:
Thanks Erik!

PR #10, opened 2019-05-08
Bootstrax

jorana:
Changed the code to also put info in a new collection (```usage```) of the MongoDB. This information comprises information on the cpu and memory usage of the node where ```Bootstrax``` lives and of the one or more subprocesses (PIDs).

In order to obtain the PID specific information ```psutil.Process``` is used, this allows one to look only one PID.


bin/bootstrax
```diff
@@ -185,10 +186,10 @@ def main():
 
     elif args.process:
         number = args.process
-        rd = run_coll.find_one({'number': number})
+        rd = consider_run({'number': number})
         if rd is None:
             raise ValueError(f"No run numbered {number} exists")
-        process_run(rd, send_heartbeats=False)
+        process_run(rd, send_heartbeats=True)
```
JelleAalbers: send_heartbeats=True is the default, so you don't have to specify this here

bin/bootstrax
```diff
@@ -556,14 +572,14 @@ def process_run(rd, send_heartbeats=True):
                 # Since chunks can be empty, and we don't want to crash,
                 # this has to be done with some care...
                 t_covered = timedelta(seconds=(
-                    max([x.get('last_endtime', 0)
+                    max([x.get('last_endtime', 0)
```
JelleAalbers: Here you're only adding trailing whitespaces. You can configure pycharm to autoremove them, or check in gitkraken when you make the commit. Of course trailing whitespace doesn't really matter, but this would make the git blame a bit harder to read.

bin/bootstrax
```diff
@@ -734,4 +750,4 @@ def cleanup_db():
 
 
 if __name__ == '__main__':
-    main()
+    main()
```
JelleAalbers: Same here with the newline at the end of the file

PR #12, opened 2019-05-10
minor bugfix with the time-information check

jorana:
Bootstrax was failing on these lines of code. They have now been adapted and run. Please note the manual division of nanoseconds -> microseconds in line 580.


jorana:
@JelleAalbers, I have also changed plugins.py. We noticed during the strax workshop that with the low datarate that the digitizers are now outputting, not all chucks get information from all readout threads. 

With the latest commit (https://github.com/XENONnT/straxen/pull/12/commits/beb5210d8916cda0053e94b3792d0894833256f6), it is first checked if the run has comleted succesfully, and if so, still accepts this run.

bin/bootstrax
```diff
@@ -278,16 +278,18 @@ def kill_process(pid, wait_time=None):
         os.kill(pid, sig)
 
 
-def set_state(state, pid_process = None, run_number = None):
+def set_state(state, pid_process = None, run_number = None, hostname = None):
     """Inform the bootstrax collection we're in a different state
 
     if state is None, leave state unchanged, just update heartbeat time
     if a pid_process and run_number are provided, update the 'usage' collection
```
JelleAalbers: I think you also require hostname now for this

bin/bootstrax
```diff
@@ -278,16 +278,18 @@ def kill_process(pid, wait_time=None):
...
     """
-    if state_doc_id is not None and pid_process is not None and run_number is not None:
+    if (state_doc_id is not None and pid_process is not None and
```
JelleAalbers: Perhaps it's clearer to add a new method (update_monitoring or something) and call it directly from line 544 (just under get_status). That keeps get_status simple and you don't need this big if statement anymore (which can cause confusion: if people forget one of the arguments, the usage update is silently skipped)

JelleAalbers:
I just realized we already flagged the messy end-of-run data here https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:dsg:nt_test_debrief (under 'Strax output at the end of run is messy').

jorana:
> I just realized we already flagged the messy end-of-run data here https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:dsg:nt_test_debrief (under 'Strax output at the end of run is messy').
I see, would you like another solution than? I'm not sure if manually adding files would make more sense than simply going on with the smaller ammount of files in the folder. 

I will submit the changes as soon as I'm able to acces Midway.

JelleAalbers:
Thanks Joran, looks good! I just mentioned the earlier note to show we actually knew about the problem before, but at least I had forgotten about it until you and Dan tackled this in Chicago.

PR #13, opened 2019-05-14
Modified max_pmt_area so that the unit becomes PE

k-hiraide:


JelleAalbers:
We eventually decided to just make area_per_channel itself have units of PE, which achieves the same goal as this.

PR #18, opened 2019-06-28
Update common.py

petergaemers:
Adds json and json.gz to the availible files to be read in by get_resource.
This is, currently, needed to read some Xenon1t config files needed by fax


petergaemers:
Ofcourse!

PR #20, opened 2019-07-05
Update common.py

petergaemers:
This lets get_resource load in files which are pickled

### DIFF
--- a/straxen/common.py
+++ b/straxen/common.py
@@ -40,7 +40,7 @@ cache_dict = dict()
 
 # Placeholder for resource management system in the future?
 @export
-def get_resource(x, fmt='text'):
+def get_resource(x, fmt='text', pickle = False):
     """Return contents of file or URL x
     :param binary: Resource is binary. Return bytes instead of a string.
     """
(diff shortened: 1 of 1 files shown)


JelleAalbers:
Sorry, can you clarify how this is used? If you want to store/load a pickle, why wrap it in a numpy array and go through np.load?

petergaemers:
As it is now, the afterpulse distribution files are stored as numpy arrays containing a dictionary. To load them you need to pass allow_pickle = True to np.load. Another file (krypton83m map) is stored as a pickled dataframe. 

This was the easiest way to read them all. 
Is is a better idea to save the arrays as pickles and do pickle.load?

JelleAalbers:
> afterpulse distribution files are stored as numpy arrays containing a dictionary

OK, if this helps load legacy XENON1T files, fine: it's only a small addition. 

But let's avoid using too many different formats like these in the future: if we have a list/array of dicts, I think we can just use pickle directly to save it. Perhaps I'm missing some deep reason why we did it like this in the past.

Instead of adding a new option to get_resource that only applies to one format, could you add a new format (e.g. 'npy_with_pickles' or 'npy_pickle')?

petergaemers:
I agree, it is needlessly confusing. I changed it to npy_pickle, and fixed the error

Issue #22, opened 2019-09-16
Peak processing plugin has insufficient condition for TF2 checking


decibelcooper:
Hello,

In attempting to walk through beginning tutorial steps, I came across a tensorflow failure to find graph elements at `straxen/plugins/peak_processing.py", line 222`.  E.g.,
```shell
ValueError: Tensor Tensor("dense_6/BiasAdd:0", shape=(?, 2), dtype=float32) is not an element of this graph.
```

Upon investigation, I see that special cases were put into place for the `peak_processing` plugin to use tensorflow v2.  Unfortunately, it looks like the condition for checking for v2 is insufficient.
https://github.com/XENONnT/straxen/blob/62d964d277729e48d5648d04669ed7fa927b8d62/straxen/plugins/peak_processing.py#L176

Perhaps one could use `parse_version(tf.__version__) >= parse_version('2.0.')`, since 2.0 is still pre-release?

decibelcooper:
Scratch that suggestion, and replace with `parse_version(tf.__version__) >= parse_version('2.0.a')`.  I really don't know if there is a convention for catching `*-rc*`, but this seems to work.

I have confirmed that with this change, tf v1.14.0 works fine for the tutorial data.

JelleAalbers:
Welcome David! Thanks for spotting this, I naively thought checking > 1.9 would do it, but of course 1.14 > 1.9.

decibelcooper:
Excellent.  Thank you for the quick response!

PR #23, opened 2019-09-16
Fix tensorflow check, see #22

JelleAalbers:
Thanks to @decibelcooper for the fix.


decibelcooper:
BTW, I'm not sure if "see" is a keyword for github, but if one puts "resolves #22" in the PR, github should automatically close the referenced Issue (https://help.github.com/en/articles/closing-issues-using-keywords).  But I'm satisfied, so I will close my issue.  Thanks again!

PR #25, opened 2019-09-26
Update common.py

petergaemers:
Fixes the electron lifetime to return a placeholder when providing a run_id not corresponding to an actual run

### DIFF
--- a/straxen/common.py
+++ b/straxen/common.py
@@ -153,7 +153,8 @@ def get_elife(run_id,elife_file):
     if not len(run_index):
         # Gains not known: using placeholders
         e = 623e3
-    else: e = x[run_index[0]]['e_life']
+    else:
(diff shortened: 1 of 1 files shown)


zhut19:
This is great for simulations, further more, can the value have an option to be pulled from config input? As the elifetime will change soon once we actually start moving to nT

petergaemers:
To get a different electron lifetime change the option "elife_file", which is 'https://raw.githubusercontent.com/XENONnT/strax_auxiliary_files/master/elife.npy' by default.
For fax simulation the lifetime is pulled from the default config

jorana:
I think this would be great indeed as it will also solve running that the runsDB does not have this info for XENONnT (obviously) and it does cause e.g. bootstrax to chrash

PR #27, opened 2019-10-07
Change the plugins needed by bootstrax.

jorana:
Boostrax currently crashes as it does not have the daqreader plugin. Hence, I propose to add all that is needed to go up to the default of ```--target``` (currently ```event_info```). 

We might also opt to change this default since 'high level' changes in e.g. event building cause bootstrax to fail (which we don’t want) whereas they are not necessarily the type of data what we intent to put into Rocio.

### DIFF
--- a/bin/bootstrax
+++ b/bin/bootstrax
@@ -157,10 +157,7 @@ def new_context():
             runid_field='number',
             new_data_path=output_folder,
             mongo_dbname=dbname),
-        register_all=[straxen.plugins.daqreader,
-                      straxen.plugins.pulse_processing, 
-                      straxen.plugins.peak_processing,                       
-                      straxen.plugins.event_processing])
+        **straxen.contexts.common_opts)
     return st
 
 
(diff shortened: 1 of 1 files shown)


JelleAalbers:
Good catch, no idea how this worked before if only peak processing plugins were selected. Perhaps this got introduced during some autorefactor I did.

Would it not be better though for the context that bootstrax uses to be defined in straxen.contexts? Even if we don't do that, I imagine bootstrax's context would want to use `straxen.contexts.common_opts`, so it will have all necessary plugins registered: https://github.com/XENONnT/straxen/blob/master/straxen/contexts.py#L6 and we don't need to maintain a second list of plugin files.

jorana:
That would look neater indeed. Since the context options ```store_run_fields``` and ```check_available``` at the moment do not used loading ```straxen.contexts.common_opts``` would be fine. Might this change if ```scan_run``` would be performed (https://github.com/AxFoundation/strax/blob/master/strax/context.py#L42)? 

I'll look into it if it might be better to add another context as soon as I work with bootstrax on the production database. For now, I think this fix you proposed will do indeed.

Issue #17, opened 2019-06-19
Convert 'daq times' to epoch times


coderdj:
The DAQ provides timestamps in ns since run start with resolution of the ADC sample size. This needs to be converted to 'ns since unix epoch'. This is hard to do in the DAQ but maybe it can be done in the DAQReader plugin at the stage where all the sub-files from each readout thread are combined.

JelleAalbers:
I guess this would involve:
  * Add an option such as run_start_time to the DAQReader;
  * Add the epoch time in ns of the run start to the DAQ times in the records here https://github.com/XENONnT/straxen/blob/master/straxen/plugins/daqreader.py#L112;
  * Modify bootstrax to pass the run doc's start time as an argument to run_strax here https://github.com/XENONnT/straxen/blob/master/bin/bootstrax#L534 and eventually to strax itself here https://github.com/XENONnT/straxen/blob/master/bin/bootstrax#L469;
  * Testing this by processing a run

jorana:
This issue is addressed by this commit (https://github.com/XENONnT/straxen/commit/569c80334b8373ed419be83d54c78b6d22e18ee0) as part of PR https://github.com/XENONnT/straxen/pull/29.

PR #29, opened 2019-10-18
Bootstrax

jorana:
I added a feature to read the compression method used for the live data and from the run-doc and pass that on to the DAQ-reader. Since for some of the runs do not have this in the run-doc, I set the default to compression method to blosc.

Also added some info on the eb_monitor collection to bootstrax although that is not related.


jorana:
@JelleAalbers indicated that some lineage problems I encountered due to this PR could be solved by disabling the tracking of the new option.

bin/bootstrax
```diff
@@ -126,7 +137,9 @@ timeouts = {
 }
 
 # Fields in the run docs that bootstrax uses
-bootstrax_projection = 'name start end number bootstrax data.host data.type data.location ini.processing_threads'.split()
+bootstrax_projection = f"name start end number bootstrax " \
```
JelleAalbers: You're not using the f-string functionality here: it's not harmful but not needed either.

jorana:
Also added the unix timestamp in ns as requested in this issue: https://github.com/XENONnT/straxen/issues/17. 

I also tested it for run 661:
time_since_epoch

PR #32, opened 2019-12-05
Update peak_processing.py

DonNabla:
Implementation of a new Peak classification algorithm similar to the one used in pax and based on the rise time/ tight coincidence variables of the peak. I added two new variables in the "PeakBasics" plugin (rise_time and range_90p_area) and a new plugin "TightCoincidence" to compute the tight coincidence variable. Both are used in the new version of the "PeakClassification" plugin. You can see below a comparison between the current strax classification plugin and the pax-like peak classification using background data. As we can, with the current classification peaks corresponding to Single Electron (SE) are identified as S2, S1 and unknown peaks... But with the pax-like classification, they are properly tagged as S2 signal.

pax_like_vs_default_PC_background_data

We test the validity of this new peak classification with the [Peak validation notebook using the WFsim](https://github.com/XENONnT/analysiscode/tree/master/PeakFinderTest) and we get the following S1 acceptance comparison between the current classification and my pax-like classification:

acceptance_comparison

In the end the result is pretty clear:

Comparison_results

### DIFF
--- a/straxen/analyses/mpl_helpers.py
+++ b/straxen/analyses/mpl_helpers.py
@@ -1,0 +1,42 @@
+import numpy as np
+import matplotlib
+import matplotlib.pyplot as plt
+
+
+def log_y(a=None, b=None, scalar_ticks=True, tick_at=None):
+    """Make the y axis use a log scale from a to b"""
+    plt.yscale('log')
+    if a is not None:
+        if b is None:
+            a, b = a[0], a[-1]
+        ax = plt.gca()
+        plt.ylim(a, b)
+        if scalar_ticks:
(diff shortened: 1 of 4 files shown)


jorana:
Great to see the code on the pax-like classification. One comment to keep in mind concerning the validation notebook:

The energies set in that notebook are (at the moment) quite low (https://github.com/XENONnT/analysiscode/blob/master/PeakFinderTest/peak_classification/wfsim_utils.py#L33). 

```
nphotons = np.random.uniform(1, 40, n)
nelectrons = np.random.uniform(10, 100, n)
```

DonNabla:
Thank you Jelle for the tweaks !

PR #36, opened 2020-01-03
Gap-size-based clustering in two steps (peaklets)

jhowl01:
Adds an additional, intermediate data-kind called "peaklets", with the same dtype as peaks. Peaks are then constructed from peaklets via "merging". This is performed by a new function, using the attributes of several constituent peaklets to construct a peak, rather than reverting to hits.

Using two data structures allows clustering in two steps, with some pseudo-classification performed in between. Thus, we can apply different clustering criteria to S1-like and S2-like waveforms. This is beneficial, since these waveforms look very different: we really want to cut S1's off quickly at the tail, but small S2s can be sparse and stochastic. This method was used in pax, but breaking it up into multiple strax plugins allows us more play in the intermediate pseudo-classification and merging (e.g. using the sum-waveform).

Before merging, a few issues remain that I wasn't sure quite how to handle (see inline comments):

  * L60: Making a smarter choice for the down-sampling threshold during the merging
  * L64: Not assuming standard time-per-sample to start with.
  * L57: It's pretty essential to numba-fy this, but my numba implementation wound up slower than this, I think since np.repeat() isn't supported so I made it uglier.

Potential other issues:
  * L28: Vectorize to avoid the for-loop or numba-fy a double-loop? The latter was much slower for me, and not sure how to do the former. Currently, profiling shows `merge_peaks()` is way more dominant, but this might change when its numba'ed.
  * L89: Do we want to do something smarter (interpolate) than just repeating downsampled samples?


jorana:
_for completeness I'm adding the gitter message below also to this PR_

Thanks a lot for the great work! At the moment, I have two questions I ran into while trying out your improvements to the clustering. 

1. On lines https://github.com/XENONnT/straxen/pull/36/files#diff-5f23c37a64ca0040b8b2e5b954ae9ca8R48 - 51, the 0's and 1's in the ``merge_with_next`` are converted to the indices where you want to start and end merging. However, at the moment for (for example ```merge_with_next = [0,1,0,0,1,0]```) I got ```end_merge = [3, 6]```, ```start_merge = [4]```. I think this is not what you want. On the other hand, setting the first ``start_merge`` to 1 (if the first ```merge_with_next``` is 1) is required. Hence, would we have to change these lines to something like below?
 ```
     if merge_with_next[0] == 1:
         start_merge[0] = merge_with_next[0]
 
     end_merge_at = np.where(end_merge)[0] + 2
     start_merge_at = np.where(start_merge)[0] + 1
     if merge_with_next[0] == 1 and start_merge_at[0] == 1:
         start_merge_at[0] = 0
 ``` 
2. Are you only passing through the re-merged peaklets here: https://github.com/XENONnT/straxen/blob/8240d4dfb234541adbee9e7741ae117c1b13a96b/straxen/plugins/peak_processing.py#L285 I would suspect you’d still have to add the peaklets that are not involved in the merges there too. I think this might be why, if I throw it at the peakfinder test, it gives no S1s.

At the moment (due to 2.) the peakfinder test results are not so helpful but I will add some figures later if you want.

jhowl01:
Thanks @jorana ! The commit I just pushed should fix your comment (2) - you're right, I was stupidly only returning the merged peaks. Annoyingly, I couldn't figure out how to concatenate merged and unmerged without rebuilding the unmerged ones column by column...

jorana:
@jhowl01, I've changed the merge_index handing a bit to solve my first question. For completeness https://github.com/XENONnT/straxen/pull/36/commits/67b79c9b6a99c3c7f9df0f8dad7f191294148664 is a test added to straxen to ckeck that ```get_start_end``` returns the output correctly.

JelleAalbers:
Thanks for this Joey! It is a key step in getting good peakfinding performance for deep low-energy S2, which are currently often fragmented. Without this, we'd have to set a really high gap size and also cannot effectively use other clustering algorithms like natural breaks.

I'm assuming that the peaklet clustering is configured so as not fragment 
single electrons (except perhaps splitting of a stray photon from the tail). If it did, there would be an awful lot of splitting and merging to do, and since the SE fragments could get tagged as S1s, it will be quite tricky.

I made a few changes:

  1. **Restructure the plugins** to (a) avoid saving two copies of all un-merged peaks, which would double our peaks-level datasize, and (b) make sure the result does not depend on how peaklets are chunked. Currently the results can change if you save peaklets in differently-sized files (or process it online), since parts of a to-be-merged cluster can be in different chunks.
      * After `Peaklets` (unchanged), `PeakletClassification` classifies the peaklets.
      * Next, `MergeS2s` creates merged S2s from peaklets. This is an [OverlapWindowPlugin](https://strax.readthedocs.io/en/latest/developer/overlaps.html#overlap-window-plugins) to avoid issues near chunk boundaries.
      * Finally, `Peaks` combines peaklets with merged S2s. It throws away peaklets that touch merged S2s, and puts the merged S2s in their place. The peaks output is never stored, since it's trivial to compute from its two constituents.
  2. Changed the peak **merging decision** routine. Of course we can put in something more advanced in the future, which is part of the point of this PR.
      * We still consider clusters of peaklets with gaps between them < 3.5 us -- but we consider all peaks at first, regardless of type. 
      * We only merge if the following criteria are met. Unless all are met, we do not merge, not even part of the cluster. All numbers are configurable.
         * The first peaklet must be an S2. Others can be of any type, since a big afterpulse could get flagged as an S1. Very shallow events will have an S2 right after their S1, so these do not get merged -- I think this is OK, as their S2s are narrow enough that the base clustering should work. If not, here is a point to improve.
         * Total cluster area < 5000 PE. Above this we assume no S2s are appreciably split by the base clustering.
         * Total cluster length < 10 us. This avoids clustering a photoionization tail together.
      * Some simplifications to avoid the merge_with_next to start/end indices mapping. This means the tests Joran wrote for this mapping no longer apply -- sorry Joran!
  3. **Classification is done once,** on peaklets. The peak type is simply copied from the peaklets if there is no merge, and set to S2 if there is a merge. The `peak_classification` data type is now gone; instead a `type` field has been added to the peaklets, peaks, and peak_basics datatypes. In peaklets this field will be always 0; peaklet_classification will override it.

You can disable peak merging by setting `s2_merge_max_gap` to a negative value. I had to put in an ugly hack to enable strax to work in this case (or generically for datasets where no merging happens): emit a 'fake merged S2' if there are no merged S2s in a chunk. Without this, strax will wait forever for merged_s2s to combine with peaklets. It will be quite difficult to fix this properly in strax: the no-data case would have to be handled much better.

jhowl01:
The changes are beautiful, thanks for doing all the hard stuff y'all ;). So I guess we decided we want afterpulses split from S1s (this avoids mis-classification as S2) but included in S2s, but we want photo-ionization signals split from S2s. I think this makes sense. To me it looks great.

jorana:
Using the [https://github.com/XENONnT/analysiscode/blob/master/PeakFinderTest/Peak_Classification_Tester-Two_step_clustering.ipynb](Peak classification tester ) I tested Joeys changes just before https://github.com/XENONnT/straxen/pull/36/commits/0ecd11bcf673695768092fd7c8f489ee718ad1b2.

The following results are obtained for simulating 5000 S1 and S2s with:
```
    nphotons = np.random.uniform(2, 100, n)
    nelectrons = np.random.uniform(1, 1000, n)
    enable_pmt_afterpulses=False
    enable_electron_afterpulses=False
```

Below the acceptance as function of the n_electron/n_photon are given for the S2/S1. The 'custom' classification applies the clustering as implemented in this PR. One can see that it performs very good compared to the 'default' classification/clustering. This clustering method has very little to no effect on the S1-classifcation. However for the S2s the clustering is greatly improved. As the temporal spacing in the simulated events is quite large no overlap between pulses is simulated.
s2_clus_acceptance
s1_clus_acceptance

To see why this acceptance is so much higher the following two figures compare the default versus the custom clustering where it is clear that the number of split S2s is greatly reduced due to the clustering in this PR.

cus_s2_clus_acceptance
![def_s2_clus_acceptance](https://user-images.githubusercontent.com/22295914/72432927-82f6f200-3798-11ea-86ec-b6ced0c9c8ef.

Finally, the acceptances for these settings are:
default
S1 eff  = 0.651 -0.013 + 0.013
S2 eff  = 0.840 -0.010 + 0.010
custom
S1 eff = 0.651 -0.013 + 0.013
S2 eff = 0.997 -0.002 + 0.001

jorana:
After Jelle's merge I've ran the PeakFinderTest again with similar results. These difference between the results below and above are due to the commits made by Jelle in commit https://github.com/XENONnT/straxen/commit/0ecd11bcf673695768092fd7c8f489ee718ad1b2 and later.
cus_s2_clus_acceptance

custom
S1 eff = 0.653 -0.013 + 0.013
S2 eff = 0.948 -0.007 + 0.006

PR #37, opened 2020-01-09
Resurrect fake daq, fix DAQReader time conversion

JelleAalbers:
This resurrects the fake_daq script, used to mimic the DAQ on a regular computer. You must point it to a folder with `raw_records` data and tell it which run to read; it will then create "live records", if desired at a specific write rate.

Perhaps more importantly, this fixes a bug in the DAQReader's absolute time conversion (https://github.com/XENONnT/straxen/issues/17). A bare 1e9 was used, but that gives a float. Our times are 64-bit integers, so doing float math with them leads to inaccuracies. In this case, it led to records with fractional-sample times, which caused strange crashes later in strax.

(I also remove some unused old straxen contexts)


JelleAalbers:
Merging to help onsite work. The build failure also affects master and is due to tensorflow/keras changes upstream; will look into this soon.

PR #38, opened 2020-01-13
Saturation correction from pax

zhut19:
Like the correction updated at [pax 6.10.0](https://github.com/XENON1T/pax/pull/712), the same procedure can be applied to raw_records, from records onwards, `data` field is corrected from certain saturation effects.

This is still a working progress. Also, a part of the changes needs to be done in strax upon implementation.

### DIFF
--- a/straxen/plugins/pulse_processing.py
+++ b/straxen/plugins/pulse_processing.py
@@ -133,6 +133,13 @@ class PulseProcessing(strax.Plugin):
         else:
             veto_regions = np.zeros(0, dtype=strax.hit_dtype)
 
+        ##
+        # Saturation correction
+        ##
+        records_correction = np.zeros(len(r), dtype=np.dtype(rc_dtype))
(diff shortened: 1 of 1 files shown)


zhut19:
Additional thought: This doesn't need to be inside `pulse_processing`. It was the case that the `get_full_pulse_record_index` function fails after tail veto (as it was relying on `pulse_length`), this led to my decision of putting the whole thing at its current place. As I rewrote this function after facing many failure cases, functionally this can be moved to be after `pulse_processing` and be an independent plugin, which should be the desired way to go.

JelleAalbers:
Yes, the issue https://github.com/AxFoundation/strax/issues/180 is irritating, I'm hitting it many times myself too. Since it always seems to be the same one that gives problems, I am going to try to disable numba's caching for this function in the next strax release. Maybe that will already help.

zhut19:
This approach, that correct saturation at record level is abandoned. See the peak level approach in https://github.com/XENONnT/straxen/pull/244.

PR #42, opened 2020-01-17
Led plugin

GiovanniVolta:
Given the similarities between SPE and gain calibration a common plugin, to be integrated into strax (the new nT processor), will be used in XENONnT to access the LEDs raw data and return the variables needed for both of them.
The plugin requires having the 'raw_records' available.
To perform these analyses, a datasets of LED signals and a dataset of background are required. The latter is acquired with LED on in a setting with voltage low enough to not produce visible signals in the PMTs, i.e. 1.0V in XENON1T, compared to the ~3.5V for the SPE and ~4.2V for the gain. These values will probably change for nT.
Two functions are defined in the plugin, to get the amplitude and the area in a given window. The window is defined as a configuration parameter that is set to a default value in the plugin, but can be changed in the analyses by defining a new configuration setting. The functions do the following:

1) get_area, needed for the gain computation, sum the data in the defined window to get the area. 

2) get_amplitude, needed for the SPE computation, take the maximum in two different regions, where there is the signal and where there is not.

N.B.: remember to set the number of PMTs to 494 when nT data will be available.


straxen/plugins/led_calibration.py
```diff
@@ -0,0 +1,114 @@
...
+#export, __all__ = strax.exporter()
+# QUESTION: what does 'export, __all__ = strax.exporter()' do?
+# ANSWER: [fill in]
+
+@strax.takes_config(
+    strax.Option('LED_window',
+                 default=(125, 250),
+                 help="Window (samples) where we expect the signal in LED calibration"),
+    strax.Option('noise_window',
+                 default=(350, 475),
```
ershockley: I fear afterpulses may be a concern if the noise window follows the LED window. Have you checked any differences if you use before/after the LED window?

straxen/plugins/led_calibration.py
```diff
@@ -0,0 +1,114 @@
...
+    Preliminary version, several parameters to set during commisioning.
+    LEDCalibration returns: channel, time, dt, lenght, Area, amplitudeLED and amplitudeNOISE.
+    The new variables are:
+    - Area: Area computed in the given window, averaged over 6 windows that have the same starting sample and different end samples.
+    - amplitudeLED: peak amplitude of the LED on run in the given window.
+    - amplitudeNOISE: amplitude of the LED on run in a window far from the signal one.
+    '''
+    
+    __version__ = '0.0.8'
+    print(__version__)
```
ershockley: Do you want the version printed everytime?

straxen/plugins/led_calibration.py
```diff
@@ -0,0 +1,114 @@
...
+    depends_on = ('raw_records',)
+    # Options below copied from other plugins, need to be reviewed by an expert
+    data_kind = 'led_cal_0'
```
ershockley: Why the 0?

straxen/plugins/led_calibration.py
```diff
@@ -0,0 +1,114 @@
...
+    print(__version__)
+    depends_on = ('raw_records',)
+    # Options below copied from other plugins, need to be reviewed by an expert
+    data_kind = 'led_cal_0' 
+    compressor = 'zstd'
+    parallel = 'process'
+    rechunk_on_save = False
+    
+    dtype = [('area', np.int32, 'Area averaged in integration windows'),
+             ('amplitudeLED', np.int32, 'Amplitude in LED window'),
```
ershockley: I know I used this notation for the 1T code, but maybe we should remove the all-caps? Maybe Jelle has a stylistic preference.

straxen/plugins/led_calibration.py
```diff
@@ -0,0 +1,114 @@
...
+# ANSWER: [fill in]
+
+@strax.takes_config(
+    strax.Option('LED_window',
+                 default=(125, 250),
+                 help="Window (samples) where we expect the signal in LED calibration"),
+    strax.Option('noise_window',
+                 default=(350, 475),
+                 help="Window (samples) to analysis the noise"),
+    strax.Option('light_level',
```
ershockley: I feel like `light_level` isn't the best name for this config option. Maybe `channel_list`?

straxen/plugins/led_calibration.py
```diff
@@ -0,0 +1,114 @@
...
+             ('amplitudeNOISE', np.int32, 'Amplitude in off LED window'),
+             ('channel', np.int16, 'Channel'),
+             ('time', np.int64, 'Start time of the interval (ns since unix epoch)'),
+             ('dt', np.int16, 'Time resolution in ns'),
+             ('length', np.int32, 'Length of the interval in samples')
+            ]
+    
+    def compute(self, raw_records):
+        
+        r = raw_records[(raw_records['channel'] >= self.config['light_level'][0])&(raw_records['channel'] <= self.config['light_level'][1])]
```
ershockley: On second thought I'm not sure if we want to make a cut on the light_level (or channel_list) here anyway. Maybe it would be best to calculate for all channels at processing time and make selections like this in the analysis script? Having higher occupancy could be useful at some point maybe? It would take longer though...

straxen/plugins/led_calibration.py
```diff
@@ -0,0 +1,114 @@
...
+        temp['amplitudeLED'] = on['amplitudeLED']
+        temp['amplitudeNOISE'] = off['amplitudeNOISE']
+
+        area = get_area(r, self.config['LED_window'])
+        temp['area'] = area['area']
+
+        
+        return temp
+
+# QUESTIONS: can some nice functions of numba.njit be used? What does @export mean?
```
ershockley: Maybe you could do a speed test comparing this function using numpy and using njit like in this notebook [here](https://github.com/XENONnT/workshop_strax_chicago/blob/master/projects/LED_calibration/LEDcal_evan.ipynb) (In [106])

straxen/plugins/led_calibration.py
GiovanniVolta: Not yet but it is on my TODO list. I didn' t think about it untill yesterday when Shingo pointed out. I will check

straxen/plugins/led_calibration.py
GiovanniVolta: No no, it was a check for me. I forgot to remove it

straxen/plugins/led_calibration.py
GiovanniVolta: There is not any particular reason. I can easily change to 'led_cal'

straxen/plugins/led_calibration.py
GiovanniVolta: For the name of the variable you mean? I can add an underscore and avoid the capitol letters

straxen/plugins/led_calibration.py
GiovanniVolta: Got it! I will change it

straxen/plugins/led_calibration.py
GiovanniVolta: Chiara wanted to have the possibility to process the plug-in only for certain PMTs. This is way it has been added.

straxen/plugins/led_calibration.py
GiovanniVolta: I can try it

straxen/plugins/led_calibration.py
chiaracapelli: The light level is set to select the channels we want to analyse for each datasets. In 1T we had 3 levels corresponding to bottom, top inner and top outer rings. So in case we want to run it over all the pmts just leave the default setting, but for the standard calibration we can just pass the list of pmts we want to run with the given light levels and reduce the time o amount of data (in 1T case by a factor 3).

skazama:
https://github.com/XENONnT/straxen/blob/2f5b45e6a4448a058c72a6a9fd62156e9c3d731f/straxen/plugins/led_calibration.py#L110
@GiovanniVolta @chiaracapelli  Do you know where the baseline is calculated? Is baseline is already corrected for raw_records['data']?

straxen/plugins/led_calibration.py
GiovanniVolta: TypingError: Failed in nopython mode pipeline (step: nopython frontend)
Invalid use of Function(<built-in function zeros>) with argument(s) of type(s): (int64, dtype=list(tuple(unicode_type x 2)))
 * parameterized
In definition 0:
    All templates rejected with literals.
In definition 1:
    All templates rejected without literals.
This error is usually caused by passing an argument of a type that is unsupported by the named function.
[1] During: resolving callee type: Function(<built-in function zeros>)
[2] During: typing of call at /home/gvolta/XENONnT/straxen/straxen/plugins/led_calibration.py (88)


File "../straxen/straxen/plugins/led_calibration.py", line 88:
def get_amplitude(raw_records, led_window, noise_window):
    <source elided>
    '''
    on = np.zeros((len(raw_records)), dtype=[('channel','int16'),('amplitude_led', '<i4')])
    ^

This is not usually a problem with Numba itself but instead often caused by
the use of unsupported features or an issue in resolving types.

To see Python/NumPy features supported by the latest release of Numba visit:
http://numba.pydata.org/numba-doc/dev/reference/pysupported.html
and
http://numba.pydata.org/numba-doc/dev/reference/numpysupported.html

For more information about typing errors and how to debug them visit:
http://numba.pydata.org/numba-doc/latest/user/troubleshoot.html#my-code-doesn-t-compile

If you think your code should work with Numba, please report the error message
and traceback, along with a minimal reproducer at:
https://github.com/numba/numba/issues/new

I got this error after adding @njit. Am I using in the wrong way?

skazama:
@GiovanniVolta @chiaracapelli I implemented the baseline correction in the area calculation as shown below. This should give us consistent gain values between my calculation and Chiara's notebook.
https://github.com/XENONnT/straxen/blob/7b512d978980acae9e6f3a72f3da302e248c13cf/straxen/plugins/led_calibration.py#L113

GiovanniVolta:
> 
> 
> @GiovanniVolta @chiaracapelli I implemented the baseline correction in the area calculation as shown below. This should give us consistent gain values between my calculation and Chiara's notebook.
> 
> https://github.com/XENONnT/straxen/blob/7b512d978980acae9e6f3a72f3da302e248c13cf/straxen/plugins/led_calibration.py#L113

May I ask you Shingo to explain me how it works?

skazama:
> > @GiovanniVolta @chiaracapelli I implemented the baseline correction in the area calculation as shown below. This should give us consistent gain values between my calculation and Chiara's notebook.
> > https://github.com/XENONnT/straxen/blob/7b512d978980acae9e6f3a72f3da302e248c13cf/straxen/plugins/led_calibration.py#L113
> 
> May I ask you Shingo to explain me how it works?

sorry that I forgot to attach some plots here to show you the changes.
Here, you can see two plots of Gain vs PMT ID for the previous one (without baseline correction calculated by @chiaracapelli ) and the new one with baseline correction.
It's clear that this correction gives the consistent estimation.

Gain_vs_PMT_old
Gain_vs_PMT_new

skazama:
> > > @GiovanniVolta @chiaracapelli I implemented the baseline correction in the area calculation as shown below. This should give us consistent gain values between my calculation and Chiara's notebook.
> > > https://github.com/XENONnT/straxen/blob/7b512d978980acae9e6f3a72f3da302e248c13cf/straxen/plugins/led_calibration.py#L113
> > 
> > 
> > May I ask you Shingo to explain me how it works?
> 
> sorry that I forgot to attach some plots here to show you the changes.
> Here, you can see two plots of Gain vs PMT ID for the previous one (without baseline correction calculated by @chiaracapelli ) and the new one with baseline correction.
> It's clear that this correction gives the consistent estimation.
> 
> Gain_vs_PMT_old
> Gain_vs_PMT_new

What I did is as follows;

1. raw_records['data'][:, left_noise:right_noise].sum(axis=1): integrating baseline inside the noise window
2. , and then divide by (right_noise-left_noise) samples to calculate average baseline amplitude for 1 sample 
3. , and then multiply by (right - left) samples, which is signal integration length

skazama:
> > > > @GiovanniVolta @chiaracapelli I implemented the baseline correction in the area calculation as shown below. This should give us consistent gain values between my calculation and Chiara's notebook.
> > > > https://github.com/XENONnT/straxen/blob/7b512d978980acae9e6f3a72f3da302e248c13cf/straxen/plugins/led_calibration.py#L113
> > > 
> > > 
> > > May I ask you Shingo to explain me how it works?
> > 
> > 
> > sorry that I forgot to attach some plots here to show you the changes.
> > Here, you can see two plots of Gain vs PMT ID for the previous one (without baseline correction calculated by @chiaracapelli ) and the new one with baseline correction.
> > It's clear that this correction gives the consistent estimation.
> > Gain_vs_PMT_old
> > Gain_vs_PMT_new
> 
> What I did is as follows;
> 
> 1. raw_records['data'][:, left_noise:right_noise].sum(axis=1): integrating baseline inside the noise window
> 2. , and then divide by (right_noise-left_noise) samples to calculate average baseline amplitude for 1 sample
> 3. , and then multiply by (right - left) samples, which is signal integration length

the notebook to make the above plot can be seen here;
https://github.com/XENONnT/analysiscode/blob/master/PMTs/Gain/calc_gain.ipynb

chiaracapelli:
Thanks Shingo!

skazama:
Hi @GiovanniVolta, for your SPE analysis, I think baseline also needs to be corrected. Maybe [0, 50] samples can be used for calculating baseline, and then [50, 125] samples can be used for calculating the maximum amplitude of noise.

https://github.com/XENONnT/straxen/blob/e598d73dd0a93a50e83a7a596c4bf66f0e59edde/straxen/plugins/led_calibration.py#L92

straxen/contexts.py
```diff
@@ -101,6 +101,18 @@ def xenonnt_online(output_folder='./strax_data',
     return st
 
 
+def xenon1t_led(**kwargs):
```
jorana, line 104: Don't you want  `xenonnt_led` here?

GiovanniVolta: I think so! I think it is a typo

jorana: Alright, I'll push it with bootstrax if that is ok

jorana:
@GiovanniVolta , I checked it and I processed a run succesfully as if it were an LED-run.

GiovanniVolta:
> 
> 
> @GiovanniVolta , I checked it and I processed a run succesfully as if it were an LED-run on one of our eventbuilders.

I am very happy! Thank you :)

PR #45, opened 2020-01-28
Natural breaks clustering

JelleAalbers:
This changes the splitting algorithm for peaklets from a natural minimum finder to low-split natural breaks, described in https://github.com/AxFoundation/strax/pull/225.

This should lead to a better separation of the Kr S1 peaks, while Joey's peak merging (https://github.com/XENONnT/straxen/pull/36) should prevent the splitting of low-energy S2s.

The goodness of split threshold is a function of area and rise time. Basically there are two thresholds: one for S1s (low rise time) and another for S2s (high rise time), which we interpolate between using a sigmoid function.

The thresholds were chosen by this procedure:
  * Simulating 'bare' S1 and S2 peaks, without PMT afterpulses or single electrons.
  * Disabling essentially all peak splitting
  * Find out how the natural breaks goodness of split is distributed vs area. Choose by hand a curve that lies comfortably above the distribution.

More details and tests will follow in a note.

### DIFF
--- a/straxen/plugins/peaklet_processing.py
+++ b/straxen/plugins/peaklet_processing.py
@@ -17,12 +17,21 @@ export, __all__ = strax.exporter()
                  help="Include this many ns right of hits in peaks"),
     strax.Option('peak_min_pmts', default=2,
                  help="Minimum contributing PMTs needed to define a peak"),
-    strax.Option('peaklet_split_min_height', default=25,
-                 help="Minimum height in PE above a local sum waveform"
-                      "minimum, on either side, to trigger a split"),
-    strax.Option('peaklet_split_min_ratio', default=4,
-                 help="Minimum ratio between local sum waveform"
-                      "minimum and maxima on either side, to trigger a split"),
+    strax.Option('peak_split_gof_threshold',
+                 default=(
+                     None,  # Reserved
+                     ((0.5, 1), (3.5, 0.25)),
+                     ((2, 1), (4.5, 0.4))),
+                 help='Natural breaks goodness of fit/split threshold to split '
+                      'a peak. Specify as tuples of (log10(area), threshold).'),
+    strax.Option('peak_split_filter_wing_width', default=70,
+                 help='Wing width of moving average filter for '
+                      'low-split natural breaks'),
+    strax.Option('peak_split_min_area', default=40.,
+                 help='Minimum area to evaluate natural breaks criterion. '
+                      'Smaller peaks are not split.'),
+    strax.Option('peak_split_iterations', default=20,
+                 help='Maximum number of recursive peak splits to do.'),
     strax.Option('diagnose_sorting', track=False, default=False,
                  help="Enable runtime checks for sorting and disjointness"),
     strax.Option('to_pe_file',
@@ -43,7 +52,7 @@ class Peaklets(strax.Plugin):
     compressor = 'zstd'
     rechunk_on_save = True
 
-    __version__ = '0.2.1'
+    __version__ = '0.3.0'
 
     def infer_dtype(self):
         self.to_pe = straxen.get_to_pe(self.run_id, self.config['to_pe_file'])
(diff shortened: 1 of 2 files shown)


jorana:
Hi Jelle,
Thanks for this pull request and the nice explanation in https://github.com/AxFoundation/strax/pull/225. I'd like to start with a question on the latter PR. In that PR the Vanilla NB is not further elucidated in detail, I guess this is the product of Normalized variance and Low Split and that it is the implementation used in straxen in this PR (#45).

I've tested this branch and on the PeakFinder with WFSim (v0.0.3) and I do not find differences. The WFSim options I used:
```
enable_pmt_afterpulses=True, 
enable_electron_afterpulses=True
nelectrons = np.random.uniform(1, 200, n)
nphotons = np.random.uniform(1, 30, n)
```
For 5000 simulated S1s and 5000 simulated S2s the efficiencies are the same. Errors indicate the binomial 95% confidence interval.
'default' strax (before nat.-breaks):
 - S1 eff 0.1894 -0.0108 + 0.0111
 - S2 eff 0.9506 -0.0064 + 0.0058

custom (including nat-breaks #45):
 - S1 eff 0.1894 -0.0108 + 0.0111
 - S2 eff 0.9506 -0.0064 + 0.0058.



There is a very minor difference between the classification of the S2s:
'default' strax (before nat.-breaks):
 - **2.38% b'chopped'**
 - **1.20% b'split'**
 - 95.06% b'found'
 - 0.56% b'merged'
 - 0.02% b'misid_as_s1'
 - 0.60% b'missed'
 - 0.18% b'split_and_misid'

custom (including nat-breaks #45):
 - **2.50% b'chopped'**
 - **1.08% b'split'**
 - 95.06% b'found'
 - 0.56% b'merged'
 - 0.02% b'misid_as_s1'
 - 0.60% b'missed'
 - 0.18% b'split_and_misid'

Finally, I'm not so sure what the ```natural_breaks_threshold``` (https://github.com/XENONnT/straxen/pull/45/files#diff-c62206fd1f9601dc1502975520b1182bR128) does exactly and how the parameters therein are tuned. For example, why is there a reserved item at location [0] for the peak_split_gof_threshold? For this, I look forward to your note.

JelleAalbers:
Hi Joran, Thanks for looking into this!

> Vanilla NB is not further elucidated in detail

The goodness of split function for ordinary natural breaks is just under the big equation in https://github.com/AxFoundation/strax/pull/225, i.e. it is `s = sum[ w(t) (t - m)^2 ]` with `m = sum[t w(t)]/sum[w(t)]`, `w(t)` the sum waveform, and `t` the time labels of the sum waveform samples.

In this PR, we use "low split" natural breaks, since it seems to work best on the example waveforms (and a few others I looked at) in https://github.com/AxFoundation/strax/pull/225.

> I've tested this branch and on the PeakFinder with WFSim (v0.0.3) and I basically do not find differences

Great, thank you! This result makes sense for these isolated simulated low-energy peaks: we ignore peak fragments below 40 PE (`peak_split_min_area`) in natural breaks, and for low-energy S2s, the peak merging currently reverse anything we do to split up S2s below 5000 PE ~ 180 electrons (`s2_merge_max_area`).

> I'm not so sure what the natural_breaks_threshold (https://github.com/XENONnT/straxen/pull/45/files#diff-c62206fd1f9601dc1502975520b1182bR128) does exactly and how the parameters therein are tuned

Yeah, I should put this in the note... but let me give you a short preview:

The numbers are (log10_area, threshold) pairs between which we interpolate linearly, as shown in the figure below. Item 0 has no meaning, 1 applies for peaks with a low rise time (like S1s) and 2 for peaks with a high rise time (like S2s). So the reserved item is just so we can use indices starting from 1 in the code, otherwise 1 would refer to S2s.

To get these thresholds I simulated S1s and S2s without afterpulsing, photionization, or clustering. Then I plotted the peaks in (area, goodness of split) and chose lines that runs well above them, as shown in the figure below (blue is S1s, green S2s):

clustering_calibration

jorana:
Thanks Jelle for the explanation, it is most helpful. Maybe we could add your note to be to the plugin later so that it might facilitate the maintenance of these functions.

straxen/plugins/peaklet_processing.py
```diff
@@ -17,12 +17,21 @@
...
-                      "minimum, on either side, to trigger a split"),
-    strax.Option('peaklet_split_min_ratio', default=4,
-                 help="Minimum ratio between local sum waveform"
-                      "minimum and maxima on either side, to trigger a split"),
+    strax.Option('peak_split_gof_threshold',
+                 default=(
+                     None,  # Reserved
+                     ((0.5, 1), (3.5, 0.25)),
+                     ((2, 1), (4.5, 0.4))),
+                 help='Natural breaks goodness of fit/split threshold to split '
```
jorana: A comment pointing one to your note might help future maintenance.

straxen/plugins/peaklet_processing.py
```diff
@@ -76,16 +85,19 @@ def compute(self, records):
         # which is asserted inside strax.find_peaks.
         lone_hits = hits[strax.fully_contained_in(hits, peaklets) == -1]
 
+        # Compute basic peak properties -- needed before natural breaks
         strax.sum_waveform(peaklets, r, self.to_pe)
+        strax.compute_widths(peaklets)
 
-        # Split peaks based on local minima
+        # Split peaks using natural breaks
```
jorana: It could be an idea to explicitly state that only split_low is used.

straxen/plugins/peaklet_processing.py
JelleAalbers: Thanks, added one a few lines above

jorana:
Just to allow for easier lookup. The note is here:
https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:analysis:strax_clustering_classification#b_natural_breaks_splitting

PR #57, opened 2020-02-21
Deleting records which contain empty data.

WenzDaniel:
By default only fragments of the order 1 or higher are delete during this process.

Was faster then including this in my other PR.

### DIFF
--- a/straxen/plugins/pulse_processing.py
+++ b/straxen/plugins/pulse_processing.py
@@ -150,6 +150,9 @@ class PulseProcessing(strax.Plugin):
 
             # Probably overkill, but just to be sure...
             strax.zero_out_of_bounds(r)
+            
+            # Deleting fragments which contain zero data:
+            r = _del_empty(r, 1)
 
         return dict(records=r,
                     diagnostic_records=diagnostic_records,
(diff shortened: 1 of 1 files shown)


WenzDaniel:
Update: I just checked with the run 181027_2044 which fraction of fragments is actually empty and it is about 1%.

JelleAalbers:
Thanks Daniel! 1% is still a nice data reduction if it's free.

This would also delete an empty record in between two filled records in the same pulse, right? I guess this is harmless at this stage of processing, though I'm not 100% sure. Maybe we can start doing this in the next release of straxen, so if it causes any problems, we can isolate them from the bigger hitfinder and strax changes.

WenzDaniel:
> This would also delete an empty record in between two filled records in the same pulse, right?

Yes, but I can modify it if necessary. 

> Maybe we can start doing this in the next release of straxen, so if it causes any problems, we can isolate them from the bigger hitfinder and strax changes.

I agree, thanks.

WenzDaniel:
In addition to your comment Jelle, I just learned with the nVETO straxen plugins that there is an other issue which I have not thought of. 
I do not know how big the chances are to get something like this, but I was (un)lucky to find a hit with a right extension of a single sample into a higher fragment. This single sample was 0... strax.zero_out_of_bounds set also the rest of the fragment to zero and so it was cut. A bit later in the chain my higher order plugins wondered why they cannot find the fragment because it was supposed to contain real data....
I will try to rewrite this small function and include all comments.

jorana:
@WenzDaniel can we/you delete this branch?

PR #63, opened 2020-03-07
added elifetime

Jianyu010:
Can we add one entry about electron lifetime correction to event_info for S2(x,y) correction?

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 0.4.1
+current_version = 0.5.0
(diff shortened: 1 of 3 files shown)


JelleAalbers:
Hi Jianyu, it looks like you accidentally made a commit to our `stable` branch, which was meant to update on releases only. I've changed the permissions to prevent this now. Your PR also had a typo (causing the Travis test build to fail) and some unintended changes (changing the default context to point to your home directory). The idea of your change seemed fine, happy to consider this -- but please test it first, and have a look through the code comparison before you make a PR.

PR #64, opened 2020-03-08
Sort out subdetector outputs in DAQReader

JelleAalbers:
This:
 * Lets DAQReader sort/split raw records data from different subdetectors into different data types.
 * Makes a start with splitting off XENON1T and XENONnT contexts explicitly.

After this, DAQReader will be a multi-output plugin, producing `raw_records`, `raw_records_lowgain`, `raw_records_mv`, and `raw_records_aqmon`. If any of them has no entries (e.g. MV for a non-linked TPC run) the data types will still exist, but all chunks will be empty. On disk, they get a folder with only a metadata.json file (no empty files).

Sorting out the different subdetectors in DAQReader seems preferable over several alternatives (to me -- happy to hear other perspectives!):
  * **Having multiple DAQReaders** (what I had initially in mind) would put the burden on redax to sort things in different (sub)folders on ceph. Strax would also have a big headache trying to piece together the many independent source plugins producing data with not-perfectly aligned chunks (because final chunk breaks are set by where the data shows gaps).
  * **Sorting in PulseProcessing** is what we currently do. This means PulseProcessing shows up in the muon veto and acquisition monitor dependency chain, so when we e.g. change the TPC hitfinder threshold all muon veto processing would have to be redone from scratch. That can't be right.
  * **Let the first plugin of each subsystem select a channel range** would still mean you still have to (download and) go through the full raw_records, even to reprocess very light data such as the acquisition monitor or muon veto.
  * **Having separate channel-selector plugins** that do no processing for each subsystem would be almost equivalent to this PR does -- except that you would save the intermediate unsorted raw_records unnecessarily. Having many plugins depend on a big monolithic raw_records could also increase strax's memory usage, depending on under what conditions numpy arrays get copied internally.

This does not consider the neutron veto yet. I'm starting to agree with @darrylmasson that the most elegant solution would be a single DAQReader plugin that can handle everything. We would have to make a few modifications to deal with the different digitizer time resolution, and (if I understood correctly) the fact that the data will appear in a different folder. We'd also have to consider if the data is still sparse enough to look for simultaneous chunk breaks.

### DIFF
--- a/straxen/plugins/daqreader.py
+++ b/straxen/plugins/daqreader.py
@@ -67,7 +67,6 @@ class DAQReader(strax.Plugin):
         'raw_records',
         'raw_records_lowgain',
         'raw_records_aqmon',
-        'raw_records_aqmon_nv',
         'raw_records_mv')
 
     data_kind = frozendict(zip(provides, provides))
--- a/straxen/plugins/daqreader.py
+++ b/straxen/plugins/daqreader.py
@@ -288,6 +288,12 @@ class DAQReader(strax.Plugin):
         # Convert to strax chunks
         result = dict()
         for i, subd in enumerate(self.config['channel_map']):
+
+            # Ignore data from the 'blank' channels, corresponding to
+            # channels that have nothing connected
+            if subd.endswith('blank'):
+                continue
+
             result_name = 'raw_records'
             if subd != 'tpc':
                 result_name += '_' + subd
--- a/straxen/plugins/daqreader.py
+++ b/straxen/plugins/daqreader.py
@@ -3,7 +3,9 @@ import os
 import shutil
 import warnings
 
+from frozendict import frozendict
 import numpy as np
+import numba
 
 import strax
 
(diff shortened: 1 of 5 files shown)


straxen/plugins/daqreader.py
```diff
@@ -22,45 +24,63 @@ class ArtificialDeadtimeInserted(UserWarning):
...
+                 help="time of start run (s since unix epoch)"),
     strax.Option('daq_chunk_duration', track=False,
                  default=int(5e9), type=int,
                  help="Duration of regular chunks in ns"),
     strax.Option('daq_overlap_chunk_duration', track=False,
                  default=int(5e8), type=int,
                  help="Duration of intermediate/overlap chunks in ns"),
-    strax.Option('daq_input_dir', type=str, track=False,
-                 help="Directory where readers put data"),
+    strax.Option('daq_compressor', default="blosc", track=False,
```
darrylmasson: Is there a specific reason we're still specifying blosc as the default compressor for live data? Redax uses lz4 by default, and has for several months at this point.

JelleAalbers: Thanks, should be fixed now

straxen/plugins/daqreader.py
```diff
@@ -259,9 +279,84 @@ def compute(self, chunk_i):
...
+        for r in result.values():
+            print(f"\t{r}")
         return result
+
+
+@export
+class Fake1TDAQReader(DAQReader):
+    provides = (
+        'raw_records',
+        'raw_records_diagnostic',
```
jorana: @JelleAalbers I think this does not work with the xnt_common_config as above since this will result in a key error.

straxen/plugins/daqreader.py
```diff
@@ -259,9 +279,84 @@ def compute(self, chunk_i):
...
+                continue
+
+            result_name = 'raw_records'
+            if subd != 'tpc':
+                result_name += '_' + subd
+            result[result_name] = self.chunk(
+                start=self.t0 + break_pre,
+                end=self.t0 + break_post,
+                data=result_arrays[i],
+                data_type=result_name)
```
jorana: Shouldn't this be raw_records?

straxen/plugins/pulse_processing.py
```diff
@@ -62,13 +62,8 @@
              'Specify as a tuple of length n_tpc_pmts, or a number.'),
 
     strax.Option(
-        'n_tpc_pmts',
-        default=straxen.n_tpc_pmts,
+        'n_tpc_pmts', type=int,
```
jorana: For some reason I'm not able to acces this option. I'm not able to get it to the ```self.config``` by e.g. passing on these options: ```{'daq_input_dir': '/live_data/xenonnt/007079', 'daq_compressor': 'lz4', 'run_start_time': 1583745091.405, 'record_length': 500, 'n_readout_threads': 30, 'n_tpc_pmts': 493, 'channel_map': <frozendict {'tpc': (0, 493), 'lowgain': (500, 752), 'aqmon': (799, 807), 'tpc_blank': (999, 999), 'mv': (1000, 1083), 'mv_blank': (1999, 1999), 'diagnostic': (3000, 4000)}>}```

straxen/plugins/daqreader.py
JelleAalbers: True: the Fake1TDAQReader is only meant to be run on XENON1T data through the fake_daq. XENONnT can use DAQReader.

straxen/plugins/daqreader.py
JelleAalbers: We are looping over the different subdetectors here and creating a chunk for each. So this will take on several values; raw_records is one of them.

straxen/plugins/pulse_processing.py
JelleAalbers: Interesting: since no default is given, something must be getting to this option to allow the Travis build to succeed (at least in the XENON1T condition). But you must be passing the config correctly, otherwise the daqreader would have complained about missing daq_input_dir... Not sure what is going on.

straxen/plugins/daqreader.py
jorana: Ah okay, I thought it wasn't creating the other datakinds correctly. However, I see that the error I ran into is due to the omission of the aqmon_nv here. ttps://github.com/XENONnT/straxen/blob/ff98a67e42d922f2cc62d36cfdcfbf84a7635233/straxen/contexts.py#L35

straxen/plugins/pulse_processing.py
jorana: Could it be due to strax? If I add a default everyting works. I'm not sure what happens if no default is given here:
https://github.com/AxFoundation/strax/blob/master/strax/config.py

straxen/plugins/daqreader.py
jorana: Thanks, I did a register all instead of register for daqreader rather than for DAQReader

straxen/plugins/pulse_processing.py
JelleAalbers: I see you're only adding the `xnt_common_config` options inside `st_make`, rather than in `new_context`, maybe this is an issue. Otherwise, could you post the exact error you're getting?

I'm happy to dig for bugs in strax's config handling. It's just curious that `daq_input_dir` also doesn't have a default, but seems to be passed fine, and the fake DAQ has no trouble finding `n_tpc_pmts`.

jorana: > I see you're only adding the `xnt_common_config` options inside `st_make`, rather than in `new_context`, maybe this is an issue. Otherwise, could you post the exact error you're getting?
> I'm happy to dig for bugs in strax's config handling. It's just curious that `daq_input_dir` also doesn't have a default, but seems to be passed fine, and the fake DAQ has no trouble finding `n_tpc_pmts`.

Jelle you are a genius! It does work now. I must admit I'm quite surprised that as you say the ```daq_input_dir``` can be passed as an argument this way whereas the ```n_tpc_pmts``` apparantly cannot?

straxen/plugins/daqreader.py
JelleAalbers: Good catch, I removed the line

straxen/plugins/pulse_processing.py
JelleAalbers: Great! I think n_tpc_pmts is probably already used in some infer_dtype, so crash will complain about that at an earlier stage. Do you want to push the bootstrax fix to this PR? And are you OK to merge after that?

jorana: Sure, I'm merging after Travis approves

PR #66, opened 2020-03-10
Example is so out of date that it's not useful

sophiaandaloro:
(Chris also typing here)


tunnell:
Ah, thanks.  It's also version control so people can always see what was here.

Everybody in my group is running strax in different settings and doing comparisons, so we're just working on some machinery that takes a context list of strax configs then batch queues that.  We stumbled across this but it's easy enough to do something small, flexible, and nice that can generalize.

PR #65, opened 2020-03-09
KrDoubleS1Scatter plugins

DonNabla:
Implementation of the four plugins of the KrDoubleS1Scatter events processing. Those plugins are used for Kr events selection and analysis. The main features are the time ordering of the two main S1s and S2s peaks. Furthermore, several new variables, useful for the Kr events selection, have been implemented such as:

- **s1_a_area:** Area of the first main S1 in time.
- **s1_b_area:** Area of the second main S1 in time.
- **cS1_a and cS1_b:** Corrected area of S1_a and S1_b using the event position.
- **cS2_a and cS2_b:** Corrected area (position and electron lifetime) of the main and secondary S2 using their corresponding position.
- **s1_a_n_contributing:** The number of PMTs contributing to the first S1.
- **s1_b_n_contributing:** The number of PMTs contributing to the second S1.
- **ds_s1_b_n_distinct_channels:** The number of PMTs contributing to the second S1 (S1_b) distinct from the one contributing to the first.
- **ds_s1_dt:** The delay time between S1_a and S1_b peaks (using the center time of each peak).
- **ds_s2_dt:** The delay time between S2_a and S2_b peaks (using the center time of each peak).

Summary note on Kr events selection with strax is under construction.

PR #67, opened 2020-03-12
Move Maxime's improvements to EventBasics

JelleAalbers:
This:
  * Upgrades EventBasics to include the extra fields suggested by Maxime in https://github.com/XENONnT/straxen/pull/65.
  * Adds the computation of the number of channels contributing to the secondary but not primary S1 in a separate plugin.
  * Modifies Maxime's event_info plugin to load the existing event_basics, but change the field name convention to the conventional one for double scatter analyses -- e.g. `s1_xxx`->`s1_a_xxx`, `alt_s1_xxx` -> `s1_b_xxx`.
    * I'm actually not sure if it is a good idea to maintain two conventions long-term. But at least this should allow you and others to test it and transition more smoothly, if you want.

The reason for separating the distinct channel computation is that this requires going back to relatively low-level information only available in peaks. An alternative would be to add e.g. `does_channel_contribute` bools for each channel to peak_basics, but that means peak_basics would become of a comparable size to peaks itself. Since we have ~500 PMTs, the per-channel info is the part of `peaks` that takes the most space.

@DonNabla Do you agree with this approach, and if so, could you test it on a Kr dataset? I only tried it on a one-minute background set. Feel free to make other changes as well: this PR is only to make a modification to your branch in https://github.com/XENONnT/straxen/pull/65.

### DIFF
--- a/straxen/contexts.py
+++ b/straxen/contexts.py
@@ -11,6 +11,7 @@ common_opts = dict(
         straxen.peaklet_processing,
         straxen.peak_processing,
         straxen.event_processing,
+        straxen.double_scatter,
         straxen.cuts],
     store_run_fields=(
         'name', 'number',
--- a/straxen/plugins/KrDoubleS1Scatter.py
+++ b/straxen/plugins/KrDoubleS1Scatter.py
@@ -1,436 +1,72 @@
-import strax
+import numpy as np
 
-from straxen.common import pax_file, get_resource, get_elife, first_sr1_run
-from straxen.itp_map import InterpolatingMap
+import strax
 export, __all__ = strax.exporter()
 
+
 @export
-@strax.takes_config(
-    strax.Option('n_top_pmts', default=127,
-                 help="Number of top PMTs"))
-class DoubleS1Scatter_PeakBasics(strax.Plugin):
-    __version__ = "0.1.0"
-    parallel = True
-    depends_on = ('peaks',)
-    provides = 'DoubleS1Scatter_peak_basics'
+class DistinctChannels(strax.LoopPlugin):
+    __version__ = '0.1.0'
+    depends_on = ('event_basics', 'peaks')
     dtype = [
-        (('Start time of the peak (ns since unix epoch)',
-          'time'), np.int64),
-        (('End time of the peak (ns since unix epoch)',
-          'endtime'), np.int64),
-        (('Weighted center time of the peak (ns since unix epoch)',
-          'center_time'), np.int64),
-        (('Peak integral in PE',
-            'area'), np.float32),
-        (('Number of PMTs contributing to the peak',
-            'n_channels'), np.int16),
-        (('PMT number which contributes the most PE',
-            'max_pmt'), np.int16),
-        (('Area of signal in the largest-contributing PMT (PE)',
-            'max_pmt_area'), np.int32),
-        (('Width (in ns) of the central 50% area of the peak',
-            'range_50p_area'), np.float32),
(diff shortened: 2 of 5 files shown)


DonNabla:
Hi Jelle, I implemented a few things in your proposition, the main idea was to implement the time ordering of S1s and S2s peaks in a way that it will benefit to the others events class ( event_ position, corrected_areas) without unnecessary changes in their code. To do so, the ordering in time must be done in the event_basics class, as it is illustrated below:

Kr_Double_Scatter

In this way, the corrected areas will be properly derived from the position of each peak ordered in time.

I would like to know if there is a way to make understand strax that when someone is asking for the event_info_double class:

```python
data = st.get_df(['170206_1355'],['event_info_double'])
```
It implies that he has to set  **time_ordering = True** in the config.

```python
data = st.get_df(['170206_1355'],['event_info','distinct_channels'],config = dict(time_ordering = True))
```

JelleAalbers:
Hi Maxime, thanks for this! It looks like an innovative solution, which indeed means we would need to add very little extra code.

There is currently no way to tell strax to set a configuration option whenever a user loads a plugin, though technically everything is possible of course. A more explicit way would be to have a "Krypton context" for each analysis context, or a switch that toggles 'krypton mode' for a context.

However, regardless of how we implement this, it would mean the main and krypton analyses become quite separate. A cut defined in one (say on s2_area_fraction_top) could not be used in the other without changing meaning (and thus acceptance and rejection power). Since the definition of what "main S1" means changes, so does the meaning of most variables in event_basics and beyond.

I think there are two possible ways to deal with this:
  1. We keep the option you introduced in `EventBasics` to swap the main and alternate S1 if the alternate S1 occurs first. However, it would be only for experimental/exploratory use, and no official cuts would be defined with the option turned on. This way you can still enforce time ordering to make plots to distinguish populations in a Kr analysis, but afterwards, you would turn the option off again so you can use cuts from other analyses without them changing meaning (and other analyses can use your cuts). After all, the large S1 should come first in a good Kr event.
  2. Alternatively, we could go back to my original proposal, i.e. have the s1_a/b swap as an option for `EventInfoDouble`. It can now be left on by default: since `EventInfoDouble` and `EventBasics` use different names for their variables (s1_a/b vs s1/alt_s1), their computations are independent, so you can mix selections defined on both safely. The main drawback of this, as you pointed out, is that we would have to define main-S1 and alternate-S1 versions of variables in all plugins after `EventBasics` (such as `EventPositions`). The variables might be useful anyway, but if one plugin forgets to define them, it has the potential to mislead to the Kr analyst.

I think (1) would be the most elegant, unless time ordering is fundamental to getting a good Kr selection in a way I'm not appreciating.

How would you feel about this? Do you agree with the problem, or see an alternate solution?

DonNabla:
Hi Jelle, you raise a good point here. I agree that (1) is the best option. In addition, it will allow me to do a comparatives study of events selection with time-ordering vs area-ordering, which could help us to state the equivalence between them.

Therefore, I propose to merge this branch (in its current state) in my PR.

JelleAalbers:
Thanks Maxime! If you permit, I made two further changes:
  * Simplify the logic a bit. Rather than swapping all the properties (which got tricky for the drift times), we swap the main and alternate indices after they are assigned. 
  * Rename `time_ordering` to `force_main_before_alt`. 'Time ordering' could be confused with the requirement that S1s occur before S2s, or by taking the very first S1 in the event as the main S1 regardless of its size.

If you agree, let's merge both branches.

PR #68, opened 2020-03-14
Move bootstrax's context into contexts.py

JelleAalbers:
This:
  * Moves the online processing context created in bootstrax out to the main context file. This is mainly so you can load nT data in notebooks on the eb machine after `st = straxen.contexts.xenonnt_online()`.
  * Fixes some name/number handling in rundb.py.

I tested it briefly on eb0; bootstrax still seems to work.

### DIFF
--- a/bin/bootstrax
+++ b/bin/bootstrax
@@ -225,6 +225,7 @@ def new_context():
     return straxen.contexts.xenonnt_online(
         allow_multiprocess=args.cores > 1,
         allow_shm=args.cores > 1,
+        we_are_the_daq=True,
         max_messages=args.max_messages)
 
 st = new_context()
--- a/bin/bootstrax
+++ b/bin/bootstrax
@@ -222,19 +222,10 @@ def new_context():
     """Create strax context that can access the runs db"""
     # We use exactly the logic of straxen to access the runs DB;
     # this avoids duplication, and ensures strax can access the runs DB if we can
-    st = strax.Context(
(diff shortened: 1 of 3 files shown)


bin/bootstrax
```diff
@@ -222,19 +222,11 @@ def new_context():
...
-            runid_field='number',
-            new_data_path=output_folder,
-            mongo_dbname=dbname),
-        config=straxen.contexts.xnt_common_config,
-        **context_opts)
-    st.context_config['allow_multiprocess'] = True if args.cores > 1 else False
-    st.context_config['allow_shm'] = True if args.cores > 1 else False
-    st.context_config['max_messages'] = args.max_messages
-    return st
+    return straxen.contexts.xenonnt_online(
```
jorana: If we don't specify the ```output_folder``` the data isn't going to a place where datamanager can fetch it.

jorana:
Other than the minor comment it looks good Jelle!

bin/bootstrax
JelleAalbers: Good catch, thanks. Should be fixed now.

jorana:
Thanks Jelle, I've removed some superfluous DB calls that were now obsolete (
76b02f2). Since these commands were now a little scattered around, I bundled them again.

Also changed the reference you mentioned last Thursday for the nT-simulator (f03fe85) as you are editing context.py anyway.

PR #69, opened 2020-03-16
fix n_tpc_pmts option

jorana:
Double check that the number of PMTs in the nT-configuration match up with the to_pe file (which is currently not there).

Also the reference previously pointing to ```straxen.n_tpc_pmts``` was returning 248. Hence, I propose to rename this variable to avoid confusion.

@JelleAalbers would you say it makes sense to make a mock ```nt_to_pe``` in the aux. files and parse the reference thereto here:
https://github.com/XENONnT/straxen/blob/master/straxen/contexts.py#L32

### DIFF
--- a/bin/fake_daq
+++ b/bin/fake_daq
@@ -39,7 +39,7 @@ parser.add_argument('--sync_chunk_duration', default=0.2, type=float,
 args = parser.parse_args()
 
 n_readout_threads = 8
-n_channels = straxen.n_tpc_pmts
+n_channels = straxen.n_tpc_pmts_1T
 channels_per_reader = np.ceil(n_channels / n_readout_threads)
 
 if args.shm:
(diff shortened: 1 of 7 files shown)


JelleAalbers:
Hi Joran, thanks for finding these issues! This would solve the problem, but I would propose a slightly different solution in #71:
  * Removing straxen.common.n_tpc_pmts (and referencing the config) seemed cleaner than introducing a new common variable. The different mini-analyses that use the number should now also work for nT -- though some use other settings that are still hardcoded to 1T, such as the PMT positions.
  * The run -> to_pe convention is fine for XENON1T since no more new runs are added. If we wanted to use this for nT if would be hard to experiment with new gains. As you suggested over gitter, a more flexible option would be useful.

jorana:
Sure as #71 seems a smarter solution let's close this PR.

PR #71, opened 2020-03-16
Gain model specification, remove 1T defaults

JelleAalbers:
This:
  * Adds the `gain_model` and `hev_gain_model` options to describe PMT gain models (and replace the old `to_pe_file` option). These must be set to a tuple: (a model type string, a configuration or version specific to that model). The software high-energy veto has a separate option, because we may want to set this during online processing and never change it. Currently we only support a few basic model types; in the future time- or voltage-dependent models could be added:
    * `to_pe_constant`: configuration is a single number, the ADC -> PE conversion factor to use for all channels.
    * `to_pe_file`: configuration is a URL (or filename) pointing to a .npy file mapping run_id's to to_pe arrays.
    * `disabled`: configuration should be None. This is a placeholder for hev_gain_model to denote that the software-HEV should be disabled. (You can also disable the software-HEV by setting its threshold to a False value, but then it would be odd to still have to specify a gain model.)
  * Replace the XENON1T-like defaults in PulseProcessing by values more sensible for XENONnT first data (no pulse filter, no high-energy veto, generous hit extension, uniform hitfinder threshold and gain). The XENON1T settings are moved to `straxen.common.x1t_common_config`, which is auto-applied for 1T contexts. I hope we can move all 1T-specific defaults here eventually so it doesn't clutter straxen later on.
  * Removes `straxen.n_tpc_pmts`. The number of PMTs has for some time now been specified as part of the config.
  * Lets mini-analyses takes `config` as an argument. They could already get it through `context.config`, but this seems more convenient.

These changes are inspired by Joran's #69.

### DIFF
--- a/straxen/analyses/holoviews_waveform_display.py
+++ b/straxen/analyses/holoviews_waveform_display.py
@@ -64,7 +64,7 @@ def hvdisp_plot_pmt_pattern(*, records, to_pe, array='bottom'):
     return pmts
 
 
-def _records_to_points(*, records, to_pe, t_reference):
+def _records_to_points(*, records, to_pe, t_reference, config):
     """Return (holoviews.Points, time_stream) corresponding to records
     """
     import holoviews as hv
@@ -84,7 +84,7 @@ def _records_to_points(*, records, to_pe, t_reference):
         kdims=[hv.Dimension('time', label='Time', unit='sec'),
                hv.Dimension('channel',
                             label='PMT number',
-                            range=(0, straxen.n_tpc_pmts))],
+                            range=(0, config['n_tpc_pmts']))],
         vdims=[hv.Dimension('area', label='Area', unit='pe')])
 
     time_stream = hv.streams.RangeX(source=rec_points)
@@ -92,7 +92,7 @@ def _records_to_points(*, records, to_pe, t_reference):
 
 
 @straxen.mini_analysis(requires=['records'], hv_bokeh=True)
-def hvdisp_plot_records_2d(records, to_pe,
+def hvdisp_plot_records_2d(records, to_pe, config,
                            t_reference, width=600, time_stream=None):
     """Plot records in a dynamic 2D histogram of (time, pmt)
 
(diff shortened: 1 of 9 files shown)


straxen/analyses/holoviews_waveform_display.py
```diff
@@ -84,15 +84,15 @@ def _records_to_points(*, records, to_pe, t_reference):
...
+                            range=(0, config['n_tpc_pmts']))],
         vdims=[hv.Dimension('area', label='Area', unit='pe')])
 
     time_stream = hv.streams.RangeX(source=rec_points)
     return rec_points, time_stream
 
 
 @straxen.mini_analysis(requires=['records'], hv_bokeh=True)
-def hvdisp_plot_records_2d(records, to_pe,
+def hvdisp_plot_records_2d(records, to_pe, config,
```
jorana: If you are passing on the config, you could be a little lazier and omit the to_pe requirement and simply read it from the config. Might also prevent one from accidently using incompatible options.

straxen/analyses/records_matrix.py
```diff
@@ -8,7 +8,7 @@
 @straxen.mini_analysis(requires=('records',),
                        warn_beyond_sec=5e-3,
                        default_time_selection='touching')
-def records_matrix(records, time_range, seconds_range, to_pe):
+def records_matrix(records, time_range, seconds_range, config, to_pe):
```
jorana: Same holds here

straxen/analyses/holoviews_waveform_display.py
JelleAalbers: True, but that would mean adding
```
to_pe = straxen.get_to_pe(run_id, config['gain_model'], config['n_tpc_pmts'])
```
to every mini-analyses that uses `to_pe`. This syntax could also change if we introduce voltage-dependent gain models, since then we'd need an extra input (unless get_to_pe gets special powers, e.g. to fetch from the run doc or slow control info). Currently we have this line in mini_analysis.py, so if you require `to_pe` in a mini-analysis it is just an alias for the line above.

PR #72, opened 2020-03-19
switch naming to high energy

jorana:
The DAQ-group would prefer a consistent naming convention. As such it might be better to change this naming to be compliant with naming it:
 * high energy (for 0vbb search, the low gain channels that is)
 * low energy (for dark matter search, the high gain channels that is)

### DIFF
--- a/straxen/contexts.py
+++ b/straxen/contexts.py
@@ -49,7 +49,7 @@ xnt_common_config = dict(
     channel_map=frozendict(
          # (Minimum channel, maximum channel)
          tpc=(0, 493),
-         high_e=(500, 752),
+         high_energy=(500, 752),
          aqmon=(799, 807),
          tpc_blank=(999, 999),
          mv=(1000, 1083),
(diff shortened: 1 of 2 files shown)


tunnell:
Can you at least spell out the words such that it's easier for new people to understand?  raw_records_high_energy. Code is read more than written :)

jorana:
Thanks for the suggestion @tunnell

JelleAalbers:
@tunnell we discussed it, but most people preferred raw_records_he over raw_records_high_energy. We also have raw_records_mv, raw_records_aqmon, and a few others that would be long to write out. The abbreviated suffixes might also be used for data types further on in the chain.

tunnell:
Just makes it harder for newer people by increasing amount of jargon... but sure

tunnell:
Also, I hope that both raw_records_he and le are used when estimating S2 areas etc.  This is how it works in most experiments...

PR #74, opened 2020-03-25
Update bootstrax

jorana:
Update infer_run_mode introduced in #70 on the basis of:
https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:dsg:daq:eb_speed_tests_update

Most importantly we can process faster than previously anticipated (https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:dsg:daq:eb_speed_tests) due to three things:
  * Update to python v. 3.7
  * Updates in strax/straxen and https://github.com/JelleAalbers/npshmex
  * No pulse filtering: as for nT-data we don't know how the PMT-response will look like. As such we also don't have the appropriate filter. As this is a very CPU intensive step this is expected to change the results significantly. As such, if a filter will be implemented for nT, it might be good to redo these tests (partially).
  * Only process with the older ebs if the new ebs are falling behind (fix of https://github.com/XENONnT/straxen/issues/75) in cbfa793.

Additionally, whereas in older strax version the timeout for mailboxes was set to 300 s, this was lowered to 60 s. That is fine for low data rates. However, for higher datarates 60 s will unnecessarily result in failures while processing the run.

### DIFF
--- a/bin/bootstrax
+++ b/bin/bootstrax
@@ -647,8 +647,8 @@ def set_status_finished(rd):
         pass
     else:
         # Do not override this field for runs already uploaded in admix
-        raise ReferenceError(f'Trying to set set the status {rd.get("status")} to '
-                             f'{ready_to_upload}! One should not override this field.')
+        raise ValueError(f'Trying to set set the status {rd.get("status")} to '
+                         f'{ready_to_upload}! One should not override this field.')
 
 
 def abandon(*, mongo_id=None, number=None):
--- a/bin/bootstrax
+++ b/bin/bootstrax
@@ -457,11 +457,11 @@ def infer_run_mode(rd, input_dir):
     print(f'infer_run_mode::\tWorking with an estimated data rate of {data_rate} MB/s')
     run_settings = {
         'new_eb':
-            {'low_rate': dict(cores=12, max_messages=20, timeout=60),
-             'med_rate': dict(cores=12, max_messages=20, timeout=100),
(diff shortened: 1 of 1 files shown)


jorana:
In 5c46be3 a request from @lucascottolavina and @ershockley was added to mark the field 'status' as 'eb_ready_to_upload' as soon as processing was finished successfully. This field is what they plan to use for datamanager/admix to keep track of the readiness/status of the data.

bin/bootstrax
```diff
@@ -569,6 +589,25 @@ def set_run_state(rd, state, return_new_doc=True, **kwargs):
...
+        run_coll.find_one_and_update(
+            {'_id': rd['_id']},
+            {'$set': ready_to_upload})
+    elif rd.get('status') == ready_to_upload.get('status'):
+        # This is strange, boostrax already finished this run before
+        print('WARNING: bootstax has already marked this run as ready for upload. Doing nothing.')
+        pass
+    else:
+        # Do not override this field for runs already uploaded in admix
+        raise ReferenceError(f'Trying to set set the status {rd.get("status")} to '
```
JelleAalbers: The ReferenceError built-in exception apparently means something complicated about weak references; I had never seen it before. Raising a RuntimeError or a ValueError would be a bit more conventional.

jorana: I changed it in 641f12d

jorana:
Only process with the older ebs if the new ebs are falling behind (fix of #75) in cbfa793.

PR #76, opened 2020-03-27
Strax data to json microservice

JelleAalbers:
This adds a microservice `microstrax` based on [hug](https://www.hug.rest/), which serves arbitrary strax data over HTTP as json. To run it, just type `microstrax` in the command line, then go to e.g. `http://localhost:8000/get_data?run_id=XXXX&target=raw_records`. Hopefully this will help @darrylmasson @jorana with the online monitoring.

Features:
  * `--port` sets the HTTP port to listen to
  * `--context` sets which context in straxen.contexts.... we should use; 
  * API supports start, end, and selection_str arguments to customize data selection;
  * API supports `max_n` to limit the number of rows returned (defaults 1000);
  * Requests aren't allowed to load more than 1 GB of data;
  * Requests won't convert to jsons larger than 10 MB;
  * Requests aren't allowed to build new data. This requires the latest strax master (for [86552f](https://github.com/AxFoundation/strax/commit/86552f5ab64e8cbfc25eae4083afe5f5fe3f3346)).

This also moves some XENONnT-rundb authentication specifics from straxen.contexts into the RunDB class, so that we can run on the eventbuilders without claiming to be bootstrax.


bin/microstrax
```diff
@@ -0,0 +1,105 @@
...
+                # All of this chunk will be loaded
+                load_at_least_n += chunk_info['n']
+
+            if load_at_least_n > max_n:
+                # No point loading any more, we already exceeded max_n
+                time_range[1] = chunk_info['end']
+                break
+
+    if load_bytes > int(1e9):
+        raise ValueError(f"Cannot load {target} for run {run_id}: "
```
darrylmasson: Is raising an error the most robust way to signal to the caller that this doesn't work, or should we rather return something like `{'message' : "Cannot load..."}`, so the the caller knows why it didn't get anything when it asked? I've spent more hours than was probably necessary tracing 500 errors because the server code had a bug, and unanswered requests seems like a great way to generate these.

JelleAalbers: There is an exception handler defined, so if there is an error you will get {'error': 'something failed'} already. We could consider also returning the exception type and traceback though.

darrylmasson: Ok, that should be fine

bin/microstrax
```diff
@@ -0,0 +1,113 @@
...
+                         f"Try lowering max_n.")
+
+    return [dict(zip(row.dtype.names, row)) for row in x]
+
+
+if __name__ == "__main__":
+
+    parser = argparse.ArgumentParser(
+        description='Start a microservice to return strax data as json')
+    parser.add_argument('--context', default='xenonnt_online',
```
jorana: @JelleAalbers could we also specify one or more data-directories? It would be ideal if we could, for example, add five or six directories for the different event builders. Maybe this can help with this by making another field in the rundoc from the bootstrax-side.

JelleAalbers: Sure, I added an --extra_dirs argument which allows you to list as many directories as you like (e.g. `microstrax --extra_dirs /some/dir /some/other/dir`).

straxen/contexts.py
```diff
@@ -62,33 +62,26 @@
...
-                runid_field='number',
                 readonly=not we_are_the_daq,
+                runid_field='number',
                 new_data_path=output_folder),
             strax.DataDirectory(
                 '/dali/lgrandi/xenonnt/raw',
                 readonly=True,
-                take_only='raw_records')],
+                take_only='raw_records'),
+            strax.DataDirectory(output_folder),
```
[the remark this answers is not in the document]
JelleAalbers: Good point, can fix. But I didn't expect that particular error: you got this from bootstrax or microstrax?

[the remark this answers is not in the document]
jorana: It was bootstrax. I also wouldn't be expecting microstrax to process raw_records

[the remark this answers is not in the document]
JelleAalbers: Yes, microstrax shouldn't do any processing, so that's good. I tried to fix now, but feel free to fix it more / commit to master if this doesn't cut it.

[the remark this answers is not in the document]
jorana: Thanks Jelle, looks good.

PR #80, opened 2020-04-01
Separate LED contexts, numba for get_amplitude

JelleAalbers:
This:
  * Make separate contexts `xenon1t_led` and `xenonnt_led` for LED processing and analysis. They take the same arguments as `xenon1t_dali` and `xenonnt_online`, and return versions of these parent contexts that have all plugins deregistered except the ones providing led_calibration and raw_records.
  * Changes the led plugin a little bit (it did not work when I tried it) so `get_amplitude` now uses numba. I haven't really looked at what the code does, just tried to make small changes that leave the meaning invariant.

@GiovanniVolta this PR just goes to your branch, so it would be added on top of #42 if you agree

### DIFF
--- a/straxen/contexts.py
+++ b/straxen/contexts.py
@@ -100,13 +100,16 @@ def xenonnt_online(output_folder='./strax_data',
     return st
 
 
-def xenonnt_led(**kwargs):
+def xenon1t_led(**kwargs):
     st = xenonnt_online(**kwargs)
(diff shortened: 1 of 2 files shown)


JelleAalbers:
Hm, I thought github would only show my commits, but somehow it's repeating the entire LED branch...

JelleAalbers:
Hm, something went wrong here. For some reason it's showing a lot of commits to master as though they are part of this PR...

PR #82, opened 2020-04-01
aq mon starts at 800

jorana:
@JelleAalbers , I think the aqmon starts at 800:
https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:dsg:daq:channel_groups


JelleAalbers:
Yes, but https://github.com/XENONnT/straxen/blob/master/straxen/plugins/daqreader.py#L18

jorana:
Okay, that's clear. Thanks

PR #89, opened 2020-04-14
Deleting free_option in xenon1t_led

GiovanniVolta:
The context is not working, it seems there is a conflict with free_option. Is it indispensable?


jorana:
It is already specified here:
https://github.com/XENONnT/straxen/blob/master/straxen/contexts.py#L184

So you don't have to specify it again. You are right

JelleAalbers:
Thanks Giovanni!

PR #92, opened 2020-04-18
Temporarily discard mis-assigned data on chunk boundary

JelleAalbers:
This is a temporary fix to be able to build data affected by https://github.com/coderdj/redax/pull/82. After that is deployed and the data is built, we should reverse this hack.

### DIFF
--- a/straxen/plugins/daqreader.py
+++ b/straxen/plugins/daqreader.py
@@ -156,6 +156,18 @@ class DAQReader(strax.Plugin):
         first_start, last_start, last_end = None, None, None
         if len(records):
             first_start, last_start = records[0]['time'], records[-1]['time']
+
+            if last_start == end:
+                # TEMPORARY HACK: Discard data assigned to wrong chunk due to
+                # https://github.com/coderdj/redax/pull/82
+                print(
+                    f"Bad data from DAQ: chunk {path} should contain data "
+                    f"that starts in [{start}, {end}), but we see start times "
+                    f"ranging from {first_start} to {last_start}.")
+                print('-> discarding offending data for now!!')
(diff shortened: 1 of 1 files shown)


JelleAalbers:
The build failure on python 3.6 also affects master, even though master built fine yesterday. So this must be due to a dependency update. Monday's round of automatic updates on base_environment will hopefully tell us more. Anyway, not a relevant issue for this I think.

jorana:
@JelleAalbers should we maybe add it as an option that changes the lineage so that we can easily distinguish between runs with this option and those without?

I've just checked out this PR on eb3 (py3.7) and tested it on run 7251; it works.

straxen/plugins/daqreader.py
```diff
@@ -156,6 +156,18 @@ def _load_chunk(self, path, start, end, kind='central'):
         first_start, last_start, last_end = None, None, None
         if len(records):
             first_start, last_start = records[0]['time'], records[-1]['time']
+
+            if last_start == end:
```
jorana, line 160: Maybe do it only when a tracked option == ``True`` in order to change the lineage for now?

JelleAalbers:
Thanks for testing! If we change the lineage people still won't be able to load these runs. We could add a non-tracked option for now, but I think we'll be turning it off forever on Monday once Darryl deploys https://github.com/coderdj/redax/pull/82.

jorana:
Makes sense! I'll merge and update the ebs.

jorana:
I cannot merge because of travis, I'll simply checkout this branch, we then also don't have to remove this temporary change from master again.

JelleAalbers:
Sorry, hit the button before reading your second sentence :-)

PR #96, opened 2020-04-19
Temporary bootstrax warnings for out of place timestamps

jorana:
**Disable check to allow processing on the DAQ**
At the moment there are some timestamp issues in the daq so we temporary disable this the check if the timestamps are in accordance with the duration of the run. As soon as this is fixed we should again implement this check rather than simply warning about it.

**Push deleting the live_data to a subprocess**
It was noted that deleting live_data from ceph was rather slow. Commits https://github.com/XENONnT/straxen/pull/96/commits/8b3a7a304947623bcc8852b8ff1b2e3b18bcba50  and https://github.com/XENONnT/straxen/pull/96/commits/6790c8205f6494e3406b74e0af475624d4adb6ed push the task of deleting the /live_data to a new thread such that one does not have to wait for the files of the previous run to be removed.

### DIFF
--- a/bin/bootstrax
+++ b/bin/bootstrax
@@ -574,7 +574,11 @@ def delete_live_data(rd, live_data_path):
 
     if os.path.exists(live_data_path) and not args.keep_live:
         print(f'Deleting data at {live_data_path}')
+        os.system(f'du -h -d0 {live_data_path}')
+        del_start = time.time()
         shutil.rmtree(live_data_path)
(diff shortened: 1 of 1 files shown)


jorana:
@JelleAalbers , if you approve, could you also merge?

JelleAalbers:
I hope Travis is fixed now, so you can also merge yourself :-) But happy to do it otherwise

PR #101, opened 2020-04-22
Restore checks on DAQ data

JelleAalbers:
This restores some checks bootstrax straxen does on the DAQ data (disabled in 5d73ee9dc0ceb13a4042873f923e5fdec1261ed6, https://github.com/XENONnT/straxen/pull/96). Now that the timestamp and overwriting issues seem to be fixed, normal data should pass these checks.


jorana:
Looks good Jelle!

PR #107, opened 2020-04-29
event_scatter colorbar label fix

sophiaandaloro:
Event scatter plot does not have an override setting for when the colorbar does not indicate S1_area_fraction_top. This will return the default colorbar label, unless user specifies the colorbar to be something else (say, n_peaks.) Then the colorbar label will just be the colormap indicator used.


sophiaandaloro:
Lol sure! I figured you could always do plt.gca() but since I missed it I figured others may also do the same, so a good catch before saving/sending around plots with the wrong label! :) Thanks for reviewing!

PR #108, opened 2020-04-30
update cuts for new strax version

jorana:
For the new strax version (https://github.com/XENONnT/straxen/pull/60) each plugin needs to have a time and endtime. This PR adds it to the xenon1t cuts.


jorana:
Good idea. Makes more sense than having the same exact lines in every plugin as I did now.

jorana:
In https://github.com/jorana/strax/pull/18 a CutPlugin is proposed that should somewhat simplify adding cuts as one doesn't have to copy paste the same lines of code for each plugin

PR #111, opened 2020-05-04
Keeps the live data entry in the rundoc

darrylmasson:
Rather than removing all trace of the live data existing once live processing concludes and the data on disk is deleted, its entry in the rundoc is kept, but its location is changed to "deleted".


jorana:
Thanks Darryl. I've fixed a missing comma and also replaced one other instance where we delete dat to be consistent with what you propose.

As soon as Travis finishes I'll merge it with master.

PR #113, opened 2020-05-08
Fix coordinate flipping

zhut19:
Thanks, @terliuk for bringing this issue back in https://github.com/XENONnT/WFSim/issues/47
This negligence of mine being carried all the way here.


terliuk:
Just a small question. I think @petergaemers mentioned that 1T simulations seemed to be fine. Did you check there as well?

petergaemers:
Ah there it is!

@terliuk, 1T configuration doesn't use the interpolating map. It uses a krypton map so it doesn't call this function

PR #114, opened 2020-05-12
Use world accessible mirror for RunDB

tunnell:
Suggest using fried rice, even though singularity doesn't access internet at all.


ershockley:
Hey Chris, we cannot use fried rice only until the Midway admins whitelist that site. I sent them an email yesterday so hopefully they respond soon. It's not a singularity problem.

Can't we just include both xenon1t-daq and fried rice? Or is that not recommended?

tunnell:
It depends on where you think people will run.  If default is LNGS then straxen doesn’t work out of the box on somebody laptop.  I agree just to wait on this since not high priority, but just smart default.  The default can be both rice and Chicago mirrors.

tunnell:
This is still pending

darrylmasson:
I'm not sure what permissions the LNGS firewall has, but `xenon1t-daq` is fully open on 27017. Also, given that we're now back to the nominal 3 member replica set, we should put this into a "final" configuration.

PR #115, opened 2020-05-13
Bootstrax db interactions

jorana:
Update bootstrax to:

- Write to the updated adresses of the Mongo dbs
- Only write to the 'eb_monitor' collection for monitoring bootstrax. This renders the 'bootstrax' collections in both the runs database and the daq database obsolete. 
- Remove functions that were doing the same kind of things as the system monitor does.
- Change the philosophy of how the bootstrax/eb_monitor collection is used. In the past, we were writing one single entry per host to the 'bootstrax' collection. However, to have a log for the website to read from we are now inserting new entries into the 'eb_monitor' collection rather than updating.

The ‘eb_monitor’ collection will now log what parameters it is running with. For example:
```
'host': 'eb0.xenon.local', 
'pid': 29272, 
'time': datetime.datetime(2020, 5, 13, 12, 11, 47, 769000),
'state': 'idle',
'target': 'event_info', 
'max_cores': 4, 
'max_messages': 4, 
'undying': False, 
'production_mode': True
```

when it starts a run it updates fields like:
```
'run_id': 000000
'target': 'raw_records', 
'max_cores': 40, 
'max_messages': 10, 
```


bin/bootstrax
```diff
@@ -328,25 +320,19 @@ def main():
 
 def main_loop():
     """Infinite loop looking for runs to process"""
-    global state_doc_id
 
     # Wait for mongo connection
     ping_dbs()
 
     # Ensure we're the only bootstrax on this host
     any_running = list(bs_coll.find({'host': hostname}))
```
darrylmasson, line 328: Depending on how large the collection cap size is, this could give a cursor with a lot of documents. A possible issue this could cause (other than wasted CPU cycles) is that we could spend time trying to kill a task that's already been killed. Possibly sort by descending `_id` and take the most recent 2 or 3?

bin/bootstrax
```diff
@@ -474,31 +435,45 @@ def override_target(rd):
...
-        update['state'] = state
-    bs_coll.update_one(
-        {'_id': state_doc_id},
-        {'$set': update})
-
+    if not state:
+        # Find the last message of this host
+        previous_entry = bs_coll.find_one({'host': hostname,
+                                           'time': {'$lt': now(-timeouts['bootstrax_presumed_dead'])}},
+                                          sort=[('time', pymongo.DESCENDING)])
```
darrylmasson, line 449: I've had issues with `find_one` and the `sort` argument, I got around it via `find(query).sort().limit(1)`

bin/bootstrax
jorana: Good point. How about this:
I add a extra time field to the eb_montitor. If the current time is within say 60 s I'll update the latest _id in the collection. Otherwise I start a new one. This of course depends on what we want to show exactly on the website.

By the way the collection is capped to 36288000 B (36 MB) and 181440 files.

bin/bootstrax
jorana: Thanks, I have tested this syntax and it did seem to work. I can double check.

darrylmasson: I tested this call on the runs db: `client.xenonnt.runs.find_one({'detector' : 'tpc'}, sort=[('time', -1)])['number']` and got `0`, which is obviously not correct.

bin/bootstrax
darrylmasson: Documents in a capped collection can't be updated, unless we implement the cap as a TTL index (which is how the system monitor and readout status collections are handled).

Also, does it make sense to update the query to only look for other pids? `{'pid' : {'$ne' : os.getpid()}}`?

bin/bootstrax
```diff
@@ -474,31 +435,45 @@ def override_target(rd):
...
+        if not previous_entry:
+            raise ValueError('No previous entries nor state set. Are you running a rogue bootstrax?')
```
darrylmasson, line 451: If the main loop responds slowly for some reason, is there a possibility that we reach this condition and crash?

bin/bootstrax
jorana: I have set 'eb_monitor' as a TTL collection (very short time of now to check it is working as desired) when confirmed I'll increase the time. 

Thx for the query suggestion!

bin/bootstrax
jorana: The reason why `client.xenonnt.runs.find_one({'detector' : 'tpc'}, sort=[('time', -1)])['number']` does not work is that there is no 'time' field in the rundoc. You can however sort by `_id` and that is what your original suggestion is doing when you:
`find(query).sort('_id',-1).limit(1)`

bin/bootstrax
jorana: This is indeed not the best place to raise ValueErrors for something we foresee happening. I'll change it.

bin/bootstrax
darrylmasson: We actually don't have any indexes in the rundocs. We might want to think about that at some point, it'll help performance in a few years once there's many more thousands of runs.

jorana:
Thanks Darryl for your review. I've implemented most in 8ad0d0d. Additionally I have given each object in the eb_monitor collection an effective lifetime of approx. 60 seconds. After that a new document is inserted in the collection.

bin/bootstrax
```diff
@@ -683,19 +682,26 @@ def _delete_data(rd, path, data_type):
...
         ddoc.update({'at': now(), 'by': hostname})
+        log.info(f'update with {ddoc}')
         run_coll.update_one({'_id': rd['_id']},
                             {"$addToSet": {'deleted_data': ddoc},
-                             "$pull" : {"data" :
-                                    {"data.type" : data_type,
-                                     "data.host" : hostname}}})
+                             "$pull": {"data":
+                                           {"type": data_type,
+                                            "host": {'$in':['daq', hostname]}}}})
```
darrylmasson, line 704: What about `ddoc["host"]`?

jorana: I contemplated that too but I want to make sure we only delete stuff that is either from the daq or this machine. What if for some reason we are deleting data that is also e.g. on dali, we should never touch those, bootstrax may only say something about data that is either from itself or from the readers.

darrylmasson: Good point. We can't trust that data won't have already been shipped out when we call this routine.

PR #117, opened 2020-05-13
fried contexts

tunnell:
Adding a context for fried accessing xenon1t and xenonnt data.  Since this is JupyterHub, it might be useful for demos.


tunnell:
Found another solution

PR #121, opened 2020-05-18
Tweaks how clean_run works

darrylmasson:
If a run fails in processing bootstrax tries to clean up. This will inadvertently delete things we don't want. This PR should fix this.


jorana:
@darrylmasson I've moved this under a single umbrella of _delete_data where we also remove these items from the 'data' entry and write them under the 'deleted_data' entry in the rundoc.

PR #125, opened 2020-05-26
Giovanni volta patch 1

GiovanniVolta:
I have modified **channel_list** option, implementing it as list. Now we can pass the information of which PMT we want to build the led_calibration plugin.
This is very helpful for the light levels defintion. Futhermore it speeds up the gainand SPE acceptance script.

I have tested with the HV scan data.

### DIFF
--- a/straxen/plugins/led_calibration.py
+++ b/straxen/plugins/led_calibration.py
@@ -2,7 +2,7 @@
 Dear nT analyser, 
 if you want to complain please contact: chiara@physik.uzh.ch, gvolta@physik.uzh.ch, kazama@isee.nagoya-u.ac.jp
 '''
-
+import datetime
 import strax
 import numba
 import numpy as np
@@ -15,7 +15,7 @@ channel_list = [i for i in range(494)]
 @export
 @strax.takes_config(
     strax.Option('baseline_window',
-                 default=(0,50),
+                 default=(0,40),
                  help="Window (samples) for baseline calculation."),
     strax.Option('led_window',
                  default=(75, 105),
@@ -26,6 +26,7 @@ channel_list = [i for i in range(494)]
     strax.Option('channel_list',
                  default=(tuple(channel_list)),
                  help="List of PMTs. Defalt value: all the PMTs"))
+
 class LEDCalibration(strax.Plugin):
     """
     Preliminary version, several parameters to set during commisioning.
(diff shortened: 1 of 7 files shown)


skazama:
Hi Giovanni, thanks for the modification! I think you can also change the led_window (75, 110) to (70, 105) based on the plot you showed me.
[https://github.com/XENONnT/straxen/blob/GiovanniVolta-patch-1/straxen/plugins/led_calibration.py#L21](url)

[wf_100ns_1400HV_6_0V.pdf](https://github.com/XENONnT/straxen/files/4694240/wf_100ns_1400HV_6_0V.pdf)

straxen/plugins/led_calibration.py
```diff
@@ -11,20 +11,20 @@
...
     strax.Option('led_window',
-                 default=(75, 110),
+                 default=(75, 105),
                  help="Window (samples) where we expect the signal in LED calibration"),
     strax.Option('noise_window',
                  default=(20, 55),
                  help="Window (samples) to analysis the noise"),
     strax.Option('channel_list',
-                 default=(0,494),
+                 default=(channel_list),
```
jorana, line 27: Hi Giovanni, maybe a naive question but wouldn't it be faster to add another plugin on top of this one to extract this information? Guessing from your description it looks like you'd want to reprocess this plugin using various versions of the channel_list. My guess would be that having this plugin doing all the channels as a default and extracting the required channels later would be much faster as you don't have to go back to raw_records each time?

GiovanniVolta: Thank you for the comment Joran, all the suggestions are very welcomed! As you pointed out it should be already kind of implemented. For every raw records chunk, which are quite fast to load, I am selecting only those PMTs belong to the **channel_list**.

JelleAalbers:
Thanks Giovvanni. Could you make the default value a tuple instead of a list? So `tuple(channel_list)` instead of `(channel_list)`. Mutable defaults are a bit scary, see e.g. [here](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments) or [here](https://florimond.dev/blog/articles/2018/08/python-mutable-defaults-are-the-source-of-all-evil/).

straxen/plugins/led_calibration.py
chiaracapelli: Hi, this would be a good idea for the stable calibration during the science data taking, once we established which channels are associated with the relative light level. I think something to associate the n-datasets of a single calibration to a defined list of pmts each, based on the filename of the dataset.

PR #128, opened 2020-06-05
Fix deep matplotlib waveform plot

JelleAalbers:
@tunnel [reported](https://gist.github.com/tunnell/2f8c90cc26fe779bd23043a2a6f8a5ae) that plotting large time ranges (> few ms) with `st.plot_waveform(..., deep=True)` can result in an empty records plot.

index

I suspect the cause is that the records matrix became very large (many GBs), causing rendering goofs.

This modifies `st.records_matrix` (used by the waveform plot) to create downsampled matrices if the requested time axis exceeds 20_000 samples (configurable through the `max_samples` argument). This solves the problem:

```
import straxen
st = straxen.contexts.xenon1t_dali()
run_id = '170204_1710'
df = st.get_array(run_id, "event_info")
event = df[4]
st.plot_waveform(run_id, time_within=event, deep=True)
```

bla

It also has other benefits:
  * We actually fill a smaller matrix with downsampled waveforms, rather than downsampling a huge matrix. Thus preparing the plot (not just the rendering itself) is much faster and takes far less memory.
  * You can now plot very large time ranges with the matplotlib display -- though the color scale inference could use improvement -- e.g.

```
st.plot_waveform('180215_1029', seconds_range=(0, 10), deep=True)
```
bla_long

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 0.8.1
+current_version = 0.9.0
 files = setup.py straxen/__init__.py
 commit = True
 tag = True
--- a/bin/bootstrax
+++ b/bin/bootstrax
@@ -18,7 +18,7 @@ You can run more than one bootstrax instance, but only one per machine. If you s
 
 Philosophy
 ----------------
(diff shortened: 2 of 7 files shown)


skazama:
Hi, this is just a bug fix, so we can merge now.

PR #129, opened 2020-06-10
Bootstrax  updates and removing data from ebs

jorana:
In this PR several bootstrax/bin changes are implemented mostly aimed at safely removing data from the eventbuilders.

Most importantly, I envisage a new script ``ajax`` for deleting old data from the eventbuilders. I think that as bootstrax is dynamically increasing quite some functionality is now aimed at removing data from the eventbuilders. By moving this functionality to a separate script bootstrax may become more clearly aimed at processing new data.

Furthermore there is a small change was made in the logic when to start up the old event builders (in case the new ones are swamped), we saw this recently if the buffers were full.

Please note that bootstrax is not using ajax at the moment (this is still a draft). I used it for cleaning stuff of the very old runs in the incorrect format we had when the buffers were filled up.

Additionally Darryl added several improvements as explained below.

### DIFF
--- a/bin/ajax
+++ b/bin/ajax
@@ -18,6 +18,15 @@ import shutil
 import pymongo
 import pytz
 
+# TODO
+#  Ajax is aimed at making bootstrax simpler and more devoted to the processing while Ajax
+#  is clearing the data behind it.
+#  To this purpose I want to add:
+#  Delete the failed runs from the /live_data partition
+#  Delete old high level plugins
+#  Move/Delete old raw_records_nv
+#  Delete live_data if the run has been successfully processed and uploaded
+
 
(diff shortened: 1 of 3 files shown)


jorana:
In c8737d9 I further added the ajax script. As written at the top of the script. This tool keeps the event builders clean by:
 - deleting data associated to a run if it has been abandoned (delete live_data and data from eb)
 - deleting data if there is old high level data on the machine
 - re)moving old raw_records_nv as we do not plan to store this data long term
 - check that the live_data is deleted if a run has been successfully processed and uploaded

In principle I plan to have a screening session open with:
```
ajax --mode clean_all --execute
```

PR #130, opened 2020-06-11
Rechunk pulse_counts and veto_regions

JelleAalbers:
This uses the new strax feature in https://github.com/AxFoundation/strax/pull/272 to indicate pulse_counts and veto_regions should be rechunked. 

I tested this by building 180215_1029 from fake_daq-generated fake-live data. Indeed pulse_counts and veto_regions now have just one file, as opposed to 28 for records and raw_records.

### DIFF
--- a/straxen/contexts.py
+++ b/straxen/contexts.py
@@ -1,5 +1,3 @@
-import warnings
-
 from immutabledict import immutabledict
 import strax
 import straxen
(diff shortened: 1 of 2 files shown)


JelleAalbers:
Thanks Joran! Indeed, since rechunking happens [in the saver](https://github.com/AxFoundation/strax/blob/master/strax/storage/common.py#L544) after the data has passed through the last mailbox, I'd be surprised if this is related to any mailbox timeouts. But you never know...

PR #132, opened 2020-06-14
Update gain model

GiovanniVolta:
I have added the adc_to_pe arraies for PMT at 1.3, 1.4, 1.5, 1.6 kV. Futhermore, I have added the arrayis also for the DC data taken in nitrogen and in vacuum.
I am not sure if this is the right way for doing it so, please, let me know. Moreover let me know if the format is okay or you prefer less digits.
The notebook which I have used to make the arrays is here: https://github.com/XENONnT/analysiscode/blob/master/PMTs/Gain/XENONnTGain/notebook/adc_to_pe.ipynb.


petergaemers:
There was a syntax error in the dict. It should be fixed now. Lets wait for what travis will say

tunnell:
@ahiguera-mx thoughts here?  Keeping gains this way from PR makes a bit of a mess, but not sure if it's too soon for corrections infrastructure?

feigaodm:
@tunnell do you think this is something you want to improve now? Or we can merge it and improve later?

ahiguera-mx:
@feigaodm we are woking on the implementation of correction management tool (CMT), tomorrow we will discuss it with members of the pmt group

tunnell:
@skazama I guess Shingo decided :)  I think merging now and improving later is fine.

PR #133, opened 2020-06-16
add filecount for admix checksums

jorana:
Luca has pointed out that it would be good to have the number of files for each of the folders for a given plugin in the runs DB. I think we can add that info to the runs DB with this line.


JelleAalbers:
Thanks Joran! I remember we used to (and would guess that we still) register data to the rundb as we start processing, not when it is done. If so, this would always create the field with None in it, right?

jorana:
Thanks Jelle, you are right! I will ask you more in the upcoming straxferno course! 

I'll add some lines to bootstrax instead to solve this problem.

straxen/rundb.py
```diff
@@ -153,10 +153,17 @@ def _find(self, key: strax.DataKey,
                 doc = self.collection.find_one(run_query, projection={'_id'})
                 if not doc:
                     raise ValueError(f"Attempt to register new data for non-existing run {key.run_id}")   # noqa
+
+                # Count the number of files in the output path
+                file_count = None
+                if os.path.exists(output_path) and len(os.listdir(output_path)):
```
darrylmasson, line 159: What's the logic behind `and len(os.listdir(output_path))`? If the folder exists but is empty, wouldn't we want to store `0` rather than `None`?

jorana: Good point changed it

jorana:
In c8737d9 I changed it to have bootstrax fill the required number.

PR #140, opened 2020-06-29
Added average baseline and average baseline rms values to pulse_counts

WenzDaniel:
For a quickly monitoring the overall PMT rates `pulse_counts` is already quite handy and fast to load. However if one wants to check the stability of the baseline one has to go back all the way down to records where data loading becomes slow. To overcome this problem I added an additional `baseline_mean` field as well as an `baseline_rms_mean` field which stores the average baseline and baseline rms value per PMT.

### DIFF
--- a/straxen/plugins/pulse_processing.py
+++ b/straxen/plugins/pulse_processing.py
@@ -343,7 +343,6 @@ def rough_sum(regions, records, to_pe, n, dt):
 ##
 # Pulse counting
 ##
-
 @export
 def pulse_count_dtype(n_channels):
     # NB: don't use the dt/length interval dtype, integer types are too small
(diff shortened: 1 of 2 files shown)


straxen/plugins/pulse_processing.py
```diff
@@ -384,20 +387,23 @@ def _count_pulses(records, n_channels, result):
     # Array of booleans to track whether we are currently in a lone pulse
     # in each channel
     in_lone_pulse = np.zeros(n_channels, dtype=np.bool_)
-
+    baseline_buffer = np.zeros(n_channels, dtype=np.float64)
```
jorana, line 390: I think you could do with integers here as https://github.com/AxFoundation/strax/blob/master/strax/dtypes.py#L50 is stored in integers.

WenzDaniel: Thanks, this is what I tried first and the final result is also stored as int16. However when computing the mean value I had to use a float64 since otherwise the results deviated in some cases from the expected value:

`np.mean(records_of_first_fragment_in_one_channel['baseline'])`

jorana:
Hi Daniel, maybe we should also up the version to make the lineage change explicit.

JelleAalbers:
You're right Joran, though since Shingo just merged another PR updating the PulseProcessing version is should be fine if we merge now.

Issue #138, opened 2020-06-28
Lower number of cores and max messages while increasing timeout for failing runs


jorana:
If bootstrax fails a run, it retries it several times. If for some reason it continues to fail for a given run we change the target to raw_records to prevent us from trying using a broken event_basics for many times. Additionally, we should give it more time and less cores/max mailbox messages (as that makes it more reliable).

jorana:
Fixed in #143

PR #143, opened 2020-07-03
lower bootstrax resources and use xenon_sectets

jorana:
Two minor additions are made to bootstrax:
  - **lower the resources available to process a run to be more lenient on CPU/RAM when a run has failed 10 times or more.**  Was raised in https://github.com/XENONnT/straxen/issues/138. 
I could have implemented a roll-of where each fail gradually lowers the amount of resources. I choose not to do this as this might waste too much time on a problem that actually needs a proper fix or is simply not related to the CPU/RAM usage. 
Please do note that sub-optimal performance of the eventbuilders may stay undetected if this happens regularly! For example a few days I restarted the eventbuilders as their swap and RAM were clogged up with all kind of stuff. One could imagine that if this happens very often one would not notice after this PR. However, especially comming year, I will be paying close attention to the rundb and often query the runsdb with the following syntax to check that this doesn't happen too often: 
```
{"bootstrax.n_failures":{"$gt":1}, "number":{"$gt":8000}}
```

  - **use xenon_sectrets.py as a resource for rundb passwords.** This is a technical issue of not being able to restart bootstrax as the environment variables are not available over ```ssh -t 'bootstrax'```. Using this syntax allow us to store the passwords in the xenon_sectrets.py without having to mingle with the sshd_config of the eventbuilders.
 - Finally I've also **upgraded the default of the number of cores and max messages** as these values were ridiculously low.

### DIFF
--- a/bin/bootstrax
+++ b/bin/bootstrax
@@ -568,26 +568,14 @@ def eb_can_process():
     return False
 
 
-def infer_mode(rd, input_dir):
+def infer_mode(rd):
     """Infer a safe operating mode of running bootstrax based on the size of the first
     chunk. Estimating save parameters for running bootstrax from:
     https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:dsg:daq:eb_speed_tests_update
     returns: dictionary of how many cores and max_messages should be used based on an
     estimated data rate.
     """
-    nap_time, i = 10, 0
-    default_opt = result = dict(cores=args.cores, max_messages=args.max_messages, timeout=300)
-
-    # min. of 3 chunks with each pre_, chunk and post_
-    min_chunks = 3
-    while len(os.listdir(input_dir)) < min_chunks * 3:
-        if i >= 10:
-            # Too few files to be sure the first chunk is completely written
-            log.info(f'Waited {i * nap_time} s but less than {min_chunks} were written. '
-                     f'Could not infer run mode. Using default: {result}.')
-            return default_opt
-        i += 1
-        time.sleep(nap_time)
+    default_opt = dict(cores=args.cores, max_messages=args.max_messages, timeout=300)
 
     # Get data rate from dispatcher
     try:
@@ -598,7 +586,7 @@ def infer_mode(rd, input_dir):
         data_rate = sum([d['rate'] for d in docs])
     except Exception as e:
         log_warning(f'infer_mode ran into {e}. Cannot infer mode, using default mode.')
-        return default_opt
+        data_rate = None
 
     # Find out if eb is new (eb3-eb5):
     is_new_eb = int(hostname[2]) >= 3  # ebX.xenon.local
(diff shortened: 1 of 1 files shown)


bin/bootstrax
```diff
@@ -200,12 +200,18 @@ dead_state = 'dead_bootstrax'
 
 # The maximum time difference (s) allowed between the timestamps in the data and the
 # duration of the run (from the runs metadeta). Fail if the difference is larger than:
-max_timetamp_diff = 2
+max_timetamp_diff = 5
```
darrylmasson, line 203: Do we have sufficient confidence that we know the edge cases where this large difference is necessary and handle them individually? One is noise data, and one would involve busy/HEV at the very end of a run.

Alternately, do we leave this generally large and assume that the analysts will find the issues?

bin/bootstrax
```diff
@@ -200,12 +200,18 @@ dead_state = 'dead_bootstrax'
...
 
 # The maximum number of retries for processing a run. After this many times of retrying
 # to process a run, the DAQ-group has to either manually fix this run or manually fail it.
 max_n_retry = 20
 
+# Bootstrax retries runs multiple times. If there have been this many fails we could lower
+# the resources to be somewhat more lenient on the CPU and RAM of the eventbuilders. Use
+# this option with care as we might be running in a sub-optimal mode that may not be
+# noticed and eventbuilders may be spending more time on trying to reprocess failed runs.
+lower_resources_after_n_failures = 10
```
darrylmasson, line 213: A nice-to-have feature would be something that can extract some useful information from the error message and decide if reducing resources would actually solve the issue. Very far outside the scope of this PR, but maybe something for a rainy week.

bin/bootstrax
```diff
@@ -624,7 +630,17 @@ def infer_mode(rd, input_dir):
         result = run_settings[eb]['very_high_rate']
     else:
         result = run_settings[eb]['max_rate']
-    log.info(f'Override processing mode for run {rd["number"]} changing to:\t {result}')
+    log.info(f'infer_mode::\tOverride processing mode for run {rd["number"]} changing to:'
+             f'\t {result}')
+
+    # Lower the settings if there are continuous failures on this run:
+    if rd['bootstrax'].get('n_failures', 0) >= lower_resources_after_n_failures:
+        result['cores'] //= 2
```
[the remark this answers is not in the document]
jorana: Well you make a new ```run_settings``` every time overwriting any cashed ```run_settings``` so no this will not happen.
afbeelding

bin/bootstrax
jorana: > Do we have sufficient confidence that we know the edge cases where this large difference is necessary and handle them individually? One is noise data, and one would involve busy/HEV at the very end of a run.

I'm not entirely sure. You probably know better than I do. 

> Alternately, do we leave this generally large and assume that the analysts will find the issues?
This is indeed the approach I'm adopting here. You can consider 5 seconds reasonably large.

bin/bootstrax
jorana: Thanks for the idea! Of course this is somewhat hard to know a-priori as CPU and RAM issues have been throwing some quite funny messages because of multiprocessing. Might be nice to indeed keep in mind though.

bin/bootstrax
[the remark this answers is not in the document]
darrylmasson: If the memory model is what we'd normally expect, then there shouldn't be any issues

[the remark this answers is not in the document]
jorana: To be absolutely sure I've changed it in 882c8b5. Additionally I changed some (old) code to also do this lowering of the resources if we end up with the default number of cores and messages (because the aggregate status isn't available or so)

PR #146, opened 2020-07-09
check all records for endtime determination as it affects chunking

jorana:
This is the straxen counterpart of https://github.com/AxFoundation/strax/pull/281. Please see the problem explained there. 

**The problem**
See https://github.com/AxFoundation/strax/pull/281.

**Symptoms**
During a high rate run we didn't observe any problems for `raw_records` and `records` however for `lone_hits` we ran into:
``` 
ValueError: Attempt to create chunk [008543.raw_records: 1594281342sec 999999000 ns - 1594281348sec 499999000 ns, 5836752 items, 258.9 MB/s] whose data ends late at 1594281348499999220
```

This was presumably caused by the fact that while the (long) pulse was in the middle of  `raw_records` and `records` but was classified as the last `lone_hit` leading to the error above.

**Daqreader implementation**
As illustrated nicely below by Daniel we need to make sure we have all the latest possible wfs of each channel as raw_records may overlap. In a worse case scenario we need to look for all the channels and each of the channels has a 1 sample long pulse (which is unrealistic but lets assume this for the sake of argument).

This means we have to look back for a single channel the length of each of the samples. Next we treat all the channels known to the daqreader as an independent channel with its own timestamps (which is also very conservative). Hence we have to look back:
```
n_samples * length_of_samples
```

### DIFF
--- a/straxen/plugins/daqreader.py
+++ b/straxen/plugins/daqreader.py
@@ -156,8 +156,9 @@ class DAQReader(strax.Plugin):
         first_start, last_start, last_end = None, None, None
         if len(records):
             first_start, last_start = records[0]['time'], records[-1]['time']
-            # Records are fixed length, so we also know the last end:
-            last_end = strax.endtime(records[-1])
+            # Records are sorted by (start)time and are of variable length.
+            # Their endtimes can differ.
+            last_end = strax.endtime(records).max()
             if first_start < start or last_start >= end:
                 raise ValueError(
                     f"Bad data from DAQ: chunk {path} should contain data "
--- a/straxen/plugins/daqreader.py
+++ b/straxen/plugins/daqreader.py
@@ -157,8 +157,12 @@ class DAQReader(strax.Plugin):
         if len(records):
             first_start, last_start = records[0]['time'], records[-1]['time']
             # Records are sorted by (start)time and are of variable length.
-            # Their endtimes can differ.
-            last_end = strax.endtime(records).max()
+            # Their end-times can differ. In the most pessimistic case we have
+            # to look back one record length for each channel.
+            tot_channels = np.sum([np.diff(x)+1 for x in
+                                   self.config['channel_map'].values()])
+            look_n_samples = self.config["record_length"] * tot_channels
+            last_end = strax.endtime(records[-look_n_samples:]).max()
             if first_start < start or last_start >= end:
                 raise ValueError(
                     f"Bad data from DAQ: chunk {path} should contain data "


straxen/plugins/daqreader.py
```diff
@@ -156,8 +156,9 @@ def _load_chunk(self, path, start, end, kind='central'):
         first_start, last_start, last_end = None, None, None
         if len(records):
             first_start, last_start = records[0]['time'], records[-1]['time']
-            # Records are fixed length, so we also know the last end:
-            last_end = strax.endtime(records[-1])
+            # Records are sorted by (start)time and are of variable length.
+            # Their endtimes can differ.
+            last_end = strax.endtime(records).max()
```
WenzDaniel, line 161: Maybe my logic is flawed, but would not it be enough for records to check the last len(data) - 1 * channel records since they are fixed length? We cannot have two record spanning the same time range for the same channel, hence we have only len(data) - 1  * channel possibilities. Maybe use 110 + 1 * 494  in case I did a mistake in my calculation :D

Example:
len(data) = 5 dashes 

ch4: |-----| 
ch3: _|-|    
ch2: __|-|
ch1: ___|-|
ch6: ___|-|
ch8: ___|--|
ch7: _____|-| <- same end time as ch4.
etc.

WenzDaniel: Might be equivalent to testing all raw_records anyhow.

jorana: Yeah thanks Daniel, indeed I think the worse case scenario is:
len(data) = 5 dashes
```
ch1: |-|-|-|-|_
ch2: .|-|-|-|-|_
ch3: ..|-----|
```
Sorted by start time all records of ch1 and ch2 precede ch3. So we would indeed need (len(data) -1 ) * n_channels to get the the record of channel 3.

Although the time needed is not that large, the relative times for one of the latest 200 MB/s runs for a given chunk is significant:
afbeelding

I'll change it

darrylmasson: Yeah, I'm not sure that there's much difference between checking the last 50,000 records and the whole chunk

darrylmasson: Also, if I read those benchmarks correctly, it's faster to do the whole chunk than part of it, by a factor of nearly 3

jorana: This would be the implementation I'd go for which is like 500 us. In the most pessimistic case we'd have to consider all known channels.

```
            # Records are sorted by (start)time and are of variable length.
            # Their end-times can differ. In the most pessimistic case we have
            # to look back one record length for each channel.
            tot_channels = np.sum([np.diff(x)+1 for x in
                                   self.config['channel_map'].values()])
            look_n_samples = self.config["record_length"] * tot_channels
            last_end = strax.endtime(records[-look_n_samples:]).max()
```

PR #150, opened 2020-07-16
change S1 split threshold main contexts

jorana:
For the data we are currently taking we see that too many peaks are split. This is due to the fact that we are detecting a lot of large S1s, the split threshold for these S1-like signals is to low. This PR solves this.

For more info see [Raise peak splitting requirement for S1-like signals in GXe on the XENONnT wiki](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:analysis:meetings:peak_splitting_gxe) as this repo is public.

**Implications of merging this PR**
If we merge this PR, that means that effectively **all** nT data we currently have is out of date and could not be able to be loaded using this context until we rebuild from records. Only the the datatypes in the figure below would be loadable of the data we currently have. This includes simulation data!

afbeelding
Everything above this we would have to reprocess (see https://straxen.readthedocs.io/en/latest/reference/datastructure.html#event-positions).

1T data / context / analyses would not be affected.

**Alternative(s)**
We could alternatively change only the xenonnt_online context by adding lines and inserting ```context``` 
[here](https://github.com/XENONnT/straxen/blob/92aaf23f6635cfab06ad9d0de13ccb4cadbf47da/straxen/contexts.py#L55):
```python
 config = straxen.contexts.xnt_common_config
 config['peak_split_gof_threshold'] = (
        None,  # Reserved
        ((0.5, 1), (4, 0.4)),
        ((2, 1), (4.5, 0.4)))
```
I choose not to do this as the simulation- and data-contexts would diverge and we couldn't compare the two easily anymore.

**Related PRs and notes**
- https://github.com/AxFoundation/strax/pull/225
- https://github.com/XENONnT/straxen/pull/45
- [Peak splitting note](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:analysis:meetings:peak_splitting_gxe)
- [Natural breaks splitting](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:analysis:strax_clustering_classification#b_natural_breaks_splitting)
- [Example wfs](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:analysis:meetings:20200713:fast_response_gxe:alpha_s1_wf)

**Unrelated**
I fixed some error in the simulation context.

### DIFF
--- a/straxen/contexts.py
+++ b/straxen/contexts.py
@@ -26,7 +26,11 @@ xnt_common_config = dict(
          aqmon=(790, 807),
          tpc_blank=(999, 999),
          mv=(1000, 1083),
-         mv_blank=(1999, 1999)))
+         mv_blank=(1999, 1999)),
+    peak_split_gof_threshold=(
+        None,  # Reserved
+        ((0.5, 1), (4, 0.4)),
+        ((2, 1), (4.5, 0.4))))
 
 
 ##
--- a/straxen/contexts.py
+++ b/straxen/contexts.py
@@ -26,11 +26,7 @@ xnt_common_config = dict(
          aqmon=(790, 807),
          tpc_blank=(999, 999),
          mv=(1000, 1083),
-         mv_blank=(1999, 1999)),
-    peak_split_gof_threshold=(
-        None,  # Reserved
-        ((0.5, 1), (4, 0.4)),
-        ((2, 1), (4.5, 0.4))))
+         mv_blank=(1999, 1999)))
 
 
 ##
@@ -140,7 +136,11 @@ x1t_common_config = dict(
     peak_right_extension=30,
     peak_min_pmts=2,
     save_outside_hits=(3, 3),
-    hit_min_amplitude='XENON1T_SR1')
+    hit_min_amplitude='XENON1T_SR1',
+    peak_split_gof_threshold=(
+        None,  # Reserved
+        ((0.5, 1), (3.5, 0.25)),
+        ((2, 1), (4.5, 0.4))))
 
 
 def demo():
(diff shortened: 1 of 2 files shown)


jhowl01:
Thanks @jorana ! I was going to ask for a detection efficiency validation, but I think NaturalBreaks is a small effect for those smaller S1s. I think you've demonstrated the modification makes sense.

petergaemers:
Also with deconvolution s1 peaks get split. I've seen this multiple times with simulated data. 
s1_splits

The peaks in the middle are the main S1s which get cut off. The parts at the left are the remainders of the cut peaks

jorana:
> Thanks @jorana ! I was going to ask for a detection efficiency validation, but I think NaturalBreaks is a small effect for those smaller S1s. I think you've demonstrated the modification makes sense.

Yes, this will have to be calculated at some point. But as Jelle said let's not over tune the clustering before we have appropriate data.

> I previously set the XENONnT settings in the plugin defaults

This makes a lot of sense. I'll commit the change

>  if you enable deconvolution later you'd have to redo the clustering tuning.

Thanks, fully agree. 

> Also with deconvolution s1 peaks get split. I've seen this multiple times with simulated data.

I'm actuality not that surprised see e.g. this peak taken from https://github.com/AxFoundation/strax/pull/225. This is really not that far off from the 0.25 set for these kind of energies (see left y-axis and purple line).

afbeelding

JelleAalbers:
Indeed; some splitting could be beneficial since there are often PMT afterpulses and single electrons after the S1. A bias/linearity study would give the final verdict. But clearly there was too much splitting for the current data.

jorana:
@ACs this is ready for merge

feigaodm:
I think it's good to make the clustering change in this PR, we will need to retune a lot when we see liquid data anyway later.

PR #158, opened 2020-07-23
Update contexts.py

petergaemers:
This pull request:
  * Sets the position reconstruction to a XENONnT neural network to the common config


jorana:
@feigaodm this can be merged. It will only affect data > peaks_positions for xenonnt which we don't really have (or should have).

PR #161, opened 2020-07-28
High energy channel processing

petergaemers:
This pull request adds the first high energy channel processing

Plugins added:
- PeakProcessingHe
- PeakletsHe
- PeakletClassificationHe
- MergedS2sHe
- PeaksHe
- PeakBasicsHe

Actual computation is based on the normal version of the plugin. Any changed needed to be made to the conf is done in the setup of infer_dtype functions


straxen/plugins/high_energy_processing.py
```diff
@@ -0,0 +1,121 @@
...
+        help='Minimum hit amplitude in ADC counts above baseline. '
+             'See straxen.hit_min_amplitude for options.'
+    )])
+
+
+@export
+@strax.takes_config(
+    *HITFINDER_OPTIONS_he)
+class PulseProcessingHe(straxen.PulseProcessing):
+    __version__ = '0.2.5'
```
jorana, line 21: You could also start with version 0.0.0 right?

straxen/plugins/high_energy_processing.py
```diff
@@ -0,0 +1,121 @@
...
+    def compute(self, raw_records_he, start, end):
+        result = super().compute(raw_records_he, start,end)
+        return result['records']
+
+    
+@export
+@strax.takes_config(
+
+    strax.Option(
+        'n_tpc_pmts_all',track=False,default=800
```
jorana, line 44: Perhaps you also want to change some of the options like min_area et cetera

straxen/plugins/high_energy_processing.py
```diff
@@ -0,0 +1,121 @@
...
+    data_kind = 'peaklets_he'
+    __version__ = '0.0.3'
+    parrallel=False
+    
+    def infer_dtype(self):
+        return strax.peak_dtype(
+                        n_channels=self.config['n_tpc_pmts_all'])
+
+    def setup(self):
+        self.config['hit_min_amplitude'] = self.config['hit_min_amplitude_he']
```
jorana, line 60: Is the lineage fine like this or does it become very confusing if you look at the metadata?

straxen/plugins/high_energy_processing.py
jorana: Also for other plugins

straxen/plugins/high_energy_processing.py
```diff
@@ -0,0 +1,121 @@
...
+    """Classify peaklets as unknown, S1, or S2."""
+    provides = 'peaklet_classification_he'
+    depends_on = ('peaklets_he',)
+    __version__ = '0.0.1'
+
+    def compute(self, peaklets_he):
+        return super().compute(peaklets_he)
+
+
+FAKE_MERGED_S2_TYPE = -42
```
jorana, line 84: This can be omitted I think otherwise you import it from peaklet processing

straxen/plugins/high_energy_processing.py
```diff
@@ -0,0 +1,121 @@
...
+                        n_channels=self.config['n_tpc_pmts_all'])
+
+    def setup(self):
+        self.config['hit_min_amplitude'] = self.config['hit_min_amplitude_he']  
+        self.to_pe = straxen.get_to_pe(self.run_id,
+                                       self.config['gain_model'],
+                                       n_tpc_pmts=self.config['n_tpc_pmts'])
+        buffer_pmts = np.zeros(500)
+        self.to_pe = np.concatenate((buffer_pmts,self.to_pe))
+        self.to_pe *= 20
```
jorana, line 66: better add this to an option ?

straxen/plugins/high_energy_processing.py
```diff
@@ -0,0 +1,121 @@
...
+    __version__ = '0.0.1'
+
+    def compute(self, peaklets_he):
+        return super().compute(peaklets_he)
+
+
+FAKE_MERGED_S2_TYPE = -42
+
+@export
+class MergedS2s(straxen.MergedS2s):
```
jorana, line 87: Forgot He

straxen/plugins/high_energy_processing.py
```diff
@@ -0,0 +1,121 @@
...
+    
+    def infer_dtype(self):
+        return strax.unpack_dtype(self.deps['peaklets_he'].dtype_for('peaklets_he'))
+
+    def compute(self, peaklets_he):
+        return super().compute(peaklets_he)
+
+
+@export
+class Peaks(straxen.Peaks):
```
jorana, line 102: forgot He

straxen/plugins/high_energy_processing.py
darrylmasson: In principle this is something we could store somewhere once we've calibrated the per-channel gain difference. It's going to be roughly 20, but we don't know the spread.

straxen/plugins/high_energy_processing.py
WenzDaniel: In the lineage you will find both options `hit_min_amplitude` and  `hit_min_amplitude_he` both set to the same value.

I think I found a solution which does not require a lot of changes, I am currently trying it.

straxen/plugins/high_energy_processing.py
darrylmasson: Also shouldn't we be dividing by 20 rather than multiplying?

straxen/plugins/high_energy_processing.py
petergaemers: You can change them by setting min_area. When we'll have a new default value we should add them

straxen/plugins/high_energy_processing.py
petergaemers: I'm not a paying attention to details kind of guy

straxen/plugins/high_energy_processing.py
petergaemers: Yes this is not used

straxen/plugins/high_energy_processing.py
petergaemers: The signal is 20 times smaller. So 1 pe in a he channel is 20 real pe's

petergaemers:
The most "to be improved" part is the number of channels used in peaklets. The area_per_channel field now has a length of 800  to store ~250 numbers

darrylmasson:
The context does store information about which channels are HE, maybe this information could be leveraged somehow?

jorana:
By the way, apart from tests don't forget to add it to the init:
https://github.com/XENONnT/straxen/blob/master/straxen/plugins/__init__.py

straxen/plugins/high_energy_processing.py
WenzDaniel: Could you make this number an option? Kind of similar to the hitfinder threshold? There you can specify either an array or just a single number. https://github.com/AxFoundation/strax/blob/0f90d8c6dc520d5747e5108957ac7044a580274d/strax/processing/pulse_processing.py#L194

straxen/plugins/high_energy_processing.py
jorana: yeah, I meant that if we know anyway that the parameters should be different it's maybe easier to add them now (and make educated guesses like for the coincidence requirement which would be ~2 times lower since there are twice as little HE channels)

straxen/plugins/high_energy_processing.py
```diff
@@ -0,0 +1,126 @@
...
+        return result['records']
+
+    
+@export
+@strax.takes_config(
+
+    strax.Option(
+        'n_tpc_pmts_all',track=False,default=800
+        ),
+    strax.Option(
```
WenzDaniel, line 46: Is not it possible to take the `channel_map` as config here? Allows to simplify lines 69++

WenzDaniel:
One additional minor things. Do we want to keep everything in a single file, split it like for the TPC or maybe add the he plugins to the very same files as their parents?

darrylmasson:
I would argue that the easier maintenance will be to have the HE plugins in the same file as their low-energy parent.

straxen/plugins/high_energy_processing.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+@export
+@strax.takes_config(
+    strax.Option(
+        'n_tpc_pmts_all',track=False,default=800
+        ),
+    *HITFINDER_OPTIONS_he)
+class PulseProcessingHe(straxen.PulseProcessing):
+    __version__ = '0.0.1'
```
WenzDaniel, line 24: One thing which just came to my, I think we must also inherit the version number from the parent. Otherwise changes in the parent wont be visible in the child

WenzDaniel: Okay never mind I think the best thing would be to have individual version numbers. But we should leave in the parent plugins a comment next to the version number that we also have to increase the versions of the he plugins.

straxen/plugins/high_energy_processing.py
petergaemers: this is already done

straxen/plugins/high_energy_processing.py
```diff
@@ -0,0 +1,126 @@
+import strax
+import straxen
+import numpy as np
+
+export, __all__ = strax.exporter()
+
+# These are also needed in peaklets, since hitfinding is repeated
+HITFINDER_OPTIONS_he = tuple([
+    strax.Option(
```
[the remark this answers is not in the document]
petergaemers: That did not work. But maybe it was because of something els

straxen/plugins/high_energy_processing.py
petergaemers: Better way is to add the parent in the depends on, then it will track changes in version of the parent in the lineage

petergaemers:
One way to do it is to add a channel offset to find_peak https://github.com/AxFoundation/strax/blob/0f90d8c6dc520d5747e5108957ac7044a580274d/strax/processing/peak_building.py#L13 
For normal data it's 0, for this 500. Then subtract it from the channel number when writing to the array. 

As for same file or seperate file.. Same file might be a bit more easy? Not really convinced of it either way

straxen/plugins/high_energy_processing.py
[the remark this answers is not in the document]
WenzDaniel: Can you poste the error?

straxen/plugins/high_energy_processing.py
WenzDaniel: Okay I will add this.

petergaemers: Even better is to add the version of the parent as an option

jorana: @petergaemers , that would make it quite confusing no? You can't actually choose what version of the parent you use

WenzDaniel: In https://github.com/AxFoundation/strax/pull/297 the version of the parent is directly retrieved from the registered plugins in the currently used context. This ensures that if you change the context you get the correct version of the parent.

petergaemers: It is about tracking where the data comes from

straxen/plugins/high_energy_processing.py
```diff
@@ -0,0 +1,155 @@
...
+        return dtype
+
+    def setup(self):
+        self.hev_enabled=False
+        self.config['n_tpc_pmts'] = self.config['n_tpc_pmts_he']
+        self.config['hit_min_amplitude'] = self.config['hit_min_amplitude_he']
+
+    def compute(self, raw_records_he, start, end):
+        result = super().compute(raw_records_he, start,end)
+        return result['records']
```
WenzDaniel, line 44: Could you also check if some of the other outputs would also make sense for the he channels? I guess something like pulse_counts_he would be nice to have  too.

straxen/plugins/high_energy_processing.py
```diff
@@ -0,0 +1,155 @@
...
+        ),
+    strax.Option(
+        'parent_version',track=True,default=straxen.PulseProcessing.__version__
+        ),
+    *HITFINDER_OPTIONS_he)
+class PulseProcessingHe(straxen.PulseProcessing):
+    __version__ = '0.0.1'
+    rechunk_on_save = False
+    depends_on = 'raw_records_he'
+    provides = 'records_he'
```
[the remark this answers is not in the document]
petergaemers: Ofcourse. This is the "minumum working example". It's better to just do provides = ('records_he','pulse_counts_he'). Doing it like that makes sense if we would make an abstract base class do it. Now it's just less readable :P

straxen/plugins/high_energy_processing.py
```diff
@@ -0,0 +1,155 @@
...
+        self.config['hit_min_amplitude'] = self.config['hit_min_amplitude_he']  
+        self.to_pe = straxen.get_to_pe(self.run_id,
+                                       self.config['gain_model'],
+                                       n_tpc_pmts=self.config['n_tpc_pmts'])
+        buffer_pmts = np.zeros(self.config['he_channels_offset'])
+        self.to_pe = np.concatenate((buffer_pmts,self.to_pe))
+        self.to_pe *= self.config['amplification']
+
+    def compute(self, records_he, start, end):
+        result = super().compute(records_he, start,end)
```
WenzDaniel, line 83: I had just an idea how we can avoid having the 800 channel arrays.

You could modify all the compute in the following way e.g.:

```
def compute(self, records_he, start, end):
    records_he['channel'] = records_he['channel'] - self.config['channel_map']['he'][0]
    result = super().compute(records_he, start,end)
    records_he['channel'] = records_he['channel'] + self.config['channel_map']['he'][0]
```

Its very similar to your he_channel offset, but maybe a bit simpler.

straxen/plugins/high_energy_processing.py
```diff
@@ -0,0 +1,155 @@
...
+        'he_channels_offset',track=False,default=500
+        ),
+    strax.Option(
+        'amplification',track=False,default=20
+        ),
+    strax.Option(
+        'parent_version',track=True,default=straxen.Peaklets.__version__
+        ),
+    *HITFINDER_OPTIONS_he
+)
```
WenzDaniel, line 62: Like Joran already commented it, could you also add the other options for peakelts here. You can keep the parameters as they are for the normal channels. But I think in the future we will need those options. So better to add them right in the beginning.

WenzDaniel: Same for the other plugins, I just checked their options and I think at some point they will be all needed.

jorana:
@petergaemers PR https://github.com/AxFoundation/strax/pull/297 is merged and there are some open requests. Can you let us know when this PR is ready for merging as far as you are concerned?

straxen/plugins/peaklet_processing.py
```diff
@@ -186,6 +186,47 @@ def clip_peaklet_times(peaklets, start, end):
...
                 p['length'] = (end - p['time']) // p['dt']
+                
+                
+@export
+@strax.takes_config(
+    strax.Option(
+        'n_tpc_pmts_he',track=False,default=800
+        ),
+    strax.Option(
+        'channel_offset_he',track=False,default=500
```
jorana, line 197: you can use https://github.com/XENONnT/straxen/blob/master/straxen/contexts.py#L25 first argument. No need for an extra option.

straxen/plugins/peaklet_processing.py
```diff
@@ -186,6 +186,47 @@ def clip_peaklet_times(peaklets, start, end):
                 p['time'] = start
             if strax.endtime(p) > end:
                 p['length'] = (end - p['time']) // p['dt']
+                
+                
+@export
+@strax.takes_config(
+    strax.Option(
+        'n_tpc_pmts_he',track=False,default=800
```
jorana, line 194: you can use https://github.com/XENONnT/straxen/blob/master/straxen/contexts.py#L25 second argument

straxen/plugins/peaklet_processing.py
```diff
@@ -186,6 +186,47 @@ def clip_peaklet_times(peaklets, start, end):
...
+        ),
+    strax.Option(
+        'channel_offset_he',track=False,default=500
+        ),
+    strax.Option(
+        'amplification',track=False,default=20
```
jorana, line 200: want to track this right?

straxen/plugins/pulse_processing.py
```diff
@@ -18,6 +18,13 @@
              'See straxen.hit_min_amplitude for options.'
     )])
 
+HITFINDER_OPTIONS_he = tuple([
+    strax.Option(
+        'hit_min_amplitude_he',track=False,
```
jorana, line 23: My guess is that you want to track this right?

straxen/plugins/pulse_processing.py
```diff
@@ -199,6 +206,42 @@ def compute(self, raw_records, start, end):
                     pulse_counts=pulse_counts,
                     veto_regions=veto_regions)
 
+    
+@export
+@strax.takes_config(
+    strax.Option(
+        'n_tpc_pmts_he',track=False,default=800
```
jorana, line 213: see earlier

straxen/plugins/pulse_processing.py
```diff
@@ -476,3 +519,5 @@ def _check_overlaps(records, last_end):
             return r['channel'], r['time']
         last_end[r['channel']] = strax.endtime(r)
     return -9999, -9999
+
+
```
jorana: This is code factors issue

straxen/plugins/peaklet_processing.py
petergaemers: No you cannot. This is executed during the init, you do not have acces yet to straxen.whatevers. Also relative paths do not work (from ..contexts import whatever)

straxen/plugins/pulse_processing.py
petergaemers: This should already be fixed

straxen/plugins/peaklet_processing.py
jorana: Hmm I see, odd. Can you set the top to 752 than at least (to make utterly sure you are not doing any aqmon monitoring stuff) and give a description of the option

straxen/plugins/pulse_processing.py
```diff
@@ -199,6 +206,42 @@ def compute(self, raw_records, start, end):
...
+    rechunk_on_save = immutabledict(
+        records_he=False,
+        pulse_counts_he=True)
+    depends_on = 'raw_records_he'
+    parallel = 'process'
+    compressor = 'lz4'  
+    
+    def infer_dtype(self):
+        dtype=dict()
+        dtype['records_he'] = strax.record_dtype(strax.DEFAULT_RECORD_LENGTH)
```
jorana, line 232: Shouldn't this be something like
```python
record_length = strax.record_length_from_dtype(
            self.deps['raw_records_he'].dtype_for('raw_records_he'))
dtype['records_he'] = strax.record_dtype(record_length )
```

straxen/plugins/pulse_processing.py
```diff
@@ -17,6 +17,13 @@
              'See straxen.hit_min_amplitude for options.'
     )])
 
+HITFINDER_OPTIONS_he = tuple([
+    strax.Option(
+        'hit_min_amplitude_he',track=True,
+        default=tuple(np.ones(800, dtype=np.int16) * 15),
```
jorana, line 23: I think you'll have to add this to the https://github.com/XENONnT/straxen/blob/petergaemers-patch-2/straxen/hitfinder_thresholds.py as:

1. Strax cannot save the metadata like this as the saver is complaining that we cannot json this
2. We also don't want 800 times 15 in the metadata.

straxen/plugins/pulse_processing.py
petergaemers: No, thats another way of doing the same

straxen/plugins/peaklet_processing.py
petergaemers: Aqm channels are already split off

straxen/plugins/pulse_processing.py
jorana: okay but what if the record-length isn't the default?

straxen/plugins/pulse_processing.py
petergaemers: fixed

straxen/plugins/pulse_processing.py
jorana: ok I propose something in https://github.com/XENONnT/straxen/pull/176 to override this if we are doing some special run

jorana:
Thanks Peter! I think this is almost ready. I added a few things in #176 that exploit Daniels additions to strax. If you approve that PR will be added to this one and we can do the release

straxen/plugins/pulse_processing.py
```diff
@@ -197,11 +206,47 @@ def compute(self, raw_records, start, end):
                     pulse_counts=pulse_counts,
                     veto_regions=veto_regions)
 
+    
+@export
+@strax.takes_config(
+    strax.Option('n_he_pmts', track=False, default=752,
+                 help="Maximum channel of the he channels"),
+    strax.Option('record_length', default=110, track=False, type=int,
```
WenzDaniel, line 214: I think this can be removed. The linked mode will require the same length for all detectors. Why would you expect the he channels to be different?

jorana: First we had the Default.Recordlen

then I tried the ```
record_length = strax.record_length_from_dtype(
            self.deps['raw_records_he'].dtype_for('raw_records_he'))
dtype['records_he'] = strax.record_dtype(record_length )```

PR #163, opened 2020-07-29
Replace xenon_secrets with utilix config file

ershockley:
_**This should not be merged yet**_

This PR is a proposal to replace the xenon_secrets python file with the configuration file handled by utilix. This is helpful for running strax on e.g. OSG or anywhere else where a xenon_secrets.py file might not be present. I also prefer this type of text file configuration, which one could modify and/or have multiple copies of in case you want to change anything (e.g. I have a 1T config and an nT one, and just need to change XENON_CONFIG env variable to switch between the two). 

In addition, it uses the utilix wrapper around pymongo, which increases readability and centralizes the runDB stuff. 

I'm still testing some things, but wanted to get people's thoughts on something like this. If we don't want to go this route I need a different solution for OSG. 

As an example, my utilix config file now looks like this (removing the actual content to be overly cautious)

```
[RunDB]
rundb_api_url =
rundb_api_user =
rundb_api_password = 
pymongo_url = 
pymongo_user = 
pymongo_password = 
pymongo_database = 

[straxen]
s3_access_key_id = 
s3_secret_access_key = 
```
In this way we can combine configuration of many things into a single file. For more info on utilix config, see [here](https://github.com/XENONnT/utilix).

_**This should not be merged yet**_


jorana:
Hi Evan, thanks for the proposal. A couple of ideas:
- Why can't we keep the xenon-secrets in case we don't have access to the environment variables? I for instance utilize this on the ebs as when I initialize bootstrax over SSH I don't have access to the environment variables. I think adding it can be useful but I don't see a reason not to keep the xenon-secrets.
- Great if utilix improves reliability of the rundb access. Is it also easier to initialize the runsdb via utilix if not on dali (e.g. your own laptop)? Briefly looking at it it doesn't seem fit for running bootstrax right as I want admin rights to push data to the rundb.
- How would the utilix wrapper change the way the user interacts with the runDB storage? Can it access more/have more rights/need to do other commands?

ershockley:
Hi Joran, 

- I'm not totally opposed to keeping xenon_secrets also -- I just think would be nice to have a single xenon config file with categories for different pieces of software. We could use xenon_secrets for that also but I personally just prefer a config file type thing rather than importing from a python file, super minor point though. '

- I don't understand your point on the environment variables. By default utilix looks for a config file in your home directory, so if XENON_CONFIG isn't set it will try to load ~/.xenon_config. Does this resolve your issue? 

- You can definitely push data to the runDB with utilix. All it is doing is putting the boilerplate collection setup into a single place. You set the user/passsword/database/url in the runDB so it should be pretty flexible, I think. 

Does this answer your questions?

jorana:
> but I personally just prefer a config file type thing rather than importing from a python file

On the other hand keeping it as a backup option wouldn't be a bad idea.

> I don't understand your point on the environment variables. By default utilix looks for a config file in your home directory, so if XENON_CONFIG isn't set it will try to load ~/.xenon_config. Does this resolve your issue?

Yeah sorry it's some edge case I ran into I fixed with the xenon-sectrets but adding a ~/.xenon_config seems similar to the xenon_secrets so that sounds like it work, guess would have to try.

> You can definitely push data to the runDB with utilix. All it is doing is putting the boilerplate collection setup into a single place. You set the user/passsword/database/url in the runDB so it should be pretty flexible, I think.

OK, I also just looked very briefly and saw that the secondary was preferred which is readonly usually right?
https://github.com/XENONnT/utilix/blob/master/utilix/rundb.py#L345

I guess I'd be nice to add the required kwargs into this part of the straxen RunDB.py but this would presumably be one of those things we could fix if people agree on your proposal.

ershockley:
I will test some of these things more thoroughly and get back to Joran (also with more examples etc). I will be mostly offline for the next ~ week or more so it might be a bit. 

Thanks for the comments!

ershockley:
I would like to revive this PR, as this change is important for any future reprocessing campaign. There are two main goals here:

1. Replace the xenon_secrets.py file with what I refer to as a xenon_config file (see [here](https://github.com/XENONnT/utilix). As it stands now this change is required in order to process data on OSG (though there are ways around this if we want to keep xenon_secrets). This xenon_config file can be used by other software outside of straxen.

2. Streamline our analysis tools so that they talk to each other more. If all of our tools access the runsDB using utilix, then any future change in runDB settings (urls, passwords, etc), as happened a few times in XENON1T, can be easily applied by just updating utilix, and we don't have to make hardcoding changes in many different packages. I also think the general xenon_config file will be useful in the future. 

@jorana thoughts? Not sure I addressed your concerns from before or not.

ershockley:
Maybe I should add that, to make this work, one needs a xenon_config file of the following syntax:

```
[RunDB]
pymongo_url = 
pymongo_user =
pymongo_password = 
pymongo_database = 

[straxen]
s3_access_key_id = 
s3_secret_access_key = 
```

This is what is replacing the xenon_secrets file. When you enter a singularity container on midway, by default the config loaded is in a file called xenon.config, in same place as current xenon_secrets file. But if you are running somewhere else (DAQ machines, e.g.) you will need to create your own config file.

ershockley:
Now added GitHub actions to automatically update the context collection of the RunDB whenever there is a new straxen release. I've tested a slightly different version of this with [my personal fork of utilix](https://github.com/ershockley/utilix) but have not tested this exact one.

jorana:
@ershockley , let me know when this is ready for review by 'un-drafting' the PR.

tunnell:
Wouldn't this be easy to do with the corrections system?  This could be for strax settings beyond physics.  @ahiguera-mx thoughts?

ershockley:
@tunnell I don't understand what you mean, sorry. This change would replace e.g. how straxen sets up the runDB, which is needed to access correction system in the first place right?

ershockley:
@jorana I finally made travis happy, though I think this could use some more testing. 

For an example config, see `/project2/lgrandi/xenonnt/xenon.config` on midway. To address #268, you would need to create a config file on the DAQ machine where you change the pymongo_url accordingly. 

I still need to test the Github Actions some more, but that should not affect normal straxen analyses.

straxen/common.py
```diff
@@ -188,31 +191,12 @@ def get_secret(x):
...
-            else:
-                raise ValueError(
-                    message + ' nor could we load the secrets module from '
-                              f'{path_to_secrets}, even though you seem '
-                              'to be on the midway analysis hub.')
-
-        else:
-            raise ValueError(
-                message + 'nor are you on the midway analysis hub.')
+        config = utilix.config.Config()
```
jorana, line 194: I don't like that this would throw out the mechanism that was in place before without any deprecation warning or so. Let's push for using utilix but phase out the xenon_secrets, i.e. throw a deprecation warning such that users should be well aware that they should start using utilix and the corresponding file.

We will then phase out the xenon-secrets.

straxen/rundb.py
```diff
@@ -74,16 +65,7 @@ def __init__(self,
...
-            mongo_url = get_mongo_url(self.hostname)
-
-        self.client = pymongo.MongoClient(mongo_url)
-
-        if mongo_dbname is None:
-            mongo_dbname = default_mongo_dbname
-        if mongo_collname is None:
-            mongo_collname = default_mongo_collname
-        self.collection = self.client[mongo_dbname][mongo_collname]
+        self.collection = utilix.rundb.pymongo_collection(**kwargs)
```
jorana, line 68: The kwargs are  already used for the super().__init__

straxen/rundb.py
```diff
@@ -74,16 +65,7 @@ def __init__(self,
...
-        if mongo_url is None:
-            mongo_url = get_mongo_url(self.hostname)
-
-        self.client = pymongo.MongoClient(mongo_url)
-
-        if mongo_dbname is None:
-            mongo_dbname = default_mongo_dbname
-        if mongo_collname is None:
-            mongo_collname = default_mongo_collname
-        self.collection = self.client[mongo_dbname][mongo_collname]
```
jorana, line 86: @ershockley why delete this? This is not a good idea as one cannot intuitively change which user/collection one is using. E.g. on the eventbuilder we have a production and testing mode where we use different access to the database for obvious reasons.

jorana: Also testing with another database becomes quite inflexible

straxen/scada.py
```diff
@@ -22,12 +24,15 @@ def __init__(self, context=None):
...
+            self.SCLastValue_URL = uconfig.get('scada', 'sclastvalue_url')
+            self.SCADA_SECRETS = dict(QueryType=uconfig.get('scada', 'querytype'),
+                                      username=uconfig.get('scada', 'username'),
+                                      api_key=uconfig.get('scada', 'api_key')
+                                      )
         except ValueError:
-            raise ValueError('Cannot load SCADA information, from xenon'
-                             ' secrets. SCADAInterface cannot be used.')
+            raise ValueError(f'Cannot load SCADA information, from xenon'
+                             ' config at {uconfig.config_path}. SCADAInterface cannot be used.')
```
jorana, line 35: If uconfig is None, this will become quite an unintuitive traceback, this is very well possible. Please keep this case in mind.

straxen/contexts.py
```diff
@@ -2,6 +2,14 @@
 import strax
 import straxen
 
+# list of contexts that gets tracked in runDB context collection
+# needs to be maintained for each straxen release
+context_list = ['xenonnt_led',
```
jorana, line 7: Would it not make sense to put this somewhere else? If a user would do e.g.  straxen.contexts.context_list, I would be expecting it would list all the available contexts or so, which this isn't.

straxen/contexts.py
```diff
@@ -2,6 +2,14 @@
 import strax
 import straxen
 
+# list of contexts that gets tracked in runDB context collection
+# needs to be maintained for each straxen release
+context_list = ['xenonnt_led', 
+                'xenonnt_online',
+                'xenonnt_initial_commissioning',
+                'xenonnt_simulation'
```
jorana, line 10: Not sure if you want to track this at the moment as it's more of a placeholder which is adapted at the moment by people wanting to simulate stuff. 

This of course may change if we are going to make a centralized simulation context with also shared files, but at the moment this is not used as such.

straxen/contexts.py
jorana: moving it to the github/scripts would make much more sense I think

jorana (changes requested):
Hi Evan, thanks a lot for adding this. Already in #268 we would have benefited from this way of doing things. Also hardcoding values is a tricky thing as we have seen in the past with mirrors being changed which made backward compatibility hard.

I'm very happy that you are solving this before it's an issue.

Concerning the PR, this first review is just on how/what you implemented. Here I have some concerns with respect to the "this is how we do it now, crash your notebook if you haven't set it up like utilix expects"- change and I much rather like to phase out the xenon-secrets mechanism by warning users a bit beforehand. We can fully shift to utilix soon but I would strongly advocate a transition period. I also have some concerns about the flexibility, please find them in the review comments.

I haven't been able to test the PR itself yet.

straxen/__init__.py
```diff
@@ -1,5 +1,12 @@
 __version__ = '0.11.1'
 
+# load configuration file using utilix
+try:
+    from utilix.config import Config
+    uconfig = Config()
+except:
```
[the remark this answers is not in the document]
jorana: internal utilix errors should not go unnoticed

straxen/scada.py
```diff
@@ -22,12 +24,15 @@ def __init__(self, context=None):
             if you would like to query data via run_ids.
         """
         try:
-            self.SCData_URL = straxen.get_secret('SCData_URL')
-            self.SCLastValue_URL = straxen.get_secret('SCLastValue_URL')
-            self.SCADA_SECRETS = straxen.get_secret('SCADA_SECRETS')
+            self.SCData_URL = uconfig.get('scada', 'scdata_url')
```
jorana, line 27: These values should be None right otherwise it's unclear to the user why the url does not work I image

straxen/corrections_services.py
```diff
@@ -35,12 +36,12 @@ def __init__(self, username='nt_analysis', password=None, is_nt=True):
...
             self.password = password
         elif self.username.endswith('analysis'):
-            self.password = straxen.get_secret('rundb_password')
+            self.password = uconfig.get('RunDB', 'pymongo_password')
         else:
             raise ValueError(f'No password for {username}')
 
         # Get the readonly account for the rundb using hostname = ''
-        runsdb_mongo_url = straxen.rundb.get_mongo_url(hostname='')
+        runsdb_mongo_url = uconfig.get('RunDB', 'pymongo_url')
```
jorana, line 44: this should also be more flexible, bootstrax may be running with admin privileges but I would not want to be running with those here.

jorana:
@ershockley I also had a look at the /project2/lgrandi/xenonnt/xenon.config file but from experience we know that the way `pymongo_url` is specified does not work see the related PR: 
https://github.com/XENONnT/straxen/pull/213

straxen/common.py
ershockley: Oh are users using their own xenon secrets file? I thought it was just the single global one? If it's just replacing the global secrets file with a different global file then I don't see the need for a warning, but sure will do.

jorana: It has to be in your straxen/xenon_sectres.py file. People can have different ones, especially when not running on dali.

straxen/rundb.py
ershockley: Sorry maybe I misunderstood, but I thought the goal was to remove this type of hardcoding? Changing an ini file is much more flexible than changing the source code, no?

.github/scripts/update-context-collection.py
```diff
@@ -0,0 +1,34 @@
...
+
+def main():
+    for context in context_list:
+        # get these from straxen.contexts.*
+        st = eval("%s()" % context)
+        hashes = get_hashes(st)
+        hash_dict = {dtype: h for dtype, h in hashes}
+
+        doc = dict(name=context,
+                   date_added=datetime.datetime.utcnow(),
```
WenzDaniel, line 23: Wont be `.utcnow()` different depending on your machines location? Will this be an issue?

straxen/contexts.py
ershockley: Sure we can move it to .github/scripts, just need to have ACs (or whoever) remember that's where the list is maintained. I thought it might be useful for the list to be importable, but that's fine.

straxen/common.py
```diff
@@ -188,31 +191,12 @@ def get_secret(x):
...
+        return config.get('straxen', x)
 
-    if hasattr(xenon_secrets, x):
-        return getattr(xenon_secrets, x)
-    raise ValueError(message + " and the secret is not in xenon_secrets.py")
+    except:
+        message += (f"nor was there a valid field in your utilix configuration file, ")
+        raise
```
WenzDaniel, line 200: Are not you missing here something after the raise?

straxen/contexts.py
ershockley: You're talking about the simulation one specifically right? If so I will remove. I don't want to be in charge of maintaining this list, just was proposing the existence of one

straxen/scada.py
ershockley: I'm not sure I follow. The configparser `get` is a bit different from the standard dictionary `get` -- is that what you're asking? This line will look for a section called 'scada' in the config, and variable named 'scdata_url'. If one doesn't exist it will throw an error.

jorana: oops, I indeed was thinking too much about a dict operation. Thanks for the clarification!

straxen/contexts.py
jorana: Indeed, I was only mentioning `xenonnt_simulation`

straxen/corrections_services.py
ershockley: Yeah so the bootstrax config would look different -- or is this not the functionality you want? In my view this should be handled by a config file, not hardcoded. Maybe I'm missing something though.

straxen/rundb.py
jorana: I thought is was only the hosts/passwords/usernames we were trying to tackle right? Now looking at the source code of utilix, I see where my misunderstanding comes from. All these things may be specified in the kwargs to utilix. Perhaps it's good to add a comment or so what we can/need to specify in the kwargs.

ershockley:
> @ershockley I also had a look at the /project2/lgrandi/xenonnt/xenon.config file but from experience we know that the way `pymongo_url` is specified does not work see the related PR:
> #213

@jorana I just tested on midway and using all 3 works fine

```
>>> from utilix.rundb import pymongo_collection

>>> coll = pymongo_collection()

>>> coll.find_one({'number': 10000}, {'number': 1})
{'_id': ObjectId('5fa30d218471e9d11dfd5276'), 'number': 10000}

>>> from utilix.config import Config
>>> cfg = Config()
>>> cfg.get('RunDB', 'pymongo_url')
'xenon-rundb.grid.uchicago.edu:27017,xenon1t-daq.lngs.infn.it:27017,fried.rice.edu:27017/xenonnt'
```

I do remember you had issues, but I think that was on your personal machine? Not sure we ever figured out what was causing that, but in general having all three is not a problem, I don't think.

.github/scripts/update-context-collection.py
ershockley: utcnow does not depend on the machine location -- it always returns the current time in UTC. I just tested comparing a machine in San Diego and a machine in Chicago

straxen/common.py
ershockley: Yes thanks!

straxen/scada.py
ershockley: Good point, thanks @jorana. Will fix.

straxen/common.py
ershockley: Ok I see, then yeah a phase out period makes sense.

straxen/__init__.py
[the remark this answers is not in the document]
ershockley: This is the change I made to make travis happy :) But yes I agree, I'll fix.

straxen/corrections_services.py
jorana: So perhaps we can find a solution. So it seems to me that in principle, not all these usages of `pymongo_url` have to be the same. Like in bootstrax I want to have write permissions in https://github.com/XENONnT/straxen/blob/ca8dc1af05dbab89c8cbb2a4f14ad5bd5ba74bda/straxen/online_monitor.py#L29 while having readonly access here. 

Like this, just specifying the `pymongo_database` in the config is not able to distinguish right? You will load all with the same authentication/privilages. One could argue that one with write access should be comfortable doing everything with write access but I'm just trying to see if we can minimize things that we know one should not touch (if that makes sense).


There are some things possible, let me know what you think is best / you have a better idea:
 - Create `pymongo_database_RunDB`, `pymongo_database_CMT`, `pymongo_database_...` in the ini file, write those at the appropriate places. The ini file would be specifying the same things under a different name many times but at least it would be super transparent.
 - Shift all this to where one loads these different classes like for instance here: https://github.com/XENONnT/straxen/blob/master/straxen/contexts.py#L103 to ask for a different password/username and database. I'm not sure if it's transparent though.
- Create an `pymongo_admin_username`, `pymongo_admin_password`, `pymongo_admin_url`, `pymongo_admin_database` in the ini file. Then, using `uconfig.get` one should explicitly write down when trying to start any of these classes with the write access. This means changing this line to `if runsdb_mongo_url is None: ....` and allowing one to specify it in the init of this class. 

I think the last options seems to make most sense. Sorry for the long story here.

jorana:
> I do remember you had issues, but I think that was on your personal machine? Not sure we ever figured out what was causing that, but in general having all three is not a problem, I don't think.

Jup, it was when you were on any location that was not dali or LNGS (I think). We could not determine which was primary because none was. At the moment it's fine because all the databases are up (also on my machine).

straxen/common.py
ershockley: This should be addressed now

straxen/rundb.py
ershockley: This should be addressed now.

straxen/rundb.py
ershockley: Ok i added more comment strings

straxen/scada.py
ershockley: Fixed now

straxen/contexts.py
ershockley: Done

straxen/corrections_services.py
ershockley: Just to recap @jorana and my conversation offline, we will have bootstrax use a different configuration parameter to prevent unwanted db calls on the EBs with admin privileges.

ershockley:
@jorana Sorry we're working on two repos at once which is making things unstable. 

To satisfy travis we need to be able to import straxen and do basic tests without a valid config file. If we merge utilix [#8](https://github.com/XENONnT/utilix/pull/8) then the most recent changes should work, I think. This will have utilix throw an error, but straxen does try-except to circumvent it.

jorana:
If one ever runs into problems using three mirrors. Please replace:
`pymongo_url = 'xenon-rundb.grid.uchicago.edu:27017,xenon1t-daq.lngs.infn.it:27017,fried.rice.edu:27017/xenonnt'`
with
`pymongo_url = 'xenon-rundb.grid.uchicago.edu:27017/xenonnt'`
or
`pymongo_url = 'fried.rice.edu:27017/xenonnt'`

In the config file

PR #164, opened 2020-07-30
Changes to the RunDB frontend for rucio

SidAhmedMa:
Hello, this is the long-awaited version that allows access to Rucio's data on Dali directly (I hope it works perfectly 🤞).  Knowing that obviously I have already done a lot of tests.

### DIFF
--- a/straxen/rundb.py
+++ b/straxen/rundb.py
@@ -40,7 +40,7 @@ class RunDB(strax.StorageFrontend):
                  local_only=False,
                  new_data_path=None,
                  reader_ini_name_is_mode=False,
-                 rucio_path = '/dali/lgrandi/rucio/',
+                 rucio_path=None,
                  *args, **kwargs):
         """
         :param mongo_url: URL to Mongo runs database (including auth)
@@ -109,8 +109,9 @@ class RunDB(strax.StorageFrontend):
         self.backends = [
             strax.S3Backend(**s3_kwargs),
             strax.FileSytemBackend(),
-            strax.rucio(rucio_path),
         ]
+        if self.rucio_path is not None:
+            self.backends.append(strax.rucio(self.rucio_path))
 
         # Construct mongo query for runs with available data.
         # This depends on the machine you're running on.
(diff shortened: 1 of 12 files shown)


straxen/rundb.py
```diff
@@ -105,8 +105,9 @@ def __init__(self,
         self.collection = self.client[mongo_dbname][mongo_collname]
 
         self.backends = [
-            strax.S3Backend(**s3_kwargs),
-            strax.FileSytemBackend(),
+            #strax.S3Backend(**s3_kwargs),
```
WenzDaniel, line 108: Why not removing those lines?

WenzDaniel:
Hej Sid, thanks a lot. I just briefly skimmed over the code. Could you a share with us some examples?

straxen/rundb.py
SidAhmedMa: These lines correspond to backends that still exist, so you have the choice between removing the comments, which will make you have 3 backends in all, or leaving it like that, which will make it go faster.

SidAhmedMa:
> Hej Sid, thanks a lot. I just briefly skimmed over the code. Could you a share with us some examples?

Very good suggestion, you should already know that this is my first pr :). However I also have to submit a draft on strax (as you suggested to me), so the changes will be complete.

straxen/rundb.py
WenzDaniel: Okay I do not get it why do we have to comment out those existing backends? Either they exist and we use them then we should not comment them out, or they do not exist and we do not use them, in this case we can remove those lines.

SidAhmedMa: You are right, I will remove the comments

SidAhmedMa:
Hello everyone, given the particular structure of rucio, which for each file matches a sub-sub-folder, it was mandatory to create a new backend to manage this new structure,

In the new structure, we use strax.rucio as the backend and the database frontend, you will find a notebook below which shows the tests I performed.

Notebook: /dali/lgrandi/sahmedma/rucio-backend-test.ipynb

WenzDaniel:
Hi Sid, thanks a lot. Could you also upload this notebook to our analysis repo? This makes it easier for people to look into it. You could use the following directory: https://github.com/XENONnT/analysiscode/tree/master/StraxTests

SidAhmedMa:
> Hi Sid, thanks a lot. Could you also upload this notebook to our analysis repo? This makes it easier for people to look into it. You could use the following directory: https://github.com/XENONnT/analysiscode/tree/master/StraxTests

It's done. Thank you by teaching me how it works :)

jorana:
This feature is tested and ready for merge

straxen/rundb.py
```diff
@@ -40,6 +40,7 @@ def __init__(self,
                  local_only=False,
                  new_data_path=None,
                  reader_ini_name_is_mode=False,
+                 rucio_path = '/dali/lgrandi/rucio/',
```
WenzDaniel, line 43: Not sure if we have anywhere in straxen dali related paths except for the contexts. Maybe you could change this into None and add this information simple to the context?

jorana: There are some references to specific places/hosts even in this file:
https://github.com/XENONnT/straxen/blob/b176d08eec923876994f20fad328e714d070c22c/straxen/rundb.py#L28
https://github.com/XENONnT/straxen/blob/b176d08eec923876994f20fad328e714d070c22c/straxen/rundb.py#L91

But indeed perhaps it's nicer to put it in the context 1ac9c56

Issue #154, opened 2020-07-21
(boot)strax handling of last steps


jorana:
Luca pointed out the a problem with run 8675 where the number of files was uploaded to the rundb by eb5. However eb4 claimed to be the one that correctly processed the run but didn't include the filecount.

**Reconstruction of events**
It's really a conglomerate of bad screw-ups and quite unlikely events. I'm going to make corresponding issues on bootstrax/strax. Let me summarize what happened:
   - Eb5 finished but on the very last chunk it failed saving because of rsync was hammering it.
   - The exception was caught just while strax was renaming <folder>_temp to <folder> . This means the folders have the appropriate names.
   - In fact the processing proceeded so far the the call was made in bootstrax to count the files and update the database.
   - Eb3 tried and failed like eb5 because of rsync.
   - Eb4 then started and completed processing. However before it made the call to count the number of files bootstrax was killed as we ran out of diskspace.

All of these things seem very unlikely but somehow it all happened to this one run.

Bottomline: it was bootstraxs' fault and didn't update the filecount on eb4.

**Issues to fix**
 - Strax: don't start renaming things from <folder>_temp to <folder> while savers are not finished.
 - bootstrax: 
    - Check that the filecount is in the data after stop (crash first recover later)
    - Don't crash this hard if we don't have space to process data. Also wait longer.
    - Can we check that all the savers succeded? We need to do so somehow.

jorana:
fixed in https://github.com/XENONnT/straxen/pull/206

PR #165, opened 2020-07-31
Bootstrax/Ajax/Microstrax updates

jorana:
**Preamble**
This is a aggregation of the scripts running on the eventbuilders to process data and maintain the hosts/rundatabse. Below the changes to each script are highlighted.

Also for all of the scripts, we have changed the disks on eb0 and eb1 to also read `/data`. Also, the scripts are changed such that eb2 is not contributing at all to live processing.

**Ajax**
- Delete old lineage-data from the host. Remove any high-level data if the lineage hash for that datatype is not in correspondence with the latest lineage hash.
- Clean the runsdatabase. We had some entries that were listed in the database but they appeared non-existent. Therefore we add a task to ajax to check that the data listed in the runs-database actually exist.

**Bootstrax**

- Prepare bootstrax for the linked mode. At the moment we have some muon-veto data that we should be able to open with bootstrax. **Please be aware.** We are now just trying to open the data with the defaults from redax but it's much better if these are properly set. 
- Reinstate a check for 'dead' bootstrax instances, mark as dead if bootstrax hasn't checked in for a while.
- Update the data-document in the rundoc with metadata concerning the strax(en) version and the data-size. See for example:
afbeelding.

**Microstrax**
- Have eb2 now write a document every minute that it is hosting microstrax.

### DIFF
--- a/bin/ajax
+++ b/bin/ajax
@@ -81,7 +81,7 @@ def main_ajax():
         remove_run_from_host(args.number, delete_live=args.delete_live, force=args.force)
 
     if args.clean:
-        _modes = {'ceph': clean_cehp,
+        _modes = {'ceph': clean_ceph,
                   'high_level_data': clean_high_level_data,
                   'non_latest': clean_non_latest,
                   'unregistered': clean_unregistered,
@@ -101,7 +101,7 @@ def main_ajax():
         elif args.clean == 'all':
             while True:
                 if hostname == eb_can_clean_ceph:
-                    clean_cehp()
+                    clean_ceph()
                     clean_abandoned(delete_live=True)
                     clean_unregistered(delete_live=True)
                     clean_database(delete_live=True)
@@ -120,13 +120,13 @@ def main_ajax():
             raise ValueError(f'Unknown cleaning mode {args.clean}')
 
 
-def clean_cehp():
+def clean_ceph():
     """
     Look for old data on ceph. If found, delete it. Recursively iterate until no old data
     is found. This function only works on the the host that is eb_can_clean_ceph.
     """
     if not hostname == eb_can_clean_ceph:
-        log.info(f'clean_cehp::\tfor cleaning ceph, go to {eb_can_clean_ceph}')
+        log.info(f'clean_ceph::\tfor cleaning ceph, go to {eb_can_clean_ceph}')
         return
 
     rd = run_coll.find_one({'bootstrax.state': 'done',
@@ -140,16 +140,16 @@ def clean_cehp():
         return
     else:
         run_number = rd['number']
-        log.info(f'clean_cehp::\tremove data associated to run {run_number}')
+        log.info(f'clean_ceph::\tremove data associated to run {run_number}')
         remove_run_from_host(run_number, delete_live=True, force=args.force)
-        log.info(f'clean_cehp::\tfinished for {run_number}, take a'
+        log.info(f'clean_ceph::\tfinished for {run_number}, take a'
                  f' {ajax_thresholds["short_nap"]} s nap')
         time.sleep(ajax_thresholds['short_nap'])
 
         # Repeat.
         if args.execute:
             wait_on_delete_thread()
-            clean_cehp()
+            clean_ceph()
 
     # Finally, check that there is no old data on ceph that is also not in the rundoc.
     clean_unregistered(delete_live=True)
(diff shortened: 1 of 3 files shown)


jorana:
TODO:
  - add to to rundoc the strax(en) versions
  - add filesize to data-doc

bin/ajax
```diff
@@ -81,41 +80,44 @@ def main_ajax():
...
                 if not args.execute:
                     break
                 log.info(f'Loop finished, take a {ajax_thresholds["nap_time"]} s nap')
                 time.sleep(ajax_thresholds['nap_time'])
         else:
-            raise ValueError(f'Unknown mode {args.mode}')
+            raise ValueError(f'Unknown cleaning mode {args.clean}')
 
 
 def clean_cehp():
```
darrylmasson, line 123: `:%s/clean_cehp/clean_ceph/gc` ?

bin/ajax
```diff
@@ -284,7 +287,43 @@ def clean_unregistered(delete_live=False):
...
+        remove_if_unregistered(run_number, delete_live=delete_live)
+
+
+def clean_old_hash():
+    """
+    Query the database for data that was processed with other settings of straxen that
+    lead to a different linage hash. If we can find something on this host that is not in
+    correspondence with the latest hash: delete it.
+    """
+    rds = run_coll.find({'data.type.$': {'$nin': low_level_data}, 'data.host': hostname})
```
darrylmasson, line 300: If we expect this to run more than once, we'll want to index this field, and I think that this field would be a messy and inefficient index. If we made `data` its own collection, this would be rather simpler.

I'd almost suggest doing this via the filesystem rather than the database.

bin/bootstrax
```diff
@@ -176,7 +168,7 @@ timeouts = {
 # hence, we can check if there is sufficient diskspace and if not, wait a while.
 # Below are the max number of times and number of seconds bootstrax will wait.
 wait_diskspace_max_space_percent = 90
-wait_diskspace_n_max = 100  # times
+wait_diskspace_n_max = 60 * 24 * 7 # times
```
darrylmasson, line 171: An entire week?

bin/bootstrax
```diff
@@ -1095,14 +1122,27 @@ def process_run(rd, send_heartbeats=args.production):
...
 
-        thread_info = rd.get('daq_config', dict()).get('processing_threads', dict())
-        n_readout_threads = sum([v for v in thread_info.values()])
         try:
-            daq_chunk_duration = int(rd['daq_config']['strax_chunk_length'] * 1e9)
-            daq_overlap_chunk_duration = int(rd['daq_config']['strax_chunk_overlap'] * 1e9)
+            # Fetch parameters from the rundoc. If not readably, let's use redax' default
+            # values (that are hardcoded here).
+            if 'daq_conig' not in rd:
+                log_warning('No daq_config in the rundoc!')
```
darrylmasson, line 1129: I'd argue that this should be a fatal error rather than a warning.

bin/bootstrax
```diff
@@ -1095,14 +1122,27 @@ def process_run(rd, send_heartbeats=args.production):
...
+            dq_conf = rd.get('daq_config', dict())
+            to_read = ('processing_threads', 'strax_chunk_length','strax_chunk_overlap',
+                       'strax_fragment_payload_bytes')
+            for conf in to_read:
+                if conf not in dq_conf:
+                    log_warning(f'{conf} not in rundoc for {run_id}! Using default values.')
```
darrylmasson, line 1135: The default value for `processing_threads` is complicated.  The default is 8 threads per instance of redax, but finding _that_ value is rather more involved.

bin/bootstrax
```diff
@@ -1302,21 +1341,17 @@ def cleanup_db():
...
-    #          'status': {'$ne': dead_state},
-    #          'time': {'$lt': now(-timeouts['bootstrax_presumed_dead'])}
-    #          },
-    #         {'$set': {'state': dead_state}})
-    #     if bd is None:
-    #         break
-    #     log.info(f"Cleaned old boostrax entry on host {bd['host']}. Now presumed dead. Rest in peace")
+    # Check for all the ebs if their last state message is not longer ago than the time we assume that the eb is dead.
+    for eb_i in range(6):
+        bd = bs_coll.find_one(
```
darrylmasson, line 1346: We should index this query, and also remove the old collections.

bin/microstrax
```diff
@@ -119,6 +124,49 @@ def get_data(
...
+    Inform the bootstrax collection we're running microstrax
+    """
+    # DAQ database
+    daq_db_name = 'daq'
+    daq_db_password = straxen.get_secret('MONGO_DAQ_PASSWORD'.lower())
+    daq_db_username = straxen.get_secret('MONGO_DAQ_USERNAME'.lower())
+    daq_client = pymongo.MongoClient(f"mongodb://{daq_db_username}:{daq_db_password}@xenon1t-daq:27017/admin")
+    daq_db = daq_client[daq_db_name]
+    bs_coll = daq_db['eb_monitor']
+    while True:
```
darrylmasson, line 142: For threads like this that constantly run in the background, I like to add a signal handler so it exits cleanly via ctrl-c.

bin/bootstrax
jorana: Well, it doesn't do anything until it simply dies. It just waits until you free up some space. We could even set it to infinite.

bin/ajax
jorana: Sorry I'm not sure if I follow this comment

bin/bootstrax
jorana: will change it

bin/bootstrax
jorana: at the moment we would get None or 0 using this logic if it's not specified

bin/ajax
jorana: This would only be done once each hour although we can up this time as most of the time you wouldn't really expect a lot of old data lying around. 

I'm not sure if doing it via the filesystem would be of much help given that we in the end want to sync the rdb and the filesystem.

bin/ajax
darrylmasson: It's the vim command to replace `clean_cehp` with `clean_ceph`.

bin/ajax
darrylmasson: That query is going to run increasingly slowly as more and more runs accumulate, while the filesystem will generally have a "constant" number of items (certainly less than the number of runs). It will probably be faster to query the disk, perform whatever changes are necessary, and then update the database, rather than the other way around. If we reduce the frequency of this (1/day?) it'll be less of an issue.

bin/bootstrax
darrylmasson: I'm just wondering if there's a cleaner way to handle this field, because using the "default" value is not as straightforward as for the other fields.

bin/microstrax
jorana: Thanks, actually this is not really needed in python:
https://stackoverflow.com/questions/56144570/how-to-stop-a-python-daemon-thread-with-infinite-loop-at-shutdown

bin/bootstrax
jorana: If I'm not mistaken it is indexed in order to be TTL right?

bin/microstrax
darrylmasson: Fancy.

I've had less-than-perfect success with `pymongo.errors.NotMasterError` - sometimes it doesn't throw that error, so I just wrapped a generic `try...except` rather than one specifically targeting that one.

bin/bootstrax
darrylmasson: `time` is indexed (TTL), but `host` isn't (or wasn't, until I wrote this comment). Also, index intersection doesn't always work as desired or expected, but if we use separate indexes for `host` and `time` it should be fine.

bin/ajax
jorana: thx done!

bin/ajax
jorana: Hmm okay, actually doing it over the fs is easier. Thanks for the suggestion!

jorana:
Thanks for the feedback Darryl!

bin/microstrax
jorana: Good point will track other errors and add those as we go. I think I did the same for the bootstrax --undying mode

jorana:
Several items in #154 are addressed in this PR

PR #166, opened 2020-08-04
Remap old runs.

jorana:
**Required strax-implementation**
This PR only works if https://github.com/AxFoundation/strax/pull/304 is merged.

As discussed in https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:dsg:daq:sector_swap some channels were swapped. In this implementation we fix this by remapping all channels as soon as the data is loaded. This is done only for 'old' data.

[like in https://github.com/AxFoundation/strax/pull/304] when one loads data you do either st.get_array or st.get_df (which then calls st.get_array for you). In the lines changed in this PR, you see that the results that is usually returned directly can now be changed by a function. 

In this PR we use this feature to:
1. Check if the data is older than the timestamp where we changed the cabling
2. If so, remap all data that is associated to channel-numbers. 

A (very elaborate) example can be found here https://github.com/XENONnT/analysiscode/blob/master/DAQ/remap_sectors/Software_remap.ipynb (please note option 3 is what we are doing here).

**Cons**
 - The extra step naturally takes time and memory. At the moment I consider this an acceptable loss considering the low likelihood that a lot of analyses will be done on this data. The other option would be to reprocess all the data we have taken so far but this sounds like a lot larger effort.
 - I haven't included the he-channels. This can easily be done if we also update the reference to the auxiliary files accordingly after we merge https://github.com/XENONnT/straxen/pull/161.

### DIFF
--- a/straxen/common.py
+++ b/straxen/common.py
@@ -244,16 +244,23 @@ def get_livetime_sec(context, run_id, things=None):
             return (md['end'] - md['start']).total_seconds()
 
 @export
-def remap_channels(data, verbose=True, _tqdm=False):
+def remap_channels(data, verbose=True, safe_copy=False, _tqdm=False, ):
     """
     There were some errors in the channel mapping of old data as described in
         https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:dsg:daq:sector_swap
         using this function, we can convert old data to reflect the right channel map
         while loading the data. We convert both the field 'channel' as well as anything
         that is an array of the same length of the number of channels.
+
     :param data: numpy array of pandas dataframe
+
     :param verbose: print messages while converting data
+
+    :param safe_copy: if True make a copy of the data prior to performing manipulations.
+        Will prevent overwrites of the internal references but does require more memory.
+    
     :param _tqdm: bool (try to) add a tqdm wrapper to show the progress
+
     :return: Correctly mapped data
     """
     # This map shows which channels were recabled. We now have to do the same in software
@@ -272,21 +279,32 @@ def remap_channels(data, verbose=True, _tqdm=False):
                 pass
         return x
 
-    def get_dtypes(dat):
+    def get_dtypes(_data):
         """
         Return keys/dtype names of pd.DataFrame or numpy array
-        :param dat: 
-        :return: 
+
+        :param _data: data to get the keys/dtype names
+
+        :return: keys/dtype names
         """
-        if isinstance(dat, np.ndarray):
-            _k = dat.dtype.names
-        elif isinstance(dat, pd.DataFrame):
-            _k = dat.keys()
+        if isinstance(_data, np.ndarray):
+            _k = _data.dtype.names
+        elif isinstance(_data, pd.DataFrame):
+            _k = _data.keys()
         return _k
 
-    def convert_channel(_data, replace=('channel',)):
-        """Given an array, replace the 'channel' entry if we had to remap it according to 
-        the map"""
+    def convert_channel(_data, replace=('channel', 'max_pmt')):
+        """
+        Given an array, replace the 'channel' entry if we had to remap it according to the
+            map of channels to be remapped
+
+        :param _data: data whereof the replace entries should be changed according to the
+            remapping of channels
+
+        :param replace: entries (keys/numpy-dtypes) that should be changed in the data
+
+        :return: remapped data where each of the replace entries has been replaced
+        """
         # loop over the things to replace
         for _rep in replace:
             _k = get_dtypes(_data)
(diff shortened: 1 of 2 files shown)


straxen/common.py
```diff
@@ -243,7 +247,185 @@ def get_livetime_sec(context, run_id, things=None):
...
+        :param _data: data whereof the replace entries should be changed according to the
+            remapping of channels
+
+        :param replace: entries (keys/numpy-dtypes) that should be changed in the data
+
+        :return: remapped data where each of the replace entries has been replaced
+        """
+        # loop over the things to replace
+        for _rep in replace:
+            _k = get_dtypes(_data)
```
WenzDaniel, line 314: Can be outside of the loop

straxen/common.py
```diff
@@ -243,7 +247,185 @@ def get_livetime_sec(context, run_id, things=None):
...
+            if _rep not in _k:
+                # Apparently this data doesn't have the entry we want to replace
+                continue
+            # Make a buffer we can overwrite and replace with an remapped array
+            buff = np.array(_data[_rep])
+            _n_replacements = 0
+            for _, _row in wr_tqdm(remap.iterrows()):
+                pmt_new, pmt_old = _row['PMT_new'], _row['PMT_old']
+                mask = _data[_rep] == pmt_old
```
WenzDaniel, line 323: Boolean indexing is super slow. Especially on large arrays (if I remember correctly). Maybe it would be better to change this in a simple numba function?

straxen/common.py
```diff
@@ -243,7 +247,185 @@ def get_livetime_sec(context, run_id, things=None):
...
+            _data[_rep] = buff
+            if verbose:
+                print(f'convert_channel::\tchanged {_n_replacements}/{len(_data)} items')
+        # Not needed for np.array as the internal memory already reflects it but it is
+        # needed for pd.DataFrames.
+        return _data
+
+    def remap_single_entry(_data, _entry):
+        """
+        Remap the data of a single entry.
```
WenzDaniel, line 335: Could you add some more information to this function. In the beginning I had some hard times to see the difference to the previous function.

straxen/common.py
```diff
@@ -243,7 +247,185 @@ def get_livetime_sec(context, run_id, things=None):
...
+        :param _entry: key or dtype of the data.
+
+        :return: correctly mapped data
+        """
+        _k = get_dtypes(_data)
+        if _entry not in _k:
+            raise ValueError(f'remap_single_entry::\tcannot remap {_entry} in data with'
+                             f' fields {_k}')
+        buff = np.array(_data[_entry])
+        for _, _row in remap.iterrows():
```
WenzDaniel, line 349: Do you have to loop here. Can you not do simply new[:, [1, 2, 3, 4, 5]] = old[:, [6, 7, 8, 9, 10]], might be faster.

straxen/common.py
```diff
@@ -243,7 +247,185 @@ def get_livetime_sec(context, run_id, things=None):
...
+        """Wrap input x with tqdm"""
+        if _tqdm:
+            try:
+                return tqdm.tqdm_notebook(x)
+            except (AttributeError, ModuleNotFoundError, ImportError):
+                # ok, sorry lets not wrap but return x
+                pass
+        return x
+
+    def get_dtypes(_data):
```
WenzDaniel, line 286: Might be also helpful for other people, maybe put it into common.py as a regular function idl.

straxen/common.py
jorana: Sure will do

straxen/common.py
jorana: Looks more elegant what you propose anyway, let me check

straxen/common.py
```diff
@@ -29,6 +30,9 @@
 n_tpc_pmts = 494
 n_top_pmts = 253
 
+# See https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:dsg:daq:sector_swap
+last_miscabled_run = 8796
```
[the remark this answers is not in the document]
jorana: I'm fine with either but pep8 seems to agree with you:
https://www.python.org/dev/peps/pep-0008/#constants

straxen/common.py
jorana: Done

straxen/common.py
jorana: Done

straxen/common.py
jorana: interestingly enough,  it is the fasted.
```python
%%timeit
remap_single_entry(peaks, 'area_per_channel')
```
The implementation as above:
1.84 s ± 44.9 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
What you propose (``buff[:, np.array(remap['PMT_new'].values)] = _data[_array_entry][:, np.array(remap['PMT_old'].values)]``)
2 s ± 61.8 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
Numba (first run once to initialize, after that):
2.78 s ± 46.7 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

straxen/common.py
jorana: Thanks, indeed you are right:
```python
%%timeit
convert_channel(rr)
```
The implementation above:
3.28 s ± 14.3 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

With Numba (after first initiation):
568 ms ± 10.2 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

straxen/common.py
WenzDaniel: That is indeed interesting. Looks a bit odd to me that numba is slower here. I mean the function would be the same except that the for-loop should get a boost.

jorana: one thing to keep in mind is that there are only ~20 items in this for loop. The looping here might go faster but is a rather small part of the computation time

PR #170, opened 2020-08-13
Update windows settings

GiovanniVolta:
Update windows settings after hardware components installations. 
The new seetting have been tested with succes. The summary of gain and SPE acceptance analysis with the new seeting can be found here: https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:giovo:gain_spe_update#gain_and_spe_acceptance_results.

### DIFF
--- a/straxen/plugins/led_calibration.py
+++ b/straxen/plugins/led_calibration.py
@@ -18,10 +18,10 @@ channel_list = [i for i in range(494)]
                  default=(0,40),
                  help="Window (samples) for baseline calculation."),
     strax.Option('led_window',
-                 default=(75, 105),
+                 default=(78, 116),
                  help="Window (samples) where we expect the signal in LED calibration"),
     strax.Option('noise_window',
-                 default=(20, 50),
+                 default=(10, 48),
                  help="Window (samples) to analysis the noise"),
     strax.Option('channel_list',
                  default=(tuple(channel_list)),
@@ -37,7 +37,7 @@ class LEDCalibration(strax.Plugin):
     - amplitudeNOISE: amplitude of the LED on run in a window far from the signal one.
     """
     
-    __version__ = '0.2.3'
+    __version__ = '0.2.4'
     depends_on = ('raw_records',)
     data_kind = 'led_cal' 
     compressor = 'zstd'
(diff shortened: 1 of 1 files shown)


jorana:
Hi Giovanni, thanks for the PR. You don't have to up the version if you change the option. The lineage will change accordingly. Usually you only up the version if you change the software.

@WenzDaniel 
> Not sure if it is the correct approach to always update the setting defaults and plugin version when something changes. Would not it be easier to simply update the context?

I would argue that this is the correct approach as you stay as close as possible to the relevant code. At least that is how Jelle and I have been doing it so far as much as possible.

straxen/plugins/led_calibration.py
```diff
@@ -37,7 +37,7 @@ class LEDCalibration(strax.Plugin):
     - amplitudeNOISE: amplitude of the LED on run in a window far from the signal one.
     """
     
-    __version__ = '0.2.3'
+    __version__ = '0.2.4'
```
jorana, line 40: I think you can leave it at 0.2.3 as the code itself hasn't changed

GiovanniVolta:
For me it is okay change the context but since the plugin is built at the event builder side I would like that the context over there has the right setting already.
Can I change the context in the eb machine? Otherwise I have to download the raw_records and built the led_calibration plugin by my self.

jorana:
@GiovanniVolta thanks for aa9082e . At the moment you can more or less assume that as soon as your PR gets approved (preferably by an AC) it's changed at the event-builder side as I `git checkout master` on a close to daily basis.

GiovanniVolta:
> 
> 
> @GiovanniVolta thanks for [aa9082e](https://github.com/XENONnT/straxen/commit/aa9082e38a66db4ed4eafa1c86d5dc571629ac47) . At the moment you can more or less assume that as soon as your PR gets approved (preferably by an AC) it's changed at the event-builder side as I `git checkout master` on a close to daily basis.
> 
> Daniels point was that we could also change it as an option in `straxen.contexts.xenonnt_online()` where it would have exactly the same effect. However, as I explained above, it would be neater to change the setting right at the place we use them (like you did in this PR). The result for the event-builders would be the same.

I also agree with Daniel's point. Indeed I don't mind also to donwload the raw_records, build the led_calibration compute the gaina nd delete the raw_records. With the new hardware components that you guys installed for the pulse generator the data are way less heavy than before.
Please let me know how do you want proceed.

jorana:
I'll make sure that this PR is merged before tonight and that the ebs are updated accordingly. Although I don't think its required per se - having an AC approve would be nice. Perhaps @feigaodm can approve.

Issue #149, opened 2020-07-11
Error in time range handling waveform plot


jorana:
**Problem**
Something goes wrong when plotting small time ranges in st.plot_peaks. 

**Symptons**
See below:
![afbeelding](https://user-images.githubusercontent.com/22295914/87221874-bd200580-c36f-11ea-9e79-221b83030424.png)

(start time is 1594353486000000000)

**Traceback**
https://gist.github.com/jorana/c0b8e36486250ca38e25b96d59237624

**Software**
![afbeelding](https://user-images.githubusercontent.com/22295914/87221918-f35d8500-c36f-11ea-9820-a93d43934d12.png)
running on dali

jorana:
@cheryonthetop would you think you could have a look?

JelleAalbers:
Could be related to https://github.com/AxFoundation/strax/issues/247?

jorana:
It can be related. Unfortunately the mentioned work around does not work for the runs where I ran into this problem as the datarate is so high that there are no evident slightly wider time ranges that allow me to load the data.

jorana:
Should be fixed in AxFoundation/strax#345 rather than #261

PR #172, opened 2020-08-14
add wf plot to minianalyses

jorana:
Make functions to plot wf + hitpattern 

Extend on currently implemented methods to make a combined WF plot

**Issues**
As https://github.com/AxFoundation/strax/issues/247 and https://github.com/XENONnT/straxen/issues/149 haven't been solved there are all kinds of unpreferable exception-clauses in this PR.


straxen/analyses/waveform_plot.py
```diff
@@ -211,15 +247,113 @@ def plot_peak(p, t0=None, **kwargs):
...
+    :param time_fmt: format fo the timestamp (datetime.strftime format)
+    :param kwargs: kwargs for plot_peaks
+    """
+    p = containers  # usually peaks
+
+    if isinstance(runs, str):
+        print(f'Assuming {runs} is a run id')
+        runs = pd.DataFrame({'name': [runs]})
+
+    def get_run_start(peak):
```
WenzDaniel, line 288: I think I do not fully catch the idea behind this function. Can you describe the users use case for this plot function and how to call it? To me `containers` implies somehow I can specify an iterable of things, while when reading the code I get the impression it is just for a single event peak etc.

I understand it in the following way: The user works with peaks/events from several runs in his/her analysis and spots maybe some odd behavior and would like to plot the waveform for the corresponding peaks. Hence he/she puts the corresponding list of e.g. `events` into the argument `containers` and specifies in `runs` which `run_ids` were used. Since the user does not know to which run the events belongs to, you call this function `get_run_start`? `get_run_start` searches now for the last run_id which end_time is smaller than the last containers (events) start time.   

A bit further down you plot then all peaks which are within this run_id and the corresponding time of the containers. But what happens if the containers (events) are distributed over more than one run? I have the impression you would chop all peaks which do not belong to the very last run?

WenzDaniel (commented):
I need some clarification here, sorry.

straxen/analyses/waveform_plot.py
```diff
@@ -78,16 +82,38 @@ def plot_peaks(peaks, seconds_range, t_reference, show_largest=100,
...
+                 vmin=None,
+                 log_scale=False,
+                 label=None,
+                 single_figure=False,
+                 figsize=(10, 4), ):
+    if single_figure:
+        plt.figure(figsize=figsize)
+    if len(peaks) > 1:
+        print(f'warning showing total area of {len(peaks)} peaks')
+    straxen.plot_pmts(np.sum(peaks['area_per_channel'], axis=0),
```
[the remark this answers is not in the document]
jorana: Yes, it's in a different file right

straxen/analyses/waveform_plot.py
```diff
@@ -152,22 +178,27 @@ def plot_records_matrix(context, run_id,
...
     plt.xlim(*seconds_range)
     ax = plt.gca()
     ax.ticklabel_format(useOffset=False)
     xticks = plt.xticks()[0]
     if not len(xticks):
         return
-    #xticks[0] = seconds_range[0]
-    #xticks[-1] = seconds_range[-1]
+
+    #     xticks[0] = seconds_range[0]
```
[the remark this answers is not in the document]
jorana: whoops, you are right

straxen/analyses/waveform_plot.py
```diff
@@ -211,15 +247,113 @@ def plot_peak(p, t0=None, **kwargs):
...
     """Return (x, y) numpy arrays for plotting the waveform data in p
     using 'steps-pre'.
-
     Where x is the time since t0 in seconds (or another time_scale),
       and y is intensity in PE / ns.
     :param p: Peak or other similar strax data type
     :param t0: Zero of time in ns since unix epoch
     """
-    n = p['length']
+    n = int(p['length'])
```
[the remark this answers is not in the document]
jorana: I don't think it can, can indeed remove this

straxen/analyses/waveform_plot.py
jorana: matbe I should indeed update the docstring. We are assuming that you are plotting peaks that are close to one another (why would you want to have peaks over multiple runs anyway). They are plotted in a single figure, I think that is the key information you are missing.

We need to have the run_id, so you either specify it or you just give a list that the function itself with figure out where the group of peaks belongs to (assuming all are in the same).

WenzDaniel: Okay thanks for clarification. Now it makes indeed more sense. I was in the beginning more thinking about some sort of interactive plot where you can move your view back and forth in time. Then it make sense.

jorana: Ok, just simplified a bit

WenzDaniel (approved):
Looks fine

PR #192, opened 2020-08-24
Add docs to all plugins

jorana:
Since #185, #187 and #189 we are now uploading the nT data-structure to the documentation. With this PR @WenzDaniel and I are adding brief descriptions to the plugins to contribute to the documentation.


straxen/plugins/peaklet_processing.py
```diff
@@ -53,6 +53,20 @@
...
 )
 class Peaklets(strax.Plugin):
+    """
+    Split records into:
+        -peaklets
+        -lone_hits
+
+    Peaklets are very aggressively split peaks such that we are able to find S1-S2s even
+        if they are close to each other. (S2) Peaks that are split into too many peaklets
+        will be merged later on.
```
WenzDaniel, line 63: Maybe also add one line for lone_hits?

straxen/plugins/pulse_processing.py
```diff
@@ -90,8 +90,9 @@
 class PulseProcessing(strax.Plugin):
     """
     1. Split raw_records into:
-     - tpc_records
+     - (tpc) records
      - aqmon_records
+     - pulse_counts
```
WenzDaniel, line 95: I think also for pulse_counts we should write 1-2 sentences.

WenzDaniel (approved):

straxen/plugins/peaklet_processing.py
WenzDaniel: Added something

straxen/plugins/pulse_processing.py
WenzDaniel: Done

WenzDaniel:
I think I added all the missing parts. Unfortunately I was not able to get graphviz running to check whether the format is correct.

jorana:
Hi Daniel! Thanks for the help! I just went over all of them a last time and noticed we had quite a different style (in line-breaks). I tried smoothing it a little by abiding the PEP8 convention on docstrings https://www.python.org/dev/peps/pep-0008/#maximum-line-length (72 characters).

WenzDaniel:
Hi, I thought I would use this convention already (at least for the nveto part). At least I thought to remember that I set-up guiding line in PyCharm which follows this rule. The other docs I wrote "free-hand" directly in github.

jorana:
> 
> 
> Hi, I thought I would use this convention already (at least for the nveto part). At least I thought to remember that I set-up guiding line in PyCharm which follows this rule. The other docs I wrote "free-hand" directly in github.

Therefore I also didn't alter things for the nveto part (in d2662b2) but more on the descriptions I wrote earlier.

PR #193, opened 2020-08-24
Pytest all plugins

jorana:
We don't test new plugins using travis (at least, until this PR). We do test some test data pulling for a 1T context (the demo in test_basics.py) but that will not catch all registered plugins.

For each plugin, it tries to get some data and check if it's stored (as long as the plugin is actually saving it's results).

One important thing I had to do is disable the init of the rundb as that wouldn't work while building Travis. I think this is an okay work around but if someone knows a better work around I'd be most interested because if people don't know what they are doing and disable the rundb in their context that might lead to unwanted behavior (although the same is true for the ``we_are_the_daq`` option). 

```bash
(strax) joran@DESKTOP-URE1BBI:~/software/straxen$ pytest tests/test_plugins.py
```

### DIFF
--- a/straxen/contexts.py
+++ b/straxen/contexts.py
@@ -46,6 +46,7 @@ xnt_common_config = dict(
 
 def xenonnt_online(output_folder='./strax_data',
                    we_are_the_daq=False,
+                   _database_init=True,
                    **kwargs):
     """XENONnT online processing and analysis"""
     context_options = {
@@ -53,17 +54,17 @@ def xenonnt_online(output_folder='./strax_data',
         **kwargs}
 
     st = strax.Context(
-        storage=[
-            straxen.RunDB(
-                readonly=not we_are_the_daq,
-                runid_field='number',
-                new_data_path=output_folder,
-                rucio_path='/dali/lgrandi/rucio/'),
-        ],
         config=straxen.contexts.xnt_common_config,
         **context_options)
     st.register([straxen.DAQReader, straxen.LEDCalibration])
 
+    st.storage = [
+        straxen.RunDB(
+            readonly=not we_are_the_daq,
+            runid_field='number',
+            new_data_path=output_folder,
+            rucio_path='/dali/lgrandi/rucio/'), ] \
+        if _database_init else []
     if not we_are_the_daq:
         st.storage += [
             strax.DataDirectory(
--- a/straxen/contexts.py
+++ b/straxen/contexts.py
@@ -58,13 +58,11 @@ def xenonnt_online(output_folder='./strax_data',
         **context_options)
     st.register([straxen.DAQReader, straxen.LEDCalibration])
 
-    st.storage = [
-        straxen.RunDB(
-            readonly=not we_are_the_daq,
-            runid_field='number',
-            new_data_path=output_folder,
-            rucio_path='/dali/lgrandi/rucio/'), ] \
-        if _database_init else []
+    st.storage = [straxen.RunDB(
+        readonly=not we_are_the_daq,
+        runid_field='number',
+        new_data_path=output_folder,
+        rucio_path='/dali/lgrandi/rucio/'), ] if _database_init else []
     if not we_are_the_daq:
         st.storage += [
             strax.DataDirectory(
--- a/straxen/plugins/pulse_processing.py
+++ b/straxen/plugins/pulse_processing.py
@@ -222,7 +222,6 @@ class PulseProcessingHe(PulseProcessing):
         records_he=False,
         pulse_counts_he=True)
     depends_on = 'raw_records_he'
-    parallel = 'process'
     compressor = 'lz4'
     child_ends_with = '_he'
 
--- a/tests/test_plugins.py
+++ b/tests/test_plugins.py
@@ -33,12 +33,14 @@ class DummyRawRecords(strax.Plugin):
         return chunk_i < N_CHUNKS
 
     def compute(self, chunk_i):
-        r = np.zeros(recs_per_chunk, self.dtype['raw_records'])
         t0 = chunk_i + self.config['secret_time_offset']
-        r['time'] = t0
-        r['length'] = r['dt'] = 1
-        r['channel'] = np.arange(len(r))
-
+        if chunk_i < N_CHUNKS - 1:
+            r = np.zeros(recs_per_chunk, self.dtype['raw_records'])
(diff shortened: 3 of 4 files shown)


straxen/contexts.py
```diff
@@ -46,24 +46,25 @@
...
         config=straxen.contexts.xnt_common_config,
         **context_options)
     st.register([straxen.DAQReader, straxen.LEDCalibration])
 
+    st.storage = [
+        straxen.RunDB(
+            readonly=not we_are_the_daq,
+            runid_field='number',
+            new_data_path=output_folder,
+            rucio_path='/dali/lgrandi/rucio/'), ] \
```
WenzDaniel, line 66: I very much dislike these "\\"-line breaks does not make the code very readable. Since st.storage is a simple list could not you do simply:

```
if _database_init:
    st.storage.append(....)
    # or st.storage.insert(0, ...) in case it shall be listed first
```

straxen/plugins/pulse_processing.py
```diff
@@ -222,7 +222,6 @@ class PulseProcessingHe(PulseProcessing):
         records_he=False,
         pulse_counts_he=True)
     depends_on = 'raw_records_he'
-    parallel = 'process'
```
WenzDaniel, line 225: Why did you remove this? I am asking because it was previously the default for PulseProcessing. Hence I used it for the nveto version of the plugin as well.

tests/test_plugins.py
```diff
@@ -0,0 +1,143 @@
...
+    """
+
+    with tempfile.TemporaryDirectory() as temp_dir:
+        st.storage = [strax.DataDirectory(temp_dir)]
+
+        # As we use a temporary directory we should have a clean start
+        assert not st.is_stored(run_id, 'raw_records'), 'have RR???'
+
+        # # 1Ts NN seems funky, let's ignore it for now?
+        # target = {'nT': 'event_info', '1T': 'event_info'}[context_name]
```
WenzDaniel, line 68: Needed?

tests/test_plugins.py
```diff
@@ -0,0 +1,143 @@
...
+        assert st.is_stored(run_id, target), f'Could not make {target}'
+
+        # I'm only going to do this for nT because:
+        #  A) Doing this many more times does not give us much more
+        #     info (everything above already worked fine)
+        #  B) Most development will be on nT, 1T may get less changes
+        #     in the near future
+        if make_all:
+            # First make some _he stuff for multiprocessing
+            st.make(run_id, 'peaks_he', **proces_kwargs)
```
WenzDaniel, line 84: Why do you need this line here? Should not his be covered when you loop over the plugin registry

tests/test_plugins.py
```diff
@@ -0,0 +1,143 @@
...
+        return chunk_i < N_CHUNKS
+
+    def compute(self, chunk_i):
+        r = np.zeros(recs_per_chunk, self.dtype['raw_records'])
+        t0 = chunk_i + self.config['secret_time_offset']
+        r['time'] = t0
+        r['length'] = r['dt'] = 1
+        r['channel'] = np.arange(len(r))
+
+        res = {p: self.chunk(start=t0, end=t0 + 1, data=r, data_type=p)
```
WenzDaniel, line 42: I think we can have here easily two birds with one stone. Maybe just add an additional chunk which is empty to test this here as well.

WenzDaniel (approved):

straxen/contexts.py
jorana: I agree with you on the ''\'' in general. I don't feel strongly about it but the reason I think it's justifiable here is that for any non-oddly specific use cases (like this PR) you'd read it like you want it to be most of the time (you just init the rundb). 

I will change it though as I was also in doubt if this was really a good enough reason to write it like this.

straxen/plugins/pulse_processing.py
jorana: It's obscuring the fact that we are already copying this option from the parent so this change is not of any relevance for the nveto.

tests/test_plugins.py
jorana: Good idea, will do.

jorana: In 0d09964 I indeed found a bug (which I might have induced) in the Ledplugin using an empty chunk, thanks Daniel!

tests/test_plugins.py
```diff
@@ -0,0 +1,141 @@
...
+        # Create event info
+        target = 'event_info'
+        st.make(run_id=run_id,
+                targets=target,
+                **proces_kwargs)
+
+        # The stuff should be there
+        assert st.is_stored(run_id, target), f'Could not make {target}'
+
+        # I'm only going to do this for nT because:
```
WenzDaniel, line 78: I do not get this comment. You are calling this function in test_1T.

tests/test_plugins.py
```diff
@@ -0,0 +1,141 @@
...
+def test_nT(ncores=1):
+    if ncores == 1:
+        print('-- nT lazy mode --')
+    st = straxen.contexts.xenonnt_online(_database_init=False)
+    _update_context(st, ncores)
+    _run_plugins(st, make_all=True, max_wokers=ncores)
+    print(st.context_config)
+
+
+def test_nT_mutlticore():
```
WenzDaniel, line 134: Do you see any reason to test things also in lazy mode? I do not, but on the other hand these tests should be super fast why not simply run them.

WenzDaniel (approved):

WenzDaniel:
Thanks Joran, this will be a super nice and helpful addition.

tests/test_plugins.py
jorana: Done

tests/test_plugins.py
jorana: I'd argue that we want to know if we only have issues in lazy/mutlicore mode as that would allow us to debug software faster. Also I don't see any particular valid reason not to test lazy mode

tests/test_plugins.py
jorana: I'm calling this fucntion always for 1T with this argument:
https://github.com/XENONnT/straxen/blob/8a5ea4034126ab3c7dcee4bde95fd7170798859e/tests/test_plugins.py#L121

Therefore the ```if make_all``` is always ``False`` for 1T

PR #199, opened 2020-08-28
Fix documentation styling

jorana:
Fix https://github.com/XENONnT/straxen/pull/192 for https://straxen.readthedocs.io/en/latest/reference/datastructure.html. As one can see some Plugins have funny formats. Let's hopefully fix it like this.


WenzDaniel (approved):
Looks good.

PR #200, opened 2020-08-28
Patch he names

WenzDaniel:
Explicitly call high energy Plugins ...HighEnergy instead of ...HE

### DIFF
--- a/straxen/plugins/peak_processing.py
+++ b/straxen/plugins/peak_processing.py
@@ -96,7 +96,7 @@ class PeakBasics(strax.Plugin):
 
 
 @export
-class PeakBasicsHe(PeakBasics):
+class PeakBasicsHighEnergy(PeakBasics):
     __doc__ = PeakBasics.__doc__
     __version__ = '0.0.1'
     depends_on = 'peaks_he'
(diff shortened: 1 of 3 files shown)


jorana:
Hi Daniel, in principle looks fine. Will test it on Monday if we don’t change anything unintendedly. 

Also in d951245  I added a small preamble for all HE plugins. I assume you are fine with that.

jorana:
@WenzDaniel I just checked and I don't see any problems. I think we do need approve of @feigaodm as the lineage of ``records_he``  and beyond changes. @feigaodm, if we change this now (for readability) we need to reprocess the following plugins:
afbeelding

jorana (approved):
Software wise it looks completely fine. The lineage changes hence I think the merge approval should be an ACs call.

feigaodm:
Thanks @WenzDaniel and @jorana , it looks fine to me. I guess we will need to reprocess every run when this is merged. So far I see everything from normal and HE boards are separately processed and analyzed. We shall think about how can we connect the two together nicely, but this might be more natural when we have good S2s, especially when it's related to positions. These comments are not directly related to this PR though, but I want to put it here as a reminder for us ;-)

feigaodm (approved):

Issue #205, opened 2020-09-01
Data management on eventbuilders


jorana:
Yesterday runs 9236, 9237, 9238 were irrecoverably lost from the event builders. The big question is how it can be that we at some point seem to have lost the data stored on `/data/xenonnt_processed` or have somehow passed [line 586](https://github.com/XENONnT/straxen/blob/fd7c2c9301408056c05b8e17ea44ed4a548fc712/bin/ajax#L586) in `Ajax`.

The reason turned out to be a faulty logic in [this line](https://github.com/XENONnT/straxen/blob/fd7c2c9301408056c05b8e17ea44ed4a548fc712/bin/ajax#L518) that said that that are less than two hours old should be deleted. 

**RECONSTRUCTION**
These are the events that happened (focusing on runs 9236).

**"2020-08-31T18:05:47.776Z" - Processing finished**
`Bootstrax` successfully processed the run and then deleted the [live_data at this time ](https://github.com/XENONnT/straxen/blob/3b9a7ae1694f6b4dc22163e02c8883c5eab37087/bin/bootstrax#L1290) which is only done after we have successfully stored all the data (there is a [check in](https://github.com/XENONnT/straxen/blob/3b9a7ae1694f6b4dc22163e02c8883c5eab37087/bin/bootstrax#L901) ```set_status_finished``` that makes sure the data has been written to disk). Furthermore, one can see from the deleted entries that we have saved .
![afbeelding](https://user-images.githubusercontent.com/22295914/91857607-65639380-ec68-11ea-817b-7ea75b7822a0.png)

**"2020-08-31T19:.." - `Ajax` deletes the 'unregistered' data**
[The bug was here](https://github.com/XENONnT/straxen/blob/fd7c2c9301408056c05b8e17ea44ed4a548fc712/bin/ajax#L518)
```
7208899 MainThread root clean_unregistered::    found 398 runs stored on/data/xenonnt_processed/. Checking that each is in the runs-database
7209768 MainThread root remove_if_unregistered::        run 9236 is NOT registered in the runDB
7209768 MainThread root No data for 009236 found! Double checking /data/xenonnt_processed/!
7209769 MainThread root Cleaning /data/xenonnt_processed/009236-raw_records_nv-rfzvpzj4mf
7209770 MainThread root Cleaning /data/xenonnt_processed/009236-raw_records_aqmon-rfzvpzj4mf
```

**"2020-08-31T20:18:53.934Z" - `Ajax` removes entries from runs-database**
[...]

PR #206, opened 2020-09-01
Eb data management

jorana:
The issue is detailed in #205 . Although the error is[ only in a single line ](https://github.com/XENONnT/straxen/blob/fd7c2c9301408056c05b8e17ea44ed4a548fc712/bin/ajax#L518) it's of such severity that we should safeguard ourselfs from making the same mistake again.

Furthermore in this PR we re-base the logging of error messages by `bootstrax ` to the same collection as the readers. 
Input would be appreciated if [these kind of levels ](https://github.com/XENONnT/straxen/blob/8ec44042d6dd54833e2ad0377fc45c94e74d59a4/bin/bootstrax#L510) are in line with the readers and if not what would be appropriate levels.

- Disable multiple modes that shouldn't be run on a regular basis.
- Add checks (even if they don't make any sense at the moment) to doubly check that it's okay to delete data.
- Redirect warning messages to the daq_db collection and add a run_id wherever possible.

### DIFF
--- a/bin/ajax
+++ b/bin/ajax
@@ -15,7 +15,7 @@ that may delete the live_data such that we prevent multiple attempts to perform 
 action.
 """
 
-__version__ = '0.2.4'
+__version__ = '0.2.5'
 
 import argparse
 from datetime import datetime, timedelta
@@ -258,6 +258,7 @@ def check_sn_trigger():
     Get the information of a supernova trigger from some place. If occurred we should
     crash immediately to prevent data loss
     """
+    raise NotImplementedError("Don't have SN checks yet")
     # Check some database for supernova triggers
     super_nova = False
     # super_nova = True if run_db['sn_trigger'].find_one({'time':{'$exists':True}}) else False
@@ -268,6 +269,8 @@ def check_sn_trigger():
 
 def clean_raw_records_nv():
     """Based on the runsDB, remove the raw_records_nv if they are old"""
+    raise NotImplementedError("NV integration not yet part of AJAX")
+
     check_sn_trigger()
     rd = run_coll.find_one({'bootstrax.state': 'done',
                             'status': 'transferred',
@@ -290,7 +293,7 @@ def clean_raw_records_nv():
 def clean_unregistered(delete_live=False):
     """
     Clean data that is not in the database. To do this check the output folder and remove
-    files if there is no corresponding entry in the rundoc
+    files if there is no corresponding entry in the rundoc.
     :param delete_live: bool, if true delete unregistered live data.
     """
     folder_to_check = output_folder if delete_live is False else ceph_folder
@@ -333,6 +336,9 @@ def clean_old_hash():
                                     'data.type': data_type,
                                     'data.host': hostname,
                                     'number': int(run_id)})
+            if 'raw_records' in data_type:
+                raise ValueError(f'You did some sloppy regex on the data-type, almost '
+                                 f'deleted {data_type}!')
             if rd:
                 delete_data(rd, loc, data_type, test=not args.execute)
             else:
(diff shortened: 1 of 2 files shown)


bin/bootstrax
```diff
@@ -490,22 +490,27 @@ def send_heartbeat(update_fields=None):
...
         return
     getattr(log, priority)(message)
-    log_coll.insert_one({
+    warning_message = {
         'message': message,
         'user': f'bootstrax_{hostname}',
-        'priority': dict(warning=2, info=1).get(priority, 3)})
+        'priority': dict(warning=2, info=1).get(priority, 3)}
+    if run_id:
+        warning_message.update({'run_id': run_id})
```
darrylmasson, line 512: Redax uses `runid` as the key (see [here](https://github.com/coderdj/redax/blob/master/MongoLog.cc#L141)). Also, I think pep8 recommends the explicit `if run_id is not None:`, but that's stylistic. Also, is it worth adding the full hierarchy of error messages from `debug` up to `fatal`? It seems like just two could be too limiting.

bin/bootstrax
```diff
@@ -659,7 +667,7 @@ def sufficient_diskspace():
             send_heartbeat(dict(state='disk full'))
     set_state(dead_state)
     message = f"No disk space to write to. Kill bootstrax on {hostname}"
-    log_warning(message, priority='warning')
+    log_warning(message, priority='Error')
```
darrylmasson, line 670: We should be consistent with capitalization of message levels. Also, this is something that redax would log with 'fatal' priority, because it's this process's last dying gasp before it dies.

bin/bootstrax
```diff
@@ -1173,7 +1188,9 @@ def process_run(rd, send_heartbeats=args.production):
                        'strax_fragment_payload_bytes')
             for conf in to_read:
                 if conf not in dq_conf:
-                    log_warning(f'{conf} not in rundoc for {run_id}! Using default values.')
+                    log_warning(f'{conf} not in rundoc for {run_id}! Using default values.',
+                                priority='info',
+                                run_id=run_id)
```
darrylmasson, line 1193: Rather than potentially log several messages in this for-loop, it might be cleaner to accumulate them and log outside the loop.

darrylmasson (commented):
Dan wrote a brief treatise on logging levels in [redax](https://github.com/coderdj/redax/blob/master/MongoLog.hh#L22), which I will summarize here.

- LOCAL: the lowest logging level that contains most of the program's thoughts as it runs. Sometimes called TRACE in other software. Only written to file and kept for a week.
- DEBUG: debug stuff. Stuff useful during commissioning, but once you enter "stable" operation it's largely useless. Only written to disk.
- INFO: something you want a user of the website to be able to see, but not be notified of. Things that aren't necessarily run-of-the-mill but aren't causing issues.
- WARNING: a little oops. Nothing crashed, but something might have broken somewhere. Either way, the end user should be notified because all is probably not well.
- ERROR: a more significant issue. Something stopped, and expert intervention is probably required.
- FATAL: the last thing the program gasps out before it dies ignominiously. This message is what the operator should expect to find on the process' tombstone.

jorana:
Thanks @darrylmasson , I believe I've addressed the issues you pointed out.

bin/ajax
```diff
@@ -335,6 +356,8 @@ def clean_abandoned(delete_live=False):
     # runs to be deleted simultaneously.
     if not delete_live:
         rd = run_coll.find_one({'bootstrax.state': 'abandoned',
+                                'bootstrax.time':
+                                    {"$lt": now(-ajax_thresholds['wait_after_processing'])},
```
darrylmasson, line 360: Is this intended to protect us against accidental abandonment? I'm not sure if this is particularly likely to happen.

bin/ajax
```diff
@@ -371,6 +396,38 @@ def clean_abandoned(delete_live=False):
...
+    """
+    Remove entries from the database if the data is not on this host anymore
+    """
+    # We need to grep all the rundocs at once as simply doing one at the time (find_one)
+    # might get stuck as the query may yield the same result the next time it is called.
+    if not delete_live:
+        rds = run_coll.find({'bootstrax.state': 'done',
+                             'bootstrax.time':
+                                 {"$lt": now(-ajax_thresholds['wait_after_processing'])},
+                             'data.host': hostname})
```
darrylmasson, line 409: Do we want to add `status: 'transferred'` here as well, just to be explicit, or do we think it isn't necessary?

bin/ajax
```diff
@@ -442,6 +532,23 @@ def delete_data(rd, path, data_type, test=True):
...
+    """
+    Count the number of files of live_data (cannot be >1) and raw_data (transfers can get
+    it greater than 1)
+    :param rd: rundoc
+    :return: length 2 tuple with n_live_data and n_raw_records being the number of files
+    in the rundoc for the live data and raw records respectively.
+    """
+    n_live_data, n_raw_records = 0, 0
+    for dd in rd['data']:
+        if dd['type'] == 'live' and dd['location'] != 'deleted':
```
darrylmasson, line 545: The second half of this boolean shouldn't do anything, once we delete data it isn't in `rd['data']` anymore.

darrylmasson (approved):
I like it.

bin/ajax
jorana: True, it's some legacy code. I just queried and I think we indeed have no more `{"data.location":"deleted"}` in the runsdb

jorana: Removed in 983958e

bin/ajax
```diff
@@ -869,6 +865,7 @@ if __name__ == '__main__':
     daq_client = pymongo.MongoClient(f"mongodb://{daq_db_username}:{daq_db_password}@xenon1t-daq:27017/admin")
     daq_db = daq_client[daq_db_name]
     log_coll = daq_db['log']
+    daq_db.command('ping')
```
darrylmasson, line 868: Splitting hairs here, but both databases run on the same server instance, so I don't see a situation where we can ping one but not the other.

bin/ajax
```diff
@@ -858,9 +853,10 @@ if __name__ == '__main__':
     run_db_username = straxen.get_secret('RUNDB_USERNAME'.lower())
     run_db_password = straxen.get_secret('RUNDB_PASSWORD'.lower())
     run_client = pymongo.MongoClient(
-        f"mongodb://{run_db_username}:{run_db_password}@xenon1t-daq:27017,old-gw:27017/admin")
+        f"mongodb://{run_db_username}:{run_db_password}@xenon1t-daq:27017/xenonnt")
```
darrylmasson, line 856: I'm mildly confused that this worked in the past. Authenticating against the wrong database should fail.

jorana: Sorry Darryl. This was a faulty commit that didn't work. I reverted it in d96c33c .

bin/ajax
jorana: If we use the wrong credentials for one we will be able to ping the other.

bin/ajax
darrylmasson: Fair enough

jorana: I was mixing the read-only account with the admin password. Of course we want to use the admin account if we need to update the database.

bin/ajax
darrylmasson: This is true, but credential-handling comes with clean error message, and I wouldn't expect it to be a lasting issue.

Issue #211, opened 2020-09-07
Throw a warning if fried rice is down insteat of crashing everything.


WenzDaniel:
In our current implementation of the rundb.py straxen raises an error when fried rice is down:

```
~/mymodules/straxen/straxen/rundb.py in _find(self, key, write, allow_incomplete, fuzzy_for, fuzzy_for_options)
    153                         'protocol': 'rucio'}}}
    154             doc = self.collection.find_one({**run_query, **dq},
--> 155                                            projection=dq)
    156             if doc is not None:
    157                 datum = doc['data'][0]

# (some long mongo trace back)

ServerSelectionTimeoutError: fried.rice.edu:27017: timed out
```

This issue was introduced with https://github.com/XENONnT/straxen/pull/164/files .

In most cases analyzers will work with locally stored data and hence do not need to access anything via the rundb storage system. Hence I propose rather to throw a warning and to drop the rundb storage system from the registered storage rather than raising an error and stopping any ongoing analysis.

WenzDaniel:
Evans PR makes this issue obsolete

jorana:
@WenzDaniel #164 is not related to this issue. Rather https://github.com/XENONnT/straxen/pull/114 is the cull-pit.

jorana:
An alternative solution is proposed in #213 using the other two mirrors as backup in order not to prevent analyzers to work outside of dali (see cited PR)

PR #212, opened 2020-09-07
Use all three mongo mirrors

ershockley:
This should address #211. I'm being a bad collaborator and have not tested it -- @WenzDaniel can you give it a try maybe? Sorry really busy. 

Allow for all three mongo mirrors to be used instead of just fried rice. 

Same as before, just new string.

Nope, haven't tested it.

### DIFF
--- a/straxen/rundb.py
+++ b/straxen/rundb.py
@@ -12,7 +12,7 @@ import straxen
 export, __all__ = strax.exporter()
 
 
-default_mongo_url = 'fried.rice.edu:27017/xenonnt'
+default_mongo_url = 'xenon-rundb.grid.uchicago.edu:27017,fried.rice.edu:27017,xenon1t-daq.lngs.infn.it:27017/xenonnt'
 default_mongo_dbname = 'xenonnt'
 default_mongo_collname = 'runs'
 


WenzDaniel (approved):
Hi Evan, thanks a lot. I tested it and it works.

straxen/rundb.py
```diff
@@ -12,7 +12,7 @@
 export, __all__ = strax.exporter()
 
 
-default_mongo_url = 'fried.rice.edu:27017/xenonnt'
+default_mongo_url = 'xenon-rundb.grid.uchicago.edu:27017,fried.rice.edu:27017,xenon1t-daq.lngs.infn.it:27017/xenonnt'
```
jorana, line 15: Would it make sense to put fried.rice first? I'm not an expert but I can imagine that it tries connecting sequentially?

jorana (changes requested):
Hi Evan, thanks! I think we also discussed this in:
https://github.com/XENONnT/straxen/pull/114

By the way here is a minimal working example:
```python
import strax
import straxen
st = strax.Context(
        **straxen.contexts.common_opts)
username = straxen.get_secret('rundb_username')
password = straxen.get_secret('rundb_password')
_mongo_url = f'mongodb://{username}:{password}@fried.rice.edu:27017,xenon-rundb.grid.uchicago.edu:27017,xenon1t-daq.lngs.infn.it:27017/xenonnt'
st.storage = [straxen.RunDB(
    mongo_url= _mongo_url,
        readonly=True,
        runid_field='number',
        new_data_path='./strax_data',
)
             ]
st.storage[0].client['xenonnt'].command('ping')
```

It at the moment does not work if you are not on dali.

ershockley:
@jorana do you mean not on dali but also not at LNGS? Both Chicago and fried rice are world accessible, so we could also just put those two, as discussed in #114.

straxen/rundb.py
ershockley: If it goes sequentially I think it makes the most sense to have the Chicago one first since that is the closest to midway and on same uchicago network.

jorana:
@Evan, I'm closing this bevause of #213 if that is alright.

PR #213, opened 2020-09-08
Add backup mongo URLs 

jorana:
Use Chicago and xenon1t-daq.lngs.infn.it as backup mongo instances for fried-rice.

Use fried-rice always as the first default as that is accessible outside dali

Try to connect to fried-rice, if it doesn't use a ```backup_mongo_url``` 

```python
import strax
import straxen

RunDB = <see PR>

st = strax.Context(
        **straxen.contexts.common_opts)
username = straxen.get_secret('rundb_username')
password = straxen.get_secret('rundb_password')
st.storage = [RunDB(
        readonly=True,
        runid_field='number',
        new_data_path='./strax_data',
)
             ]
st.storage[0].client['xenonnt'].command('ping')
```

**Difference with #212**
Rather than having all tree in a single URL we split it to a to stage initialization. Otherwise if not on dali it cannot find the primary host.

### DIFF
--- a/straxen/rundb.py
+++ b/straxen/rundb.py
@@ -113,7 +113,6 @@ class RunDB(strax.StorageFrontend):
                             raise pymongo.errors.ServerSelectionTimeoutError(
                                 'Cannot connect to any mongo url')
 
-
         self.client = pymongo.MongoClient(mongo_url)
 
         if mongo_dbname is None:
--- a/straxen/rundb.py
+++ b/straxen/rundb.py
@@ -12,8 +12,9 @@ import straxen
 export, __all__ = strax.exporter()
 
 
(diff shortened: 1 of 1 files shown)


ershockley:
If we want to go this route it makes more sense to use the Chicago mirror as the default IMO -- it's much closer to midway/dali. I also don't get why this is a preferred method to #212?

jorana:
Evan and I had a chat, here is [an example showing ](https://gist.github.com/jorana/54f93cb89c6e4a154329b5335b913f49) that the approach in #212 fails because mongo cannot decide on who is primary. This PR (213) is a patch rather than the solution to the problem that is shown in the appended gist.

ershockley (approved):

PR #216, opened 2020-09-11
add online monitor

jorana:
Preamble: https://github.com/AxFoundation/strax/pull/315 should be merged first before this PR makes sense.

Add a new frontend to the main context.

It allows certain specified plugins to be writing data to the online_monitor collection of the database. This data can then be loaded elsewhere and displayed on the website 

The output will look like:
https://gist.github.com/jorana/d3176c4e561ccca4e8ca2fc2420acf7c

For code see:
XENONnT/analysiscode#27


straxen/contexts.py
```diff
@@ -62,7 +62,10 @@ def xenonnt_online(output_folder='./strax_data',
         readonly=not we_are_the_daq,
         runid_field='number',
         new_data_path=output_folder,
-        rucio_path='/dali/lgrandi/rucio/'), ] if _database_init else []
+        rucio_path='/dali/lgrandi/rucio/'),
+        straxen.OnlineMonitor(readonly=not we_are_the_daq,
```
jorana, line 66: @darrylmasson we could move this line under an ``if we_are_the_daq`` to prevent a multitude of queries. Any expert strax user could later do 

```python
st.storage += [OnlineMonitor(readonly=True, take_only=('pulse_counts', 'pulse_counts_he')]
```

jorana:
Since https://github.com/XENONnT/straxen/pull/207 it is also possible to output the deadtime to the website. This will be added soon to this PR after https://github.com/AxFoundation/strax/pull/315 has been merged.

straxen/__init__.py
```diff
@@ -3,6 +3,7 @@
 from .common import *
 from .itp_map import *
 from .rundb import *
+from  .online_monitor import *
```
jorana, line 6: Why is there an extra space here?

straxen/online_monitor.py
```diff
@@ -0,0 +1,31 @@
+from straxen.rundb import get_mongo_url, default_mongo_dbname
+from strax import MongoFrontend, exporter
+from socket import getfqdn
+
+export, __all__ = exporter()
+
+default_online_collection = 'online_monitor'
```
jorana, line 7: There is no collection for this yet. At the moment it is envisaged as a TTL with a duration of about a week, indexed by run-number.

straxen/online_monitor.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+@export
+class OnlineMonitor(MongoFrontend):
+    """Online monitor Frontend for Saving data temporarily to the the database"""
+    def __init__(self,
+                 uri=None,
+                 take_only=None,
+                 database=default_mongo_dbname,
+                 col_n=default_online_collection,
```
jorana, line 17: rename to col_name (i.e. convention in strax)

jorana:
@darrylmasson, based on your request I've changed the init of the online monitor backend. This is now only available if one is the DAQ or adds the storage by hand.

Also, what plugins would you say we would want to display on the website? I would argue the AQMon-plugin written by @AlexElykov and also the pulse-counts as you can really see the rates changing while also allowing e.g. a PMT expert to load the associated data.

AlexElykov:
> 
> Also, what plugins would you say we would want to display on the website? I would argue the AQMon-plugin written by @AlexElykov and also the pulse-counts as you can really see the rates changing while also allowing e.g. a PMT expert to load the associated data.

I think it is a good idea, especially if we will able to deploy this before the upcoming light yield tests, this will give us a better handle on the DAQ/veto performance.

What is the envisioned implementation of this on the user side (@darrylmasson), is it gonna be another page on the DAQ website (NODIAQ..jade, javascript based) like I've made for the deadtime chart? or you has something else in mind?

darrylmasson:
I don't have anything specific envisioned for the visualization, but we already have the page that interfaces with microstrax, and that seems like a pretty good place to put things.

straxen/online_monitor.py
darrylmasson: Which DB do we envision this going into?

jorana: RunsDB makes most sense. It is accesable if needed via dali without the need to share DAQ database-cridentials.

darrylmasson: DAQ DB credentials don't work outside of the DAQ anyway.

jorana:
After https://github.com/AxFoundation/strax/pull/323, there is an entry in each document: `write_time`. 

To this end, I've created a new collection `online_monitor` in the xenonnt database which is TTL and expires after 7 days.

straxen/rundb.py
```diff
@@ -295,3 +273,30 @@ def run_metadata(self, run_id, projection=None):
         if self.reader_ini_name_is_mode:
             doc['mode'] = doc.get('reader', {}).get('ini', {}).get('name', '')
         return doc
+
+
+def get_mongo_url(hostname):
```
darrylmasson, line 278: Was splitting this function out of the class strictly necessary? I could see if we were exporting it for other modules, but we aren't (at least there's no `@export`)

straxen/rundb.py
```diff
@@ -295,3 +273,30 @@ def run_metadata(self, run_id, projection=None):
...
+    if hostname.endswith('xenon.local'):
+        username = straxen.get_secret('mongo_rdb_username')
+        password = straxen.get_secret('mongo_rdb_password')
+        url_base = 'xenon1t-daq:27017,old-gw:27017/admin'
```
darrylmasson, line 282: `old-gw` is dead (long live `old-gw`), probably best not to include it in the URI.

straxen/online_monitor.py
```diff
@@ -0,0 +1,35 @@
+from straxen.rundb import get_mongo_url, default_mongo_dbname
```
darrylmasson, line 1: Ok, here we're importing `get_mongo_url`, but neither of these quantities get exported from `rundb.py`

darrylmasson (changes requested):
Missing `@export`s, I think? Otherwise I'm missing something.

straxen/rundb.py
jorana: We need it in the online_monitor. I don't like duplicating code so therefore I recon it's better to split it.

straxen/rundb.py
jorana: Let's do some voodoo to reincarnate old-gw soon. But good point let me change it

straxen/online_monitor.py
jorana: You don't need to export it. The @export allows you to e.g. do:
```straxen.DAQReader`` 
instead of
``straxen.plugins.daqreader.DAQReader`` 

But either works. For these functions I say there is no need for e.g. a `` straxen.get_mongo_url.``

jorana:
After implementing https://github.com/AxFoundation/strax/pull/327 we successfully ran this PR on the event builder. The entries for run 9158 are now in the rundoc. Please notice that in the case of crashes (like the first document in the 'online_monitor' collection) are not automatically removed but can be recognized by the omission of 'writing ended'.

WenzDaniel (approved):
Looks fine to me, but I am not a mongo expert at all. Maybe just add an export and some doc-string to get_mongo_url.

jorana:
Hi all,
Thanks, I don't think exporting `get_mongo_url` is a good idea as people might for some reason start printing it in a notebook and upload our cridentials for the mongo-database somewhere. There is limited reason why people should want to get the mongo-url outside the instances provided in this PR (RunDB and OnlineMonitor).

I will add a docstring, thanks for the good suggestion Daniel!

PR #221, opened 2020-09-23
HE veto length clipped at chunk end time

sophiaandaloro:
Software HE veto helps veto tails of large peaks to reduce unneeded data. The duration of this veto should be a configurable setting, which defaults to the end of the chunk if the length is larger than the chunk length. 

The issue is that the length is clipped at the end of records length right now, which is not always the same as the chunk endtime.

np.clip used to clip values above chunk endtime. 

```
    veto_end = np.clip(peaks['time'] + veto_length,
                       None,
                       chunk_end)
```
will return `peaks['time'] + veto_length` or `chunk_end` time, whichever is smaller.

  - Update the docstring(s) DONE
  - Update the documentation DONE
  - Tests to check the (new) code is working as desired. TESTING WITH PR 
  - Does it solve one of the open issues on github? NO (but I refrained from opening one just to solve it!)

### DIFF
--- a/straxen/plugins/pulse_processing.py
+++ b/straxen/plugins/pulse_processing.py
@@ -176,7 +176,7 @@ class PulseProcessing(strax.Plugin):
 
         if len(r) and self.hev_enabled:
             r, r_vetoed, veto_regions = software_he_veto(
-                r, self.to_pe,
+                r, self.to_pe, end,
                 area_threshold=self.config['tail_veto_threshold'],
                 veto_length=self.config['tail_veto_duration'],
                 veto_res=self.config['tail_veto_resolution'],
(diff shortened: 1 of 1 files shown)


sophiaandaloro:
One quick question: do I need to increment the version for something this minor?

feigaodm:
If it triggers reprocessing, we may consider to increment the version later but it's better to remember it.

WenzDaniel (commented):
Hi Sophia, thanks a lot for spotting and fixing this. Do you mind to update to doc string of software_he_veto as well? 
A version increment might here indeed not be necessary since in the current settings are not using the he_veto anyhow,

sophiaandaloro:
Yes agree that since the current settings disable the veto, reprocessing will not be affected at this time. Docstring updated to include new parameter passed through for chunk endtime.

PR #223, opened 2020-09-28
Fix comments cable mapping (and PEP8)

jorana:

- Add more info about remapping channels and provide reference. See https://github.com/XENONnT/straxen/pull/166 and https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:dsg:daq:sector_swap for reference. @tunnell pointed out that the description in the context is minimal. This PR adds some more info (in addition to the readily available info already in context.py).
- There were also some PEP8 and indentation issues with the context that are fixed on the fly.

### DIFF
--- a/straxen/contexts.py
+++ b/straxen/contexts.py
@@ -17,26 +17,25 @@ common_opts = dict(
         'name', 'number', 'tags.name',
         'start', 'end', 'livetime', 'mode'))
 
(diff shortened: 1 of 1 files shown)


WenzDaniel (approved):
PEP 8 <3 :)

PR #225, opened 2020-09-29
Use CutPlugins for the cuts ported from 1T

jorana:
The cuts have a lot of copy pasting in the code. With this PR we utilize the CutPlugin-plugintype intoduced in https://github.com/AxFoundation/strax/pull/274.

Please refer to the just linked strax PR


zhut19:
Nice changes! Mind I have two suggestions? 
1. It would be more consistent if the provided field can match up with the one generated by `camel_to_snake`. e.g. 
```
class CutS2SingleScatter(strax.CutPlugin)
CutS2SingleScatter().provides == 'cut_s2_single_scatter'
```
2. How about adding a class property listing all the fields required by the cut algo. also take `S2SingleScatter` for example with something like. Although this will be dangling around for the time being, I think it might come in handy later when someone (me) trying to downsize the data fields before the cuts.
```
depends_on_fields = ['alt_s2_area', 's2_area']
```

straxen/plugins/x1t_cuts.py
```diff
@@ -121,39 +114,30 @@ def other_s2_bound(cls, s2_area):
 
         return rescaled_s2_0 * another_term_0 + rescaled_s2_1 * another_term_1
 
-    def compute(self, events):
+    def cut_by(self, events):
```
zhut19, line 117: Here's an old bug `events['s2_largest_other']` should now be `events['alt_s2_area']`

jorana:
@Tianyu , is 1a13d32 what you had in mind: CamelCases for the cut names and snake_names for the provides arguement? I agree, let's agree on a consistent convention.

Concerning point 2 it sounds like a nice feature but I don't think it falls under the scope of this PR. If we'd want something like this I would start a strax issue as this is arguably not only usefull for cut plugins but for any plugin. If we'd be adding it manually it sounds like more work than it is to checkout the code itself.

zhut19:
It looks better, and almost there. Do you think it's better to have the class names changed to `cut_name` as well? `S2Width` -> `CutS2Width`. So it will always be consistent with [strax/plugin.py#L668](https://github.com/AxFoundation/strax/blob/640ed96d5f1062e4e9c5c51cbe2aa7dee9959ef8/strax/plugin.py#L668).

As for point 2, you are right. A new strax issue would be more appropriate.

jorana:
Thanks Tianyu, you are right, too long ago already that I wrote that class.

PR #226, opened 2020-09-29
Include python3.8

jorana:
Like in https://github.com/AxFoundation/strax/pull/320, we can also test straxen for python 3.8 with travis. This has been shown to increase performance (see figure 5: https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:dsg:daq:eb_speed_tests).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ dist: xenial
 python:
 - '3.6'
 - '3.7'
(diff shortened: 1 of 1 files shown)


darrylmasson (approved):
Let's hope for improved speed and stability without unexpected changes

PR #231, opened 2020-10-08
Temporary LXe for 2e6 gain HV map. Cut-off: 1310V

GiovanniVolta:
New adc_to_pe map for the data acquired from October 8th, 2020.


straxen/contexts.py
```diff
@@ -22,7 +22,7 @@
     n_nveto_pmts=120,
     n_tpc_pmts=straxen.n_tpc_pmts,
     n_top_pmts=straxen.n_top_pmts,
-    gain_model=('to_pe_constant', 'TemporaryGXe_1500V_PMT116_1300_PMT195_1300'),
+    gain_model=('to_pe_constant', 'TemporaryGXe_1500V_PMT116_1300_PMT195_1300', 'TemporaryLXe_2e6HVmap_cutoff_1310'),
```
jorana, line 25: Sadly it does not work likt this, you can specify just one in this 'to_pe_contstant' setting. There are other options but I don't think it makes too much sense to exploit them as we hopefully have a CMT soon.

jorana (changes requested):
@GiovanniVolta , thanks for adding this. @skazama can comment if we actually want all of the data we have taken so far to use this gain model but I suspect not. It would be only these specific runs you want a different gain for. It's clear we need something like a CMT soon.

jorana (approved):
Thanks Giovanni, lets not update the context for now with these gains but do add it to straxen.

Issue #233, opened 2020-10-08
Remove HE plugins from 1T context


jorana:
@SerenaDiPede pointed out the command below does not work. The reason is that the He Plugins should not be in the 1T context. 

```python
st=straxen.contexts.xenon1t_dali()
st.search_field('cs1') 
```

PR #236, opened 2020-10-11
fix issue 233

jorana:
See https://github.com/XENONnT/straxen/issues/233

The problem is that there were plugins in the 1T contexts that were depending on data-kinds that were not provided by the RecordsFromPax.

I considered fixing the `register_all` in the common_opts and registering all the 'nT' datakinds by hand (as we do with the DAQreader). However, as we need to do this anyway because of the High Energy plugins, I opted for removing it from the 1T contexts.

**Edit 12-10**
In the end I decided it's cleaner to do d5e2007 where I split out the nT only files like the neutron-veto associated plugins. I do keep the function `strip_nt_plugins` as the peaks/peakltes/pulses-processing are in the same files as their High Energy counterparts. Here I now rely on our naming convention to keep the plugins with `provides == *_he` out of the 1T contexts.

### DIFF
--- a/straxen/contexts.py
+++ b/straxen/contexts.py
@@ -5,11 +5,20 @@ import numpy as np
 
 common_opts = dict(
     register_all=[
-        straxen.pulse_processing,
-        straxen.peaklet_processing,
-        straxen.peak_processing,
+        # straxen.pulse_processing,
+        # straxen.peaklet_processing,
+        # straxen.peak_processing,
         straxen.event_processing,
         straxen.double_scatter],
+    register=[
+        straxen.PulseProcessing,
+        straxen.Peaklets,
+        straxen.PeakletClassification,
+        straxen.MergedS2s,
+        straxen.Peaks,
+        straxen.PeakBasics,
+        straxen.PeakPositions,
+        straxen.PeakProximity],
     check_available=('raw_records', 'peak_basics'),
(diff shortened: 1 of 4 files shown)


WenzDaniel (commented):
Hmm I see the problem and how this change solves the issue, but I am not 100 % sure if I like the fact, that we first register all plugins within a file and then strip unneeded plugins off. This makes things more ugly, we could use `register` instead of `register_all` as alternative. This would be a bit more typing work, but more explicit.

jorana:
@WenzDaniel did you look at the last commit? I changed it to only remove the HE plugins.

WenzDaniel:
> 
> 
> @WenzDaniel did you look at the last commit? I changed it to only remove the HE plugins.

Yes I saw it. I think in general it is fine, I can just hear very silent the Python-Zen in the back of my head which is complaining :D

jorana:
My guess the cleanest way would be to remove the He plugins from e.g. pulse_processing.py but that reduces readability.

jorana:
@WenzDaniel , sorry, my zennes was also being affected. How would you like it now? I also added a small test ~2 s to test all the contexts for funny behaviour in the future.

WenzDaniel (approved):
Looks perfect now. I also like the test. Thanks

PR #237, opened 2020-10-12
Early commisioning context

jorana:
After having the discussion here about [the gains for the coming weeks ](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:analysis:commissioning:telecons:20201009:gains_stragegy)we need to update the main context to the right values. 
These gains are different from the gains we have been using for the GXe phase where the gains were dictated by the setting of the voltages to ~1500 V. As these two sets of data require different settings, we want to split these contexts. 

@GiovanniVolta will be providing the right gains for the new runs when available.


GiovanniVolta:
Using the information from PMT calibration of October 12th, I have updated the LXe gain model and included it in the main context.
I have also changed the name, removing the _temporary_.

straxen/contexts.py
```diff
@@ -74,6 +76,7 @@ def xenonnt_online(output_folder='./strax_data',
 
     st.storage = [straxen.RunDB(
         readonly=not we_are_the_daq,
+        minimum_run_number=_minimum_run_number,
```
WenzDaniel, line 79: One question will this affect also data which is stored on disk? If I use this context and I ask for a run before 9271, the only protection the user has is the "forbid_creation_of"-setting is not it? Otherwise I would just make the new data without knowing it.

How about a user warning which pops up when starting the context. We can remove this warning with the next release again.

WenzDaniel (approved):

petergaemers (approved):
Looks good Joran

straxen/contexts.py
jorana: Good question Daniel, it is jus the st.select_runs() where you scan the database that will be affected. It's not a problem at all to load old runs (if the lineage hash matches).

PR #238, opened 2020-10-12
add min-gap for acqmon-processing

jorana:
The acqmon processing used the `find_peak_groups` and used a `max_gap_threshold` for that. However the function assumes it's a minimum gap threshold leading to finding improperly defined peak-durations. Therefore, we add an extra option to the plugin. This value should be at least 1 ns otherwise the extensions don’t work.

@AlexElykov sorry for the PEP8-stuff pycharm sometimes is a little overenthusiastic.


straxen/plugins/acqmon_processing.py
```diff
@@ -77,56 +76,57 @@ def infer_dtype(self):
...
         return res
-    
+
     @staticmethod
     def merge_vetos(channels, gap, dtype, t):
         if len(channels):
-            start, stop = strax.find_peak_groups(channels, gap_threshold = gap)
-            result = np.zeros(len(start),dtype=dtype)
-            result['time'] = start 
+            start, stop = strax.find_peak_groups(channels, gap_threshold=gap)
```
jorana, line 121: @AlexElykov this was the culprit

AlexElykov:
Thanks! I was about to start an issue :).

straxen/plugins/acqmon_processing.py
AlexElykov: Yeah I 've seen it from the bootstrax error message in the run json, but had no time to look at it proper. I will check this update later today on some benchmark runs and see maybe I can spot some more bugs in my code.

straxen/plugins/acqmon_processing.py
```diff
@@ -6,69 +6,68 @@
...
 # The start/stop signals for the HEV are generated by the V1495 board
-  
+
 @export
 @strax.takes_config(
-    strax.Option('max_veto_gap', default = int(5e8),
-                 help = 'Maximum separation between veto stop and start pulses [ns]'),
-    strax.Option('channel_map', track = False, type = immutabledict, 
-                 help = 'immutabledict mapping subdetector to (min, max)'
+    strax.Option('min_veto_gap', default=1, type=int,
```
AlexElykov, line 54: I've checked the plugin on some recent runs and seems to find the correct deadtimes. I would request one change though. Before you merge, can you please set the min_veto_gap  =  int(1e6) in the default.

PR #241, opened 2020-10-12
fix_mongo4.4_projections

jorana:
Mongo 4.4 causes issues with the projections being used in straxen. Therefore this PR replicates the way https://github.com/AxFoundation/strax/pull/325 altered the projections needed in the RunDB class.

Please review carefully before merging.


straxen/contexts.py
```diff
@@ -19,7 +19,7 @@
         straxen.PeakProximity],
     check_available=('raw_records', 'peak_basics'),
     store_run_fields=(
-        'name', 'number', 'tags.name',
```
jmosbacher, line 22: Is it necessary to remove this if we are fixing the rundb frontend to be able to handle it? or is this just outdated and being removed anyway?

straxen/rundb.py
```diff
@@ -266,9 +266,13 @@ def _scan_runs(self, store_fields):
             query = {'number': {'$gt': self.minimum_run_number}}
         else:
             query = {}
+        projection = strax.to_str_tuple(list(store_fields))
+        projection = [f1 for f1 in projection
```
jmosbacher, line 270: Can we change this to: projection = [f1 for f1 in projection
                     if not any([f2.startswith(f1+".") for f2 in projection])]?
That would cover the edge case of two fields that just start with the same characters but are not related.

jmosbacher (commented):
I think it is correct to make this change here instead of of in strax. Not sure about removing the fields and also would like a small implementation fix.

straxen/contexts.py
jorana: It's hardcoded here:
https://github.com/AxFoundation/strax/blob/master/strax/run_selection.py#L80

So we don't need this one, the correct info is stored under 'tags'

straxen/rundb.py
jorana: Yeah good catch, let me add it

jorana:
> I think it is correct to make this change here instead of of in strax. Not sure about removing the fields and also would like a small implementation fix.

Ah yeah, I think you are right. It should be in 2fea669 , right?

jmosbacher (approved):
Looks good to me :)

Issue #227, opened 2020-10-01
Rucio does not check options


petergaemers:
Issue:

- Loading data from the rucio catalogue does not check if the versions and options match the strax configuration.

More specifically, the function ```st.is_stored(run_id,data_type)``` will always return true if there is any data of that data_type and run in the rucio database. The lineage is not checked.

Example:
```
import straxen
st = straxen.contexts.xenonnt_online()
st.set_config(dict(s1_max_rise_time=50000,
                  peak_split_filter_wing_width=1000000))
st.is_stored('009149','peak_basics')
st.get_array('009149','peak_basics')
```
 
The data belonging to this configuration does not exist. However since is_stored will return true anyway, get_array will retrieve the data from rucio despite it having different settings

jorana:
Fixed in  #242

PR #242, opened 2020-10-12
check lineage hash for rucio

jorana:
Fix rucio backend to check for the lineage hash. Fix https://github.com/XENONnT/straxen/issues/227

We were not looking for the lineage hash. As such, rucio was eager to claim to have data it did not have.

```
import straxen
st = straxen.contexts.xenonnt_online()
print(st.is_stored('009149','peak_basics'))
st.set_config(dict(s1_max_rise_time=50000,
                  peak_split_filter_wing_width=1000000))
print(st.is_stored('009149','peak_basics'))
```

### DIFF
--- a/straxen/rundb.py
+++ b/straxen/rundb.py
@@ -165,17 +165,25 @@ class RunDB(strax.StorageFrontend):
             dq = {
                 'data': {
                     '$elemMatch': {
+                        # TODO can we query smart on the lineage_hash?
                         'type': key.data_type,
                         'protocol': 'rucio'}}}
             doc = self.collection.find_one({**run_query, **dq},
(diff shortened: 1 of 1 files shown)


WenzDaniel (approved):
Very good catch, thanks a lot. Not sure how large this extension would be, but we are still not able to use the "fuzzy_for" option with the runDB. I think this will become more important soon, so maybe it is a good time to implement it.

jorana:
Thanks Daniel, it's on the to do list, let me make an explicit issue: https://github.com/XENONnT/straxen/issues/243

As it's not per se the scope of this PR, I'm merging this one now and open a new PR whenever I've time to fix #243 .

PR #244, opened 2020-10-12
Saturation correction (from pax)

zhut19:
**Correct for saturation in peaklets**

Right after peaklets are built, we take a look at those containing saturated records. Replacing part of the 'bad' raw records with the summed waveform of the non-saturated channels. The cartoon illustration shows the effect of the `peak_saturation_correction` function, replacing the correction region black with red, thus the peaklet can be reconstructed with reduced bias.

desat_example


straxen/plugins/peaklet_processing.py
```diff
@@ -207,6 +225,141 @@ def clip_peaklet_times(peaklets, start, end):
...
+    if not len(records):
+        return
+    if not len(peaks):
+        return
+
+    # Search for peaks with saturated channels
+    mask = peaks['n_saturated_channels'] > 0
+    if use_classification:
+        mask &= peaks['type'] == 2
+    peak_list = np.where(mask)[0]
```
WenzDaniel, line 253: FYI, this is interesting first I was thinking why you are not directly using something like `peaks[mask]` instead of your `peaks[np.where(mask)[0]]`. It turned out that the latter is about 10 % faster.

straxen/plugins/peaklet_processing.py
```diff
@@ -207,6 +225,141 @@ def clip_peaklet_times(peaklets, start, end):
...
+    record_ranges = _touching_windows(
+        records['time'],
+        strax.endtime(records),
+        peaks[peak_list]['time'],
+        strax.endtime(peaks[peak_list]))
+
+    # Create temporary arrays for calculation
+    dt = records[0]['dt']
+    n_channels = len(peaks[0]['saturated_channel'])
+    len_buffer = np.max(peaks['length'] * peaks['dt']) // dt
```
WenzDaniel, line 264: The buffer length should be one sample longer is not it? E.g. `length = 1` and `dt_peak = 0.2` would lead to `1*0.2//1 = 0`

straxen/plugins/peaklet_processing.py
```diff
@@ -146,6 +155,15 @@ def compute(self, records, start, end):
             min_area=self.config['peak_split_min_area'],
             do_iterations=self.config['peak_split_iterations'])
 
+        # Saturation correction using non-saturated channels
+        # similar method used in pax
+        # see https://github.com/XENON1T/pax/pull/712
+        if self.config['saturation_correction_on']:
+            corrected_peaklet_list = peak_saturation_correction(
```
WenzDaniel, line 162: I am a bit confused, it looks like that you are modifying the peaks in place, also the return is not used anywhere. So why do you define it in the first place? Will this not just occupy memory?

WenzDaniel (commented):
Hej, nice code. I have not had time to look through it in detail, but here are some minor comments of things I noticed while skimming through your code.

straxen/plugins/peaklet_processing.py
jorana: This should indeed go in the compute function

straxen/plugins/peaklet_processing.py
```diff
@@ -207,6 +225,141 @@ def clip_peaklet_times(peaklets, start, end):
                 p['length'] = (end - p['time']) // p['dt']
 
 
+@numba.jit(nopython=True, nogil=True, cache=True)
+def peak_saturation_correction(records, peaks, to_pe,
```
jorana, line 229: I'd say this would be better in strax right, it's quite a general thing

straxen/plugins/peaklet_processing.py
```diff
@@ -207,6 +225,141 @@ def clip_peaklet_times(peaklets, start, end):
...
+                               use_classification=False,
+                               ):
+    """Correct the area and per pmt area of peaks from saturation
+    :param records: Records
+    :param peaks: Peaklets / Peaks
+    :param reference_length: Maximum number of reference sample used
+    to correct saturated samples
+    :param min_reference_length: Minimum number of reference sample used
+    to correct saturated samples
+    :param use_classification: Choice to use peak type
```
jorana, line 241: perhaps clarify to `only use S2-types`

straxen/plugins/peaklet_processing.py
```diff
@@ -207,6 +225,141 @@ def clip_peaklet_times(peaklets, start, end):
...
+    """
+
+    if not len(records):
+        return
```
jorana, line 245: is this None the right dtype?

straxen/plugins/peaklet_processing.py
```diff
@@ -207,6 +225,141 @@ def clip_peaklet_times(peaklets, start, end):
...
+        records['time'],
+        strax.endtime(records),
+        peaks[peak_list]['time'],
+        strax.endtime(peaks[peak_list]))
+
+    # Create temporary arrays for calculation
+    dt = records[0]['dt']
+    n_channels = len(peaks[0]['saturated_channel'])
+    len_buffer = np.max(peaks['length'] * peaks['dt']) // dt
+    max_nrecord = len_buffer // strax.DEFAULT_RECORD_LENGTH + 1
```
jorana, line 265: This is tricky, it's like hardcoding `dt = 10`. Please do
 ```python
rec_length =len records[0]['data'])
max_nrecord = len_buffer // rec_length + 1
```

File: straxen/plugins/peaklet_processing.py [hunk 8]
```diff
@@ -207,6 +225,141 @@ def clip_peaklet_times(peaklets, start, end):
...
+
+    # Buff the sum wf [pe] of non-saturated channels
+    b_sumwf = np.zeros(len_buffer, dtype=np.float32)
+    # Buff the records data [ADC] in saturated channels
+    b_pulse = np.zeros((n_channels, len_buffer), dtype=np.int16)
+    # Buff the corresponding record index of saturated channels
+    b_index = np.zeros((n_channels, len_buffer), dtype=np.int32)
```
jorana, line 272: perhaps small but doesn't np.int32 kind of overdo the max index of let's say ~500?

jorana (changes requested):
Hi Tianyu for adding this. I think we need to very carefully look at where this is used especially given the fact that we might be computing the sumwf/build peak properties in several places. For instance if we mergeS2s, could the saturation correction in peaklets might get undone if we don't set to the p['n_saturated_channels']  to 0 (I think we should anyway store somewhere the info that this has been desaturated). 

Also I think that some of these tools migth be usefull for strax too. 

Perhaps it's good to provide some exampary peaklets and check that it works correctly, e.g. by comparing the wf with the HighEnergy-data in that time-range.

straxen/plugins/peaklet_processing.py
zhut19: Right, it is safer to +1. I am under the assumption that dt_peak is multiple of dt_records.

straxen/plugins/peaklet_processing.py
zhut19: The return is a relic from when the code wasn't optimized. (When I was doing sum waveform outside the correction function). Will just delete the return

straxen/plugins/peaklet_processing.py
zhut19: Right, I remember being surprised finding a bug due to using int16, and changed to int32. Maybe switching both to int64 just to be safe.

straxen/plugins/peaklet_processing.py
```diff
@@ -207,6 +225,141 @@ def clip_peaklet_times(peaklets, start, end):
...
+
+    # Main
+    for ix, peak_i in enumerate(peak_list):
+        b_sumwf[:] = 0
+        b_pulse[:] = 0
+        b_index[:] = -1
```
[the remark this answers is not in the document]
zhut19: Hmm, I am not sure the np.ones will undo the buffer? I was hoping to just keep one set of buffer arrays, they don't need to be recreated in each loop. And reassigning constants operation should be much faster.

straxen/plugins/peaklet_processing.py
zhut19: Interesting insight, my intention was just for that deprecated return at the end. But 10% faster is a bonus?

straxen/plugins/peaklet_processing.py
zhut19: I am in favor of this being part of straxen. As this seems to be a feature, specific to XENON detectors. It's hard to say this method can be universally applied to other non-XENON ones.

straxen/plugins/peaklet_processing.py
zhut19: Now it should be in https://github.com/XENONnT/straxen/pull/244/commits/2198316c7b312165cd2e21a584d14e950a0ed23d. With the last return removed

zhut19:
Thank you, @jorana @WenzDaniel, for the review! I made some minor changes. Let me know if those are good. I also checked again for the peak classification; unfortunately, to the disappointment of @feigaodm, even with the gap-size clustering happening before this step, classification comes later.

straxen/plugins/peaklet_processing.py
jorana: I'm not so sure, most experiments will have ADCs. E.g. a small scale TPC like XAMS or XEBRA will both potentially also see ADC-saturation as the hardware is similar to the nT experiment.

jorana (approved):
Thanks a lot Tianyu, as far as I can see it looks good!.

@feigaodm @skazama we could merge this PR as far as I'm concerned. We would have changing lineage-hashes for peaklets and above as well as peaklets_he and above.

skazama:
@zhut19 @jorana @WenzDaniel Thanks all for your efforts to implement the saturation correction. I also think that we can merge this PR.

feigaodm:
@zhut19 is there a way that you can apply peak classification in peaklet so that you can fully avoid correct S1s?

straxen/plugins/peaklet_processing.py
```diff
@@ -266,13 +267,14 @@ def peak_saturation_correction(records, peaks, to_pe,
 
     # Buff the sum wf [pe] of non-saturated channels
     b_sumwf = np.zeros(len_buffer, dtype=np.float32)
-    # Buff the records data [ADC] in saturated channels
-    b_pulse = np.zeros((n_channels, len_buffer), dtype=np.int16)
+    # Buff the records 'data' [ADC] in saturated channels
+    b_pulse = np.zeros((n_channels, max_nrecord), dtype=np.int16)
```
zhut19, line 271: @jorana I am not sure what do you mean by changing from len_buffer to max_nrecords. The b_pulse is supposed to save the whole pulse with 10ns resolution.

jorana: Sorry Tianyo, my bad, should have been the line below: https://github.com/XENONnT/straxen/pull/244/commits/9863e2b6ddf9a525b074ebcf3fc81c7f93593876

zhut19:
@feigaodm, sorry I haven't found a way to apply peak classification. I got stuck at getting the context inside the Peaklet plugin, thus making it impossible to call the `compute` of `PeakletClassification` during the desaturation step.

straxen/plugins/peaklet_processing.py
```diff
@@ -207,6 +225,143 @@ def clip_peaklet_times(peaklets, start, end):
...
+
+
+@numba.jit(nopython=True, nogil=True, cache=True)
+def _peak_saturation_correction_inner(channel_saturated, records, p,
+                                      b_sumwf, b_pulse, b_index,
+                                      reference_length=100,
+                                      min_reference_length=20,
+                                      ):
+    """Would add a third level loop in peak_saturation_correction
+    Which is not ideal for numba, thus this function is written
```
WenzDaniel, line 318: Just out of curiosity do you have a reference for this? Why is not it ideal?

WenzDaniel (approved):
Sorry, for the delay busy days. Thanks for this nice addition. I do not really have something to add. The only thing I would like to mention I am also slightly in favor to move this into strax since as Joran pointed out this is a general digitizer issue which also other TPCs may encounter.

jorana:
@feigaodm just approved this PR, I'm going to merge it as soon as travis finishes

straxen/plugins/peaklet_processing.py
zhut19: Sorry, Daniel, I tried to find something to support my argument but failed. Now I am not sure if I really read this somewhere. I was running into issues that numba failing to recognize value type in the deep loops but perfectly capable of doing so if I move that loop into a separate function.

PR #246, opened 2020-10-15
First commissioning phase context

jorana:
After having the discussion here about [the gains for the coming weeks ](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:analysis:commissioning:telecons:20201009:gains_stragegy)we need to update the main context to the right values. 
These gains are different from the gains we have been using for the first phase where the gains were dictated by the setting of the voltages to ~1500 V. As these two sets of data require different settings, we want to split these contexts. 

@GiovanniVolta added the gains and @WenzDaniel provided a review.


straxen/contexts.py
```diff
@@ -59,7 +60,9 @@
 
 def xenonnt_online(output_folder='./strax_data',
                    we_are_the_daq=False,
+                   _minimum_run_number=9271,
```
jorana, line 63: @WenzDaniel , this would only affect the `st.select_runs` you can still use st.get_array if the data is stored (with the right lineage) in one of the frontends.

PR #251, opened 2020-10-18
Simplify rundb init for CMT

jorana:
Sligthly simplify the rundb init used in the CMT ( #220) using the same checks as for the RunDB.

9c973a3 also fixed a circular import that I had created when merging #220 into master.


WenzDaniel (approved):
Looks good

PR #253, opened 2020-10-19
Fix devide by zero for zero-area peaks

jorana:
Dividing by 0 is a bad idea. We did not check if zero area peaks. We should, therefore only compute center-times (division by peak['area']) only for positive area peaks.


darrylmasson (approved):
Pretty straightforward fix

PR #257, opened 2020-10-20
add online-monitor plugin for area vs width

jorana:
There is a request for online monitoring.  This PR provides the user with the data to very easily make area vs widht plots. See the demo here for an example:
https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:angevaare:onlinemonitor#demo

Make the output in a histogram, write that to the mongo database

https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:angevaare:onlinemonitor#demo


WenzDaniel (approved):
From a technical point of view I think it is fine, though I do not understand the logic of the online monitoring yet. This is what I think I understood so far:

* You produce this data on one of our ebs and store it temporarily in our mongoDB? Somehow the storing is done before the run has finished yet. 
* Then you load the data on dali to make this nice self updating plot via interactive matplotlib plots. 

Here is my question: peak_basics is so small just a view MB would not it be possible to do the same without this plugin? Or are there some technical details I have not understood yet.

jorana:
As to your question Daniel, please see:
https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:angevaare:onlinemonitor

jorana:
@feigaodm , @skazama , are there any requests for this tool? At the moment it looks like the lone hit-rate over time but are there any others? 

Please keep in mind that the data has to be compact, I can make histograms but either using predefined cuts or none at all. As such, I like keeping the aggregation per chunk as you know you cannot make larger documents as the length is fixed to 1 data-field per chunk.

If there are no more requests at the moment, I can merge this PR and add new features later too.

WenzDaniel:
>  I can make histograms but either using predefined cuts or none at all. As such, I like keeping the aggregation per chunk as you know you cannot make larger documents as the length is fixed to 1 data-field per chunk.

Maybe we can implement some default cuts on the following parameters:

* n_channels
* area_fraction_top
* rise_time
* tight_coincidence

and

* type

This gives the analyzers alot of freedom for the online monitoring. The first four option would look by default like this ```
Option(field_name (None, None)`
```
and in the code:
```
m = np.zeros(len(data), dtype=np.bool_) 
for option in options:
    m &= data > ... 
``` 
 so that one can specify and lower and upper bound. The last one would then be invoked in an additional step.

jorana:
Hi Daniel! This unfortunately does not work! If you change the options, the user won't be able to change the config that is running on the eventbuilder. Only when we somehow add all this info we can do cuts but that then again implies the data would be too big. In fact this requires all those properties to be shipped individually. 

Of course I can add it as an option and add the implementation as we do here (https://github.com/AxFoundation/strax/blob/master/strax/context.py#L1032) but the effect is not as you describe.

WenzDaniel:
Ah I see. Not sure I am just thinking out of the blue, but could we ship the context itself via mongo?

jorana:
>Ah I see. Not sure I am just thinking out of the blue, but could we ship the context itself via mongo?
Oh for sure we can, but I don't want non-experts touching anything going on with the live-processing.

WenzDaniel:
I agree on that, but could not we have a second "online_monitoring_context" which runs the required processes?

feigaodm:
Hi Joran, thanks for adding this feature, I think it's important. At different stage, we might need to add different features. For example, during the cathode light emission test, we would add the lone-hit rate. For Kr83m, we may add Kr83 double-S1 peaks in addition to the standard peak width vs are. But I believe this example is helpful for analyzers to develop and contribute.

PR #260, opened 2020-10-22
Option to randomly override a hard veto

sophiaandaloro:
Let's try this again:) 
  * Adds an additional setting to the Software HE Veto step in pulse_processing, requested for certain commissioning & calibration applications, for us to optionally pass some random (configurable) fraction of peaks that are otherwise vetoed. This allows us to keep statistics rather than a hard cutoff at a certain PE.

  * Set the `'fractional_pass'` configuration option to non-zero (between 0 and 1) and that fraction of peaks which are vetoed in the current configuration will randomly get selected to pass anyway. 

See this [note](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:andaloro:fast_response:lxe_s1_shev_update) which uses it, especially in the plots of the peaks. (Plots not included here since the data may be sensitive at the moment)

  - Update the docstring(s) --> Done.
  - Update the documentation --> Done.
  - Tests to check the (new) code is working as desired. --> Done in study (linked above)


sophiaandaloro:
Hopefully this PR works. No rush to get it merged in the next day or anything so leave until it's convenient for you.

straxen/plugins/pulse_processing.py
```diff
@@ -250,18 +257,22 @@ def compute(self, raw_records_he, start, end):
         return dict(records_he=result['records'],
                     pulse_counts_he=result['pulse_counts'])
 
+
 ##
 # Software HE Veto
 ##
 
+# Reproducible random state for HE Veto function
+prng = np.random.RandomState(123)
```
jorana, line 266: Does it make sense to put this inside the function or did you put it here on purpose? Otherwise when you do straxen.software_he_veto(records) you still have a different result than from the processing. 

What do you think?

jorana (commented):
In principle the code looks good, please have a look at some of these comments/suggestions.

I'm fine with merging it from a software perspective, it's just that we need to discuss with the ACs/computing when is the best moment.

jorana:
@feigaodm , can you comment if this is needed in the Kr-calibration? If so, we should also active it in this PR.

The effect will be that all the `records` would become incompatible with the straxen-master unless we reprocess (I am not sure what makes most sense). 

@sophiaandaloro , did you have time to look at my comments?

tunnell:
For ACs, please comment in Slack instead of here.  This PR is just for the code.

jorana:
> For ACs, please comment in Slack instead of here. This PR is just for the code.

For bookkeeping: 
https://xenonnt.slack.com/archives/C01772B62JD/p1604411750343300

sophiaandaloro:
Hey Joran, yes I see your comments, they are all good and worth changing, if we decide to go this route but should ask ACs probably before we spend time fixing it up, in case we decide to scrap this idea. Once we reach a consensus on whether to do this, or to just change the settings as-is, I will go ahead and make a new PR. Is that alright ?

jorana:
> Hey Joran, yes I see your comments, they are all good and worth changing, if we decide to go this route but should ask ACs probably before we spend time fixing it up, in case we decide to scrap this idea. Once we reach a consensus on whether to do this, or to just change the settings as-is, I will go ahead and make a new PR. Is that alright ?

I think my comments on the technical implementation are minor. For changing the option as it is, I would open a new PR. For this route you can either add the values to this PR or to a different one, what you judge to be best.

sophiaandaloro:
New minor PR made at #266, if we decide we still need the fractional pass option to randomly pass otherwise vetoed peaks, I can implement technical changes and make a new PR, but given what we said above, since this will only be a feature used once in the lifetime of nT, probably best to avoid if we can.

jorana:
temporary converting this pr to `draft` until an AC decides on this PR

jorana:
@WenzDaniel , same as #260 . Although I must admit I like this PR better than #266 as the arbitrariness is smaller where you put a boundary

jorana:
@WenzDaniel , same as #260 . Although I must admit I like this PR better than #266 as the arbitrariness is smaller where you put a boundary

jorana:
@sophiaandaloro can we close this one as #266 is merged?

Issue #247, opened 2020-10-15
Documentation of the XENON1T cuts


HenningSE:
We could add a few lines of documentation on how to use the cuts cuts given in the x1t_cuts.py plugin.

PR #261, opened 2020-10-30
Do save peaks

jorana:
Fix long outstanding issue https://github.com/XENONnT/straxen/issues/149

Don't be fooled by the simple appearance of this PR, it's been keeping me awake at night.

So peaks is never saved, therefore the time-selection in the peak-plotting is tricky. Each of the plugins may have info that are distributed among asynchronous chunks. Therefore the simple operation of making a wf plot becomes hard. I therefore propose to remove this, there is storage-wise little sound argumentation for this as we are already 10 x smaller at peaks than at peaklets.

I think we should save our self the trouble and just store it.

**How to replicate**
I just ran the wf plot as in #149 just untill I hit this error, run the following two commands and you will see that you are now able to actually plot the wf:
```python
straxen.plugins.Peaks.save_when = strax.SaveWhen.ALWAYS 
st.make(run_id, 'peaks')
# where time_range=(9343216270, 9343224420) is a time range in simulated data 
# where from I figured out that I could run into this bug
st.plot_peaks(run_id,time_range=(9343216270, 9343224420)) 
```

### DIFF
--- a/straxen/plugins/peaklet_processing.py
+++ b/straxen/plugins/peaklet_processing.py
@@ -437,14 +437,12 @@ class MergedS2sHighEnergy(MergedS2s):
 class Peaks(strax.Plugin):
     """
     Merge peaklets and merged S2s such that we obtain our peaks
-    (replacing all peaklets that were later re-merged as S2s). As this
-    step is computationally trivial, never save this plugin.
+    (replacing all peaklets that were later re-merged as S2s).
     """
     depends_on = ('peaklets', 'peaklet_classification', 'merged_s2s')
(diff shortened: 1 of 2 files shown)


JelleAalbers:
Hi Joran, you're right, this will simplify and speed up several things. However, I hoped the peak-level data could be stored at midway. If you double the data size (since peaks are now stored both in peaks and in peaklets/merged_s2s) that may no longer be feasible. I think #247 has another fix (removing a check that does not make sense in the current chunking system when selecting time ranges), so maybe this is only needed as an interim solution.

Issue #255, opened 2020-10-20
Nans returned (if returned by CMT) cause errors


jorana:
Having Nans in to_pe will cause errors in peak processing causing up to 100x more peaks.

I think this is due to typecasting. Also it's bad if we run into this as apparently there would be information missing in the corrections-collections. As such we need to raise a value error when we get but a single None from the CMT. Otherwise we are going to have irreproducible errors/divergences later

PR #262, opened 2020-10-30
Fix strax updates into straxen and fix issues

jorana:
8dffccb  Fix deprecation issues due to https://github.com/AxFoundation/strax/pull/318
a70b630 Remove S3 from straxen https://github.com/AxFoundation/strax/issues/307
cd9fd4c Double check Nones are not returned by CMT https://github.com/XENONnT/straxen/issues/255
2492e23 fix documentation on cut-plugins https://github.com/XENONnT/straxen/issues/247 

Fix some open issues that were still lingering.

### DIFF
--- a/straxen/corrections_services.py
+++ b/straxen/corrections_services.py
@@ -55,6 +55,9 @@ class CorrectionsManagementServices():
         else:
             self.collection = client['run']['runs_new']
 
+    def __str__(self):
+        return self.__repr__()
+
     def __repr__(self):
         return str(f'{"XENONnT " if self.is_nt else "XENON1T"}'
                    f'-Corrections_Management_Services')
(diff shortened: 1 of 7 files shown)


jorana:
@ahiguera-mx , do you agree with the additional check in cd9fd4c?

ahiguera-mx:
> @ahiguera-mx , do you agree with the additional check in [cd9fd4c](https://github.com/XENONnT/straxen/commit/cd9fd4c4d7b09346eee6812862812a2705e09a36)?

Hi @jorana, yes thanks for catching this, although that should never happened if the gain model is correct, but better to be safe than sorry

ahiguera-mx (approved):

straxen/corrections_services.py
```diff
@@ -167,7 +171,7 @@ def get_pmt_gains(self, run_id, model_type, global_version):
                                  f'{type(global_version)}')
 
             # Generate an array of values and multiply by the 'global_version'
-            to_pe = np.ones(n_tpc_pmts, dtype=np.float32) * global_version
+            to_pe = np.ones(n_tpc_pmts, dtype=gain_dtype) * global_version
```
WenzDaniel, line 174: Why can the dtype of the gains change?

WenzDaniel (approved):
Looks fine just one question to be answered

straxen/corrections_services.py
jorana: Not per se something that should be changed, therefore I made it a kwarg. Could of course also have it hardcoded in the function. I like it more than saying `np.ones(n_tpc_pmts, dtype=np.float32)` twice (also some lines down), I rather define it once.

PR #263, opened 2020-11-02
thread out queries for multiple corrections in CMT

jorana:
On some machines with no direct access to the corrections database, loading the PMT gains might take a while with the CMT.
This PR is inspired by @jmosbacher 's idea on slac.

For each of the values that CMT  loads that is cast into an array, make a different tread to load the value from the correction.

Below I give the exemplary speed improvement I achieve on my own laptop.
```python
corrections = straxen.CorrectionsManagementServices(is_nt=True)
def get_to_pe(run_id, gain_model = ("CMT_model", ("to_pe_model", "ONLINE")), ntpc_pmts = 494):
    model_type, model_conf = gain_model
    to_pe = corrections.get_corrections_config(run_id, 'pmt_gains', model_conf)
    return to_pe
%%time
get_to_pe('008658')
```
Takes before: Wall time: Wall time: 3min 13s
Takes after this PR: Wall time: 8.82 s

_Edit_
On dali the time difference is much smaller:
Wall time: 8.3 s (not-threaded) vs. Wall time: 5.01 s (threaded)

_Edit2_
I also tested it on the event builders (much more important than my laptop)
Wall time: 2min 19s (not-threaded) vs. Wall time: 14.1 s (threaded)

### DIFF
--- a/straxen/corrections_services.py
+++ b/straxen/corrections_services.py
@@ -5,6 +5,7 @@ import pymongo
 import numpy as np
 from warnings import warn
 from functools import lru_cache
+from threading import Thread
 
 import strax
 import straxen
@@ -83,7 +84,8 @@ class CorrectionsManagementServices():
     #  entry for e.g. for super runs
     # cache results, this would help when looking at the same gains
     @lru_cache(maxsize=None)
-    def _get_correction(self, run_id, correction, global_version):
+    def _get_correction(self, run_id, correction, global_version,
+                        correction_dtype=np.float64):
         """
         Smart logic to get correction from DB
         :param run_id: run id from runDB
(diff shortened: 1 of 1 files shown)


tunnell:
I think this is the wrong approach.  Threading code should be avoided when possible, and sort of opaque.  

I suggest we do the following:

https://gist.github.com/tunnell/58bd1b760d233c86e40e02eb5b3fa13a

Which will be faster, allows caching, and offloads this to Mongo.  It's also sort of the same format that CMT currently parses from Mongo, so shouldn't be a big change.  The previous gist was meant to be a proof of concept, but this one is a bit cleaner.

tunnell (changes requested):
See other comment

jorana:
Thanks for the feedback

PR #264, opened 2020-11-03
Use more realistic range for area vs width

jorana:
In the OM, we are using quite a large log-log area-width range. We don't expect many 1^6 PE ms long peaks. Therefore, drop one order of magnitude 

**Note to reviewer**
If you approve, please let me merge it such that I can also add it to the eventbuilders.


jorana:
Duplicate #265

PR #266, opened 2020-11-03
option 1 changing nT context for shev

sophiaandaloro:
Given discussion in [Slack](https://xenonnt.slack.com/archives/C01772B62JD/p1604411750343300) (sensitive info left out.) we are looking to change veto abilities for the current runs planned. This code simply hard-codes settings for nT configuration, where if you aren't looking at tomorrow's planned data, you'll have to just add a line to the top of your code such as `st.set_config(<disable hev>)` at the top of analysis. 

Threshold is set at 1500 PE for reasons discussed in TPC meeting (regarding upcoming runs), with a pass fraction of 1000, meaning virtually no peaks are passed once a high-energy S1 triggers the veto in place. This results in roughly a 6x factor of data reduction on previous runs, so will be slightly lower (though comparable) for upcoming runs.

_Edit by Joran_ link to note:
https://xenonnt.slack.com/archives/C01772B62JD/p1604505610388200?thread_ts=1604411750.343300&cid=C01772B62JD


straxen/contexts.py
```diff
@@ -96,7 +101,7 @@ def xenonnt_online(output_folder='./strax_data',
             st.storage.append(
                 strax.DataDirectory(output_folder))
 
-        st.context_config['forbid_creation_of'] = straxen.daqreader.DAQReader.provides + ('records',)
```
sophiaandaloro, line 99: Actually since pulse_processing is done with we_are_the_daq, I can keep this as-is, and this will prohibit people from accidentally reprocessing records even with the new settings.

straxen/contexts.py
```diff
@@ -27,6 +27,11 @@
     n_tpc_pmts=straxen.n_tpc_pmts,
     n_top_pmts=straxen.n_top_pmts,
     gain_model=('to_pe_constant', 'gain_2e6HVmap_cutoff_1310'),
+    tail_veto_threshold=1500.,  # PE
```
jorana, line 30: Thanks Sophia, in general all looks good as far as I'm concerned. Will be an AC call (@feigaodm) when this will be merged.

_just adding the same comment for book keeping_
Could you move the config to `xenonnt_online`?

jorana (approved):

straxen/contexts.py
```diff
@@ -77,6 +78,13 @@ def xenonnt_online(output_folder='./strax_data',
     st.register_all(have_nT_plugins)
     st.register([straxen.DAQReader, straxen.LEDCalibration])
 
+    st.set_config({
```
jorana, line 81: Sorry I should have been more specific, the early commissioning context clones the config of the xenon_nt_online context. As such, could you also save the previous settings to that context in a similar fashion?

sophiaandaloro: Yep!

sophiaandaloro:
Hey I just specified the 'hev_gain_model' for each of the contexts based on the SHEV disabled if hev_gain_model(0)='disabled'. This should work... let me know if I am missing anything..

sophiaandaloro:
@feigaodm @skazama Hi Fei and Shingo. This is the PR for the Software HEV, which Joran reviewed for me. Old conflict due to it being a big outdated, but I will resolve that. 
Will you please review either here or (preferred:) on Slack, and let me know your thoughts? This could potentially save us >5x on storage of records onwards... 

Thanks!

jorana (changes requested):
Hi Sophia, I think something went wrong on git, do you need some help with git?

sophiaandaloro:
Hey Joran, I synced up with mater since there were new changes. Then had to resolve some conflicts. The goal now is to recreate a software HE veto function that can do what we discussed (not veto on alphas, as this PR currently does.) Let me know if the best course of action is to keep this open and add to this PR, or close the PR and reopen once the new function is built! Thanks!

sophiaandaloro:
I cannot tell why these builds are failing, it seems suddenly that none of the config settings are seen by the plugins. Not sure if this is an issue on the CMT side because I have only changed 1 file...

WenzDaniel:
Hi Sophia, the error report says the following: 

```
TypeError: software_he_veto() got an unexpected keyword argument 'pass_veto_extend'
```

maybe you removed this argument, but forgot it in the function call.

jorana:
You didn't update the `software_he_veto` - function. If you ever run into these issues, usually it's easier to understand than the travis logs.

Just go to straxen and type `pytest`  (you need some packages, just pip install until it stops complaining at import). Here is an example:
https://gist.github.com/jorana/0e128adff61677571cdac9d69af4ed8b

sophiaandaloro:
Oh wow, nice. Thanks. I see now. I didn't know we had this in straxen so this helps a lot.

So here are our options, this should work same as before when pulled in, unless we do one of the following for processing: 
1. I can create an entirely new context something like xenonnt_vetoed or something with the settings shown that result in the plots on the note [here](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:andaloro:20201117:lxe_s1_veto). Then users would just have to load the context up by name and it'll work. 
2. I can create a block of python code where I register the SHEV settings that result in the reduced data. Then can share with analysts and encourage them to use the settings. If they don't, they'll just default back to the non-vetoed data as we're copying peaks from both for now.

In either case, let me know which makes more sense for maintenance and transitioning between SHEV settings throughout commissioning... I'd imagine more contexts is not ideal.
Also let me know what you need me to do to get newly processed/vetoed data locally. If you just need the same context construction as dali users, I'll just paste it into a gist and share.]

Thanks again :)

jorana:
Thanks Sophia, I think @feigaodm can best answer.

Having two context seems out of place. To me it sounds like if you activate it, you want to have the option everywhere. Otherwise people start doing reprocessing records on their own which is a bad idea.

In order to add it to the main context, we'd have to reprocess records since it's lineage changing even if the option itself is not activated.

sophiaandaloro:
So pending Fei's approval, would the ideal situation to be using the same context but with different config settings? Can we introduce a min_run_number to ensure we don't reprocess previous data before the last week or so of runs?

This seems easiest for users, but not sure in practice if this makes life harder for you on the maintenance and processing side.

jorana:
temporary converting this pr to `draft` until an AC decides on this PR

jorana:
Since this option is not activated by default I think we can merge it together with #317 without much problem. @WenzDaniel do you agree?

WenzDaniel:
Yes I agree. Thanks for your work Sophia.

PR #267, opened 2020-11-04
Update monitor

jorana:
More updates on preliminary plots

### DIFF
--- a/straxen/plugins/online_monitor.py
+++ b/straxen/plugins/online_monitor.py
@@ -235,7 +235,7 @@ class OnlineVetoMonitor(strax.Plugin):
     depends_on = 'veto_intervals'
     provides = 'online_veto_monitor'
     __version__ = '0.0.1'
-    data_kind = 'online_monitor'
+    # data_kind = 'online_monitor'
 
     def infer_dtype(self):
         dtype = [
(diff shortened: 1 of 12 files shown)


straxen/plugins/online_monitor.py
```diff
@@ -228,50 +229,30 @@ def area_width_hist(self, data):
...
-        if end-start > 0:
-            res['live_time'] = 1 - np.sum(veto_intervals['veto_interval']) / (end-start)
+    def compute_loop(self, peaks, veto_intervals):
+        res = {}
+        for d in peaks.dtype.names:
+            res[d] = peaks[d]
+        dt = strax.endtime(peaks) - peaks['time']
+        assert not np.iterable(dt) or len(dt) == 1
+        if dt > 0:
+            res['live_time'] = 1 - np.sum(veto_intervals['veto_interval'])/dt
```
AlexElykov, line 255: Just a small comment. For normal operation this is correct; however, in Veto TRG mode the np.sum(veto_intervals['veto_interval']) will actually be the live-time (i.e. veto off period). That's just how I made the start and stop signals in this mode. I don't think anything needs to be changed, just don't get surprised if you get the live-time reversed.

jorana: Yeah, thanks Alex, you mentioned this. I thought it would make most sense to write it here like we expect it to be used. We can always change the interpretation on the analysts-side.

AlexElykov: Yeah I agree, hopefully we don't run with this Veto TRG mode for long :)

AlexElykov (approved):
Thank you for doing this!

PR #268, opened 2020-11-06
use gw at the DAQ

jorana:
Use gw as the host for mongo on the DAQ.

Also see the discussion here:
https://xenonnt.slack.com/archives/C016UJZ090B/p1604669100226800

### DIFF
--- a/bin/ajax
+++ b/bin/ajax
@@ -15,7 +15,7 @@ that may delete the live_data such that we prevent multiple attempts to perform 
 action.
 """
 
-__version__ = '0.2.5'
+__version__ = '0.2.6'
 
 import argparse
 from datetime import datetime, timedelta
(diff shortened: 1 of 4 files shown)


darrylmasson (approved):
I'm seeing real benefit in a `config.ini` or equivalent so we don't end up version-bumping for "trivial" changes like database addresses.

jorana:
yeah I guess we are waiting until Evan labels this one ready https://github.com/XENONnT/straxen/pull/163

tunnell:
You can use the corrections system too.  Just make some variable called 'daq_rundb_address' then put it in ONLINE.  The corrections system is a time dependent INI file, where the ONLINE collection is just a normal ini for strings.

tunnell:
@ahiguera-mx see my comment about ini

jorana:
@tunnell I'm not sure if I see the point, you still need access to the database before. For reading a simple file, we might as well use Admix # 163

tunnell:
Just the same machinery so only one configuration thing.  And track.

darrylmasson:
You can't store the database access information in the database and expect it to be useful. This is like locking your keys in the car and then asking why you're having trouble starting the engine. Also tracking this info seems pointless, we don't care what the address _was_, we only care what it is right now.

ershockley:
I agree the utilix config would solve this. I will work on that PR today

ahiguera-mx:
> @ahiguera-mx see my comment about ini

PR #270, opened 2020-11-07
Allow 2ns in daqreader

jorana:
In the DAQ, some digitizers may be reporting at a different sampling resolution. In this PR we slightly change the DAQ-reader to accomodate this change. 

We are doing this by instead of assuming a dt, we assume a maximum dt where we conservatively look for breaks in the data using a gap that is presumably larger than the gap needed for dts smaller than the dt_max.

A simple script is tested here:
https://github.com/XENONnT/analysiscode/blob/master/DAQ/neutrons_in_daqreader/load_2ns_dt.ipynb

To see that there are no obvious problems with adding it like this.


WenzDaniel (approved):
Looks, exactly like my modification beside the change of the option name and check update in L368. Thanks a lot and sorry for duplicating the work, was not my intention to trigger this.

PR #271, opened 2020-11-09
Add CMT to straxen

jorana:
Add CMT to load time-evolving values.

**What is affected**
Peaklets and above as in this chart. This means if one wants to use the straxen master one would have to reprocess starting from records. 
afbeelding

This is just activating an option discussed here:
https://github.com/XENONnT/straxen/pull/220
and to a lesser extend here:
https://github.com/ahiguera-mx/straxen/pull/2
https://github.com/XENONnT/straxen/pull/251


WenzDaniel (commented):
Looks fine to me.

tunnell:
Triggering a release not strictly necessary at this point. Does not require repeating analyses.

PR #272, opened 2020-11-10
Negative channel area as contributing PMT

WenzDaniel:
Found a mini-bug in the number of contributing channels for negative area contributing channels. Although these channels are clearly just contributing with noise , they are contributing to the peak and hence should be counted in n_channels. Otherwise one can have peaks without any channel contribution which is odd.

Example event with n_channel == 0, run_id = '009933', time_range=(1604478325669703040, 1604478325669704130)

Branch can be deleted after merge


jorana (changes requested):
Can you explain some more because it looks like this is inducing a bug rather than solving it. Why would you include negative or zero-area PMTs?

WenzDaniel:
Because a Peak with 0 channel contribution is a bit confusing is not it? Since we have an requirement on the number of contributing PMTs. And evenall hits which make up the peak are noise hits with a negative area, they are contributing to the peak are not they? Otherwise this peak would not exist.

jorana:
I think I disagree, there are in this case 0 PMTs contributing to the area of the Peak. Period. E.g. you don't want to have a 3 PMT requirement with 2 PMTs and some noise. You want to have 3 PMTs seeing something.

I see that having 0 contributing sounds strange, but I rather have that than have a wrong number later in analysis.

WenzDaniel:
Bipolar noise should be in 50% of the cases positive if the baseline correction is done properly. Hence, you would just hide this effect. The only way you can ensure to have real photon signals contributing is with a higher area/adc threshold.

jorana:
But this would render .000001 PE => peak, -0.0000001 PE => peak, and 0.0 PE => no peak. I must be missing something but I still don't understand.

WenzDaniel:
0.0 PE does not work here since this value is computed via the area per channel which is for non contributing channels by default 0. I agree that this is a loophole, but I think it is super unlikly to hit exactly 0. I can also see where you are coming from. The other option would to change the help string of the n_channel parameter. Because the thing you were pointing out:

```
I think I disagree, there are in this case 0 PMTs contributing to the area of the Peak. Period. E.g. you don't want to have a 3 PMT requirement with 2 PMTs and some noise. You want to have 3 PMTs seeing something.
```

is exactly the danger here. Analysts may think exactly that, but in reality they may get 2 PMTs and some noise.

jorana:
> is exactly the danger here. Analysts may think exactly that, but in reality they may get 2 PMTs and some noise (independent whether the noise area is slightly positive or negative).

It's just positive noise right that would be contributing. After this PR it would also be negative noise. Keeping out noise (positive or negative) should be done as much as possible should it not? 

I'm just getting the feeling that we would be solving the odd case of having 0 contributing channels by the sacrifice of taking more data (of which we know it can only be noise) to obscure the data.

Indeed, I was also contemplating that highlighting the help of n_channels might resolve the issue.

WenzDaniel:
Since there is so much ambiguity in the interpretation of this parameter I am not even sure if we should keep it after all. It is trivial to compute based on area_per_channel.

jorana:
but for that you need peaks which is rather low level when you are at events.

feigaodm:
Early experience: we can set an area threshold (0.5PE in Xe100) when we count coincidence (n_contributing channels) and tight coincidence. Based on the noise problem we see, I suggest that we do something like this with configurable thresholds. Or better, I would suggest @WenzDaniel to check possibility to cut on entropy per channel in order to decide whether that channel shall be rejected in reconstructing S1 at all.

feigaodm (changes requested):
I suggest that we try to do some smarter "counting" to reject noise contributions instead. We can also save info about these "noisy" channels.

jorana:
perhaps close and file an issue or combine with #309 ?

PR #273, opened 2020-11-12
Context-changes

jorana:
**S1-risetime**
Change `s1_max_rise_time` from `60` to `100` 

_why_
Based on the notes by @BarbaraPa and @jingqiangye:
https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:analysis:waveforms_team:20201112#featured_notes

We see that a rise-time of 82 ns would be optimal. As discussed today, we see that the simulator used might be slightly underestimating the rise time. To be on the safe side, we may increase it some more. The black population in figure 1:
https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:barbara:xenonnt:finetuning will not be affecting the classification much.

**online monitor**
Instead of writing online_peak_monitor and veto_intervals separately, write them together in the online_monitor.

I will change scripts where needed

### DIFF
--- a/straxen/contexts.py
+++ b/straxen/contexts.py
@@ -103,8 +103,7 @@ def xenonnt_online(output_folder='./strax_data',
         st.storage += [straxen.OnlineMonitor(
             readonly=not we_are_the_daq,
             take_only=('veto_intervals',
-                       'online_peak_monitor',
-                       'online_veto_monitor',))]
+                       'online_monitor',))]
 
     # Remap the data if it is before channel swap (because of wrongly cabled
     # signal cable connectors) These are runs older than run 8797. Runs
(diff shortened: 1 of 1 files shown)


feigaodm:
thanks @jorana , do we want to modify "s1_max_rise_time_post100" too? Otherwise I will approve the PR.

jorana:
> thanks @jorana , do we want to modify "s1_max_rise_time_post100" too? Otherwise I will approve the PR.

Can, but neither Jingqiangs, nor Barbara's note give me a strong motivation to do so. I would not be sure what kind of value would be good. I also checked some waveforms, 150 ns sounds ok at the moment.

feigaodm (approved):

jingqiangye:
Yeah I agree with keeping 150 ns for area > 100 PE. This part seems ok.

PR #275, opened 2020-11-13
ffill ONLINE corrections

jorana:
As pointed out by @AlexElykov , you don't want to have time dependent results of when the gains were uploaded to CMT especially when working with the ONLINE version. This means that we should fill rather than interpolate the gains.

Use: this function https://github.com/AxFoundation/strax/blob/master/strax/corrections.py#L100 to get the corrections.


ahiguera-mx (approved):
Looks ready to go

PR #279, opened 2020-11-13
add conditions to avoid zero division

zhut19:

Encountered zero division during desaturation

The `min_reference_length` should be checking on the number of samples exceeding 1 PE with the reference region. And both the saturated pulse and the summed wf model should be checked.


jorana:
Looks good, thanks Tianyu

jorana (approved):
Thanks for the changes, the extra comments are also very helpfull

Issue #280, opened 2020-11-16
Git-action need finetuning in: scripts/update-context-collection.py


jorana:
Evan is working on this but the script in ".github/scripts/update-context-collection.py" does not work yet. Needs some testing but takes time because it is only triggered when the master is updated.

jorana:
Fixed by Evan

Issue #277, opened 2020-11-13
Travis does not test databases


jorana:
As straxen is getting more heavily relying on database access through e.g. the online monitor (#216) and CMT (#220, #251) we should start exploring the opportunity to also let Travis have access to the runsdatabase or at least have access to one run or so that we store as a/multiple CSV file(s)?

Perhaps @tunnell has experience setting this up?

tunnell:
TravisCI has a buildin mongo you can spin up.  I wouldn't use our production server.  You can maybe do a mongo dump then load that on the CI.  It would take some time to setup well is all since we'd also have to make the tests.

jorana:
Finally solved in  #360 😄

PR #282, opened 2020-11-17
Use GitHub Actions to update context collection

ershockley:
Straxen is updated frequently, but Evan has not been good about updating the context collection so that tools like `admix-download` work properly. This PR automates the update of the context collection so that it happens whenever there is a new straxen release. There is also a button for dispatching the workflow manually.

Github Actions is a competitor of Travis CI, allowing for automating of a variety of tests and other tasks. Actions is nice because you can attach encrypted "secrets" to your repo that can then be accessed while the workflow is running. In our case, we include runDB information as a secret so that our job can update the context collection, while keeping the sensitive info private. 

No -- I have not found a way to debug these actions without just trying it. I tested it on my personal fork and it works for manual dispatch. We should test it more when we have another straxen release. 

This should address #280. 

Github Actions (though not this PR) may also be a solution for #277 if we moved  away from Travis, though it sounds like there might be options within Travis as well (Chris knows more).

### DIFF
--- a/.github/scripts/create-utilix-config.sh
+++ b/.github/scripts/create-utilix-config.sh
@@ -6,3 +6,4 @@ rundb_api_url = $RUNDB_API_URL
 rundb_api_user = $RUNDB_API_USER
 rundb_api_password = $RUNDB_API_PASSWORD
 EOF
+
--- a/.github/scripts/create-utilix-config.sh
+++ b/.github/scripts/create-utilix-config.sh
@@ -5,5 +5,9 @@ cat > $HOME/.xenon_config <<EOF
 rundb_api_url = $RUNDB_API_URL
 rundb_api_user = $RUNDB_API_USER
 rundb_api_password = $RUNDB_API_PASSWORD
+pymongo_url = $PYMONGO_URL
+pymongo_user = $PYMONGO_USER
+pymongo_password = $PYMONGO_PASSWORD
+pymongo_database = $PYMONGO_DATABASE
 EOF
 
--- a/.github/scripts/install_straxen.sh
+++ b/.github/scripts/install_straxen.sh
@@ -1,8 +1,5 @@
 #!/bin/bash
 
 cd straxen/straxen
-ls
-
 python setup.py install
-
 cd ~
--- a/.github/scripts/install_straxen.sh
+++ b/.github/scripts/install_straxen.sh
@@ -1,4 +1,8 @@
 #!/bin/bash
 
-cd straxen
+cd straxen/straxen
+ls
+
 python setup.py install
+
+cd ~
--- a/.github/scripts/install_straxen.sh
+++ b/.github/scripts/install_straxen.sh
@@ -1,0 +1,4 @@
+#!/bin/bash
+
+cd straxen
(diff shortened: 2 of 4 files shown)


WenzDaniel (commented):
Sounds very interesting, can you maybe say a view words about it in Team A?

.github/scripts/install_straxen.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+cd straxen/straxen
+python setup.py install
```
jorana, line 4: This won't work until we fix:
https://github.com/XENONnT/utilix/issues/12
right?

.github/scripts/create-utilix-config.sh
```diff
@@ -5,4 +5,9 @@ cat > $HOME/.xenon_config <<EOF
 rundb_api_url = $RUNDB_API_URL
 rundb_api_user = $RUNDB_API_USER
 rundb_api_password = $RUNDB_API_PASSWORD
+pymongo_url = $PYMONGO_URL
```
jorana, line 8: Perhaps we want to create a new account for this (just to be on the safe side) with only access to the required collection?

jorana (changes requested):
Like I said last time, it looks very nice!

Until we fix: XENONnT/utilix#12 I think it won't work. Perhaps we should prioritize that before we start adding the solution to doing an extra git install here.

.github/scripts/install_straxen.sh
ershockley: It works on my fork...

.github/scripts/create-utilix-config.sh
ershockley: This is going through the rundb API, not pymongo. I needed to add the pymongo variables (which are readonly in the secrets) to set up the contexts.

ershockley:
@jorana @WenzDaniel thanks for the comments -- I can show an example of how it works tomorrow at TeamA like Daniel suggested, then maybe it will be more clear.

.github/scripts/install_straxen.sh
jorana: hmm okay, perhaps git is smarter than travis, travis could not find utilix at least in the base env:
https://travis-ci.org/github/XENONnT/base_environment/builds/743581784

.github/scripts/create-utilix-config.sh
jorana: I think that wasn't my question, it was related to if we can somehow add a new user that is only used here, in the git actions.

Not saying we should, just a question/suggestion.

ershockley: Ah sorry I guess I misunderstood; yes we could absolutely make another user. But I wanted to clarify that the pymongo strings are here to be able to instantiate the different contexts, so this new user would need to have permissions to access all the needed collections (I guess there are two? Not sure exactly), not the context collection.

jorana:
I'm fine with merging it now. For me there is no reason not to. @ershockley , go ahead if you like.

jorana (approved):

PR #283, opened 2020-11-18
remove to_pe_constant option from CMT

ahiguera-mx:

The option `st.set_config(dict(gain_model=('CMT_model',("to_pe_constant",my_gains ))))` throws an error `InvalidDocument: cannot encode object: array()`, the reason is straxen does not support arrays as option for the metadata, therefore this option should be removed from CMT <br>
see this example https://github.com/ahiguera-mx/xenon/blob/master/Remove_to_pe_constant_from_CMT.ipynb
see this example https://github.com/ahiguera-mx/xenon/blob/master/Remove_to_pe_constant_from_CMT.ipynb

  - Update the docstring(s)
  - Update the documentation

### DIFF
--- a/straxen/get_corrections.py
+++ b/straxen/get_corrections.py
@@ -49,8 +49,6 @@ def get_to_pe(run_id, gain_model, n_tpc_pmts):
         to_pe = x[run_index[0]]['to_pe']
 
     elif model_type == 'to_pe_constant':
-        warn("to_pe_constant_run will be replaced by CorrectionsManagementSevices",
-             DeprecationWarning, 2)
         if model_conf in FIXED_TO_PE:
             return FIXED_TO_PE[model_conf]
 
@@ -69,7 +67,7 @@ def get_to_pe(run_id, gain_model, n_tpc_pmts):
 
 FIXED_TO_PE = {
     # just some dummy placeholder for nT gains
-    'gain_placeholder': np.repeat(0.001, 494)
+    'gain_placeholder': np.repeat(0.0085, 494)
 }
 
 @export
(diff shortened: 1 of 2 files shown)


straxen/get_corrections.py
```diff
@@ -69,7 +67,7 @@ def get_to_pe(run_id, gain_model, n_tpc_pmts):
 
 FIXED_TO_PE = {
     # just some dummy placeholder for nT gains
-    'gain_placeholder': np.repeat(0.001, 494)
```
WenzDaniel, line 72: Just a question of taste, but I think 

```
np.ones(494)*0.0085
```

is a bit simpler.

WenzDaniel (commented):
Looks fine

WenzDaniel (approved):

straxen/get_corrections.py
jorana: My taste goes the other way ;) would only think about replacing the 494 to `straxen.n_tpc_pmts`. (I know I was the one who put it here in the first place, shame on me.)

jorana (approved):
Really nice Aaron, I think this makes it much cleaner! 

Just one last request, can you change the doc-string accordingly:
https://github.com/XENONnT/straxen/blob/5cff94ebc5ae75f65bcc50a6fd986f644cddcccf/straxen/corrections_services.py#L153

jorana:
In the meanwhile, someone pointed out that without 5a01980 , we are getting some incomprehensible error when we are still trying to use keys that are not in FIXED_TO_PE (anymore). I propose to fix it like in 5a01980 such that people have a somewhat more transparent traceback.

straxen/get_corrections.py
ahiguera-mx: done

ahiguera-mx:
I've updated the docstrings accordingly and the last two suggestions `494 to straxen.n_tpc_pmts` and add the fix from [5a01980](https://github.com/XENONnT/straxen/commit/5a01980030917271aa01b35688214fd2cb481e84)

jorana (approved):
All looks good Aaron!

jorana:
@ahiguera-mx @WenzDaniel I'm merging this

PR #284, opened 2020-11-18
cache CMT values

jorana:
Store the CMT values in ./resource_cache for specific versions in CMT.

Just run CMT, the first time it will do like normally, the next time it will just load

**Why/how long do we need this**
Outside of dali CMT-gains take a long time to load. If you are developing straxen (like me) and need to reprocess peaklets, this becomes a bottleneck. This might be an interim solution until CMT experts get a nice aggregation working.

### DIFF
--- a/straxen/corrections_services.py
+++ b/straxen/corrections_services.py
@@ -140,9 +140,6 @@ class CorrectionsManagementServices():
         else:
             raise ValueError(f'model type {model_type} not implemented for electron lifetime')
 
-    # TODO create a propper dict for 'to_pe_constant' and 'global_version' as
-    #  the 'global_version' is not a version but an array/float for
-    #  model_type = 'to_pe_constant'
     def get_pmt_gains(self, run_id, model_type, global_version,
                       cacheable_versions=('ONLINE',),
                       gain_dtype=np.float32):
@@ -151,7 +148,8 @@ class CorrectionsManagementServices():
         :param run_id: run id from runDB
         :param model_type: to_pe_model (gain model)
         :param global_version: global version
-        :param cacheable_versions: versions that are allowed to be cached in ./resource_cache
+        :param cacheable_versions: versions that are allowed to be
+        cached in ./resource_cache
         :param gain_dtype: dtype of the gains to be returned as array
         :return: array of pmt gains to PE values
         """
--- a/straxen/corrections_services.py
+++ b/straxen/corrections_services.py
@@ -7,6 +7,7 @@ import configparser
 import strax
 import straxen
 from straxen import uconfig
+import os
 
 export, __all__ = strax.exporter()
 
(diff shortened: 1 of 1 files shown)


tunnell:
This will only work if you ONLINE if you check that the dataset that you are processing was from before whenever you cached, as future values are allowed to change.

jorana:
@tunnell , this is true. However CMT does not work in the future either:
https://github.com/XENONnT/straxen/blob/c317a4cd29ae91b05a8f5fc29932ac346488d4fd/straxen/corrections_services.py#L261

So you are inherently protected from doing this.

tunnell:
I mean something else.  Assuming t_i < t_i+1, then:

1. Request corrections at t_1
2. Corrections cached at t_1
3. reprocessing at t_0 is fine
4. Wait....
5. Reprocessing at t_2 will be ill defined and requires a new query.  

Therefore, you need to keep track of the time that the user populates the cache, and repopulate if runs from the future requested.  Or if this is purely for development, why not just make a local Mongo copy of this collection until CMT usage settles?

jorana:
So in this PR you'd cache per run (3kb on my machine), therefore you will not run into 5.

ahiguera-mx:
Since this is purely for development I have share a `to_pe_per_run` file to Joran, so he does not have to query the DB, this file contains exactly the same data as the DB, and is pretty much exactly what he is doing saving on file per run, all merged in one single file

jorana:
Hi Aaron, as I also send to you, this is not what we want. Someone should be maintaining this file and also changing all their scripts to get it working. Also, if it's basically the same as this PR, why wouldn't we implement just this?

I haven't heard any cons (that are correct) and I do see the pros. Please consider that I'm trying to help improve the code itself and have straxen run in a manner thats both fast and makes sense, I don't care who develops what as long as it works.

ahiguera-mx:
I don't think this is sustainable in the short-mid term, we are planing to implement n_veto and mu_veto gains as well, so having this temporally patch is not the right approach in mi opinion, we know what we need to do to optimize it just take time. 
I would like to know if @tunnell considers this patch a good option for now

WenzDaniel:
Hej guys, I followed this discussion a bit through my emails and I do not quite understand the issue. As Joran pointed out he would like to have this caching solution for his temporal work. I can also see his point that with this solution one has not to keep track of the gain file manually which can be quite tedious and would also make the nice work of the CMT obsolete. 

I do not fully understand the following argument: 

> I don't think this is sustainable in the short-mid term, we are planing to implement n_veto and mu_veto gains

since I cannot see how this would impact the caching. 

> we need to do to optimize it just take time.

Which is fully okay, but we cannot pause all the other work going on in the meantime.

jorana:
@tunnell let me know your thoughts. I think there is a lot of time going in this PR which should be trivial. We can also have a chat if you want next week?

Sooner or later its essential CMT caches because downloading ~O(100s MB) maps from a database is slow on any machine.

jorana:
@tunnell just approved in a DM. Merging this now.

PR #285, opened 2020-11-18
Update straxer

jorana:
Straxer is a reprocessing script. Using this script we can e.g. reprocess data. Using this PR, we can easily pass a json configuration file to this script to allow one to change the configurations.

1. Read JSON
2. Convert list -> tuple (we never use lists as options, or at least we shouldn't)
3. Insert the read JSON file in `st.set_config`

Also allow straxer to use an extra data-directory.

```python
cmd = f'straxer {run_id} --context xenonnt_simulation --target {target}'
cmd += f' --add_folder {data_dir}'
cmd += f' --config_from_json {conf_file}'
!$cmd # Submit using ipython magic
```


jorana:
Nobody? I'm merging this than. Looks like I'm the only one using it anyways. Have been using it the past few days. Works like a charm.

PR #286, opened 2020-11-18
Update contexts.py

petergaemers:
This pull requests:
  *Adds a needed configuration value to skip checking overlapping raw records in the simulation contexts. This is needed from waveform simulation

Wfsim does not know what pulses are generated by previous instructions. For example an S2 can cause a afterpulse to be generated before the S2 is fully finished. The afterpulse does not know the S2 happened since the data is already written out.
So we can get two records from the same channel who overlap

This skips the check if overlapping records are generated


jorana (approved):
looks good Peter!

PR #288, opened 2020-11-23
use utilix for client

jorana:
We are getting this error due to the readpreference of the mirrors:
https://xenonnt.slack.com/archives/C016DM0JPK9/p1605866361323700

Therefore we change the way we init the pymongo client for the corrections. NB: the init is now done here, on the `straxen` side (makes more sense to do it in the nT specific package). This PR requires:
https://github.com/AxFoundation/strax/pull/356

Given recent changes (https://github.com/XENONnT/straxen/pull/163), it makes more sense to use a centralized initiation for the pymongo client (i.e. `utilix`). 

All should work just as before. we now just don't run into https://xenonnt.slack.com/archives/C016DM0JPK9/p1605866361323700.

### DIFF
--- a/straxen/corrections_services.py
+++ b/straxen/corrections_services.py
@@ -7,7 +7,7 @@ import configparser
 import strax
 import straxen
 from straxen import uconfig
-
+import utilix
 export, __all__ = strax.exporter()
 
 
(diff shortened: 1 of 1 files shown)


ershockley (approved):
Looks great Joran, thanks!

PR #290, opened 2020-11-23
Add coveralls to travis

jorana:
Add https://docs.coveralls.io/ to travis testing.

Just like in strax

**My concern**
I'm just copy pasting code here. I have added coveralls for straxen: https://coveralls.io/github/XENONnT/straxen, however I'm not sure if this is going to work the first time. If someone is willing to give me the green light I can just push to master.

I think this would be great to add because it shows how good/bad your PR is and increases the incentive to write tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,9 +27,6 @@ deploy:
   on:
     tags: true
 
-script:
-  - coverage run setup.py test
-  
 # Upload code coverage information
 # Only do this if not jitting, otherwise we miss those tests.
 after_success:
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ install:
 script:
 - python setup.py test
 - pip install coveralls
-- coverage run setup.py test
+- coverage run --source=straxen setup.py test
 deploy:
   provider: pypi
   skip_existing: true
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,15 @@
 language: python
 dist: xenial
-python:
-- '3.6'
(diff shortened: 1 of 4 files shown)


jorana:
@WenzDaniel , are we running into https://github.com/AxFoundation/strax/issues/328 at line https://github.com/AxFoundation/strax/blob/master/strax/processing/pulse_processing.py#L116?\

This just happens for numbaless coverage (newly added in this PR) so you probably have never seen it before.

afbeelding

WenzDaniel:
This is interesting, but no the Issue you are linking would be the opposite I think. Back then I had the issue that my assignment 
```
data = thing['data'][thing['length']]
```
was a view onto `thing['data']`. Hence when I modified data, I indirectly modified thing['data'] which was not desired. Hence this read-only thing you are describing here would have saved me from this issue. 
I know that one can make an array immutable via `array.flag.writeable = False`, but I do not think that we are explicitly doing this somewhere. Further, I checked the requirements since v0.12.0 and numpy and numba has not change since then. Have you updated your env recently?

WenzDaniel:
The thing I do not understand though, why is it just popping up for the nveto_recorder?

jorana:
Thanks @WenzDaniel , trying to work from here now

jorana:
Ok, seems to work :) @ershockley or @WenzDaniel willing to review (is just technical so might do without)

jorana:
Ok, changed the way it looks at the package, it's now at 61 percent (apparently this test_plugins is very effective). I will continue and merge.

PR #292, opened 2020-11-26
add more tests to straxen

jorana:
Add several tests for packages missed by the current tests.

Mostly checks that things load properly.

### DIFF
--- a/tests/test_basics.py
+++ b/tests/test_basics.py
@@ -23,11 +23,15 @@ def test_straxen():
 
             print("Test processing")
             df = st.get_df(run_id, 'event_info')
+
             assert len(df) > 0
             assert 'cs1' in df.columns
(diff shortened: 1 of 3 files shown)


jorana:
2.5 % coverage increase of tests over entire straxen code. Not too bad..

WenzDaniel:
Hej, I am not sure if I should already review this PR. One question/suggestion, do you think it make sense to add for example for `test_straxen` some "truth" dictionary containing the expected outcomes. I am asking in case we will ever change the demo it might be easier to just modify the dict instead of tests itself.

jorana:
Sure, you mean like testing the number of events and so on? Could add with some reasonable error or margin. Thanks for the sugestion

jorana:
@WenzDaniel Done, can you re-review?

PR #296, opened 2020-11-30
Veto Proximity Plugin

AlexElykov:
Similarly to what we had in XENON1T, data kind produced by this plugin will be used by the DAQVetoCut, to remove events which are too close to the start or stop of a hardware veto signal.

For each event center (event as defined in strax) it will find and store the distance to the previous and next veto on/off. 

It can be illustrated (and was tested) using data obtained with the special triggered mode, where left boundary of the yellow region is "trg start" and right is "trg stop". For each event center, the plugin will find the time to the next/previous trg start/stop.
veto_prox


WenzDaniel:
Hej, I am getting a bit confused by your example. I understand it as following: 

You would like to have a plugin which computes the time difference between an "event center" and the next veto signals. But your example shows data where you used an external trigger to record the events. Hence to convert your example image according to your what you want (or at least I think you want). I have to imagine the blue dashed line as veto-start and the dashed-dotted line as veto-end, am I correct?

straxen/plugins/acqmon_processing.py
```diff
@@ -130,4 +133,82 @@ def merge_vetos(channels, gap, dtype, t):
...
+                else:
+                    veto_start_time_selection = veto_selection['endtime']
+                    
+                for t, time in enumerate(t_event_centers):
+                    inx = np.searchsorted(veto_start_time_selection, time, side = 'left')
+                    
+                    # Time to previous veto on/off
+                    # Just using maxsize as a huge value that will not fit in any potential VetoCut range
+                    if inx == 0:
+                        res[prev].append(sys.maxsize)
```
jorana, line 198: Perhaps this is slightly overkill as it might depend on your machine and not fit in the np.int64. Perhaps define some hard coded improbable value in this python file such that you can do ` veto_proximity['next_busy_on']==straxen.acqmon_processing.T_NO_VETO_FOUND` .

straxen/plugins/acqmon_processing.py
```diff
@@ -130,4 +133,82 @@ def merge_vetos(channels, gap, dtype, t):
...
+
+        return dtype
+    
+    def setup(self):
+        self.states = ['on', 'off']
+        
+    def compute(self, events, veto_intervals):
+        res = defaultdict(list)
+        
+        t_event_centers = (events['time'] + events['endtime'])//2
```
jorana, line 177: I can imagine that you want to have the end/start depending on if you have ` Time to previous ` or `Time to next ` . An event is just a bunch of peaks where the s1 and s2 may be very far apart in time. So I would have expected that this center time might not be what you need based on your description.

jorana (commented):
Hi Alex, nice plugin! I think this will indeed be useful for a lot of people that would otherwise be trying to figure it out themselves. Technically it seems fine. I'm just wondering about the definition of the following as I explain below.
```python
t_event_centers = (events['time'] + events['endtime'])//2
```

Further, I wonder if you'd want this to be a `CutPlugin`, or otherwise, if this is not a cut plugin, would you say it's worth making it a ```save_when = strax.SaveWhen.NEVER``` plugin (you can of course still load it / base subsequent plugins on it). This prevents us from writing each of the plugins to disk (not per se the size but more the associated bookkeeping is what I'm keeping in mind here). For example, `peaks` is such a plugin:
https://github.com/XENONnT/straxen/blob/master/straxen/plugins/peaklet_processing.py#L622

AlexElykov:
> Hej, I am getting a bit confused by your example. I understand it as following:
> 
> You would like to have a plugin which computes the time difference between an "event center" and the next veto signals. But your example shows data where you used an external trigger to record the events. Hence to convert your example image according to your needs (or at least what I think you want), I have to imagine the blue dashed line as veto-start and the dashed-dotted line as veto-end, am I correct?

Hi, 

These data were not taken with an external trigger but with the "special veto" mode that I've made for commissioning. In this mode, we periodically artificially render the digitizers busy (veto), allowing them to operate even under high rate conditions, and that is how we get the 17%, 20%, 80%, etc live time data.  https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenon1t:alexelykov:v1495_busy_special

It made more sense at the time to record the start and end of the periods when we have data. These times are recorded in the hev_on/off channels of the acquisition monitor, so I just used them as a proxy for real hev_on and hev_off for test purposes here. Hope this makes more sense.

straxen/plugins/acqmon_processing.py
AlexElykov: I guess I can think of something else. I was using np.inf, but strax.dict_to_rec didn't like it.

WenzDaniel:
Thanks for clarification and I think

> It made more sense at the time to record the start and end of the periods when we have data. 

goes in the direction of what I was wondering. In this case your plugin works, while I think if you revert the logic you would need an overlapping window plugin. Since in principle it would be possible  that you chunk your data at a point where you have n events and n-1 veto_regions (or vise-versa). Hence you would miss one "delta t to the next/prev veto".

AlexElykov:
> Thanks for clarification and I think
> 
> > It made more sense at the time to record the start and end of the periods when we have data.
> 
> goes in the direction of what I was wondering. In this case your plugin works, while I think if you revert the logic (dash-dotted start and dashed end) you would need an overlapping window plugin. Since in principle it would be possible that you chunk your data at a point where you have n events and n-1 veto_regions (or vise-versa). Hence you would miss one "delta t to the next/prev veto".
> 
> https://github.com/AxFoundation/strax/blob/a1b814cca74957f812b5a80eb90ee174a5e72d84/strax/plugin.py#L292-L302

In real background/calibration data, if dash-dotted is veto "start" and dashed is veto "end" there would be no data in between them - and they would indicate the veto bounds. To avoid associating anything potentially found in proximity to them is one of the reasons for having this plugin.

WenzDaniel:
I think we are not on the same page yet, maybe better to move to slack :D

straxen/plugins/acqmon_processing.py
AlexElykov: > I think we are not on the same page yet, maybe better to move to slack :D

Sure :)

AlexElykov: The resulting output from this plugin would be used in a cut plugin, which will look something like what we had in lax in XENON1T, with the "nearest_x" calculated there from the "previous_x" and "next_x" (or I might add this to this plugin). 
https://github.com/XENON1T/lax/blob/1ac8e2625d82d3fecce149f8b1e9b46959cf00f8/lax/lichens/sciencerun0.py#L147

Regarding "save_when = strax.SaveWhen.NEVER", yeah I think it is a good idea.

AlexElykov:
> I think we are not on the same page yet, maybe better to move to slack :D

Hey. After our discussion on slack, I've made some quick tests with run 10362, where the busy was firing like crazy and the plugin seems to work as it is. I might make it an OverlapWindowPlugin just in case though.

vetos

jorana:
Hi Alex, I'm for a second converting this PR to draft that makes it a little easier for me to do the bookkeeping. Does not affect anything but is just a reminder to myself that you are working on it

AlexElykov:
> Hi Alex, I'm for a second converting this PR to draft that makes it a little easier for me to do the bookkeeping. Does not affect anything but is just a reminder to myself that you are working on it. Let me know when you are ready and if we can help but I think you got this covered.

No problem. I had some other things to finish up today but I have some commits ready.

jorana:
Hi Alex, I did not notice before but there are quite some untracked options in acqmon processing that look like they should be tracked. Are you sure about those? Not really related to this  PR

AlexElykov:
> Hi Alex, I did not notice before but there are quite some untracked options in acqmon processing that look like they should be tracked. Are you sure about those? Not really related to this PR

Hey, what do you mean by untracked options?

jorana:
> 
> 
> > Hi Alex, I did not notice before but there are quite some untracked options in acqmon processing that look like they should be tracked. Are you sure about those? Not really related to this PR
> 
> Hey, what do you mean by untracked options?

Options like these where `track==False`. Now that I think about it of course make sense since it is a logic signal and these are rather low thresholds. It just at first looks like an option that should be tracked, e.g. you don't want the same data when someone sets it to 0 instead. But okay, maybe not as relevant as I thought at first glance.
https://github.com/XENONnT/straxen/blob/4e137c98d273eb87219780062b95dfecb66d14bd/straxen/plugins/acqmon_processing.py#L17

AlexElykov:
> > > Hi Alex, I did not notice before but there are quite some untracked options in acqmon processing that look like they should be tracked. Are you sure about those? Not really related to this PR
> > 
> > 
> > Hey, what do you mean by untracked options?
> 
> Options like these where `track==False`. Now that I think about it of course make sense since it is a logic signal and these are rather low thresholds. It just at first looks like an option that should be tracked, e.g. you don't want the same data when someone sets it to 0 instead. But okay, maybe not as relevant as I thought at first glance.
> https://github.com/XENONnT/straxen/blob/4e137c98d273eb87219780062b95dfecb66d14bd/straxen/plugins/acqmon_processing.py#L17

I can change them to tracked if you think it is necessary. I was planning to submit a PR anyways (there are some bugs that I've found a while ago) so I can add it there. However, thus far I simply didn't have time to update anything cause I'm trying to concentrate on the thesis thing.

AlexElykov:
Hopefully final update to the aqmon plug-in set:

- I re-did the veto interval plugin, as it had issues with finding intervals of variable length - now it seems to work well in triggered mode and for vetoes in normal runs as well.
- Made updates based on your inputs to the plugins, i.e. tracking of options, overlawindow plugin for proximity, fixed max veto distance to 1 hour.

straxen/plugins/acqmon_processing.py
```diff
@@ -51,21 +57,19 @@ def compute(self, raw_records_aqmon):
...
 @strax.takes_config(
-    strax.Option('min_veto_gap', default=int(1e6), type=int,
-                 help='Minimum separation between veto stop and start pulses [ns]'),
-    strax.Option('max_veto_gap', default=int(5e8), type=int,
+    strax.Option('max_veto_window', default=int(5e8), track = True, type = int,
                  help='Maximum separation between veto stop and start pulses [ns]'),
-    strax.Option('channel_map', track=False, type=immutabledict,
+    strax.Option('channel_map', type = immutabledict, track=False, 
                  help='immutabledict mapping subdetector to (min, max)'
-                      'channel number.'))
```
jorana, line 60: Not entirely sure but guess you don't have to change this since it used to be exactly the same as in the DAQ-reader:
https://github.com/XENONnT/straxen/blob/c84928fb4ea20bd1872405e9dcef87db9508d220/straxen/plugins/daqreader.py#L61

straxen/plugins/acqmon_processing.py
```diff
@@ -90,44 +94,125 @@ def setup(self):
...
+                prev = 'previous_%s_%s' %(name, state)
+                nxt = 'next_%s_%s' %(name, state)
+                
+                if state == 'on':
+                    veto_start_time_selection = veto_selection['time']
+                else:
+                    veto_start_time_selection = veto_selection['endtime']
+                    
+                for t, time in enumerate(t_event_centers):
+                    inx = np.searchsorted(veto_start_time_selection, time, side = 'right')
```
jorana, line 197: I think that at this point, you can also add:
`inx = np.searchsorted(veto_start_time_selection[prev_inx:], time, side = 'right')`
whereafter you set `prev_inx` to `inx` since the events are sorted in time. That way you don't have to each the first x veto intervals if you know that previous event wasn't in those first x either.

I see that most likely there is little events so it will be fast anyway we are doing a third level for loop here. Faster still would be if we do the for loop of the names and states inside the for loop of time such that you only have to traverse the time loop (longest) once. But I guess this also has it advantage that it looks rather clear what you are doing.

jorana (approved):
Look good Alex! Thanks for the PR, made some comment that might speed it up a little but I know it's not really much of a concern for this plugin.

Do you want me to also test, technically I think it looks fine.

AlexElykov:
> Look good Alex! Thanks for the PR, made some comment that might speed it up a little but I know it's not really much of a concern for this plugin.
> 
> Do you want me to also test, technically I think it looks fine.

Hi, feel free to test it. 
I made some tests on a triggered run and a normal bg run with high rate where there were some busy signals and it worked fine.
However, maybe a good idea would be to see if you can run it and then get the correct livetimes in the online monitor.

straxen/plugins/acqmon_processing.py
AlexElykov: Cheers, for the suggestions, if there would be some speed issues for the bigger runs I'd update this accordingly. :)

straxen/plugins/acqmon_processing.py
AlexElykov: Hmm...I actually wonder if that is the reason why I couldn't set track=True for this option in the aqmon plugin.

jorana (changes requested):
@AlexElykov unfortunately the test was not successful:
```python
2021-02-03T12:44:19.174546 | [INFO]     | Still processing run 011593
Process Process-1:
  File "/home/xedaq/miniconda/envs/py38/lib/python3.8/multiprocessing/process.py", line 315, in _bootstrap
    self.run()
  File "/home/xedaq/miniconda/envs/py38/lib/python3.8/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/home/xedaq/software/straxen/bin/bootstrax", line 1231, in run_strax
    st_make()
  File "/home/xedaq/software/straxen/bin/bootstrax", line 1210, in st_make
    for i, c in enumerate(st.get_iter(run_id,
  File "/home/xedaq/software/strax/strax/context.py", line 956, in get_iter
    generator.throw(e)
  File "/home/xedaq/software/strax/strax/context.py", line 933, in get_iter
    for n_chunks, result in enumerate(strax.continuity_check(generator), 1):
  File "/home/xedaq/software/strax/strax/chunk.py", line 266, in continuity_check
    for s in chunk_iter:
  File "/home/xedaq/software/strax/strax/processor.py", line 270, in iter
    raise exc.with_traceback(traceback)
  File "/home/xedaq/software/strax/strax/mailbox.py", line 254, in _send_from
    x = next(iterable)
  File "/home/xedaq/software/strax/strax/plugin.py", line 583, in iter
    yield from super().iter(iters, executor=executor)
  File "/home/xedaq/software/strax/strax/plugin.py", line 373, in iter
    raise RuntimeError(
RuntimeError: VetoProximity was unable to get time-consistent inputs after ten passess. Inputs:
{'veto_intervals': [011593.veto_intervals: 1609670504sec 0 ns - 1609670507sec 657170360 ns, 609 items, 0.0 MB/s], 'events': [011593.events: 1609670504sec 0 ns - 1609670507sec 657170360 ns, 442 items, 0.0 MB/s]}
Input buffer:
{'veto_intervals': [011593.veto_intervals: 1609670507sec 657170360 ns - 1609670507sec 687090199 ns, 5 items, 0.0 MB/s], 'events': [011593.events: 1609670507sec 657170360 ns - 1609670509sec 213960090 ns, 186 items, 0.0 MB/s]}
```

Perhaps it should be tested from processing starting at `raw_records_*`.

AlexElykov:
> @AlexElykov unfortunately the test was not successful:
> 
> ```python
> 2021-02-03T12:44:19.174546 | [INFO]     | Still processing run 011593
> Process Process-1:
> Traceback (most recent call last):
>   File "/home/xedaq/miniconda/envs/py38/lib/python3.8/multiprocessing/process.py", line 315, in _bootstrap
>     self.run()
>   File "/home/xedaq/miniconda/envs/py38/lib/python3.8/multiprocessing/process.py", line 108, in run
>     self._target(*self._args, **self._kwargs)
>   File "/home/xedaq/software/straxen/bin/bootstrax", line 1231, in run_strax
>     st_make()
>   File "/home/xedaq/software/straxen/bin/bootstrax", line 1210, in st_make
>     for i, c in enumerate(st.get_iter(run_id,
>   File "/home/xedaq/software/strax/strax/context.py", line 956, in get_iter
>     generator.throw(e)
>   File "/home/xedaq/software/strax/strax/context.py", line 933, in get_iter
>     for n_chunks, result in enumerate(strax.continuity_check(generator), 1):
>   File "/home/xedaq/software/strax/strax/chunk.py", line 266, in continuity_check
>     for s in chunk_iter:
>   File "/home/xedaq/software/strax/strax/processor.py", line 270, in iter
>     raise exc.with_traceback(traceback)
>   File "/home/xedaq/software/strax/strax/mailbox.py", line 254, in _send_from
>     x = next(iterable)
>   File "/home/xedaq/software/strax/strax/plugin.py", line 583, in iter
>     yield from super().iter(iters, executor=executor)
>   File "/home/xedaq/software/strax/strax/plugin.py", line 373, in iter
>     raise RuntimeError(
> RuntimeError: VetoProximity was unable to get time-consistent inputs after ten passess. Inputs:
> {'veto_intervals': [011593.veto_intervals: 1609670504sec 0 ns - 1609670507sec 657170360 ns, 609 items, 0.0 MB/s], 'events': [011593.events: 1609670504sec 0 ns - 1609670507sec 657170360 ns, 442 items, 0.0 MB/s]}
> Input buffer:
> {'veto_intervals': [011593.veto_intervals: 1609670507sec 657170360 ns - 1609670507sec 687090199 ns, 5 items, 0.0 MB/s], 'events': [011593.events: 1609670507sec 657170360 ns - 1609670509sec 213960090 ns, 186 items, 0.0 MB/s]}
> ```
> 
> Perhaps it should be tested from processing starting at `raw_records_*`.

Hmm...interesting, I'll check it out. What did you use to run this test?

jorana:
bootstrax, on run 011593 located at /live_data/xenonnt_tests. Let me update the bootstrax instance on master that I needed to do this test.

Actually, should work if you use 011593 from raw_records

AlexElykov:
> bootstrax, on run 011593 located at /live_data/xenonnt_tests. Let me update the bootstrax instance on master that I needed to do this test.
> 
> Actually, should work if you use 011593 from raw_records.
> 
> I'll merge #354 such that you can check:
> 
> ```shell
> bootstrax --process 11593 --targets veto_proximity --post_process raw_records
> ```

Interesting, meanwhile I downloaded the raw_records_aqmon run and it worked fine from aqmon_hits to vet_proximity, just in a notebook.

jorana:
>  Can it be caused by the fact that veto_proximity needs events being available to work, or I'm missing something?

No per se, strax would make sure to make the events first.

AlexElykov:
@jorana Hi. I've just noticed that this plugin wasn't merged in the end. I thought that the rechunking changes solved the issues that it had with the test run? However, honestly I tunned out a bit so maybe I'm missing/forgetting something?

ershockley:
@jorana @AlexElykov I took a look at this and tested it with 30 runs (using data with both raw_records_aqmon and events on Midway). Most of them worked, but one run failed with this chunk error that I've seen in older data as well. Is it possible we made a change in `raw_records` chunking that did not get propagated to `raw_records_aqmon`? 

```
>>> st.get_array('020346', 'veto_proximity', progress_bar=False)
# long trace...
ValueError: Attempt to create chunk [020346.veto_proximity: 1621947872sec 385127957 ns - 1621948430sec 407021079 ns, 13371 items, 0.0 MB/s] whose data starts early at 1621947872384229430
```

ershockley:
I did a bit more debugging. I think my ignorance of internals of strax is catching up with me 😕 

So if I first create `veto_intervals`, everything works fine. But I thought strax was supposed to do this automatically? If I clear all the strax data and try reloading `veto_proximity`, I can see that the `veto_intervals` data doesn't actually get created; instead it only has a metadata file that contains the exception for the `veto_proximity` processing I show above.

ershockley:
Ok so the above was the case for Run 20346, which I was able to get past by forcing `allow_rechunk=False`. But I hit the same error for Run 20392 which persists if I (a) for `allow_rechunk=False` and/or (b) first create the `veto_intervals`.

ershockley:
Ok it works for all runs tested if I swap the order of the `depends_on` attribute to make `events` come first. Maybe some issue of mixing different data kinds? I leave it to the real experts. 

Anyway at least from my side I think if @AlexElykov can merge [this PR](https://github.com/AlexElykov/straxen/pull/1) then it should be good to merge. Assuming @jorana and @WenzDaniel are okay with it. I think this is fairly high priority for a DAQ veto cut.

AlexElykov:
@ershockley Thanks for checking it out! Let's see what @jorana will say, would be nice if this time it actually can run on ebs.

straxen/plugins/acqmon_processing.py
```diff
@@ -90,44 +94,122 @@ def setup(self):
...
+            # follows it in time. If this is not true, our hardware does not work. 
+            if len(veto_hits_start):
+                for t, time in enumerate(veto_hits_start['time']): 
+                    # Find the time of stop_j that is closest to time of start_i
+                    inx = np.searchsorted(veto_hits_stop['time'], time, side = 'right') 
+                    
+                    if inx == len(veto_hits_stop['time']):
+                        continue
+                    else:
+                        res['veto_interval'].append(veto_hits_stop['time'][inx] - time)
```
WenzDaniel, line 118: Is not this information redundant and can be easily computed?

straxen/plugins/acqmon_processing.py
```diff
@@ -90,44 +94,122 @@ def setup(self):
...
+        for n, name in enumerate(self.veto_names):
+            veto_hits_start = channel_select_(hits, self.channel_map[name + 'start'])
+            veto_hits_stop = channel_select_(hits, self.channel_map[name + 'stop'])
+            
+            # Here we rely on the fact that for each start, there is a single stop that 
+            # follows it in time. If this is not true, our hardware does not work. 
+            if len(veto_hits_start):
+                for t, time in enumerate(veto_hits_start['time']): 
+                    # Find the time of stop_j that is closest to time of start_i
+                    inx = np.searchsorted(veto_hits_stop['time'], time, side = 'right')
```
WenzDaniel, line 113: If I follow your comment from above I would expect that I get for each index `t` (which is not used btw) the correct start and stop. So why do we have to sort here?

straxen/plugins/acqmon_processing.py
```diff
@@ -90,44 +94,122 @@ def setup(self):
...
         return result
 
-
 @numba.njit
-def channel_select(rr, ch_stop, ch_start):
-    """Return data in between stop and start channels of the acquisition monitor (AM)"""
-    return rr[((rr['channel'] == ch_stop) | (rr['channel'] == ch_start))] 
+def channel_select_(rr, ch):
+    """Return data from start/stop veto channel in the acquisition monitor (AM)"""
+    return rr[rr['channel'] == ch]
```
WenzDaniel, line 129: Not sure if this really has to be a function of its own. Is the numba version adding anything to the performance?

straxen/plugins/acqmon_processing.py
```diff
@@ -90,44 +94,122 @@ def setup(self):
...
-            result = np.zeros(len(start), dtype=dtype)
-            result['time'] = start
-            result['endtime'] = stop
-        else:
-            result = np.zeros(0, dtype=dtype)
+        res = defaultdict(list)
+        
+        for n, name in enumerate(self.veto_names):
+            veto_hits_start = channel_select_(hits, self.channel_map[name + 'start'])
+            veto_hits_stop = channel_select_(hits, self.channel_map[name + 'stop'])
```
WenzDaniel, line 106: I guess I am overlooking something, but where do you define self.channel_map?

straxen/plugins/acqmon_processing.py
```diff
@@ -90,44 +94,122 @@ def setup(self):
...
+    next_x: Time in ns between the time center of an event and the next x
+    This also considers any x inside the event. x could be either:
+        - busy_x: busy on/off signal
+        - he_x:   high energy channels busy on/off signal
+        - hev_x:  high energy veto on/off signal
+    """
+        
+    __version__ = '0.0.2'
+    depends_on = ('events', 'veto_intervals')
+    provides = ('veto_proximity')
```
WenzDaniel, line 150: I think we should make the name a bit more explicit. Otherwise analysts may confuse it with our two veto detectors.

straxen/plugins/acqmon_processing.py
```diff
@@ -90,44 +94,122 @@ def setup(self):
...
+    
+    def infer_dtype(self):
+        dtype = []
+        for n in self.veto_names:
+            dtype.append((('Time to previous %s veto start [ns]' %n, 'previous_%s_on' %n), np.int64))
+            dtype.append((('Time to previous %s veto end [ns]' %n, 'previous_%s_off' %n), np.int64))
+            dtype.append((('Time to next %s veto start [ns]' %n, 'next_%s_on' %n), np.int64))
+            dtype.append((('Time to next %s veto end [ns]' %n, 'next_%s_off' %n), np.int64))
+        
+        dtype += strax.time_fields
```
WenzDaniel, line 165: Minor comment could you reverse the order here? Then it is consistent for all data_types when you call `array.dtype.names`

straxen/plugins/acqmon_processing.py
```diff
@@ -90,44 +94,122 @@ def setup(self):
...
+    
+    def get_window_size(self):
+        return (self.config['veto_proximity_window'] * 100)
+        
+    def compute(self, events, veto_intervals):
+        res = defaultdict(list)
+        
+        t_event_centers = (events['time'] + events['endtime'])//2
+
+        for n, name in enumerate(self.veto_names):
```
WenzDaniel, line 180: I would have used here a different strategy but many roads lead to Rom. If we run into performance issues as Joran already mentioned one could use something like: 
```
touching = strax.touching_windows(vetos, events_with_an_extended_time_and_endtime)
....
```

WenzDaniel (commented):
I also have a few comments, most of them minor except for the one about the channel_map.

jorana:
@AlexElykov @ershockley @WenzDaniel let me have a look if we can get it working

straxen/plugins/acqmon_processing.py
WenzDaniel: Okay was too early, I have not seen the collapsed code.

straxen/plugins/acqmon_processing.py
AlexElykov: Actually, on a related topic. I got confused for a sec when you started to name your plugins veto_x.
However, veto_proximity describes the meaning of the data perfectly. Also, as far as I can tell, this will not be used on its own by the average analysts.

AC or we should define a cut DAQveto cut, with fixed values based on veto_proximity, and apply it to all data using cutax.

ershockley:
Sorry @jorana do you mean it doesn't work on ebs or it does? It definitely works when processing from raw_records_aqmon. At least it did yesterday before the most recent changes.

jorana:
@ershockley correct, it did not work on the ebs, with https://github.com/AxFoundation/strax/pull/450 it does

jorana (approved):
Thanks Alex, have you had the time to sanity check the changes I made? I'm happy with the current implementation, thanks a lot for the effort and sorry that it all took so long 😉

AlexElykov:
> Thanks Alex, have you had the time to sanity check the changes I made? I'm happy with the current implementation, thanks a lot for the effort and sorry that it all took so long 😉

It's fine from my side. Hope it works and we can actually use the HEV to its full potential without scaring ppl with some weird s2 relics. Thank you for checking it out, and finding the strax overlapwindow plugin issue!
Also, I like your solution with the numpy arrays better than my dict one :)

PR #298, opened 2020-12-01
Test timeselection

jorana:
See https://github.com/AxFoundation/strax/pull/345. Here Daniel suggested (https://github.com/AxFoundation/strax/pull/345#pullrequestreview-541531692) to test. This is not that easy to do in strax but very doable in straxen. It also kind of makes sense to do it here because it strongly depends on the way one sets up the `plugins`, their `dependencies` and `plugin.save_when`.

I manually extracted 13 offending time ranges. Try running into an error by forcing these time ranges. 

If all goes well, this PR will fail first try (until we merge 345).


jorana:
Great, 0ddecd3 failed, 4d2f818 was committed **after** merging AxFoundation/strax#345 and should therefore succeed, highlighting the issue that this PR is addressing.

tests/test_strax_time_selection.py
```diff
@@ -0,0 +1,60 @@
+"""
```
WenzDaniel, line 1: Maybe add this text to your assert message.

WenzDaniel (approved):
Looks nice, thank you very much.

tests/test_strax_time_selection.py
jorana: Thanks, the assert statement should simply never fail it's not really the test itself. Nonetheless added a sting just above the - would be failing - line

jorana:
Added another 6% coverage :)

PR #299, opened 2020-12-02
lower min pmt to 2

jorana:
Lower the min pmt to 2


WenzDaniel (approved):
Thanks a lot.

PR #302, opened 2020-12-02
add temporary context

jorana:
Create temporary context to load selected PMTs

Mimic xenonnt_online but as if we only have 5 PMTs


WenzDaniel (approved):
I already approve, lets wait if we can use CMT.

feigaodm:
It would be better if we don't need to copy records to dali, are we creating some duplicated peaks/lone-hits instead? I guess not.

jorana:
> It would be better if we don't need to copy records to dali
Than we have a problem. We were copying records before unless something changed

> are we creating some duplicated peaks/lone-hits instead? I guess not.
We would be, that is the point, reprocess some data and create a separate copy for the purposes you envisage.

feigaodm:
I thought that we were not copying records anymore, that's why in the other doc I was assuming that no raw_records nor records are on dali

ershockley:
We are not copying records to dali anymore. If it's a small amount of data that needs to be reprocessed we can just download it briefly and then process it. If it's a lot then we can use OSG. There was one small issue I was still working on in outsource but could probably have it ready within a few days if I focus solely on it (if it's needed).

tunnell:
The discussion above isn't about this PR but more generally about our computing infrastructure.  (Which means it might get lost since we normally use Slack for that)

The code looks fine, but hopefully small tweak requests like this get handled by CMT longer term to avoid strax version changes. Just bookkeeping doing this type of thing in XENON1T was nightmare, ask Evan.

ahiguera-mx:
Hi all,
I have added a new offline global version in CMT, you can use it to get the same configuration mode with only 5 pmt
@jorana you will need to declare this as the gain model 
`gain_model=('CMT_model',("to_pe_model", "xenonnt_temporary_five_pmts")`

PR #303, opened 2020-12-02
pytest on github workflows

jorana:
Lately, travis is taking forever to build. This PR may kindle a new discussion to (partially) switch to github actions as those seem to run much faster than travis.

See this PR as working example.

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -26,7 +26,7 @@ jobs:
       - name: Install straxen
         run: bash .github/scripts/install_straxen.sh
       - name: Install pytest and hypothesis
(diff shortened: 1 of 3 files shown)


jorana:
I think we are seeing how much faster github actions is, perhaps we can distribute as the interface between coveralls and travis is nice and none-trivial. Therefore I propose to move some tests to github actions and skimming down travis a little (it's doing 4 platforms at the moment).

Any objections?

Oh this also adds this extra message by the github bot, if we want to, we can delete it, or, we can just delete it whenever it's in the way of the PR review.

WenzDaniel (approved):
Looks nice. Good job.

PR #306, opened 2020-12-07
allow pypi manually

jorana:
Straxen v0.12.4 has been released but is not in pypi.

This is because there was some error on uploading because travis was failing. Therefore it does not upload a new real ease until we make one. However, there is nothing wrong with this release specifically, there was just some numpy dependency for python 3.6 that was causing issues in strax. Therefore I want to add a knob that allows one to make a release on pypi whenever needed.

As I aready made this available for some other packages (
https://github.com/XENONnT/utilix/blob/master/.github/workflows/Pipi.yml,
https://github.com/XENONnT/pema/blob/main/.github/workflows/pypi_install.yml) and Evan also copied this to admix (https://github.com/XENONnT/admix/blob/master/.github/workflows/Pipi.yml) I think we can safely also add it here.

### DIFF
--- a/.github/workflows/pypi_install.yml
+++ b/.github/workflows/pypi_install.yml
@@ -1,0 +1,31 @@
+# Pipy upload straxen manually. NB: Don't do this when it's not needed! This
+# should normally work fine with travis. However, sometimes something goes
+# wrong with travis upon uploading, then it will not continue until the next
+# release is made.
+# Mostly based on https://github.com/marketplace/actions/pypi-publish
+on:
+  workflow_dispatch:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/pypi_install.yml
```diff
@@ -0,0 +1,31 @@
+# Pipy upload straxen manually. NB: Don't do this when it's not needed! This
+# should normally work fine with travis. However, sometimes something goes
+# wrong with travis upon uploading, then it will not continue until the next
+# release is made.
+# Mostly based on https://github.com/marketplace/actions/pypi-publish
+on:
+  workflow_dispatch:
```
jorana, line 7: NB! This should not state on release because travis is already doing that!

jorana:
Great success:

afbeelding

PR #310, opened 2020-12-09
Release v 0 12 5

WenzDaniel:
New release v0.12.5

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 0.12.4
+current_version = 0.12.5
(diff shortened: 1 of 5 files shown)


jorana (changes requested):
Please update the requirements for strax first

WenzDaniel:
Okay, also changed the corresponding tag. I will add this step to our "How-to".

jorana:
> Okay, also changed the corresponding tag. I will add this step to our "How-to".
Thanks, this is not always required though, only if we induce incompatibility on the strax side.

jorana (approved):
Travis will fail but that is ok. We can rerun after strax is released.

PR #127, opened 2020-06-05
Update simulation contexts.py

petergaemers:
This pull request adds/updates XENON1T and XENONnT simulation contexts.

With the, upcoming, wfsim update to support high energy channels the contexts needs to be updated to register the correct plugin

PR #311, opened 2020-12-11
Add fs for storing config files

jorana:
Add classes to strax that allow configuration files to be downloaded easily.

Take files under some given name, upload those using gridfs to our mongo database.

**Where would this be used for**
Loading files like neural nets. So you can do `straxen.get_resource('neural_net_stored_in_mongo.npy', fmt = 'npy')`

Perhaps it's easier if I just give an example:
https://github.com/XENONnT/analysiscode/blob/master/StraxTests/Store%20configs%20in%20mongo.ipynb

**Documentation**
I need to document this properly before this is merged.


straxen/mongo_storage.py
```diff
@@ -0,0 +1,358 @@
...
+
+    def __init__(self,
+                 store_files_at=None,
+                 *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        # We are going to set a place where to store the files. It's 
+        # either specified by the user or we use these defaults:
+        if store_files_at is None:
+            store_files_at = ('./resource_cache',
```
WenzDaniel, line 258: Would you like to give the user the opportunity to extend this with another path?

WenzDaniel (commented):
Hi Joran, thanks for this addition it looks very nice. I have no major comments, only one question:

I think it is taken into account but I am not 100% sure, in case someone uploads my_awesome_gains.npy and someone else modifies the file/or creates a different file but with the same name, how would this case be handled? Would user 2. overwrite and upload a new version of users 1. file? Will both files coexist? How can user 1. retrieve his/her file and not the one of user 2.?  

I would also like to ask either @darrylmasson, @ershockley  or @tunnell  to check and approve this PR since I have not any expertise in  mongo/pymongo. From my very novice-like skill level it looks thorough and fine.

straxen/mongo_storage.py
jorana: Sure, one for example might really want to specify that some files are stored elsewhere. Than this is easy to do at the init of the `Downloader`. I don't think it's a problem if we allow a user to cache files wherever the user desires.

jorana:
Hi Daniel, thanks for the review and question:
**Multiple files?**
This is important to indeed notice! So in the unlikely case that one has the credentials needed to actually make the upload, all other users would be fetching the latest file. The other file(s) would still be available (if you are prepared to do some manual ` GridFs` commands) but this would be outside of the scope I've set here. It's important to understand that we actually don't assume that these files change at all (e.g. LCE maps and the lot). If you want a new version of a file in the database, you should simply add a new name that reflects the version change. I talked to Chris and this was the experience from 1T.

There is a noticeable exception though: the `fax_config.json`. This file gets changed relatively regularly. Using this PR as a tool, you would always be using the latest file with this name. In case this is undesirable, I think it's easiest to tackle this problem on the user side (i.e. don't overwrite a file if you think it would remain useful to have two versions around). At the moment we are effectively doing this by every time making a PRs like these: (https://github.com/XENONnT/straxen/pull/300, ..., https://github.com/XENONnT/straxen/pull/127/files)

In case there would be a need to really specify a version (e.g. from a given date or so) this data would not be lost. There is at the moment no functionality in place to specify this to the `Downloader`  but this is certainly something that could be added.

**Who does these uploads**
So my plan is to write some script that automatically uploads the files (if any changed) in these repositories:
- https://github.com/XENONnT/private_nt_aux_files/
- https://github.com/XENONnT/strax_auxiliary_files/

This is at the moment working in principle, however we need to modify the RunDB API to except the comments if we e.g. run this script in a github actions automatized way. Doing any mongo insertions when you are not on `Fried rice, Uchicago or LNGS` are restricted. The RunDB API is a way around this but currently not designed to work with the operations as in this PR. Until that time its relatively little work to do it manually (see the example I linked above).

_I will add this to the PR description later for bookkeeping_

darrylmasson (commented):
This seems to be the start of an overhauled resource management system, which I feel could benefit from being bundled with the CMT (I see considerable overlap in the user experience). One general-ish comment would relate to the exact database schema - it seems we keep a copy of the metadata twice? GridFS provides most of what we're probably looking for without us needing to track metadata separately.

straxen/mongo_storage.py
```diff
@@ -0,0 +1,358 @@
...
+                'url': mongo_url,
+                'user': mongo_user,
+                'password': mongo_password,
+                'database': file_database,
+            }
+            # We can safely hard-code the collection as that is always
+            # the same with GridFS.
+            collection = utilix.rundb.pymongo_collection(
+                **mongo_kwargs,
+                collection='fs.files')
```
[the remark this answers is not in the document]
jorana: This is a GridFs feature, not much I can do about that.

straxen/mongo_storage.py
```diff
@@ -0,0 +1,358 @@
...
+        return doc
+
+    def config_exists(self, config):
+        """
+        Quick check if this config is already saved in the collection
+        :param config: str,  name of the file of interest
+        :return: bool, is this config name stored in the database
+        """
+        query = self.get_query_config(config)
+        if self.collection.count_documents(query):
```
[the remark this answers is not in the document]
jorana: Thanks!

jorana:
Thanks @darrylmasson 
> This seems to be the start of an overhauled resource management system, which I feel could benefit from being bundled with the CMT (I see considerable overlap in the user experience). 

I actually discussed with @tunnell and we decided this is not required. We need to be able to load some files but these never seem to change. If people have changing files, these are reflected in their names. With the just mentioned exception of the fax_config file. Nevertheless, people should by default get the last version. If they want another one, it's easy to download an modify (as it's JSON). If at some point one would argue that we need to track a version of the fax_config file, than we can put that in CMT directly (it's a simple JSON), but also this I consider unlikely.

> One general-ish comment would relate to the exact database schema - it seems we keep a copy of the metadata twice? 

This is the grid fs feature. It creates a map between `fs.files` and `fs.chunks`. You only need to communicate with `fs.files` to know what is in there. GridFs does the matching with `fs.chunks` for you.

>GridFS provides most of what we're probably looking for without us needing to track metadata separately.

Yes, and it is exactly what I'm using 😄

straxen/mongo_storage.py
```diff
@@ -0,0 +1,358 @@
...
+            # https://github.com/XENONnT/utilix/blob/master/utilix/rundb.py
+            mongo_kwargs = {
+                'url': mongo_url,
+                'user': mongo_user,
+                'password': mongo_password,
+                'database': file_database,
+            }
+            # We can safely hard-code the collection as that is always
+            # the same with GridFS.
+            collection = utilix.rundb.pymongo_collection(
```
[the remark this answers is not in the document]
jorana: Sure, better fail hard than not having any clue what failed.

WenzDaniel:
> So in the unlikely case that one has the credentials needed to actually make the upload, all other users would be fetching the latest file.

Okay thanks, this was the point I was missing here. So only a few number of selected people can do uploads? I think in that case it is fail safe.

jorana:
> Okay thanks, this was the point I was missing here. So only a few number of selected people can do uploads? I think in that case it is fail safe.

Good point, yes, there are very few people with the power to do this.

straxen/common.py
```diff
@@ -76,134 +76,160 @@ def pmt_positions(xenon1t=False):
...
+    :param fmt: str, format of the resource x
+    :return: the opened resource file x opened according to the
+        specified format
     """
-    if straxen.uconfig is not None:
+    # 1. load from memory
+    if x in _resource_cache:
+        return _resource_cache[x]
+    # 2. load from file
+    elif os.path.exists(x):
```
darrylmasson, line 151: The `elif` construction doesn't really do anything here, but that's borderline stylistic.

darrylmasson (approved):
There's probably an edge case somewhere that we haven't considered correctly, but in lieu of finding it I think it looks pretty good.

straxen/common.py
jorana: valid

PR #312, opened 2020-12-12
n_veto and mu_veto gains into corrections

ahiguera-mx:

Adding n_veto and mu_veto gains into corrections
We can access the same way as we do for the tpc pmt gains via the function to_pe
Please look at 
https://github.com/XENONnT/analysiscode/blob/master/corrections/get_gains_for_any_det.ipynb
  - Update the docstring(s)
  - Update the documentation

### DIFF
--- a/straxen/corrections_services.py
+++ b/straxen/corrections_services.py
@@ -74,7 +74,11 @@ class CorrectionsManagementServices():
         model_type, global_version = config_model
 
         if correction == 'pmt_gains':
-            return self.get_pmt_gains(run_id, model_type, global_version)
+            return self.get_pmt_gains(run_id, model_type, global_version, 'tpc')
+        elif correction == 'n_veto_pmt_gains':
+            return self.get_pmt_gains(run_id, model_type, global_version, 'n_veto')
+        elif correction == 'mu_veto_pmt_gains':
+            return self.get_pmt_gains(run_id, model_type, global_version, 'mu_veto')
         elif correction == 'elife':
             return self.get_elife(run_id, model_type, global_version)
         else:
@@ -141,7 +145,7 @@ class CorrectionsManagementServices():
             raise ValueError(f'model type {model_type} not implemented for electron lifetime')
 
     def get_pmt_gains(self, run_id, model_type, global_version,
-                      cacheable_versions=('ONLINE',),
+                      detector, cacheable_versions=('ONLINE',),
                       gain_dtype=np.float32):
         """
         Smart logic to return pmt gains to PE values.
@@ -159,14 +163,19 @@ class CorrectionsManagementServices():
         if model_type == 'to_pe_model':
             if global_version in cacheable_versions:
                 # Try to load from cache, if it does not exist it will be created below
-                cache_name = cacheable_naming(run_id, model_type, global_version)
+                cache_name = cacheable_naming(run_id, model_type, global_version, detector)
                 try:
                     to_pe = straxen.get_resource(cache_name, fmt='npy')
                 except (ValueError, FileNotFoundError):
                     pass
 
             if to_pe is None:
-                to_pe = self._get_correction(run_id, 'pmt', global_version)
+                if detector == 'tpc':
+                    to_pe = self._get_correction(run_id, 'pmt', global_version)
+                elif detector == 'n_veto':
+                    to_pe = self._get_correction(run_id, 'n_veto', global_version)
+                elif detector == 'mu_veto':
+                    to_pe = self._get_correction(run_id, 'mu_veto', global_version)
 
             # be cautious with very early runs, check that not all are None
             if np.isnan(to_pe).all():
--- a/straxen/corrections_services.py
+++ b/straxen/corrections_services.py
@@ -166,14 +166,14 @@ class CorrectionsManagementServices():
 
             if global_version in cacheable_versions:
                 # Try to load from cache, if it does not exist it will be created below
-                cache_name = cacheable_naming(run_id, model_type, global_version, target_detector)
+                cache_name = cacheable_naming(run_id, model_type, global_versionr)
                 try:
                     to_pe = straxen.get_resource(cache_name, fmt='npy')
                 except (ValueError, FileNotFoundError):
                     pass
 
-                if to_pe is None:
-                    to_pe = self._get_correction(run_id, target_detector, global_version)
+            if to_pe is None:
+                to_pe = self._get_correction(run_id, target_detector, global_version)
 
             # be cautious with very early runs, check that not all are None
             if np.isnan(to_pe).all():
--- a/straxen/corrections_services.py
+++ b/straxen/corrections_services.py
@@ -60,11 +60,10 @@ class CorrectionsManagementServices():
         return str(f'{"XENONnT " if self.is_nt else "XENON1T"}'
                    f'-Corrections_Management_Services')
 
-    def get_corrections_config(self, run_id, correction=None, config_model=None):
+    def get_corrections_config(self, run_id, config_model=None):
         """
         Get context configuration for a given correction
         :param run_id: run id from runDB
-        :param correction: correction's name (str type)
         :param config_model: configuration model (tuple type)
         :return: correction value(s)
         """
@@ -73,19 +72,15 @@ class CorrectionsManagementServices():
             raise ValueError(f'config_model {config_model} must be a tuple')
         model_type, global_version = config_model
 
-        if correction == 'pmt_gains':
-            return self.get_pmt_gains(run_id, model_type, global_version, 'tpc')
-        elif correction == 'n_veto_pmt_gains':
-            return self.get_pmt_gains(run_id, model_type, global_version, 'n_veto')
-        elif correction == 'mu_veto_pmt_gains':
-            return self.get_pmt_gains(run_id, model_type, global_version, 'mu_veto')
-        elif correction == 'elife':
+        if 'to_pe_model' in model_type:
+            return self.get_pmt_gains(run_id, model_type, global_version)
+        elif 'elife' in model_type:
             return self.get_elife(run_id, model_type, global_version)
         else:
             raise ValueError(f'{correction} not found')
 
     # TODO add option to extract 'when'. Also, the start time might not be the best
-    #  entry for e.g. for super runs
+    # entry for e.g. for super runs
     # cache results, this would help when looking at the same gains
     @lru_cache(maxsize=None)
     def _get_correction(self, run_id, correction, global_version):
@@ -127,7 +122,7 @@ class CorrectionsManagementServices():
         Smart logic to return electron lifetime correction
         :param run_id: run id from runDB
         :param model_type: choose either elife_model or elife_constant
-        :param global_version: global version, or float (if model_type == elife_constant) 
+        :param global_version: global version, or float (if model_type == elife_constant)
         :return: electron lifetime correction value
         """
         if model_type == 'elife_model':
@@ -145,7 +140,7 @@ class CorrectionsManagementServices():
             raise ValueError(f'model type {model_type} not implemented for electron lifetime')
 
     def get_pmt_gains(self, run_id, model_type, global_version,
-                      detector, cacheable_versions=('ONLINE',),
+                      cacheable_versions=('ONLINE',),
                       gain_dtype=np.float32):
         """
         Smart logic to return pmt gains to PE values.
@@ -160,22 +155,25 @@ class CorrectionsManagementServices():
         to_pe = None
         cache_name = None
 
-        if model_type == 'to_pe_model':
+        if 'to_pe_model' in model_type:
+            # Get the detector name based on the requested model_type
+            # This also will be used to the cachable name convention
+            # pmt == TPC, n_veto == n_veto's PMT, etc
+            detector_names = {'to_pe_model': 'pmt',
+                              'to_pe_model_nv': 'n_veto',
+                              'to_pe_model_mv': 'mu_veto'}
+            target_detector = detector_names[model_type]
+
             if global_version in cacheable_versions:
                 # Try to load from cache, if it does not exist it will be created below
-                cache_name = cacheable_naming(run_id, model_type, global_version, detector)
+                cache_name = cacheable_naming(run_id, model_type, global_version, target_detector)
                 try:
                     to_pe = straxen.get_resource(cache_name, fmt='npy')
                 except (ValueError, FileNotFoundError):
                     pass
 
-            if to_pe is None:
-                if detector == 'tpc':
-                    to_pe = self._get_correction(run_id, 'pmt', global_version)
-                elif detector == 'n_veto':
-                    to_pe = self._get_correction(run_id, 'n_veto', global_version)
-                elif detector == 'mu_veto':
-                    to_pe = self._get_correction(run_id, 'mu_veto', global_version)
+                if to_pe is None:
+                    to_pe = self._get_correction(run_id, target_detector, global_version)
 
             # be cautious with very early runs, check that not all are None
             if np.isnan(to_pe).all():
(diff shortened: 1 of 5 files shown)


straxen/corrections_services.py
```diff
@@ -74,7 +74,11 @@ def get_corrections_config(self, run_id, correction=None, config_model=None):
         model_type, global_version = config_model
 
         if correction == 'pmt_gains':
-            return self.get_pmt_gains(run_id, model_type, global_version)
+            return self.get_pmt_gains(run_id, model_type, global_version, 'tpc')
```
jorana, line 77: Perhaps I'm overseeing something, but would the `global_version` not already make sure that you distinguish between tpc, MV and NV? I think you can keep the code somewhat simpler in this case by not having to add a new argument.

jorana: As in I assume it's now:
```python
#TPC 
   gain_model=("CMT_model", ("to_pe_model", "ONLINE")),
# MV
   gain_model=("CMT_model", ("to_pe_model_NV", "ONLINE")),
# NV
   gain_model=("CMT_model", ("to_pe_model_MV", "ONLINE")),
```
Otherwise I think it should be like this. If the global version is already like this, there is no need to write 3 extra `elif`-statements right?

straxen/corrections_services.py
```diff
@@ -159,14 +163,19 @@ def get_pmt_gains(self, run_id, model_type, global_version,
...
+                cache_name = cacheable_naming(run_id, model_type, global_version, detector)
                 try:
                     to_pe = straxen.get_resource(cache_name, fmt='npy')
                 except (ValueError, FileNotFoundError):
                     pass
 
             if to_pe is None:
-                to_pe = self._get_correction(run_id, 'pmt', global_version)
+                if detector == 'tpc':
+                    to_pe = self._get_correction(run_id, 'pmt', global_version)
```
jorana, line 174: See my comment above

straxen/get_corrections.py
```diff
@@ -8,7 +8,7 @@
 
 
 @export
-def get_to_pe(run_id, gain_model, n_tpc_pmts):
+def get_to_pe(run_id, gain_model, n_pmts):
```
jorana, line 11: good catch!

straxen/get_corrections.py
```diff
@@ -31,9 +31,17 @@ def get_to_pe(run_id, gain_model, n_tpc_pmts):
             raise ValueError('CMT gain model should be similar to:'
                              '("CMT_model", ("to_pe_model", "v1"). Instead got:'
                              f'{model_conf}')
-        is_nt = n_tpc_pmts == straxen.n_tpc_pmts
+        # is this the best way to do this?
+        is_nt = n_pmts == straxen.n_tpc_pmts or n_pmts == straxen.n_nveto_pmts or n_pmts == straxen.n_mveto_pmts
```
jorana, line 35: I think it's intuitive enough. For readability you might want to do something like:

```python
is_nt = n_pmts in [straxen.n_tpc_pmts, straxen.n_nveto_pmts, straxen.n_mveto_pmts]
```

straxen/get_corrections.py
```diff
@@ -31,9 +31,17 @@ def get_to_pe(run_id, gain_model, n_tpc_pmts):
...
+
         corrections = straxen.CorrectionsManagementServices(is_nt=is_nt)
-        to_pe = corrections.get_corrections_config(run_id, 'pmt_gains', model_conf)
+        if n_pmts == straxen.n_tpc_pmts:
```
jorana, line 38: See earlier comment

jorana (commented):
Hi Aaron! In principle it looks very good!

I think you can make it somewhat easier if you just use the global version to distinguish between the different detectors. As you have made that already in the CMT database it's easy to also exploit that here.

ahiguera-mx:
> Hi Aaron! In principle it looks very good!
> 
> I think you can make it somewhat easier if you just use the global version to distinguish between the different detectors. As you have made that already in the CMT database it's easy to also exploit that here.

Hi @jorana, thanks for the feedback, and just one clarification. The global version is only a collection of corrections, you still have to specify which correction (dataFrame) to look at, you can see this from when we specify  `pmt` or `elife` what is true is you just need to specify a global version for the gain model as `gain_model=('CMT_model',("to_pe_model", "ONLINE")))`. In the example, I specify one for every detector for completeness, but we can specify only one, then `corrections_services.py` would look at the particular correction

straxen/corrections_services.py
ahiguera-mx: see comment below

straxen/corrections_services.py
ahiguera-mx: see comment below

straxen/corrections_services.py
WenzDaniel: Aaron, I think I do not understand your answer to Joran comment. Why is it not possible to change the code as Joran propose? Indeed, I also think it is more elegant to have a structure such as:
```
Version, Online:
          | -> to_pe_tpc
          |-> to_pe_nv
          |-> elife
Version SR0:
          |-> to_pe_tpc .... 
```

straxen/get_corrections.py
WenzDaniel: Also this does not work, at least not for the muon-veto. In case of the muon-veto it is the same number of PMTs in 1T and nT.

WenzDaniel:
Hej Aaron, thanks a lot for the changes. I unresolved two comments by Joran though, could you comment on my questions there?

straxen/get_corrections.py
jorana: Sorry Daniel but I don't follow. This is where you get the argument that you need to be able to do the init in the first place.

>  In case of the muon-veto it is the same number of PMTs in 1T and nT.
True, but there is no such thing at present as `strax xenon1T MV data` . I'm not aware of any plans of porting this data to the strax format but perhaps you know more than I do?

jorana:
Thanks Aaron for the clarification. I understand it a bit better now. The key difference is that you anticipated that we want to do this:
```python
   gain_model=("CMT_model", ("to_pe_model", "ONLINE")),
```

To capture all the gains for all thee detectors in a single option. This is perhaps not the most locial choice right because you would have to reprocess data always whereas for a given detector it might turn out that we want to use the `ONLINE` model because it was already quite good but for another one we want to use a `v2`.

Therefore, I would argue that you want to have three separate options:


```python
   gain_model=("CMT_model", ("to_pe_model", "ONLINE")),
   gain_model_nv=("CMT_model", ("to_pe_model_NV", "ONLINE")),
   gain_model_mv=("CMT_model", ("to_pe_model_MV", "ONLINE"))
```

This would mean you only have to change these lines as follows (L163 and further):
https://github.com/XENONnT/straxen/blob/b39c2b47e79bea8c9587368ef4974ef9c1bdf6e2/straxen/corrections_services.py#L163
```python
        if 'to_pe_model' in model_type: # CHANGED
            if global_version in cacheable_versions:
                # Try to load from cache, if it does not exist it will be created below
                cache_name = cacheable_naming(run_id, model_type, global_version) # CHANGED
                try:
                    to_pe = straxen.get_resource(cache_name, fmt='npy')
                except (ValueError, FileNotFoundError):
                    pass

            if to_pe is None:
			# Get the detector name based on the requested model_type
                detector_names = {'to_pe_model': 'pmt', 
                'to_pe_model_nv': 'n_veto',  
                'to_pe_model_mv': 'mu_veto'}
                target_dectector = detector_names[model_type]
                # We will string match the configuration options in the requested 
                # version to the target_dectector. (E.g. "pmt" in "pmt_006_gain_xenonnt"->True or 
                # "pmt" in "n_veto_037_gain_xenonnt"->False)
                to_pe = self._get_correction(run_id, target_dectector, global_version)
             
``` 

Perhaps I missed one or two instances where you need to change in order to accept `to_pe_model*` but this would keep it much simpler such that you only have to specify the detector at the very last place (thereby keeping the code simpler).

straxen/get_corrections.py
ahiguera-mx: Daniel as Joran mentioned we need to do the init to look either nT or 1T runDB

ahiguera-mx:
> Thanks Aaron for the clarification. I understand it a bit better now. The key difference is that you anticipated that we want to do this:
> 
> ```python
>    gain_model=("CMT_model", ("to_pe_model", "ONLINE")),
> ```
> 
> To capture all the gains for all thee detectors in a single option. This is perhaps not the most locial choice right because you would have to reprocess data always whereas for a given detector it might turn out that we want to use the `ONLINE` model because it was already quite good but for another one we want to use a `v2`.
> 
> Therefore, I would argue that you want to have three separate options:
> 
> ```python
>    gain_model=("CMT_model", ("to_pe_model", "ONLINE")),
>    gain_model_nv=("CMT_model", ("to_pe_model_NV", "ONLINE")),
>    gain_model_mv=("CMT_model", ("to_pe_model_MV", "ONLINE"))
> ```
> 
> This would mean you only have to change these lines as follows (L163 and further):
> 
> https://github.com/XENONnT/straxen/blob/b39c2b47e79bea8c9587368ef4974ef9c1bdf6e2/straxen/corrections_services.py#L163
> 
> ```python
>         if 'to_pe_model' in model_type: # CHANGED
>             if global_version in cacheable_versions:
>                 # Try to load from cache, if it does not exist it will be created below
>                 cache_name = cacheable_naming(run_id, model_type, global_version) # CHANGED
>                 try:
>                     to_pe = straxen.get_resource(cache_name, fmt='npy')
>                 except (ValueError, FileNotFoundError):
>                     pass
> 
>             if to_pe is None:
> 			# Get the detector name based on the requested model_type
>                 detector_names = {'to_pe_model': 'pmt', 
>                 'to_pe_model_nv': 'n_veto',  
>                 'to_pe_model_mv': 'mu_veto'}
>                 target_dectector = detector_names[model_type]
>                 # We will string match the configuration options in the requested 
>                 # version to the target_dectector. (E.g. "pmt" in "pmt_006_gain_xenonnt"->True or 
>                 # "pmt" in "n_veto_037_gain_xenonnt"->False)
>                 to_pe = self._get_correction(run_id, target_dectector, global_version)
>              
> ```
> 
> Perhaps I missed one or two instances where you need to change in order to accept `to_pe_model*` but this would keep it much simpler such that you only have to specify the detector at the very last place (thereby keeping the code simpler).

straxen/corrections_services.py
ahiguera-mx: I think I(we) was(were) confused by the comment, see below

ahiguera-mx:
> To capture all the gains for all thee detectors in a single option. This is perhaps not the most locial choice right because you would have to reprocess data always whereas for a given detector it might turn out that we want to use the `ONLINE` model because it was already quite good but for another one we want to use a `v2`.
> 
> Therefore, I would argue that you want to have three separate options:
> 
> ```python
>    gain_model=("CMT_model", ("to_pe_model", "ONLINE")),
>    gain_model_nv=("CMT_model", ("to_pe_model_NV", "ONLINE")),
>    gain_model_mv=("CMT_model", ("to_pe_model_MV", "ONLINE"))
> ```
> 

Yes you are right, having every gain separately is the right approach 

> This would mean you only have to change these lines as follows (L163 and further):
> 
> https://github.com/XENONnT/straxen/blob/b39c2b47e79bea8c9587368ef4974ef9c1bdf6e2/straxen/corrections_services.py#L163
> 
> ```python
>         if 'to_pe_model' in model_type: # CHANGED
>             if global_version in cacheable_versions:
>                 # Try to load from cache, if it does not exist it will be created below
>                 cache_name = cacheable_naming(run_id, model_type, global_version) # CHANGED
>                 try:
>                     to_pe = straxen.get_resource(cache_name, fmt='npy')
>                 except (ValueError, FileNotFoundError):
>                     pass
> 
>             if to_pe is None:
> 			# Get the detector name based on the requested model_type
>                 detector_names = {'to_pe_model': 'pmt', 
>                 'to_pe_model_nv': 'n_veto',  
>                 'to_pe_model_mv': 'mu_veto'}
>                 target_dectector = detector_names[model_type]
>                 # We will string match the configuration options in the requested 
>                 # version to the target_dectector. (E.g. "pmt" in "pmt_006_gain_xenonnt"->True or 
>                 # "pmt" in "n_veto_037_gain_xenonnt"->False)
>                 to_pe = self._get_correction(run_id, target_dectector, global_version)
>              
> ```
> 

Indeed, this is a very elegant solution, which I implemented and include further, this made the code looks cleaner, with the new change we only need to read the model_type info to infer which correction

straxen/get_corrections.py
WenzDaniel: > Sorry Daniel but I don't follow. This is where you get the argument that you need to be able to do the init in the first place.

Yes, sorry I confused myself here. 

> True, but there is no such thing at present as strax xenon1T MV data . I'm not aware of any plans of porting this data to the strax format but perhaps you know more than I do?

I am not aware of this for the moment. But I think it is not unlikely that you will compare at some point Strax with PAX also for the muon-veto. 

> Daniel as Joran mentioned we need to do the init to look either nT or 1T runDB

True, I understand this, but I do not like this semi-correct solution here which works only because we think non will ever use it. I think also for other reasons it would be actually nicer if the initialization of CMT becomes part of the contexts. Maybe not needed to change it right away in this PR, but at least I think it is worth to think about it, is not it? 
As a fall back solution one could do it in a similar way as in the electron life time.

WenzDaniel:
Hej Aaron, thanks for the change. Looks good now, just think a bit about the 1T/nT thing.

straxen/corrections_services.py
```diff
@@ -156,17 +155,25 @@ def get_pmt_gains(self, run_id, model_type, global_version,
...
+            # pmt == TPC, n_veto == n_veto's PMT, etc
+            detector_names = {'to_pe_model': 'pmt',
+                              'to_pe_model_nv': 'n_veto',
+                              'to_pe_model_mv': 'mu_veto'}
+            target_detector = detector_names[model_type]
+
             if global_version in cacheable_versions:
                 # Try to load from cache, if it does not exist it will be created below
-                cache_name = cacheable_naming(run_id, model_type, global_version)
+                cache_name = cacheable_naming(run_id, model_type, global_version, target_detector)
```
jorana, line 169: Technically, you don't really need the `target_detector` as this is already in the `model_type`. I would remove it (target_detector).

straxen/corrections_services.py
```diff
@@ -156,17 +155,25 @@ def get_pmt_gains(self, run_id, model_type, global_version,
...
                 try:
                     to_pe = straxen.get_resource(cache_name, fmt='npy')
                 except (ValueError, FileNotFoundError):
                     pass
 
-            if to_pe is None:
-                to_pe = self._get_correction(run_id, 'pmt', global_version)
+                if to_pe is None:
```
jorana, line 175: Not sure if this is by accident but this is not what you want. Otherwise you cannot load `v1` configs.

straxen/get_corrections.py
jorana: I see your point. Given that there is neither 1T gains nor 1T strax data for the MV I would not worry about it too much at this stage.

You raise a valid point though that we could think of something that keeps the CMT initialized for the right detector at ~ the context level.

jorana (changes requested):

straxen/corrections_services.py
jorana: afbeelding

straxen/corrections_services.py
ahiguera-mx: yes, you are right

straxen/corrections_services.py
ahiguera-mx: not sure what do you mean here @jorana, can you elaborate more, please

ahiguera-mx: never mind, yes there was an extra indent, good catch!!

jorana: You over indented this, you should not put it under the `if global_version in cacheable_versions:` if-statement.

If you follow the logic, you want to hit the ` if to_pe is None:` (which you want to do for everything unless you can read it from a file) you cannot do this for e.g. v1 because it is not cacheable.

jorana (approved):

jorana:
Thanks Aaron!

Issue #269, opened 2020-11-07
Bootstrax wishlist 


jorana:
Run_strax in bootstrax is super messy. Should should be clearer. Just pass a config-dict rather than all these un-transparent arguments.

- [x] Fix https://github.com/XENONnT/straxen/commit/d0c35379fe212d92016fd0291b95b303a022c17a to look nicer
- [x] Rethink logic to do fall-backs
- [x] interpolate infer_mode
- [x] change scheme to go down to lower plugins for processing, don't just drop back to raw_records.
- [x] fix `MergeTargets` with inconsistent times?! 011413
- [x] Put target determination in override_target as we otherwise process high level data for led plugins
- [x] Allow eb0 and eb1 to do the tags to give more time for the other ebs to process
- [x] Do failed runs on old eventbuilders
- [x] have ajax report to the database
- [ ] save entire log of run and delete when successful. Keep track of the xth failure
- [x] disk-size should be sufficient for at least a 1 TB run
- [x] it looks like bad tagging a run causes a logic failure in the processing, crashing the eb to it 7200 s timeout
   - this might be due to actually being bad data.
- [x] add a `reason` to deleted data-field
- [ ] Bad tagging is fatal
- [x] Replace `target` with targets, and change subdetectors to `post-processing-targets`. This allows to distinguish between online targets and targets that can be computed in a second step. Will most likely be much stabler.
- [x] Don't allow target to be specified twice
- [x] Write infer_mode to a csv file because it's just a bunch of numbers
- [x] Improve NV integration
   - [x] Process up to hitlets_nv

PR #313, opened 2020-12-14
Faster bootstrax queries for new runs

darrylmasson:

The `$exists` query operator in MongoDB is slow because it can't be satisfied in the usual `log(n)` time as a standard indexed query. As the runs database continues to grow in size inefficient queries will rapidly get slower and slower. Also, bootstrax should get patched to match https://github.com/coderdj/redax/pull/109

Rather than continue to use slow operators, a minor change in the DB schema allows for more efficient index usage.

This partially implements #269, depending on the scope of changes necessary

### DIFF
--- a/bin/bootstrax
+++ b/bin/bootstrax
@@ -392,7 +392,7 @@ def main_loop():
         # Check resources are still OK, otherwise crash / reboot program
 
         # Process new runs
-        rd = consider_run({"bootstrax.state": {"$exists": False}},
+        rd = consider_run({"bootstrax.state": None},
                           test_counter=new_runs_seen)
         if rd is not None:
             new_runs_seen += 1


jorana:
Hi Darryl, thanks for starting this PR, I have a branch somewhere that started on this as well after we had our discussion. Glad to see you had the time to add this to redax. Let me open a PR after this to merge the branch I had started too.

jorana (approved):
Looks good

darrylmasson:
Scooped!

I'm just trying to clean up remaining loose ends in redax before we change repositories

darrylmasson:
We'll have to synchronize the deployment, but that's relatively trivial

jorana:
Just let me know when you are doing the new deployment I can do straxen accordingly.

PR #314, opened 2020-12-14
Bootstrax mongo interaction

jorana:
Mongo is vital to our computing infrastructure. Like in #313 we should treat it carefully and be smart about our queries.

Remove some obsolete statements and tweak others.


darrylmasson (approved):
I like it

PR #320, opened 2020-12-23
S2 CNN posrec

terliuk:
S2 CNN position reconstruction. To get uncorrected S2 position register the module as (example):

```
st.set_config(dict(s2_cnn_model_path=model_path))
st.register(straxen.CNNS2PosRec)
```
And to run the reconstruction
```
recopos = st.get_array(run_id, "CNNS2PosRec")
```
Resulting array has `x_TFS2CNN` and `y_TFS2CNN` for x and y position. For more elaborate example see: 
https://github.com/XENONnT/analysiscode/blob/s2_cnn_posrec/s2_cnn_scripts/CNNS2ExampleInStraxen.ipynb

### DIFF
--- a/straxen/plugins/__init__.py
+++ b/straxen/plugins/__init__.py
@@ -38,3 +38,6 @@ from .acqmon_processing import *
 
 from . import online_monitor
 from .online_monitor import *
+
+from . import s2cnn_processing
+from .s2cnn_processing import *
--- a/straxen/plugins/s2cnn_processing.py
+++ b/straxen/plugins/s2cnn_processing.py
@@ -1,5 +1,6 @@
 try:
-    import strax, straxen
+    import strax
+    import straxen
     straxfound=True
 except ModuleNotFoundError:
     print("WARNING! No strax found! Strax module will not be loaded.")
@@ -147,7 +148,7 @@ if straxfound:
             # renormalizing since CNNs are done normalized to PMT with the largest area   
             result['patterns'][peak_mask] = patterns
             reco_pos = self.cnn_model.predict(patterns)
-            # XXX: I assume that all the networks return cm
+            # I assume that all the networks return cm
             result['x_TFS2CNN'][peak_mask] = reco_pos[:, 0]
             result['y_TFS2CNN'][peak_mask] = reco_pos[:, 1]
             return result
--- a/straxen/plugins/s2cnn_processing.py
+++ b/straxen/plugins/s2cnn_processing.py
@@ -85,8 +85,11 @@ export, __all__ = strax.exporter()
                  default=10),
     strax.Option("s2_cnn_model_path", 
                  help="Path to the CNN model in hdf5 format. WARING, this should include the whole model file and not the weights file", 
-                 default=("CNN_dw_3L_cm_maxnorm_A_lin_5_2000__Z_const_-1.hdf5")
-                )
+                 default="CNN_dw_3L_cm_maxnorm_A_lin_5_2000__Z_const_-1.hdf5"
+                ),
+    strax.Option("store_cnn_patterns",
+                 help="Store patterns as used by CNN",
+                 default=True)
 )
 class CNNS2PosRec(strax.Plugin):
     """
@@ -97,18 +100,22 @@ class CNNS2PosRec(strax.Plugin):
         - y_TFS2CNN - reconstructed Y position in [ cm ]
         - patterns - 2D array of normalized areas as used for CNN
     """
-    dtype = [('x_TFS2CNN', np.float32,
-              'Reconstructed CNN S2 X position [ cm ] '),
-             ('y_TFS2CNN', np.float32,
-              'Reconstructed CNN S2 Y position [ cm ] '), 
-             (("Patterns after DW transformation and normalization as used in CNN model",
-                 "patterns"), np.float, (33, 19,)), 
-             ] 
-    dtype += strax.time_fields
+
     depends_on = ('peaks',)
     parallel = False
     provides = "CNNS2PosRec"
-    __version__ = '0.0.0'  
+    __version__ = '0.0.1'
+    def infer_dtype(self):
+        dtype = [('x_TFS2CNN', np.float32,
+                  'Reconstructed CNN S2 X position [ cm ] '),
+                 ('y_TFS2CNN', np.float32,
+                  'Reconstructed CNN S2 Y position [ cm ] '),
+                 ]
+        if self.config['store_cnn_patterns']:
+            dtype.append( (("Patterns after DW transformation and normalization as used in CNN model",
+                     "patterns"), np.float, (33, 19,)) )
+        dtype += strax.time_fields
+        return dtype
     
     def setup(self):
         import tensorflow as tf
@@ -120,7 +127,7 @@ class CNNS2PosRec(strax.Plugin):
             downloader = straxen.MongoDownloader()
             if self.s2_cnn_model_path in downloader.list_files():
                 self.s2_cnn_model_path = downloader.download_single(self.s2_cnn_model_path)
-                print("Path do downloaded file from database : \n\t : %s"%self.s2_cnn_model_path)
+                print("Path to downloaded file from database : \n\t : %s"%self.s2_cnn_model_path)
             else: 
                 raise RuntimeError("Model file not found (locally or in DB): %s"%self.s2_cnn_model_path) 
         self.cnn_model = keras.models.load_model(self.s2_cnn_model_path)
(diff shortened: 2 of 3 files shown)


straxen/plugins/__init__.py
```diff
@@ -38,3 +38,6 @@
 
 from . import online_monitor
 from .online_monitor import *
+
+from . import s2cnn_processing
```
jorana, line 42: Hi Andrii, can you please put this functionality in Peak processing? I think that is the appropriate place.

straxen/plugins/s2cnn_processing.py
```diff
@@ -0,0 +1,154 @@
+try:
```
jorana, line 1: I guess you wrote this to work in some other environment but this should not go in straxen, can you remove this kind of logic?

straxen/plugins/s2cnn_processing.py
```diff
@@ -0,0 +1,154 @@
...
+    @export
+    @strax.takes_config(
+        strax.Option('min_reconstruction_area',
+                     help='Skip reconstruction if area (PE) is less than this',
+                     default=10),
+        strax.Option('n_top_pmts', default=straxen.n_top_pmts,
+                     help="Number of top PMTs"),
+        strax.Option("s2_cnn_model_path", 
+                     help="Path to the CNN model in hdf5 format. WARING, this should include the whole model file and not the weights file", 
+                     default=("/project2/lgrandi/terliuk/CNNmodels/CNN_defaults/disk_const_PE/CNN_dw_3L_cm_maxnorm_A_const_100__Z_const_-1.hdf5")
```
jorana, line 91: Please open a PR to https://github.com/XENONnT/private_nt_aux_files such that we can store this file in the database.

straxen/plugins/s2cnn_processing.py
```diff
@@ -0,0 +1,154 @@
...
+            - x_TFS2CNN - reconstructed X position in [ cm ]
+            - y_TFS2CNN - reconstructed Y position in [ cm ]
+            - patterns - 2D array of normalized areas as used for CNN
+        """
+        dtype = [('x_TFS2CNN', np.float32,
+                  'Reconstructed CNN S2 X position [ cm ] '),
+                 ('y_TFS2CNN', np.float32,
+                  'Reconstructed CNN S2 Y position [ cm ] '), 
+                 (("Patterns after DW transformation normalized to max PMT area",
+                     "patterns"), np.float, (33, 19,)),
```
jorana, line 108: Better make this an option?

straxen/plugins/s2cnn_processing.py
```diff
@@ -0,0 +1,154 @@
...
+                     help='Skip reconstruction if area (PE) is less than this',
+                     default=10),
+        strax.Option('n_top_pmts', default=straxen.n_top_pmts,
+                     help="Number of top PMTs"),
+        strax.Option("s2_cnn_model_path", 
+                     help="Path to the CNN model in hdf5 format. WARING, this should include the whole model file and not the weights file", 
+                     default=("/project2/lgrandi/terliuk/CNNmodels/CNN_defaults/disk_const_PE/CNN_dw_3L_cm_maxnorm_A_const_100__Z_const_-1.hdf5")
+                    )
+    )
+    class CNNS2PosRec(strax.Plugin):
```
jorana, line 94: Have you considered just implementing this in the PeakPositions plugin? Perhaps we should have a chat because if we want to use your implementation, we should just incorporate it in that plugin.

straxen/plugins/s2cnn_processing.py
jorana, line 91: I assume this only works for nT?

straxen/plugins/s2cnn_processing.py
```diff
@@ -0,0 +1,154 @@
...
+            print("====== end of model summary =====")
+
+        def compute(self, peaks):
+            result = np.ones(len(peaks), dtype=self.dtype)
+            result['time'], result['endtime'] = peaks['time'], strax.endtime(peaks)
+            result['x_TFS2CNN'] *= np.nan
+            result['y_TFS2CNN'] *= np.nan
+            result['patterns'] *=np.nan
+
+            peak_mask = peaks['area'] > self.config['min_reconstruction_area']
```
jorana, line 141: You want to make sure you are using S2s here right?

straxen/plugins/s2cnn_processing.py
```diff
@@ -0,0 +1,154 @@
...
+                else: 
+                    raise RuntimeError("Model file not found (locally or in DB): %s"%self.s2_cnn_model_path) 
+            self.cnn_model = keras.models.load_model(self.s2_cnn_model_path)
+            self.converter = DoubleWidthConverter()
+            print("====== Loaded TF CNN model =====")
+            self.cnn_model.summary()
+            print("====== end of model summary =====")
+
+        def compute(self, peaks):
+            result = np.ones(len(peaks), dtype=self.dtype)
```
jorana, line 135: You can also do:
```python
result = np.empty(len(peaks), dtype=self.dtype)
result[:] = np.NaN
```

straxen/plugins/s2cnn_processing.py
```diff
@@ -0,0 +1,154 @@
...
+    straxfound=True
+except ModuleNotFoundError:
+    print("WARNING! No strax found! Strax module will not be loaded.")
+    print("Double width converter is still available. ")
+    straxfound=False
+import numpy as np
+from copy import deepcopy
+import os 
+
+class DoubleWidthConverter():
```
jorana, line 13: I'm not sure if I understand what this class does. Since it seems to produce a map, would you say it's easy to export this map and read it in later? That would keep the code perhaps somewhat simpler, and, if you later change how this works internally you only have to update the map file. What do you think?

jorana (changes requested):
Hi Andrii,
Thanks for this PR, we are very much looking forward to implementing this in straxen and being able to use your nice PosRec algo for xenonnt. Thanks a lot for adding the notebook that clearly demonstrates this is a must-have feature!

I do think there are some things that still require some work to implement this properly in straxen, please see some of the comments below. 

Most importantly, I think we want to add this as an option to be used in the PeakPositions-plugin such that we can use it easily when needed (at the moment I think it's only nT right)? Perhaps we should have a chat together (e.g. after christmas ;) ) such that we can discuss the easiest way to implement this in straxen.

Finally, I think the notebook clearly shows that it does an excellent job on what it's supposed to do: reconstruct the position. Do you have some idea of if its performance w.r.t. what is currently implemented and how fast this algo is (w.r.t. the current implementation)? This is important to know because of a comment Chris made about these algos being the slowest in PAX, it may determine how we should set up the processing in the straxen framework and is therefore very relevant for the question I asked above.

jorana:
@Andrii, sorry about (86e3eb0  - 39e7908). We just ran into: https://github.com/AxFoundation/strax/issues/229 and I figured better to fix it right away/

straxen/plugins/s2cnn_processing.py
terliuk: Correct, only nT has regular hexagonal "pattern", so we can only use it there.

straxen/plugins/s2cnn_processing.py
terliuk: Not entirely sure about proper location. I am a bit concerned that the file itself grows really large and I think it makes sense to make keep it separately (probably, create modules subfolder). Also, there are other position reconstruction algorithms on the way, what can make it even more complicated.

straxen/plugins/s2cnn_processing.py
terliuk: Do you mean storage of patterns?

straxen/plugins/s2cnn_processing.py
terliuk: Yes, i will add it there. It is about 800 KB, so I don't know if github is the best place to store such files.

straxen/plugins/s2cnn_processing.py
terliuk: Yes, i can remove it. The idea was to make DoubleWidthConverter available even for environments without strax installed, but probably it is not necessary there. I think i will then place the converter to a separate file.

straxen/plugins/s2cnn_processing.py
terliuk: It converts area per channel (unique for each event) into the format for 2D array understood by CNN (see first picture here https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:analysis:reconstruction_team:training_S2_CNN_on_WFsim) . So it has to stay as it is unique to each event.

straxen/plugins/__init__.py
terliuk: I am a bit afraid that peak processing file grow quiet large, especially when we will have multiple position reconstruction options. It is easy to copy-paste it there, but I think it is better to keep it separate for maintenance and further  developments.

straxen/plugins/s2cnn_processing.py
terliuk: I made PR and it was merged already : https://github.com/XENONnT/private_nt_aux_files/pull/7

straxen/plugins/s2cnn_processing.py
terliuk: Done

straxen/plugins/s2cnn_processing.py
terliuk: Yes, the main point is to run on S2s only, so I added it to the mask. Technically, it should produce output for S1s as well, but it will be inaccurate (unless happens very close to the gate).

straxen/plugins/s2cnn_processing.py
terliuk: I removed this logic from plugin file.

straxen/plugins/s2cnn_processing.py
terliuk: I added a switch that defines whether patterns are stored or not in https://github.com/XENONnT/straxen/pull/320/commits/115a4acbeb3d30532d76523e290b8a5bbe056732

straxen/plugins/s2cnn_processing.py
terliuk: I added a plugin that provides peak positions here https://github.com/XENONnT/straxen/pull/320/commits/115a4acbeb3d30532d76523e290b8a5bbe056732 , it is not entirely what you were asking. Also, it provides `peak_positions` now, so it can be used in event building etc.

jorana:
Hi Andrii, like discussed, let me know when you want me to look again. You can also convert this PR to a draft for a second until you are ready.

jorana:
@terliuk , is this obsolete after #323?

terliuk:
Yes, this is obsolete now. Did you make the same changes in other branch (https://github.com/XENONnT/straxen/commit/86e3eb006b63ffb6d08d22a2f7974394c3799ff0 - https://github.com/XENONnT/straxen/commit/39e790863cbef350afe1545253c63ad09b6e1fcc, as you posted earlier)?

jorana:
> 
> 
> Yes, @jorana , this is obsolete now. Did you make the same changes in other branch ([86e3eb0](https://github.com/XENONnT/straxen/commit/86e3eb006b63ffb6d08d22a2f7974394c3799ff0) - [39e7908](https://github.com/XENONnT/straxen/commit/39e790863cbef350afe1545253c63ad09b6e1fcc), as you posted earlier)?

Yes, thanks for asking, I figured that sooner or later anyone might run into it so I fixed it on master. 

Please close this PR (and delete this branch) at your convenience.

terliuk:
Done

PR #321, opened 2021-01-04
use FS and allow disabling of progress bar

jorana:
Use https://straxen.readthedocs.io/en/latest/config_storage.html to load the files required for scada.

Also allow users to disable the progress bar (was annoying for the online monitor).


WenzDaniel (approved):
Looks fine, thanks

PR #322, opened 2021-01-06
update tutorials & documentation

jorana:
Some of the notebooks were really out of date, those can be either deleted or updated. To this end I've checked that all are running and I also added a notebook that I wrote for 1T data: https://github.com/XENONnT/analysiscode/blob/master/Fast_Response_Analysis/Open_commisioning_data.ipynb. It should be written in a general way so as not to reveal any particular sensitive information.

**Review request**
If okay, I also will add the `Open_data` notebook to the documentation on the website: https://straxen.readthedocs.io/en/latest/setup.html


WenzDaniel (approved):
Looks fine only found two mistakes in the markdown cells:
*  first sentence I think you are missing "(...) 1T dataset/run"
* Input 8 you use 100s while in the text above you speak of 10s

feigaodm (approved):
looks very useful, thanks, I'll message you elsewhere about updating online monitor.

PR #323, opened 2021-01-07
Add NN posrec for nT

reallsx:
** Enable GCN/MLP position reconstruction for XENONnT in Straxen without changing Straxen datastructure **

NN models prepared for nT cannot be used in the same way `peak_positions` used to do position reconstruction for 1T. A few changes are made to enable this plugin to use GCN/MLP model for nT:

- Changed `xnt_common_config` : add `nn_file` and commented `nn_architecture` and `nn_weights`.
- Added loading model from database into `peak_positions` plugin.
- Added renormalizing NN model input using maximum PMT area along with (not) converting unit of NN model output.

`nn_file` is used as a flag inside `peak_positions` plugin to decide which way to use for loading model and making prediction. This is for not breaking the current position reconstruction for 1T despite Straxen probably cannot access 1T NN files anyway...

GCN model is set to default in `xnt_common_config`, while one can use MLP by simply changing the `nn_file` option.

Note on how nT NN model files work:
[https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:analysis:reconstruction_team:nn_plugins_implementation](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:analysis:reconstruction_team:nn_plugins_implementation)

Example of GCN reconstruction used separately:
[https://github.com/XENONnT/analysiscode/blob/shift/run_s2_test.ipynb](https://github.com/XENONnT/analysiscode/blob/shift/run_s2_test.ipynb)

(PR saved as a draft as it could not pass all the tests).


straxen/contexts.py
```diff
@@ -45,8 +45,9 @@
         mv_blank=(1999, 1999),
         nveto=(2000, 2119),
         nveto_blank=(2999, 2999)),
-    nn_architecture='mlp_model.json',
-    nn_weights='mlp_model.h5',
+    nn_architecture=None,
```
jorana, line 48: Better overwrite default to None I think such to make clear that this option is not used

straxen/plugins/peak_processing.py
```diff
@@ -146,12 +152,27 @@ class PeakPositions(strax.Plugin):
...
-
+        
+        # XENONnT
+        if self.config['nn_file']:
+            downloader = straxen.MongoDownloader()
+            tf.keras.backend.set_floatx('float32')
+            
+            nn_file = downloader.download_single(self.config['nn_file'])
+            
+            with tempfile.TemporaryDirectory() as tmpdirname:
```
jorana, line 168: Do we also have to take out the bad PMTs like we did for 1T, I think so right?

tests/test_plugins.py
```diff
@@ -21,6 +21,7 @@
     straxen.aux_repo + 'f0df03e1f45b5bdd9be364c5caefdaf3c74e044e/fax_files/mlp_model.json',
     nn_weights=
     straxen.aux_repo + 'f0df03e1f45b5bdd9be364c5caefdaf3c74e044e/fax_files/mlp_model.h5',
+    nn_file=None,
```
jorana, line 24: This was failing the tests as we don't have access to this NN on our public repo (as desired)

straxen/plugins/peak_processing.py
```diff
@@ -187,15 +208,25 @@ def compute(self, peaks):
...
-            _in /= _in.sum(axis=1).reshape(-1, 1)
-
-        # Output: positions in mm (unfortunately), so convert to cm
-        _out = self.nn.predict(_in) / 10
+        
+        # XENONnT
+        if self.config['nn_file']:
+            _in = peaks['area_per_channel'][peak_mask, 0:self.config['n_top_pmts']]
+            with np.errstate(divide='ignore', invalid='ignore'):
+                _in = _in / np.max(_in, axis=1).reshape(-1, 1)
```
jorana, line 216: Do you know why this is `np.sum` for 1T?

jorana (commented):
Hi Shixiao, many thanks. Look forward to using this! Hereby some technical comments. I hope that @terliuk is able to provide a review on the performance.

straxen/contexts.py
reallsx: Looks good to me. It might be ok to just keep these two options too, as long as there is a `nn_file` these are not used.

straxen/plugins/peak_processing.py
reallsx: Dividing by sum is a very common way to normalize input. I don't know exactly why it is used in 1T, but using this kinds of expected as it is in almost every machine learning tutorial.
We found that NNs works better on simulation when the input is divided by the maximum value, so models for nT are using `np.max` instead.
It is also possible that normalizing would change if we could find a method that benefits reconstruction more.

tests/test_plugins.py
reallsx: Thanks for fixing it!

straxen/plugins/peak_processing.py
feigaodm: It's because we were only using relative pattern for training since Xe100 times ;) People like yuehuan and me are not following the advancement in data science. So no real physics reasons for np.sum. I remember that Shixiao showed that nu.max is better, so we are convinced, although I don't understand ;)

straxen/plugins/peak_processing.py
reallsx: I'm not sure what's the best way to handle bad PMTs. We could just leave them as 0 in the input.

For 1T the bad PMTs are removed because the model were built that way.

jorana: But this is something we would want to be able to specify at this stage as well right? It's easier than having to go back all the way to records.

feigaodm:
Sorry @reallsx , I just checked it again and indeed it is a vectorized operation on peaks. Then I don't understand why it's slow :) and the time is proportional to # of peaks

jorana:
note to self:

- [ ] add tf >= 2.2.0 as requirement

reallsx:
Do we still want this one? :)

jorana:
@terliuk @reallsx @feigaodm , as discussed, I now moved all that info up to `event_basics`. This is the datastucture now:
afbeelding
**Please notice** `peak_positions` is not stored!

I guess time will have to tell if we really need:
```python
 's2_x',
 's2_y',
 'alt_s2_x',
 'alt_s2_y',
 's2_x_mlp',
 's2_y_mlp',
 'alt_s2_x_mlp',
 'alt_s2_y_mlp',
 's2_x_cnn',
 's2_y_cnn',
 'alt_s2_x_cnn',
 'alt_s2_y_cnn',
 's2_x_gcn',
 's2_y_gcn',
 'alt_s2_x_gcn',
 'alt_s2_y_gcn',
 ```

This mean we now have these fields for `event_basics` (and likewise `event_info` etc.):
afbeelding

**Comment by feigaodm** [comment]
in pax+hax, we were using the same logic. It worked but it looks messy. I think it's better to put it in and develop new ideas later? How do you think? @Joran

**Comment by jingqiangye** [comment]
Thanks! just out of curiosity which algorithm is the default `s2_x`?;)

**Comment by jorana** [comment]
> in pax+hax, we were using the same logic. It worked but it looks messy. I think it's better to put it in and develop new ideas later? How do you think? @joran

So what I'd actually propose is to make `event_basics` a multi-output plugin, that just outputs `'event_posrec_many'`. That would provide all the uncorrected position reconstruction positons:
```
 's2_x_mlp',
 's2_y_mlp',
 'alt_s2_x_mlp',
 'alt_s2_y_mlp',
 's2_x_cnn',
 's2_y_cnn',
 'alt_s2_x_cnn',
 'alt_s2_y_cnn',
 's2_x_gcn',
 's2_y_gcn',
 'alt_s2_x_gcn',
 'alt_s2_y_gcn',
```
Whereas `event_basics` just knows about (one of the three selected at the `peak_positions` level) such that the dataframe at `event_basics` and `event_info` does not get too large:
```
 's2_x',
 's2_y',
 'alt_s2_x',
 'alt_s2_y',
... (the rest of the event_basics stuff)
 ```
 
 Then, you can load these pos-rec data-fields easily by requesting `st.get_array(run_id, ('event_basics', 'event_posrec_many'))` that automatically adds these extra uncorrected fields.

jorana:
@jingqiangye 
> Thanks! just out of curiosity which algorithm is the default s2_x?;)

Whatever the AC desires, currently I set it at MLP but that is easily changed.

straxen/plugins/position_reconstruction.py
```diff
@@ -0,0 +1,171 @@
...
+        if os.path.exists(self.model_file):
+            print(f"Path is local. Loading {self.algorithm} TF model locally "
+                  f"from disk.")
+        else:
+            downloader = straxen.MongoDownloader()
+            if self.model_file in downloader.list_files():
+                self.model_file = downloader.download_single(self.model_file)
+            else:
+                raise RuntimeError(f'Model files {self.model_file} is not found')
+        with tempfile.TemporaryDirectory() as tmpdirname:
```
jorana, line 61: One thing I noticed is that we cannot pickle the `self.model` leading to some nasty pickle errors in multiprocessing. We should discuss the best solution, now I've done:

```python
  def setup(self):
        ...
        resource_dir = os.path.join(
            './resource_cache',
            strax.utils.deterministic_hash(self.model_file))
        if not os.path.exists(resource_dir):
            os.makedirs(resource_dir)
            tar = tarfile.open(self.model_file, mode="r:gz")
            tar.extractall(path=resource_dir)
        self.model_path = resource_dir

    def compute(self, peaks):
        import tensorflow as tf
        model = tf.keras.models.load_model(self.model_path)
        # self.model -> model
```

jorana (changes requested):

straxen/plugins/position_reconstruction.py
reallsx: That's the reason why we chose to save the model this way. TensorFlow/Keras has many not pickleable stuff (especially when custom layers/operations are involved).

I think this solution might make processing slower as it takes a few seconds to load the model. So probably a few more seconds for each chunk...

jorana: I see that but you simply cannot add non-pickable objects to the Plugin class:
https://github.com/AxFoundation/strax/blob/master/strax/plugin.py#L45

reallsx: TensorFlow does not really support pickling custom objects:
https://www.tensorflow.org/guide/keras/save_and_serialize#custom_objects

Maybe the ultimate solution is switch to another framework. It looks like PyTorch models are pickable. 

I wonder do pickle errors also appear when only the MLP model is used?

jorana: So it's a python multiprocessing/strax thing, during multiprocessing python pickles these objects so all three PosRecs plugins would be affected. I guess this is why Jelle saved the 1T model this way in two seperate files, would that be an option?

There are some hacky solutions but I'm not sure if we should go down this road, especially given the link you just provided:
http://zachmoshe.com/2017/04/03/pickling-keras-models.html

PyTorch could be an option, but it's quite a heavy dependency which would slow down straxen-testing a bit (still better than slowing straxen processing down). I agree that what I wrote down as an example is not a solution we want. Perhaps @terliuk or @tunnell has an idea on how to save the model in such a way that it allows pickling and therefore can be set as a class-attribute.

jorana:
Just did a test on a 2 minute dataset to get a feeling of things (with my clumpy attempt around the pickling stuff I described above):
afbeelding

as well for the `plugin.parallel` setting, it seems that parallelizing it does not give a large improvement but that might be because 1. the dataset is very small 2. my laptop is running windows 
afbeelding

feigaodm:
@jorana  Given the performance, I think we can keep MLP the default one for the moment and make it easy to switch to any. 
And I understood from the meeting that you were proposing sth like this instead of the earlier one. Just to remind that the "corrected" variables later will also need a switch and it's good to think about that now.

jorana:
> And I understood from the meeting that you were proposing sth like this instead of the earlier one. 
Sorry I don't understand

By the way, Andrii had the idea to allow to set some of the files to None to thereby reduce computation time. Let's add.

straxen/plugins/position_reconstruction.py
feigaodm: Can we test the PyTorch option elsewhere? Does that change the PR very significantly later?

In another thread, I know that @reallsx can improve the speed by ~30% by playing with batch size. I don't know if that helps more for CNN. But I'm not sure if it's significant thing to consider now.

reallsx:
> as well for the `plugin.parallel` setting, it seems that parallelizing it does not give a large improvement but that might be because:

There was a comment in the old `peak_positions` plugin saying parallelizing does not make NNs faster (probably tested by Jelle). I guess this might apply to all operating systems.

jorana:
> > as well for the `plugin.parallel` setting, it seems that parallelizing it does not give a large improvement but that might be because:
> 
> There was a comment in the old `peak_positions` plugin saying parallelizing does not make NNs faster (probably tested by Jelle). I guess this might apply to all operating systems.

Exactly, was also part of the reason I wanted to re-test this. Now the plugin scheme is different so I wondered if this changed.

terliuk:
do we know what is the major slowdown factor here? Is it loading of tensorflow itself or something else?

jorana (approved):

feigaodm (approved):

PR #325, opened 2021-01-07
Add TPC length

jorana:
Based on https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:analysis:coordinate_system#coordinate_system_of_xenonnt

### DIFF
--- a/straxen/common.py
+++ b/straxen/common.py
@@ -21,7 +21,7 @@ import strax
 import straxen
 
 export, __all__ = strax.exporter()
-__all__ += ['straxen_dir', 'first_sr1_run', 'tpc_r', 'aux_repo',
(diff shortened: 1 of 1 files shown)


ramirezdiego:
@jorana the current MC value is 148.6515 cm (*), after the geometry review from last month. It is of course urealistically precise, but we can go for it if aiming at 100% consistency with the values used for all simulations, including light and field maps. Otherwise we can round up to 148.7 cm.

(*) Distance between the bottom of gate and top of cathode wires.

jorana:
Alright, that's also fine. To be honest I copied that value first but decided against it in order to be consistent with the linked page. Perhaps we should also change it there then.

ramirezdiego (approved):
OK, will add that information in the coordinates system page.

PR #326, opened 2021-01-08
Add event display

jorana:
Add some code the combines previous plotting code to be bundled in a single event display. See e.g. on **simulated** data:
afbeelding

**Alternative cals**
```python
# From a function
straxen.analyses.event_display.plot_single_event(st, run_id, events, event_number=2, records_matrix =True);

# Or from the context directly
st.event_display(run_id, time_range=(events[2]['time'], events[2]['endtime']), records_matrix =True);
```

**Edit, pending requests**

- [x] peaks summary (both S1 and S2):
   - [x] n_contribition_channels
   - [x] AFT 
   - [x] width
   - [x] start & endtime
   - [x] area
   - [x] etc
- [x] Hit pattern for both S1 & S2
- [x] Top bottom boundary
- [x] Center time indication
- [x] PMT-array display
- [x] XY reconstruction 

**Previous versions**
wf
wf2


feigaodm:
Nice! Can we put more information for the S1 and S2 peaks as well? For example rise_time, n_contribition_channels, AFT etc (especially the ones used in peak classification, so that people watching waveforms can imagine how close the peak is away from boundary).

feigaodm:
adding more reviews mostly for comments/suggestions.

WenzDaniel:
I am surprised to see this so we indeed doubled the work now. About your comment: 
> Nice! Can we put more information for the S1 and S2 peaks as well? For example rise_time, n_contribition_channels, AFT etc (especially the ones used in peak classification, so that people watching waveforms can imagine how close the peak is away from boundary).
This is already implemented in the display I am preparing. 
Currently I am working on your second comment: 
> In addition, it would be good to show both top and bottom patten for both peaks,
Here I am not 100 % satisfied with the look yet. I also have the hope that I can put a button to change the plots, but this is difficult since our jupyter does not allow to run java callbacks.

feigaodm:
Thanks @WenzDaniel , I do recall that you have it, then we can simply PR these features?
> I am surprised to see this so we indeed doubled the work now. About your comment:
> 
> > Nice! Can we put more information for the S1 and S2 peaks as well? For example rise_time, n_contribition_channels, AFT etc (especially the ones used in peak classification, so that people watching waveforms can imagine how close the peak is away from boundary).
> 

Great as well!
> This is already implemented in the display I am preparing.
> Currently I am working on your second comment:
> 
> > In addition, it would be good to show both top and bottom patten for both peaks,

jorana:
@feigaodm the PMT number is already displayed. Just uploaded a higher resolution, does that work for you?

jorana:
@feigaodm , eb135b1 should fix the comments

feigaodm:
> @feigaodm the PMT number is already displayed. Just uploaded a higher resolution, does that work for you?

Thanks, I meant the "circles" that represent the position and geometry of each PMTs.

sophiaandaloro:
Just writing down my comment from analysis meeting: 
Can we check on the event time that that's the span of the x axis in the waveform plot? (so limits should be start and end times of the events, perhaps +- a margin?) I also see the benefit of including stuff happening right after the event, so perhaps we could go with a light shading on the event's start to end time, which would clarify which S1 and S2 are the main ones we're interested in (or any other ideas you have). Let me know if you need input!
Again this is very excellent and will save all of us lots of time when we are pulling up events :) Thank you very much.

jorana:
@sophiaandaloro , thanks for the suggestion. The event processing does indeed this, take a right and left extension. In principle, there should not be anything after the event: https://github.com/XENONnT/straxen/blob/master/straxen/plugins/event_processing.py#L30

Also, we do keep the time range of the event: https://github.com/XENONnT/straxen/blob/ede167894df372a034aa6697f551246cec6870d0/straxen/analyses/waveform_plot.py#L193 so the range you see is what we call the event time - event endtime.

feigaodm (approved):
I think we shall merge it for people to play with and aim for improvement further.

tests/test_basics.py
```diff
@@ -15,6 +15,8 @@ def test_straxen():
 
             print("Downloading test data (if needed)")
             st = straxen.contexts.demo()
+            # Ignore strax-internal warnings
+            st.set_context_config({'free_options': tuple(st.config.keys())})
```
WenzDaniel, line 19: Why is this needed for the event display?

straxen/analyses/waveform_plot.py
```diff
@@ -338,3 +344,271 @@ def plot_wf(st: strax.Context,
...
+                           pmt_label_size=7,
+                           edgecolor='grey',
+                           cmap=color_map)
+        # Is there a better way to do this update?
+        hp_opt.update(_common_opt)
+        _update = hp_opt.copy()
+        hp_opt.update(_update)
+
+    # S1
+    if events['s1_area'] > 0:
```
WenzDaniel, line 498: Make this optional?

straxen/analyses/waveform_plot.py
```diff
@@ -338,3 +344,271 @@ def plot_wf(st: strax.Context,
...
+        plt.sca(ax_s1)
+        context.plot_peaks(run_id,
+                           time_range=(events['s1_time'] - s1_fuzz,
```
WenzDaniel, line 501: Why do you need this  extra `fuzz` time?

straxen/analyses/waveform_plot.py
```diff
@@ -338,3 +344,271 @@ def plot_wf(st: strax.Context,
...
+                                       events['s1_endtime'] + s1_fuzz),
+                           single_figure=False)
+
+        # Hit pattern plots
+        area = context.get_array(run_id, 'peaklets',
+                                 time_range=(events['s1_time'],
+                                             events['s1_endtime']),
+                                 keep_columns=('area_per_channel', 'time', 'dt', 'length'))
```
WenzDaniel, line 509: You have now a some sort of hidden requirement. Why have you decided not to add peaks to the list of requirements for the mini-analysis?

straxen/analyses/waveform_plot.py
```diff
@@ -338,3 +344,271 @@ def plot_wf(st: strax.Context,
...
+                           log_scale=True,
+                           vmin=0.1,
+                           s=(250 if records_matrix else 220),
+                           pmt_label_size=7,
+                           edgecolor='grey',
+                           cmap=color_map)
+        # Is there a better way to do this update?
+        hp_opt.update(_common_opt)
+        _update = hp_opt.copy()
+        hp_opt.update(_update)
```
WenzDaniel, line 495: This part you have to explain, at least I do not understand it.

straxen/analyses/waveform_plot.py
```diff
@@ -338,3 +344,271 @@ def plot_wf(st: strax.Context,
...
+                              ('x', '{v:.2f} cm'),
+                              ('y', '{v:.2f} cm'),
+                              ('z', '{v:.2f} cm'),
+                              ('r', '{v:.2f} cm'),
+                              ('theta', '{v:.2f} rad'),
+                              ('drift_time', '{v} ns'),
+                              ('alt_s1_interaction_drift_time', '{v} ns'),
+                              ('alt_s2_interaction_drift_time', '{v} ns')
+                              )
+
```
WenzDaniel, line 374: Would not be a dictionary simpler?

straxen/analyses/waveform_plot.py
```diff
@@ -338,3 +344,271 @@ def plot_wf(st: strax.Context,
...
+                                                (ax_peak_info, display_peak_info)]):
+        for i, (_lab, _unit) in enumerate(labels_and_unit):
+            coord = 0.01, 0.9 - 0.9 * i / len(labels_and_unit)
+            ax.text(*coord, _lab[:24], va='top', zorder=-10)
+            ax.text(coord[0] + 0.5, coord[1],
+                    _unit.format(v=event[_lab]), va='top', zorder=-10)
+            # Remove axes and labels from panel
+            ax.set_xticks([])
+            ax.set_yticks([])
+            _ = [s.set_visible(False) for s in ax.spines.values()]
```
WenzDaniel, line 550: I hope you do not have to add additional parameters, this looks quite static, but I could be wrong.

WenzDaniel (approved):
Looks nice, there are not really any larger changes I request, but rather some comments.

straxen/analyses/waveform_plot.py
jorana: Yeah its a bit convoluted. So i want to update dict A with dict B the problem is that I cannot say A = B; B.update(A) because I want to pass the reference of A as I am looping over those. But yeah , might as well loop over the items, is a little easier to follow.

straxen/analyses/waveform_plot.py
jorana: So there is some snafau happening when there is no S1 because then there is also no S1 time. I don't think we allow main S1s to be negative (right)? This is the confusing thing about having 0s as default for some parameters. Can also change the > to == to be sure

straxen/analyses/waveform_plot.py
jorana: I decided to do it this way such that you don't have to load the peaklets for the entire event time span but only during the requested time range. Will probably not matter much computation wise but may make a slight memory difference.

straxen/analyses/waveform_plot.py
jorana: Actually, it is flexible (till some level), you just need to change the `display_event_info`  argument as in the docstring. It will just divides the panel in whatever length info you need and evenly spaces it there. Could do something smarter if needed but I guess this is fine for now.

tests/test_basics.py
jorana: Not really for this event display but I noticed that we were seeing quite some warnings about things that are really not interesting to be seeing in `pytests`.

straxen/analyses/waveform_plot.py
jorana: If you want to see how that works, you can look at the difference between two of the examples I've posted, the event-info panel has been slightly updated with more info, and the lines of attributes are scaled accordingly.

jorana:
@WenzDaniel , thanks for the review. Some of these comments I will slightly tweak but I don't think any require mayor changes. Perhaps @jpienaar13  or @skazama have any additional comments or requests? If not I will make the requested changes by Daniel tomorrow morning and merge.

straxen/analyses/waveform_plot.py
WenzDaniel: Yes this is something I have not thought of. If the user does not query any time range the memory will explode. I will also changes it in the bokeh event display.

skazama:
Hi @jorana, thanks for preparing this very nice tool! I think it's already good enough and can be merged

straxen/analyses/waveform_plot.py
jorana: Okay, I checked it out, the format would be very similar but we would have a mutable default which then requires if `display_peak_info is None:` et cetera. Therefore, I keep it as it is.
```python
PEAK_DISPLAY_DEFAULT_INFO = {}
_ = [PEAK_DISPLAY_DEFAULT_INFO.update(
    {k.format(i=i): v for k, v in
     {'cs{i}': '{v:.2f} PE',
      's{i}_area': '{v:.2f} PE',
      'alt_cs{i}': '{v:.2f} PE',
      's{i}_n_channels': '{v}',
      's{i}_area_fraction_top': '{v:.2f}',
      's{i}_range_50p_area': '{v:.1f}'}.items()}) for i in (1, 2)]
PEAK_DISPLAY_DEFAULT_INFO
EVENT_DISPLAY_DEFAULT_INFO = {
    'time': '{v} ns',
    'endtime': '{v} ns',
    'event_number': '{v}',
 'x': '{v:.2f} cm',
 'y': '{v:.2f} cm',
 'z': '{v:.2f} cm',
 'r': '{v:.2f} cm',
 'theta': '{v:.2f} rad',
 'drift_time': '{v} ns',
 'alt_s1_interaction_drift_time': '{v} ns',
 'alt_s2_interaction_drift_time': '{v} ns'}
 ```

File: straxen/analyses/waveform_plot.py (hunk 5 above)
Comment (jorana) [reply straxen/analyses/waveform_plot.py:495]: Replaced this with
```python
        # update s1 & S2 hit pattern kwargs with _common_opt if not
        # specified by the user
        for k, v in _common_opt.items():
            if k not in hp_opt:
                hp_opt[k] = v
```

straxen/analyses/waveform_plot.py
jorana: I think technically, you don't strictly need it but is just nice to show if there are some thing directly in front or after the main SX peak.

tunnell:
@jorana can you add ALL position reconstruction algorithms to display?  I thought we did this before, but otherwise have fewer tools for diagnosing issues.

jorana:
> 
> 
> @jorana can you add ALL position reconstruction algorithms to display of S2 top? I thought we did this before, but otherwise have fewer tools for diagnosing issues.

Of course Chris, not sure what you mean with  `I thought we did this before, ` because I wrote this PR before we agreed on how to handle the PosRec-processing.

tunnell:
Before = XENON100 :) And 1T :)

jorana:
Thanks, yeah did not really look there to be honest, it's in #338

PR #327, opened 2021-01-11
Update matplotlib_utils.py

GiovanniVolta:

I would like to have the chance to save the PMT arrays picture

I have added "save" variable which default value is False. If it is True the function return the figure

### DIFF
--- a/straxen/matplotlib_utils.py
+++ b/straxen/matplotlib_utils.py
@@ -16,6 +16,7 @@ def plot_pmts(
         xenon1t=False,
         show_tpc=True,
         extend='neither', vmin=None, vmax=None,
+        save=False,
         **kwargs):
     """Plot the PMT arrays side-by-side, coloring the PMTS with c.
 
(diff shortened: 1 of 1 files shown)


WenzDaniel:
While it does not hurt, I am also not sure if this is strictly needed. You can always get the last figure via 
```
fig = plt.gcf()  # gcf means get current figure
```
And then save it.

GiovanniVolta:
Ok, let me close the PR then

jorana:
I'm also not sure if I understand. What would not work that you would like to do with the syntax below?
```python
import straxen
import numpy as np
import matplotlib.pyplot as plt
straxen.plot_pmts(np.arange(494))
plt.savefig('test.png')
```

GiovanniVolta:
That wasn't working for me. The picture was empty. Sure there was some error, but Daniel's one worked at the first attempt. My PR was useless

WenzDaniel:
I would not say it was useless, the `plt.gca` and `plt.gcf` are a bit hidden features of matplotlib. Maybe we should add it to the doc strings at least.

jorana:
> That wasn't working for me. The picture was empty. 

This usually happens when you `plt.show()` before `plt.savefig('test.png')`

Issue #336, opened 2021-01-19
Dead pmts in displays


jorana:
The bokeh display has a nice feature to hide 0 gain PMTs. We should add this to the matplotlib display too.

For the bokeh display, it might be nice if we can also hoover over the 0 gain PMTs to get their position and number.

jorana:
Some other things,
this logic does not work:
https://github.com/XENONnT/straxen/blob/master/straxen/analyses/event_display.py#L93
Neither does int('raw') work:
https://github.com/XENONnT/straxen/blob/master/straxen/analyses/event_display.py#L104

Also add the other posrecs

PR #338, opened 2021-01-20
Fix plotting

jorana:
Some packages were untested and or outdated, fix that.

Also work on:
https://github.com/XENONnT/straxen/issues/336


straxen/analyses/records_matrix.py
```diff
@@ -85,7 +85,7 @@ def raw_records_matrix(context, run_id, raw_records, time_range,
     # at the start of the range due to missing zeroth fragments
     records = strax.raw_to_records(raw_records)
     strax.baseline(records, allow_sloppy_chunking=True)
-    strax.zero_outside_bounds(records)
+    strax.zero_out_of_bounds(records)
```
jorana, line 88: This was the reason to reinstate more checks. Why was there a function here for over half a year that was bugged

WenzDaniel: Now you also know how many people used this function.

straxen/analyses/event_display.py
```diff
@@ -36,7 +36,26 @@
...
+def _scatter_rec(event,
+                 recs=('mlp', 'cnn', 'gcn'),
+                 scatter_kwargs=None):
+    """Convienent wrapper to show three algorithms for xenonnt"""
+    assert len(recs) <= 3, "I only got three markers/colors"
+    if scatter_kwargs is None:
+        scatter_kwargs = {}
+    scatter_kwargs.setdefault('s', '100')
+    shapes = ('v', '^', '>', '<')
+    colors = ('brown', 'orange', 'lightcoral')
```
WenzDaniel, line 48: For color you could also use the default matplotlib color cycle if you like `plt.rcParams['axes.prop_cycle'].by_key()['color']` those are eight colors I think.

jorana: Yeah, thanks, I don't think those contrast well against the green background

straxen/analyses/records_matrix.py
jorana: haha, problem is that I ran into it so it's worth fixing right?

WenzDaniel: :D Yes no arguing with that.

straxen/analyses/event_display.py
WenzDaniel: Would be cool to have a matplotlib feature which could detect this and change the color automatically. Maybe if you have time ;)

WenzDaniel: But anyhow maybe it does not hurt to add two more colors and shapes already now, just in case.

straxen/analyses/event_display.py
```diff
@@ -93,16 +114,18 @@ def event_display(context,
     if records_matrix not in ('raw', True, False):
         raise ValueError('Choose either "raw", True or False for records_matrix')
     if ((records_matrix == 'raw' and not context.is_stored(run_id, 'raw_records')) or
-            (records_matrix and not context.is_stored(run_id, 'records'))):
+        (isinstance(records_matrix, bool) and records_matrix and not context.is_stored(run_id, 'records'))):   # noqa
```
WenzDaniel, line 117: You can simplify this here since you check already above that `records_matrix` is either 'raw' or bool. It should be
`bool(raw) == True` Hence `if records_matrix and (...)` should be enough.

straxen/analyses/event_display.py
```diff
@@ -93,16 +114,18 @@ def event_display(context,
...
-            (records_matrix and not context.is_stored(run_id, 'records'))):
+        (isinstance(records_matrix, bool) and records_matrix and not context.is_stored(run_id, 'records'))):   # noqa
         print("(raw)records not stored! Not showing records_matrix")
         records_matrix = False
     if not context.is_stored(run_id, 'peaklets'):
         raise strax.DataNotAvailable(f'peaklets not available for {run_id}')
 
-    fig = plt.figure(figsize=(25, 7 * (2 + int(records_matrix))),
+    # String convert for int(_rr_bool)
+    _rr_bool = bool(records_matrix)
```
WenzDaniel, line 124: You can also convert it in one go into an `int` is not it? I think you are not using `_rr_bool`  anymore.

WenzDaniel (approved):
Looks fine. I have not checked the tests, in great detail though.

straxen/analyses/event_display.py
jorana: True

straxen/analyses/event_display.py
jorana: Yeah, I thought so too, but what happens if we have records_matrix == 'raw' and we don't have 'records' stored. Then `bool(raw) == True and not context.is_stored(run_id, 'records')` so that is condition is True. I also confused myself a little here.

Or perhaps I'm not seeing what you mean. I think strictly speaking:
`` isinstance(records_matrix, bool) and not context.is_stored(run_id, 'records'))`

is indeed all one needs.

straxen/analyses/event_display.py
jorana: Ah good point. Thanks Daniel!

jorana:
Fixes #336

PR #340, opened 2021-01-21
Listen to utilix, remove depricated function from straxen

jorana:
In a recent utilix [update](https://github.com/XENONnT/utilix/releases/tag/v0.4.0), a deprecation print statement was added. To keep one notebooks clean, we can make the suggested change to reduce the number of these statements 😄


WenzDaniel (approved):
Thanks,

PR #343, opened 2021-01-24
Reduce n-files/run

jorana:
Create less files, store less intermediate data. Here I added some places where we could possibly take some shortcuts.

Be more selective of what we store and what not

**What is the effect**
For an arbitrary short run, I processed which lead to the table below.
- with all the changes in this PR and 
- once with all the changes in this PR **except for PulseProcessing**. This means we would keep `records, pulsecounts, veto_intervals`.

|          | before               | after (saving PulseProcessing)  | after (not saving PulseProcessing) | 
| --      | ----------------- |  --------------------------------  | ------------------------------------  | 
| #files | 1267                 |  757                                           | 585                                                 |
| Total size | 88 GB         | 88 GB                                        | 61 GB                                              | 

**Consensus concerning this PR**
Let's wrap up what is being asked in this PR, we ask if we want to (checkbox should indicate consensus):
- [x] Rechunking raw_records (especially raw_records_aqmon)
- [x] set EnergyEstimates to `SaveWhen.Target` (this is a trivial plugin that just multiplies a handful of numbers)
- [x] Set `PulseProcessing` (records) to  `SaveWhen.Target`
- [x] Set `PulseProcessingHe` (records_he) to `SaveWhen.TARGET`
- [x] Set High energy processing (peaks, peaklets etc) to `SaveWhen.TARGET`

**Data-types (before changes)**
```corrected_areas-cjno725i32
energy_estimates-7ksels3aic
event_basics-jpnilkwbfr
event_info-allwalikwi
event_positions-jqknjp7gio
event_posrec_many-jpnilkwbfr
events-zf5ybclhvk
lone_hits-ioudwhzz4n
merged_s2s_he-zxpiwi3d5z
merged_s2s-4os2w32txl
online_monitor-z74bl33e5e
online_peak_monitor-kyas5n6t74
peak_basics_he-6yfhkyq57x
peak_basics-3k6xvmmrww
peak_positions_cnn-22yx74azrt
peak_positions_gcn-bojdxvz2sw
peak_positions_mlp-kuinlrzhsl
peak_proximity-npglcajfmc
peaklet_classification_he-4hu67ymbkf
peaklet_classification-nnlrqza4wr
peaklets_he-qriha2uzhs
peaklets-ioudwhzz4n
pulse_counts_he-uoteecjda5
pulse_counts-pamvninlop
raw_records_aqmon_nv-rfzvpzj4mf
raw_records_aqmon-rfzvpzj4mf
raw_records_aux_mv-rfzvpzj4mf
raw_records_he-rfzvpzj4mf
raw_records_mv-rfzvpzj4mf
raw_records_nv-rfzvpzj4mf
raw_records-rfzvpzj4mf
records_he-uoteecjda5
records-pamvninlop
veto_intervals-7q35r23nba
veto_regions-pamvninlop
```

### DIFF
--- a/straxen/plugins/daqreader.py
+++ b/straxen/plugins/daqreader.py
@@ -88,7 +88,15 @@ class DAQReader(strax.Plugin):
     data_kind = immutabledict(zip(provides, provides))
     depends_on = tuple()
     parallel = 'process'
-    rechunk_on_save = False
+    rechunk_on_save = immutabledict(
+        raw_records=False,
+        raw_records_he=False,
+        raw_records_aqmon=True,
+        raw_records_nv=False,
+        raw_records_aqmon_nv=True,
+        raw_records_aux_mv=True,
+        raw_records_mv=False,
+    )
     compressor = 'lz4'
     __version__ = '0.0.0'
 
--- a/straxen/plugins/pulse_processing.py
+++ b/straxen/plugins/pulse_processing.py
@@ -125,6 +125,7 @@ class PulseProcessing(strax.Plugin):
 
     provides = ('records', 'veto_regions', 'pulse_counts')
     data_kind = {k: k for k in provides}
+    save_when = strax.SaveWhen.TARGET
 
     def infer_dtype(self):
         # Get record_length from the plugin making raw_records
@@ -240,6 +241,7 @@ class PulseProcessingHighEnergy(PulseProcessing):
     depends_on = 'raw_records_he'
     compressor = 'lz4'
     child_plugin = True
+    save_when = strax.SaveWhen.TARGET
 
     def infer_dtype(self):
         dtype = dict()
--- a/.gitignore
+++ b/.gitignore
@@ -60,3 +60,6 @@ docs/_build
 
 # coverage
 .coverage
+
+# Documentation files
+docs/source/reference/datastructure*
--- a/straxen/plugins/event_processing.py
+++ b/straxen/plugins/event_processing.py
@@ -466,6 +466,7 @@ class EnergyEstimates(strax.Plugin):
         ('e_charge', np.float32, 'Energy in charge signal [keVee]'),
         ('e_ces', np.float32, 'Energy estimate [keVee]')
     ] + strax.time_fields
+    save_when = strax.SaveWhen.TARGET
 
     def compute(self, events):
         el = self.cs1_to_e(events['cs1'])
--- a/straxen/plugins/led_calibration.py
+++ b/straxen/plugins/led_calibration.py
@@ -47,7 +47,8 @@ class LEDCalibration(strax.Plugin):
     data_kind = 'led_cal' 
     compressor = 'zstd'
     parallel = 'process'
-    
+    rechunk_on_save = False
+
     dtype = [('area', np.float32, 'Area averaged in integration windows'),
              ('amplitude_led', np.float32, 'Amplitude in LED window'),
              ('amplitude_noise', np.float32, 'Amplitude in off LED window'),
--- a/straxen/plugins/led_calibration.py
+++ b/straxen/plugins/led_calibration.py
@@ -47,7 +47,7 @@ class LEDCalibration(strax.Plugin):
     data_kind = 'led_cal' 
     compressor = 'zstd'
     parallel = 'process'
-    rechunk_on_save = False
+    rechunk_on_save = True
     
     dtype = [('area', np.float32, 'Area averaged in integration windows'),
              ('amplitude_led', np.float32, 'Amplitude in LED window'),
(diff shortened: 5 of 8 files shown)


straxen/plugins/pulse_processing.py
```diff
@@ -125,6 +125,7 @@ class PulseProcessing(strax.Plugin):
 
     provides = ('records', 'veto_regions', 'pulse_counts')
     data_kind = {k: k for k in provides}
+    save_when = strax.SaveWhen.TARGET
```
jorana, line 128: This is a very drastic change and will disable record matrix et cetera plotting because you cannot do time selections on data that is not stored!

WenzDaniel:
Thanks, this is really a large reduction. I was also wondering if this is not an opportune moment to discuss the whole high energy channel business. Maybe I can just simply not recall its outcome, but at some point we started the discussion whether it make sense to have these stand-alone plugins or if it would not make more sense to only include their information in case a channel saturates in the regular plugins.
While I do think it is nice to have the he-plugins, I am wondering whether we should turn all of them into only `save_when_target` plugins. ACs should comment I guess.

jorana:
> While I do think it is nice to have the he-plugins, I am wondering whether we should turn all of them into only save_when_target plugins. ACs should comment I guess.

Agree, that is also kind of what I did now right (whoops, I see that somehow my changes in the peak-processing weren't uploaded)? I by the way decided that I'm not going to do higher processing that `peaklets_he` at the eventbuilder for the same reason:
https://github.com/XENONnT/daqnt/commit/eb2af34f1f10fed6c1b7985bdac6015faac659e4

darrylmasson:
Not particularly related to this PR, but I still have concerns about the integration of the entire HE subsystem into the analysis. After all the work we did to recable it (still waiting for those beers @feigaodm), I haven't seen any work on actually using this for the saturation correction rather than the unsaturated-neighbor technique in 1T.

I recall a variety of conversations with Jelle about how rechunking affects multiprocessing, maybe we'll see some issues relating to that. Rechunking is a state-ful operation, this is a bit difficult in a multiprocessing environment.

ershockley:
Thanks Joran, this will be a huge help. Can we maybe first try including only the rechunking update to reduce the number of files, and see how that goes? Maybe that will help things enough so that we can keep records and then don't have to worry about the waveform watching concern.

ershockley (approved):
I approve, though I would propose to not include the change to pulse_processing yet to see how things go. Can always add it later.

jorana:
Hi all, thanks for the prompt feedback, I think we are on the same page. I've added a `Consensus concerning this PR` to the top of this PR to summarize.

> I recall a variety of conversations with Jelle about how rechunking affects multiprocessing, maybe we'll see some issues relating to that. Rechunking is a state-ful operation, this is a bit difficult in a multiprocessing environment.

We do [recunking at the PulseProcessing ](https://github.com/XENONnT/straxen/blob/703c1470e02d9dee45ef28ca628e1aa3463fd4e3/straxen/plugins/pulse_processing.py#L120) level for pulse-counts which is also part of the Parrallel source plugin. The raw-records re-chunked here are only quite small so I hope we aren't running into any strange issues there. At least I did not find any issues when testing this PR on livedata. I've found the associated PR in strax for bookkeeping: 
https://github.com/AxFoundation/strax/pull/272

> I approve, though I would propose to not include the change to pulse_processing yet to see how things go. Can always add it later.

I agree, also would prefer this, it's nice that we are on the same page 😄.

@feigaodm , do you approve? Can we also turn off the He processing (up to some level)?

jorana:
I've exhausted my ideas of testing it. I'm confident it's safe.

Nevertheless, I think we should have someone else also validate this, there is no going back.

straxen/plugins/daqreader.py
```diff
@@ -89,7 +89,15 @@ class DAQReader(strax.Plugin):
     data_kind = immutabledict(zip(provides, provides))
     depends_on = tuple()
     parallel = 'process'
-    rechunk_on_save = False
+    rechunk_on_save = immutabledict(
+        raw_records=False,
+        raw_records_he=False,
+        raw_records_aqmon=True,
```
AlexElykov, line 95: So each rechunked chunk will have max size of 200 MB, so as you're mentioned e.g. all rr_aqmon can fit into a single chunk?

AlexElykov (commented):
Looks like a really nice idea that will hopefully solve some of our current problems.

WenzDaniel (approved):
Thanks Joran for the changes.

straxen/plugins/daqreader.py
jorana: yeah for sure, the aqmon records are just a few MB even in the largest runs

darrylmasson (approved):
This PR is one those Rubicon moments, but I looks fairly straightforward to me.

jorana:
Alright, thanks all _alea iacta est_

PR #344, opened 2021-01-25
add function for version printing

jorana:
Add a simple function to print versions (also useful outside of dali).

afbeelding


WenzDaniel:
Nice, Idea also for debugging when people have problems.

jorana:
To use it for utilix, we do require https://github.com/XENONnT/utilix/pull/31 by the way

PR #345, opened 2021-01-25
extending the event window

feigaodm:
As in the title. We'd like to change the event window to be larger to study more features. Let's see if it's ok


jingqiangye:
Can we set 3e6 for the left/right extension just to be safe?

jorana (approved):
I'm fine with it, thanks Fei

jingqiangye (approved):
Ok now I am also fine with that. Thanks Fei for changing the window length for nT, and thanks Joran for keeping it for 1T.

PR #348, opened 2021-01-26
Update create-utilix-config.sh

ershockley:

Need to update the utilix config for GitHub actions. This is due to a recent utilix change: [#29](https://github.com/XENONnT/utilix/pull/29)


jorana (approved):
Thanks Evan, looks good.

PR #349, opened 2021-01-28
Online events monitor

reallsx:
Per [issue \#36 in analysis code repo](https://github.com/XENONnT/analysiscode/issues/36), add drift time vs reconstructed R^2 to online monitor. Add information from event to online monitor. A new plugin `OnlineEventMonitor` and corresponding data type `online_event_monitor` are added to `straxen/plugins/online_monitor.py`

See [this reply](https://github.com/XENONnT/analysiscode/issues/36#issuecomment-766662156).

`OnlineEventMonitor` has been tested on an old run, while the modified loop plugin `online_monitor` has not been tested yet.

### DIFF
--- a/straxen/plugins/online_monitor.py
+++ b/straxen/plugins/online_monitor.py
@@ -33,13 +33,15 @@ export, __all__ = strax.exporter()
         type=tuple, default=(0, 1500),
         help='Boundaries area histogram of lone hits [ADC]'),
     strax.Option(
-        'online_monitor_nbins',
+        'online_peak_monitor_nbins',
         type=int, default=100,
         help='Number of bins of histogram of online monitor. Will be used '
              'for: '
              'lone_hits_area-histogram, '
              'area_fraction_top-histogram, '
-             'near_s1_hists, '),
+             'near_s1_hists, '
+             'online_se_gain estimate (histogram is not stored), '
+    ),
     strax.Option(
         'near_s1_hists_cut_string',
         type=str,
@@ -73,6 +75,11 @@ export, __all__ = strax.exporter()
     strax.Option(
         'n_tpc_pmts', type=int,
         help='Number of TPC PMTs'),
+    strax.Option(
+        'online_se_bounds',
+        type=tuple, default=(7, 70),
+        help='Window for online monitor [PE] to look for the SE gain, value'
+    )
 )
 class OnlinePeakMonitor(strax.Plugin):
     """
@@ -96,7 +103,7 @@ class OnlinePeakMonitor(strax.Plugin):
         n_bins_area_width = self.config['area_vs_width_nbins']
         bounds_area_width = self.config['area_vs_width_bounds']
 
-        n_bins = self.config['online_monitor_nbins']
+        n_bins = self.config['online_peak_monitor_nbins']
 
         n_tpc_pmts = self.config['n_tpc_pmts']
         dtype = [
@@ -128,6 +135,8 @@ class OnlinePeakMonitor(strax.Plugin):
              (np.int64, n_bins)),
             (('Near S1 peaks area hist bounds', 'near_s1_area_bounds'),
              (np.float64, 2)),
+            (('Single electron gain', 'online_se_gain'),
+             np.float32),
         ]
         return dtype
 
@@ -137,7 +146,7 @@ class OnlinePeakMonitor(strax.Plugin):
         res['time'] = start
         res['endtime'] = end
         n_pmt = self.config['n_tpc_pmts']
-        n_bins = self.config['online_monitor_nbins']
+        n_bins = self.config['online_peak_monitor_nbins']
 
         # Bounds for histograms
         res['area_vs_width_bounds'] = self.config['area_vs_width_bounds']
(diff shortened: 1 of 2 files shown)


straxen/plugins/online_monitor.py
```diff
@@ -80,9 +80,9 @@ class OnlinePeakMonitor(strax.Plugin):
     this plugin should be small such as to not overload the runs-
     database.
 
-    This plugin takes 'peaks_basics' and 'lone_hits'. Although they are
+    This plugin takes 'peak_basics' and 'lone_hits'. Although they are
```
jorana, line 83: Thanks :)

straxen/plugins/online_monitor.py
```diff
@@ -254,24 +254,130 @@ def area_width_hist(self, data):
...
+    database.
+
+    This plugin takes 'event_basics'.
+
+    Produces 'online_event_monitor' with info on the events.
+    """
+    depends_on = ('event_basics',)
+    provides = 'online_event_monitor'
+    __version__ = '0.0.1'
+    # TODO make new datakind:
```
jorana, line 290: You can remove this todo, I thought about it and it's not needed

straxen/plugins/online_monitor.py
```diff
@@ -254,24 +254,130 @@ def area_width_hist(self, data):
...
+        default='(s2_index > -1) & (s1_index > -1)',
+        help='Selection (like selection_str) applied to data for '
+             '"drift_time_vs_R2", cuts should be separated using "&"'),
+    strax.Option(
+        'drift_time_vs_R2_bounds',
+        type=tuple, default=((0, 5e3), (0, 3e6)),
+        help='Boundaries of histogram of drift time vs R^2.'
+             'y_deges'),
+    strax.Option(
+        'online_monitor_nbins',
```
jorana, line 271: perhaps make it a little more explicit that this is at event level, I should do the same at peak level.

straxen/plugins/online_monitor.py
```diff
@@ -254,24 +254,130 @@ def area_width_hist(self, data):
...
+        'drift_time_vs_R2_cut_string',
+        type=str,
+        default='(s2_index > -1) & (s1_index > -1)',
+        help='Selection (like selection_str) applied to data for '
+             '"drift_time_vs_R2", cuts should be separated using "&"'),
+    strax.Option(
+        'drift_time_vs_R2_bounds',
+        type=tuple, default=((0, 5e3), (0, 3e6)),
+        help='Boundaries of histogram of drift time vs R^2.'
+             'y_deges'),
```
jorana, line 269: Sorry I don't follow what `y_deges` is

straxen/plugins/online_monitor.py
```diff
@@ -254,24 +254,130 @@ def area_width_hist(self, data):
...
+            self.config['drift_time_vs_R2_cut_string'], events)
+        res['drift_time_vs_R2_hist'] = self.drift_time_R2_hist(events[mask])
+        # Make a new mask, don't re-use
+        del mask
+        
+        #Other event properties?
+        
+        return res
+        
+    @staticmethod
```
jorana, line 330: Let me move this somewhere else because both plugins are using it now.

straxen/plugins/online_monitor.py
```diff
@@ -254,24 +254,130 @@ def area_width_hist(self, data):
...
+            data['drift_time'],
+            range=self.config['drift_time_vs_R2_bounds'],
+            bins=self.config['online_monitor_nbins'])
+        return hist.T
+
+
+
+
+
 class OnlineMonitor(strax.LoopPlugin):
```
jorana, line 359: I think I'm going to delete this class. Why:
- It's slow to compute because it has this LoopPlugin structure, I think it's much easier to put the Online_event monitor and the online_peak monitor in separately.
- It makes the dependencies complicated, we already see this for the [online monitor as it is now
](https://straxen.readthedocs.io/en/latest/reference/datastructure.html#online-monitor), adding event level info will further amplify this. 
- It makes it easier to accept if either of the two is not available, that way you don't have to worry if we have some runs where events are not processed online. Also, you can process the online events later.

jorana (commented):
Thanks a lot @reallsx ! I say you did a clean and nice job!

Most of the comments I'm making are probably more directed to myself rather than you because I think we need to somewhat reduce the complexity by having a separate plugin for online monitoring of peaks and events. There are some reasons for that I tried detailing below.

I hope you are fine if I already add some of those things into this PR.

straxen/plugins/online_monitor.py
reallsx: Was a typo. Maybe I intended to show what are the values in this tuple.

jorana:
@reallsx @jingqiangye , many thanks, for now, I've actually activated the storage of the `event_basics` into the runs database. It should be possible to keep on doing this, if we notice that it becomes to large, we might switch to the `online_event_monitor`. 

Reason is that I realized that actually some 100x100 2D histograms are actually quite large and not per se a lot smaller than just storing all the flat data.

Issue #350, opened 2021-01-28
Problematic try exception import block


tunnell:
https://github.com/XENONnT/straxen/blob/2b32ba5a44edbb93fed575304059c524c971e3b0/straxen/corrections_services.py#L8

If utilix isn't installed, it doesn't tell you.  However, if utilix fails to load for some reason, then it won't tell you... you can check if Travis through environmental variables, or at least print a warning that there was an exception?  You can log exceptions without reraising.

tunnell:
This is misleading to new people if .xenon_secrets can't be found for some reason.  They check if utilix installed, it is, but error is about utilix not in namespace.

jorana:
Hi Chris, I tend to agree, I've also had incomprehensible errors due to this sloppy importing. I think the root issue is how utilix is initialized, rather than doing things sloppy here, I'd argue that we have to change this on the utilix side first.

jorana:
I actually think the root issue is that we don't allow [travis / github actions to our database](https://github.com/XENONnT/straxen/issues/277).

@tunnell I think we could fix all these issues and improve our testing if we allow access to our database using a read only account, I haven't found any convincing reason not to allow this access. True, we could setup some database in travis I don't have the time to do it while I can simply add an encrypted password and user into the testing (I know how to for both Travis and github actions).

This is becoming an increasingly pressing issue, for example, we cannot test the posrec algos because we set their files in our database.

ershockley:
@jorana we do have access to the database globally right?  We can't write to the database but two of the mirrors are accessible. Or am I missing something?

darrylmasson:
Some accounts are limited in where they can be used from, but nt_analysis (for instance) has read-only access and can be used from anywhere.

jorana:
Yeah, for some reason we have not used any database integration (because of performance/privacy/fear of malicious access??) .  The problem is not that we cannot, the problem is if I'm not sure if people (@darrylmasson @tunnell ) are ok with the travis testing to have access to our database to read.

Apart from the performance I don't see any argument that makes sense. But even the performance should be fine.

tunnell:
Yes, that's fine.  Just use TravisCI encrypted environmental variables.  Write users have IP restrictions, so you'd have to use read-only.

tunnell:
[...]

PR #351, opened 2021-01-29
Use rundatabase on testing

jorana:
See https://github.com/XENONnT/straxen/issues/277, https://github.com/XENONnT/straxen/issues/350

We do give a read only account to continuous integration in order to test our real software. Otherwise we're increasingly not testing stuff because we are relying more and more on the database.

Also, let's remove some unclear imports that cause errors in places one does not expect them.

**To make this work**
We should add some read-only account here (see #350):
https://github.com/XENONnT/straxen/blob/c884945548bdd5ef9a77db972e42ebe78f4a4358/.github/scripts/create_empty_utilix_config.sh


ershockley:
@jorana in your fork can you try adding this to the utilix config file? 

```
[basic]
logging_level=debug
```

Might help with the utilix issue.

PR #355, opened 2021-02-03
Add functionality to get NN model files from CMT so we can track an…

ahiguera-mx:
Add functionality to get NN files from CMT to do a better bookkeeping of any change

please see this notebook 
https://github.com/XENONnT/analysiscode/blob/master/corrections/posrec_wCMT_example.ipynb

please see this notebook
https://github.com/XENONnT/analysiscode/blob/master/corrections/posrec_wCMT_example.ipynb

### DIFF
--- a/straxen/corrections_services.py
+++ b/straxen/corrections_services.py
@@ -76,10 +76,10 @@ class CorrectionsManagementServices():
             return self.get_pmt_gains(run_id, model_type, global_version)
         elif 'elife' in model_type:
             return self.get_elife(run_id, model_type, global_version)
-        elif 'NN_weights' in model_type:
+        elif model_type in ('mlp_model', 'cnn_model', 'gcn_model'):
             return self.get_NN_file(run_id, model_type, global_version)
         else:
-            raise ValueError(f'{correction} not found')
+            raise ValueError(f'{config_model} not found')
 
     # TODO add option to extract 'when'. Also, the start time might not be the best
     # entry for e.g. for super runs
@@ -234,21 +234,17 @@ class CorrectionsManagementServices():
         Smart logic to return NN weights file name to be downloader by 
         straxen.MongoDownloader()
         :param run_id: run id from runDB
-        :param model_type: model type and neural network type; mlp, gcn, cnn 
+        :param model_type: model type and neural network type; model_mlp, 
+        or model_gcn or model_cnn 
         :param global_version: global version
         :param return: NN weights file name
         """
-        if 'NN_weights' in model_type:
-            correction = None
-            if 'mlp' in model_type:
-                correction = 'mlp'
-            elif 'cnn' in model_type:
-                correction = 'cnn'
-            elif 'gcn' in model_type:
-                correction = 'gcn'
+        if model_type not in ('mlp_model', 'cnn_model', 'gcn_model'):
+            raise ValueError(f"{model_type} is not stored in CMT use on of 'mlp_model'"
+                             f" or 'cnn_model' or 'gcn_model'")
 
-            file_name = self._get_correction(run_id, correction, global_version)
-        
+        file_name = self._get_correction(run_id, model_type, global_version)
+
         return file_name
     def get_lce(self, run_id, s, position, global_version='v1'):
         """
--- a/straxen/corrections_services.py
+++ b/straxen/corrections_services.py
@@ -76,6 +76,8 @@ class CorrectionsManagementServices():
             return self.get_pmt_gains(run_id, model_type, global_version)
         elif 'elife' in model_type:
             return self.get_elife(run_id, model_type, global_version)
+        elif 'NN_weights' in model_type:
+            return self.get_NN_file(run_id, model_type, global_version)
         else:
             raise ValueError(f'{correction} not found')
 
@@ -112,9 +114,6 @@ class CorrectionsManagementServices():
         except KeyError:
             raise ValueError(f'Global version {global_version} not found for correction {correction}')
 
-        # for single value corrections, e.g. elife correction
-        if len(corrections) == 1:
-            return float(corrections)
         else:
             return corrections
 
@@ -159,7 +158,6 @@ class CorrectionsManagementServices():
                                  f'and should provide a float. Got: '
                                  f'{type(global_version)}')
             return float(global_version)
-
         else:
             raise ValueError(f'model type {model_type} not implemented for electron lifetime')
 
@@ -231,12 +229,32 @@ class CorrectionsManagementServices():
             np.save(cache_name, to_pe, allow_pickle=False)
         return to_pe
 
+    def get_NN_file(self, run_id, model_type, global_version='ONLINE'):
+        """
+        Smart logic to return NN weights file name to be downloader by 
+        straxen.MongoDownloader()
+        :param run_id: run id from runDB
+        :param model_type: model type and neural network type; mlp, gcn, cnn 
+        :param global_version: global version
+        :param return: NN weights file name
+        """
+        if 'NN_weights' in model_type:
+            correction = None
+            if 'mlp' in model_type:
+                correction = 'mlp'
+            elif 'cnn' in model_type:
+                correction = 'cnn'
+            elif 'gcn' in model_type:
+                correction = 'gcn'
+
+            file_name = self._get_correction(run_id, correction, global_version)
+        
+        return file_name
     def get_lce(self, run_id, s, position, global_version='v1'):
         """
         Smart logic to return light collection eff map values.
         :param run_id: run id from runDB
         :param s: S1 map or S2 map
-        :param global_version:
         :param position: event position
         """
         raise NotImplementedError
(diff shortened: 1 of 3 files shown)


straxen/plugins/position_reconstruction.py
```diff
@@ -108,12 +110,8 @@ def _get_model_file_name(self):
 
 @export
 @strax.takes_config(
-    strax.Option("file_mlp",
-                 help="Name of saved MLP model file in the aux file data base."
-                      "The file contains both structure and weights. Set to "
-                      "None to skip the computation of this plugin.",
-                 default="xnt_mlp_wfsim_20201214.tar.gz"
-                 )
+    strax.Option('mlp_model',
```
jorana, line 113: can you:

1. reinstate the None explanation? 
2. also for GCN and CNN.
3. Furthermore, we should add True to the tuple.
4. Set the default to ('mlp_model', 'ONLINE', True), we'd want to be using this from the start.

jorana (commented):
@ahiguera-mx , many thanks for adding this! It's clean and effective. Thanks for putting this together so fast. @terliuk and @reallsx are probably also happy with this PR!

I have some comments below. They can be summarized mostly by:

1. Can we allow direct paths to be input, as well as allowing `None`'s as input. Those would bypass the `get_corrections.get_NN_file` function. The `None`'s are a hard requirement, the direct paths are just me thinking it keeps things a little easier during testing.
2. Add the defaults CMT maps now, in this PR
3. `ffil` is save but for `v1` we use `interpolate`, does that also work with strings

straxen/corrections_services.py
```diff
@@ -231,12 +229,32 @@ def get_pmt_gains(self, run_id, model_type, global_version,
...
+    def get_NN_file(self, run_id, model_type, global_version='ONLINE'):
+        """
+        Smart logic to return NN weights file name to be downloader by 
+        straxen.MongoDownloader()
+        :param run_id: run id from runDB
+        :param model_type: model type and neural network type; mlp, gcn, cnn 
+        :param global_version: global version
+        :param return: NN weights file name
+        """
+        if 'NN_weights' in model_type:
```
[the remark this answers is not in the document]
jorana: okay, I think it should be 'mlp_model' etc but the point remains the same. If its indeed 'mlp_model', most likely we should update the docstring

straxen/corrections_services.py
```diff
@@ -76,6 +76,8 @@ def get_corrections_config(self, run_id, config_model=None):
             return self.get_pmt_gains(run_id, model_type, global_version)
         elif 'elife' in model_type:
             return self.get_elife(run_id, model_type, global_version)
+        elif 'NN_weights' in model_type:
+            return self.get_NN_file(run_id, model_type, global_version)
         else:
             raise ValueError(f'{correction} not found')
```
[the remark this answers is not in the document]
ahiguera-mx: good catch!

straxen/corrections_services.py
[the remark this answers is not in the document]
ahiguera-mx: great suggestion!

straxen/get_corrections.py
```diff
@@ -109,3 +109,24 @@ def get_elife(run_id, elife_conf):
...
+        raise ValueError(f'NN_model must have three elements: '
+                         f'the model type, its specific configuration '
+                         f'and detector (True = nT)')
+    model_type, model_conf, is_nt = nn_conf
+    if model_type == 'CMT_model':
+        cmt = straxen.CorrectionsManagementServices(is_nt=is_nt)
+        nn_file = cmt.get_corrections_config(run_id, model_conf)
+        nn_file = ' '.join(map(str, nn_file))    
+
+    elif model_type == 'local':
```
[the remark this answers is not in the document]
ahiguera-mx: agree

straxen/plugins/position_reconstruction.py
```diff
@@ -98,8 +99,9 @@ def compute(self, peaks):
         return result
 
     def _get_model_file_name(self):
-        config_file = f'file_{self.algorithm}'
-        model_file = self.config.get(config_file, "No file")
+        
+        config_file = f'{self.algorithm}_model'
```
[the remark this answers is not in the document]
ahiguera-mx: Indeed this is better way, I have implemented it

straxen/plugins/position_reconstruction.py
ahiguera-mx: done

straxen/corrections_services.py
```diff
@@ -231,12 +229,32 @@ def get_pmt_gains(self, run_id, model_type, global_version,
...
+        if 'NN_weights' in model_type:
+            correction = None
+            if 'mlp' in model_type:
+                correction = 'mlp'
+            elif 'cnn' in model_type:
+                correction = 'cnn'
+            elif 'gcn' in model_type:
+                correction = 'gcn'
+
+            file_name = self._get_correction(run_id, correction, global_version)
```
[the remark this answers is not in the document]
ahiguera-mx: Hi @jorana 
Indeed, `v1` fails as with the following error 
`TypeError: Cannot interpolate with all object-dtype columns in the DataFrame. Try setting at least one column to a numeric dtype.` 
I need to think how to handle this, suggestions are welcome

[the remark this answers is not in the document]
ahiguera-mx: Hi @jorana 
I add a condition for `nn models` to use`fill` only, so this fixes the issue

ahiguera-mx:
Thanks @jorana for the useful review, I've addressed your suggestions

straxen/corrections_services.py
```diff
@@ -105,16 +107,16 @@ def _get_correction(self, run_id, correction, global_version,
                         # on when something was processed therefore
                         # don't interpolate but forward fill.
                         df = self.interface.interpolate(df, when, how='fill')
+                    if correction in ('mlp_model', 'cnn_model', 'gcn_model'):
```
jorana, line 110: I could also imagine that you want to add this to the `_get_corrections` function as an optional argument and then pass it from get_NN_file. This also works and is very clear.

straxen/get_corrections.py
```diff
@@ -109,3 +109,22 @@ def get_elife(run_id, elife_conf):
...
+        raise ValueError(f'NN_model must be a tuple')
+    if not len(nn_conf) == 3:
+        raise ValueError(f'NN_model must have three elements: '
+                         f'the model type, its specific configuration '
+                         f'and detector (True = nT)')
+    model_type, model_conf, is_nt = nn_conf
+    if model_type == 'CMT_model':
+        cmt = straxen.CorrectionsManagementServices(is_nt=is_nt)
+        nn_file = cmt.get_corrections_config(run_id, model_conf)
+        nn_file = ' '.join(map(str, nn_file))
```
jorana, line 124: Sorry, this escaped my attention before, why do we need this line?

straxen/plugins/position_reconstruction.py
```diff
@@ -98,22 +99,31 @@ def compute(self, peaks):
...
 @strax.takes_config(
-    strax.Option("file_mlp",
-                 help="Name of saved MLP model file in the aux file data base."
-                      "The file contains both structure and weights. Set to "
-                      "None to skip the computation of this plugin.",
-                 default="xnt_mlp_wfsim_20201214.tar.gz"
-                 )
+    strax.Option('mlp_model',
+                 help='Neural network model.' 
+                      'If CMT, specify as (CMT_model, (mlp_NN_weights, ONLINE), True)))'
```
jorana, line 123: A little too fast a copy pasta 😄 , the help is now the same for all of the plugins.

jorana (approved):
Many thanks again for the changes! Works like a charm 😄 :
```python
>>> import straxen
>>> straxen.get_NN_file('0', ("CMT_model", ('mlp_model', "ONLINE"), True))
'xnt_mlp_wfsim_20201214.tar.gz'
>>> straxen.get_NN_file('0', ("CMT_model", ('gcn_model', "ONLINE"), True))
'xnt_gcn_wfsim_20201203.tar.gz'
>>> straxen.get_NN_file('0', ("CMT_model", ('cnn_model', "ONLINE"), True))
'xnt_cnn_wfsim_A_5_2000_20210112.tar.gz'
```

straxen/plugins/position_reconstruction.py
ahiguera-mx: yeah too much copy "pasta"... that sounds good

straxen/get_corrections.py
ahiguera-mx: not sure which line you are referring, but if is the last line, it is just for convince when I save the models into pandasDataFrame, and make sure if returns a simple string

straxen/corrections_services.py
ahiguera-mx: we could do that, lets see once we have more corrections, then we can have a better idea how we can improve the functionality of  `_get_corrections`

straxen/get_corrections.py
jorana: okay, `str(nn_file)` wasn't enough? (just curious doesn't matter at all)

jorana (approved):

WenzDaniel (approved):
Nothing to add. Thanks Aaron

PR #357, opened 2021-02-04
Fix #356

### LINKED ISSUE #356: Wrong color representation in straxen.plot_pmts()
straxen.plot_pmts(...) doesn't  attribute the right colour. The colorbar is valid only for the bottom PMT array.
I noticed plotting the results of PMT calibration. Here some example:

-  Top array between [1,2] and bottom array between [0,1]. The colour suggests that both the array are between [1,2]
``
fix = np.random.uniform(low=1, high=2, size=494)
fix[:253] = np.random.uniform(low=0, high=1, size=253)
straxen.plot_pmts(fix)
``
[top01_bot_12.pdf](https://github.com/XENONnT/straxen/files/5924821/top01_bot_12.pdf)

-  Top array between [0,1] and bottom array between [1,2]. As before, the colour suggest that both the array are between [0,1]
``
fix = np.random.uniform(low=1, high=2, size=494)
fix[:253] = np.random.uniform(low=0, high=1, size=253)
straxen.plot_pmts(fix)
``
[top12_bot_01.pdf](https://github.com/XENONnT/straxen/files/5924823/top12_bot_01.pdf)

Then, the colour representation is given always from the bottom array. Am I wrong? 

I will try to look into the function. For the moment let me open this issue. You might know already where is the problem.

jorana:
Fix issue #356

### DIFF
--- a/straxen/matplotlib_utils.py
+++ b/straxen/matplotlib_utils.py
@@ -104,7 +104,9 @@ def plot_on_single_pmt_array(
 
     kwargs.setdefault('s', 280)
     if log_scale:
-        kwargs.setdefault('norm', matplotlib.colors.LogNorm(vmin=vmin, vmax=vmax)
+        kwargs.setdefault('norm',
+                          matplotlib.colors.LogNorm(vmin=vmin,
+                                                    vmax=vmax))
     else:
         kwargs.setdefault('vmin', vmin)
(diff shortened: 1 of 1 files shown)


jorana:
Hereby proof that it works as it should
afbeelding

PR #358, opened 2021-02-05
Datarate dependent compressor

jorana:
[Based on Ivy's nice study](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:ivyli:lossless_compression), it's clear we can do a better job for computing on compressing the data at the daq. Especially if the rate is low, we can use more computationally expensive algorithms.

**What does this code do**
Use three different compressors based on:
1. Chunk size (there is a max size for some)
2. Datarate (if the datarate is high, we don't have time for large compression)

I hope @napoliion might at some point open a PR with a better motivated distinction between these different compressors. For the moment, these numbers are rough guesses and don't use potential compressors like `LZMA`.

### DIFF
--- a/bin/bootstrax
+++ b/bin/bootstrax
@@ -797,7 +797,7 @@ def infer_records_compressor(rd, datarate, n_fails):
     if datarate < 100:
         # Low datarate, we can do very large compression
         return 'bz2'
-    if chunk_size_mb > 2000:
+    if chunk_size_mb > 1000:
         # Extremely large chunks, let's use LZ4 because we know that it
         # can handle this.
         return 'lz4'
--- a/bin/bootstrax
+++ b/bin/bootstrax
@@ -202,7 +202,10 @@ timeouts = {
     # Bootstrax writes it's state to the daq-database. To have a backlog we store this
     # state using a TTL collection. To prevent too many entries in this backlog, only
     # create new entries if the previous entry is at least this old (in seconds).
-    'min_status_interval': 60
+    'min_status_interval': 60,
+    # Minimum time that the run needs to be ongoing before we can infer
+    # the datarate (s).
+    'min_time_in_run': 10,
 }
 
 # The disk that the eb is writing to may fill up at some point. The data should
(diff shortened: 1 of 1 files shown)


ershockley (approved):
Perfect, thanks @jorana! Between this and #343 I am quite happy :)

bin/bootstrax
```diff
@@ -709,19 +711,24 @@ def infer_mode(rd):
...
     # Get data rate from dispatcher
     try:
-        docs = ag_stat_coll.aggregate([
-            {'$match': {'number': rd['number']}},
-            {'$group': {'_id': '$detector', 'rate': {'$max': '$rate'}}}
-        ])
-        data_rate = int(sum([d['rate'] for d in docs]))
+        data_rate = None
+        started_looking = time.time()
+        while data_rate is None:
```
jorana, line 721: @darrylmasson I think this is why we might have seen more than usual failures at the eb lately. Before this nice aggregation we had a check to see if the data_rate actually returned something. Now we should again be in the good.

darrylmasson: Alternately, if we make sure the run hasn't started within the last 10 or 15 seconds, we can be sure that the dispatcher has been through a few update cycles.

jorana: Sure, this is what I had first but then decided against it because it would lead to unnecessary waiting time: https://github.com/XENONnT/straxen/pull/358/commits/96364e9cdbf89385237cf6442090fa9400ac9566

This was also because I set the time to 1 minute rather than 10 s :)

darrylmasson (approved):

bin/bootstrax
```diff
@@ -754,16 +761,48 @@ def infer_mode(rd):
...
+        # Cannot infer datarate or failed before, go for fast & safe
+        return 'lz4' if n_fails > 1 else 'zstd'
+
+    chunk_length = (rd['daq_config']['strax_chunk_overlap'] +
+                    rd['daq_config']['strax_chunk_length'])
+    chunk_size_mb = datarate*chunk_length
+    if datarate < 50:
+        # Low datarate, we can do very large compression
+        return 'bz2'
+    if chunk_size_mb > 1000:
```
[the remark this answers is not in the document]
jorana: Fair, you are right:
```python
import strax
a = np.zeros(int(2e8), dtype=np.int64)
print(f'Buffer of {a.nbytes/(1e9)} GB')
strax.save_file('test.test', a, compressor='zstd')
```

However, we need to keep in mind that we don't want to be running into issues where just one chunk is more chunky than the others, thereby disallowing us to save the file.

Nevertheless, I agree, let's set it to 1.8 GB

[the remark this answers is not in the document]
jorana: Solved in 
ee5277a

bin/bootstrax
jorana: Solved in 
b9c6220

PR #359, opened 2021-02-05
updates for seperate waveforms

petergaemers:

This PR add the top and bottom array waveforms to peaklets and beyond. For this small changes to the arguments of some peak_building/splitting functions where needed. Do not merge before, or much later than https://github.com/AxFoundation/strax/pull/393

See https://github.com/AxFoundation/strax/pull/393
`st.make(run_id,'peaks')

Waveforms

Ps tests fail untill the strax pr is merged

### DIFF
--- a/straxen/plugins/peaklet_processing.py
+++ b/straxen/plugins/peaklet_processing.py
@@ -61,7 +61,8 @@ export, __all__ = strax.exporter()
     strax.Option('saturation_min_reference_length', default=20,
                  help="Minimum number of reference sample used "
                       "to correct saturated samples"),
-
+    strax.Option('store_top_waveform',default=False,
+                 help='Bool for storing the top array waveform seperately'),
     *HITFINDER_OPTIONS,
 )
 class Peaklets(strax.Plugin):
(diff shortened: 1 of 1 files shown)


straxen/plugins/peaklet_processing.py
```diff
@@ -499,6 +502,8 @@ def compute(self, peaklets_he):
 
 @export
 @strax.takes_config(
+    strax.Option('n_top_pmts', type=int,
+                 help='Number of PMTs in the top array'),
```
WenzDaniel, line 506: Have you changed the option here by accident? At least the plugin itself stayed the same. Or I am blind, not unlikely on a Friday afternoon :D

WenzDaniel (commented):
Hej, Peter thanks for this awesome work. Let me also check strax before I approve.

straxen/plugins/peaklet_processing.py
petergaemers: I think you're confusing n_TOP_pmts with n_TPC_pmts

petergaemers: That was a mistake finding why merging crashed

PR #360, opened 2021-02-05
Test with database

jorana:
# Restructure CI for straxen
## This PR does the following:
- Add a readonly account to the github actions for testing.
- Make sure that we can test both with and without access to the database through utilix. The reason we need to is that to other forks, the github actions secrets are not available. Therefore, we need to be sure that we can do with and without access. To this end, I've set python 3.7 in the github actions to not run with the secrets via utilix. By the way, travis now only does python 3.7 without access to the database.
- Add some tests for the online monitor and CMT.
- Removes unclear utilix imports.
- Removes the coverage commentator option since it's obsolete.
- Move coveralls integration from Travis to github actions to allow the sectrets to be parsed via utilix.
- Move the installation of the pre-requirements (tensorflow, numpy, numba, etc) to a separate script.
- Fully cleanup all actions and travis instructions.
- Add a small function `utilix_is_configured` to `misc.py` to see if utilix is initialized in a way that we can get the secrets.
- Fixes a bug in the `FiducialEvents` cut with a mutable dtype.
- Simplify the testing of `test_plugins.py` by a lot.

## Solved issues
- https://github.com/XENONnT/straxen/issues/277, 
- https://github.com/XENONnT/straxen/issues/350. 

## About CI on different forks
As mentioned above, CI (github actions) does not gain access to our secrets on other forks than the main one (see PR #370).  Therefore I reconfigured the tests to be running with and without access to our database.

We do give a read only account to continuous integration in order to test our real software. Otherwise we're increasingly not testing stuff because we are relying more and more on the database.

Also, let's remove some unclear imports that cause errors in places one does not expect them.

# CI summary
After this PR I plan to use the following schema.

| Name | Required | Notes |  Time | 
| --- | --- | --- |  --- |
| Coveralls | Yes | Access to database & Numba disabled |  6 m |
| [coverage/coveralls status ](https://coveralls.io/github/XENONnT/straxen)| Yes | if >2% drop in coverage -> fail | 0 s after Coveralls |
| Py 3.6 | Yes | Access to database | 9 m |
| Py 3.7 | Yes | **No** access to database | 7 m |
| Py 3.8 | Yes | Access to database| 9 m
| Travis-PR py37 | No | Does releases & **No** access to database | 7 m + 0-4 h queue time |  
| Travis-Push py37 | No | Does releases  & **No** access to database | 7 m + 0-4 h queue time  |
| Codefactor | Yes | | 2 s|


jorana:
> 
> 
> @jorana in your fork can you try adding this to the utilix config file?
> 
> ```
> [basic]
> logging_level=debug
> ```
> 
> Might help with the utilix issue.

https://github.com/XENONnT/straxen/pull/351#issuecomment-774207309

@ershockley , thanks, yeah I added it in this PR

WenzDaniel (approved):
The concepts sounds thorough and good. I also read through your github actions voodoo, but I am not enough into this topic to understand it from a technical point of view.

Issue #353, opened 2021-01-31
Dropping datafields in numba from dtype to dtype


jorana:
We have a few instances in strax/straxen where we basically hardcode copy a given datatype to another (e.g. [this one](https://github.com/XENONnT/straxen/blob/master/straxen/plugins/veto_hitlets.py#L140)). We might want to use a more dynamic function like this one from the [PyHEP Numba tutorial on February 3, 2021](https://nbviewer.jupyter.org/github/jpivarski-talks/2021-02-03-pyhep-numba-tutorial/blob/main/tutorial.ipynb) from Jim Pivarski. Might be much easier to maintain than having to make a new patch every time a dtype name changes:

```python
args = [
    np.random.normal(0, 1, 1000000),
    np.random.normal(0, 1, 1000000),
    np.random.normal(0, 1, 1000000),
    np.random.normal(0, 1, 1000000),
#     np.random.normal(0, 1, 1000000),
]

code = f"""
@nb.vectorize
def sum_in_quadrature({", ".join("p%d" % i for i in range(len(args)))}):
    # compiled for exactly {len(args)} arguments
    return np.sqrt({" + ".join("p%d**2" % i for i in range(len(args)))})
"""

print(code)
exec(code)
sum_in_quadrature(*args)
```

WenzDaniel:
Sorry, by accident...

PR #361, opened 2021-02-06
allow dynamic copy of dtype

jorana:
See XENONnT/straxen#353

**Can you briefly describe how it works? & Can you give a minimal working example (or illustrate with a figure)?**
See https://github.com/AxFoundation/strax/pull/395

**Prerequisites**
https://github.com/AxFoundation/strax/pull/395

### DIFF
--- a/straxen/plugins/led_calibration.py
+++ b/straxen/plugins/led_calibration.py
@@ -96,8 +96,7 @@ def get_records(raw_records, baseline_window):
               (('Waveform data in raw ADC counts', 'data'), 'f4', (record_length,))]
 
     records = np.zeros(len(raw_records), dtype=_dtype)
-    strax.copy_to_buffer(raw_records, records, "_rr_to_r_led",
-                         field_names=records.dtype.names)
+    strax.copy_to_buffer(raw_records, records, "_rr_to_r_led")
 
     mask = np.where((records['record_i'] == 0) & (records['length'] == 160))[0]
     records = records[mask]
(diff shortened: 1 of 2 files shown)


GiovanniVolta:
It looks okay to me. It should not have an impact on the computation but maybe is better if I try it or did you already try it?

WenzDaniel (approved):
Looks, good to me thanks again. About the performance: Yes we add a little overhead but should be negligible as long as you are not copying just single value. At least this would be my guess.

jorana:
In fact, it should be faster than it used to be @GiovanniVolta . Previously, this was done without numba which slower (depends also little on your machine).

afbeelding

# Here is the full example, the number of n_copy did not change the factor ~3 I got here
```python
import numba
import strax
import numpy
import numpy as np

def copy_unjitted(records, raw_records):
    for d in ('time', 'length', 'dt', 'channel', 'pulse_length', 'record_i', 'baseline', 'data'):
        records[d] = raw_records[d]
        
n_copy = 100000
rr = np.ones(n_copy, dtype=strax.raw_record_dtype())
r = np.empty(n_copy, dtype=strax.record_dtype())

print(f'Using {rr.nbytes/1e9} GB rr and {r.nbytes/1e9} GB records')


%%timeit
copy_unjitted(r, rr)

%%time
strax.copy_to_buffer(rr, r, "_test_r_to_buffer")

%%timeit
strax.copy_to_buffer(rr, r, "_test_r_to_buffer")

%load_ext memory_profiler
%memit copy_unjitted(r, rr)

%memit strax.copy_to_buffer(rr, r, "_test_r_to_buffer")
```

GiovanniVolta:
Perfect! Thank you guys!

PR #363, opened 2021-02-09
Update contexts.py

petergaemers:
This pull request:

- removes the to_pe_nt file from the gain model and replaces it with the actual value used. 0.01 is the average to_pe value. If you want a more fancy gain model use the CMT

### DIFF
--- a/straxen/contexts.py
+++ b/straxen/contexts.py
@@ -147,7 +147,7 @@ def xenonnt_led(**kwargs):
 # This gain model is a temporary solution until we have a nice stable one
 def xenonnt_simulation(output_folder='./strax_data'):
     import wfsim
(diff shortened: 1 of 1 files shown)


jorana:
Thanks Peter, looks good. Tagging @ramirezdiego in here too because he opened this issue:
https://github.com/XENONnT/WFSim/issues/89

I'm fine with merging this.

jorana (approved):

ramirezdiego:
Hey, thanks for tagging me. I thought we still wanted to have some sort of "blessed" file as default for the gains (but of course encourage people to use the CMT and query the information of a specific run), but this makes sense as well.

ramirezdiego:
Sorry, I am too lazy to open an issue. Don't you think we should do the same for the nVeto gains model inside the nT sims context? This is, replacing CMT default by some reliable value. CC @WenzDaniel for the nVeto details.

PR #364, opened 2021-02-11
Change left/right extension for event processing

jingqiangye:
As discussed in this [note](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:event_pile_up_commissioning), we should reduce the right extension to decrease the pile-up. Also as @jpienaar13 pointed out, we should increase the left extension a bit.


jorana (approved):
Thanks, looks good!

Issue #369, opened 2021-02-17
Use daqlogger for ajax


jorana:
Use the daqnt logger for ajax to write log messages locally

jorana:
fixed in #374

PR #374, opened 2021-02-19
Update tagging

jorana:
- Fix issue #369, let ajax log
- Be consistent with https://github.com/XENONnT/nodiaq/pull/10

We need to agree which people and how can delete `bad` (`abandoned`) data.


bin/ajax
```diff
@@ -879,6 +875,18 @@ if __name__ == '__main__':
     'data, old raw_records_nv')
 
     args = parser.parse_args()
+    hostname = socket.getfqdn()
+
+    try:
+        import daqnt
+        log_name = 'ajax_' + hostname + ('' if not args.execute else 'TESTING')
```
darrylmasson, line 882: Do we maybe want `'_TESTING'` so we don't get `ajax_eb0.xenon.localTESTING`?

darrylmasson (approved):
We'll want to synchronize the deployment of this change with the website, and probably also do a mass-update of the database at the same time.

jorana:
@darrylmasson converting this PR to a draft until we get the green light on the website-side

bin/bootstrax
```diff
@@ -1302,6 +1306,8 @@ def process_run(rd, send_heartbeats=args.production):
     log.info(f"Starting processing of run {rd['number']}")
     if rd is None:
         raise RuntimeError("Pass a valid rundoc, not None!")
+    elif args.production and rd.get('bootstrax', {}).get('state', None) == 'done':
+        raise RuntimeError(f'{rd["number"]} is done already, do not make a mass')
```
darrylmasson, line 1310: Should this say `mess` rather than `mass`? Also, given that the dispatcher guarantees the existence of `bootstrax.state`, are the nested `.get()`s necessary?

bin/bootstrax
```diff
@@ -1620,8 +1626,10 @@ def cleanup_db():
     abandon_queries = [
         ({'tags': {
             '$elemMatch': {
-                'name': 'bad'}}},
-         "Run has a 'bad' tag"),
+                'name': 'abandon'}},
+             'start': {'$gt': now(-timeouts['abandoning_allowed'])}
```
darrylmasson, line 1630: We could redo this query as `{'tags.name': 'abandon', 'start': {'$gt': now(-timeouts['abandoning_allowed'])}}`. Also, do we want to consider the time when the tag was added?

darrylmasson (approved):

bin/bootstrax
jorana: Divine intervention is not appreciated. 

Thanks for spotting it! 

`are the nested .get()s necessary`, looks like not but also doesnt hurt much.

bin/bootstrax
jorana: > Also, do we want to consider the time when the tag was added?

Not per se, usually one of the older ebs catches it within 10 s

PR #379, opened 2021-02-22
Fixes/improvements for 'plot_pulses'

DCichon:
`plot_pulses`: Made it possible to pass `None` as a value for `max_plots` to explicitly remove the pulse limit for plotting. Fixed file name generation for PDF output (it attempted to extract some column called 'time' from the `time_range` tuple beforehand).


WenzDaniel (approved):
Thanks a lot for the improvement and bug fix. No clue how the filename issue could have happened.

jorana (approved):
Thanks Dominick

Issue #376, opened 2021-02-21
Function _read_and_interpolate duplicate and not used


jorana:
While working on the coverage of the straxen package, I noticed that this function is never used [L123](https://github.com/XENONnT/straxen/blob/280eac854a4329785b299a85044d38f575a11125/straxen/corrections_services.py#L123)

It looks like it's a bit duplicate with this function [L88](https://github.com/XENONnT/straxen/blob/280eac854a4329785b299a85044d38f575a11125/straxen/corrections_services.py#L88)

I've also quickly checked here but I cannot find it being used:
- https://github.com/XENONnT/analysiscode/tree/master/corrections

If @ahiguera-mx agrees, we could remove this function.

jorana:
fixed in #380

PR #380, opened 2021-02-22
Fix issue #376, remove unused code block

jorana:
Fix issue https://github.com/XENONnT/straxen/issues/376

Remove a code block that isn't used


ahiguera-mx:
Good catch @jorana

ahiguera-mx (approved):
Good catch

PR #383, opened 2021-02-23
Set secrets for scada

jorana:
Master is "failing" because I did not update some sectrets yet, fix that with this PR


WenzDaniel (approved):
Thanks a lot Joran. Do you have some readings about github actions you can recommend? I also would like to look into this a bit more.

jorana:
> 
> 
> Thanks a lot Joran. Do you have some readings about github actions you can recommend? I also would like to look into this a bit more.

To be honest I don't really, all I learned was via google and iteratively trying things 😉.  It does help that I now set it up for ~10 different packages.

PR #384, opened 2021-02-24
Replace json with commentjson

zhut19:

Some of the external configurations are saved in JSON format, however, comments are not allowed in it. This can be solved by using [commentjson](https://pypi.org/project/commentjson/) package.


jorana (approved):
Nice package Tianyu! I see no problems in going forward technically but there is some discussion in XENONnT/private_nt_aux_files#24, want to converge on that first?

PR #386, opened 2021-02-26
fix commentjson

jorana:
#384 does not work for zipped json files.

Don't use #384 for json.gz


zhut19 (commented):
Looks good.

PR #387, opened 2021-03-01
Check for raw-record overlaps

jorana:
I'm not sure why [we needed this for LED-calibrations](https://github.com/XENONnT/straxen/pull/42/commits/2a0b900a253aeb524f7cc83ac363ffa8b86b9b76) but I don't think we should have this option set to `False`.

 - I will first have to test if this works on all runs, but perhaps we should only disable it for LED if needed.


WenzDaniel:
I also do not understand why it would has to be switched off. For real data it should never ever be possible that raw_records overlap unless there is a bug. Further, it breaks certain algorithms as e.g. `strax.record_links`.

WenzDaniel (approved):

PR #390, opened 2021-03-02
Check raw-record overlaps at veto pulse processing

jorana:
Similar to the TPC raw-records (#387), we should check that the raw-records of our veto systems are non-overlapping. 

~We could do this also at the rr-nv-coincidence but this would lead to problems. If we fail this check (hopefully not) we would get the error during the building of rr-nv-coincidence. This leads to issues at the DAQ because we will delete rr-nv and only keep rr-nv-coincidence.~

After discussing with Daniel, he convinced me to do this at the rr-nv-coincidence step as otherwise the damage is already done. This further complicates the rr-nv cleaning step but let's hope we are indeed never running into this issue.

### DIFF
--- a/bin/bootstrax
+++ b/bin/bootstrax
@@ -1261,7 +1261,7 @@ def run_strax(run_id, input_dir, targets, readout_threads, compressor,
                                 daq_chunk_duration=daq_chunk_duration,
                                 daq_overlap_chunk_duration=daq_overlap_chunk_duration,
                                 readout_threads=readout_threads,
-                                check_raw_record_overlaps=False,
+                                check_raw_record_overlaps=True,
                                 )
 
             for i, c in enumerate(st.get_iter(run_id,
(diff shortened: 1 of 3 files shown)


WenzDaniel:
Hmm, would not be the best place for such a check the DAQReader? Or do you just want to crash the processing of the one detector but not the other in case of overlapping records?

jorana:
Of course in an ideal world, we would check as low level as possible, however given exactly the rationale above, this doestn't make sense. If we cannot process the live-data because of some pulse overlap, that entire run gets lost staying on the DAQ forever u-processable. Therefore, we need this check to notice something is wrong, we should however not end up with a DAQ filled with bad live-data, we should ship out the raw-records and someone should look at the raw-records to see what is going on.

WenzDaniel:
Okay sounds good. Thanks for the explanation.

WenzDaniel (approved):
Thanks a lot. I approve lets hope we never run into this.

PR #394, opened 2021-03-05
add daq-plots to minianalysies

jorana:
Code for making plots like these:
https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:dsg:daq:2021_desync#what_the_data_looks_like

Add a new ordering scheme to the record-array plots where each of the channels is regrouped to a particular DAQ-hardware ordering. This is especially usefull for things like correlated noise studies that might work across different parts of the hardware.

```python
# Plot with three pannels
st.daq_plot(run_id, seconds_range = (0.000732,0.0007374))
# Single plot, sort by link or ADC ID
st.plot_records_matrix(run_id, seconds_range = (0.000732,0.0007374), group_by='link')
st.plot_records_matrix(run_id, seconds_range = (0.000732,0.0007374), group_by='ADC ID')
```

**Why is this not tested**
I'm thinking of how to do this best, trouble is that the data we have for testing is ancient and for 1T this doesn't work at all because I don't know where this info was stored than. I'm still thinking if I can come up with something smart.

**Can you add some more comments?**
Yes, I should do that


straxen/analyses/daq_waveforms.py
```diff
@@ -0,0 +1,141 @@
...
+    """
+    Query the runs database for the config of the daq during this run.
+    Either use the context of the runs collection.
+    """
+    if not context.storage[0].__class__.__name__ == 'RunDB' and run_collection is None:
+        raise NotImplementedError('Only works with the runsdatabase')
+    if run_collection is None:
+        run_collection = context.storage[0].collection
+    daq_config = run_collection.find_one({"number": int(run_id)},
+                                         projection={config_name: 1}).get(config_name, None)
```
darrylmasson, line 59: I don't strictly agree with this line. If `find_one` returns None this will throw, because `None.get()` isn't a thing. If the document exists, then the `daq_config` field also has to exist. If we somehow manage to find a malformed rundoc, the fact that this code will error is the least of our worries.

straxen/analyses/daq_waveforms.py
```diff
@@ -0,0 +1,141 @@
...
+        try:
+            if int(bdoc['board']) == board:
+                res = f"{bdoc['host']}_link{bdoc['link']}"
+                if add_crate:
+                    res += f"_crate{bdoc['crate']}"
+                return res
+        except KeyError:
+            raise ValueError(f'Invalid DAQ config {daq_config} or board {board}')
+    # This happens if the board is not in the channel map which might
+    # happen for very old runs.
```
darrylmasson, line 77: The current map predates the oldest data, I'd rather throw because this would mean there is something going wrong, and an "unknown" would probably be ignored.

straxen/analyses/daq_waveforms.py
```diff
@@ -0,0 +1,141 @@
...
+    return 'unknown'
+
+
+def _get_cable_map(name: str = 'xenonnt_cable_map.csv') -> pandas.DataFrame:
+    """Download the cable map and return as a pandas dataframe"""
+    down = straxen.MongoDownloader()
```
darrylmasson, line 83: If we dig through the rundoc to construct the board map, why don't we do the same for the cable map, rather than going to an entirely different source?

straxen/analyses/waveform_plot.py
```diff
@@ -145,19 +147,33 @@ def plot_records_matrix(context, run_id,
                     max_samples=max_samples,
                     ignore_max_sample_warning=ignore_max_sample_warning,
                     **kwargs)
+    if group_by:
```
darrylmasson, line 150: Maybe an explicit `if group_by is not None:`? I've had linters complain at me about this

straxen/analyses/waveform_plot.py
```diff
@@ -145,19 +147,33 @@ def plot_records_matrix(context, run_id,
...
-            vmax=wvm.max(),),
+            vmax=wvm.max(), ),
         cmap=plt.cm.inferno)
     plt.xlim(*seconds_range)
 
     ax = plt.gca()
-    seconds_range_xaxis(seconds_range)
-    ax.invert_yaxis()
-    plt.ylabel("PMT Number")
+    if group_by:
```
darrylmasson, line 166: Again here

straxen/analyses/daq_waveforms.py
jorana: I *always* screw this up. Thanks

straxen/analyses/daq_waveforms.py
jorana: The daq-config is quite a messy document for these purposes, it has for each thing a totally different layout. Using the cable map, we can more easily also group by other. For example, the boards are a list of dicts, channels is a dict of  lists, registers is a list of dicts where boards are strings meaning something entirely different from the "board" in boards. I didn't feel like going that any more than I needed to.

Also the cable map allows to also easily use grouping by e.g. amplifier or other columns in the cable map:
```python
st.daq_plot(run_id, seconds_range = (0.000732,0.0007374), group_by='Amp (crate.slot.channel)')
```

straxen/analyses/waveform_plot.py
jorana: Thanks good idea

straxen/analyses/waveform_plot.py
jorana: thanks

straxen/analyses/daq_waveforms.py
jorana: For example for run 009104 we had this

straxen/analyses/daq_waveforms.py
darrylmasson: That's fair. Without DAQ DB access the csv is the only other option for this information.

straxen/analyses/daq_waveforms.py
darrylmasson: .... how?

jorana: Boards 749 and 1440 aren't in the daq config

jorana:
@darrylmasson thanks a lot for the useful comments in the review

straxen/analyses/daq_waveforms.py
darrylmasson: Ah, right, we swapped 124 and 144 out

jorana: I think 9104 are even older since these boards you mention weren't replaced by the aforementioned boards:
https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:dsg:daq:summer2020work#and_144_are_very_questionable_boards_done

darrylmasson: Originally Junji swapped 1253 and 1163 for 749 and 1440, then later we swapped 124 and 144 for 1253 and 1163. I don't remember the exact dates of when these happened, but the cable map in the rundoc does reflect the changes.

jorana:
@darrylmasson , for good measure, do you approve?

HenningSE (commented):
Thanks Joran fo adding these plots!

PR #395, opened 2021-03-07
Infer-target

jorana:
Online_peak_monitor was removed from the target list because of the str match with 'peak'. Fix that.

While at it, also add some more logging statements and go for a little simpler syntax.

Honestly, this is not the prettiest part of the code. Will do some work on it when we add some new infer-logic for the linked mode as well as NV/MV processing.

### DIFF
--- a/bin/bootstrax
+++ b/bin/bootstrax
@@ -260,7 +260,13 @@ max_queue_new_runs = 2
 # this many times. If high level data is hitting some edge case, we
 # might want to be able to keep the intermediate level data.
 # NB: string match so events applies e.g. to event_basics, peak to e.g. peaklets
-remove_target_after_fails = dict(events=2, hitlets=2, peak=3, online=4, ALL=5)
+remove_target_after_fails = dict(events=2,
+                                 hitlets=2,
+                                 peaks=3,
+                                 peak_basics=3,
+                                 peaklets=4,
+                                 online=4,
+                                 ALL=5)
 
 ##
 # Initialize globals (e.g. rundb connection)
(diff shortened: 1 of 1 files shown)


bin/bootstrax
```diff
@@ -538,11 +544,19 @@ def infer_target(rd):
 
     n_fails = rd['bootstrax'].get('n_failures', 0)
     if n_fails:
-        log.debug(f'override_target::\tdeleting targets')
+        log.debug(f'Deleting targets')
         for _rm_t, _rm_t_after in remove_target_after_fails.items():
-            _t = [__t for __t in _t if (_rm_t not in __t and n_fails < _rm_t_after)]
-            _pp = [__pp for __pp in _pp if (_rm_t not in __pp and n_fails < _rm_t_after)]
+            for __i, __t in enumerate(_t):
```
darrylmasson, line 549: I think it's worth cleaning up the variable names here, `i` is cleaner than `__i`, and `potential_targets` is cleaner than `_t`. This isn't the c++ standard library, I don't think we need _this_ many underscores.

bin/bootstrax
```diff
@@ -538,11 +544,19 @@ def infer_target(rd):
...
+                if __t in _rm_t and n_fails > _rm_t_after:
```
darrylmasson, line 550: Is the logic here the right way around? Previously this was `_rm_t in __t`

darrylmasson (commented):
One `in` seems to me to be reversed

bin/bootstrax
jorana: >Honestly, this is not the prettiest part of the code. Will do some work on it when we add some new infer-logic for the linked mode as well as NV/MV processing.

yeah, this whole thing got messier than it supposed to

bin/bootstrax
jorana: Logic is reverted, now this `if` -> deletes targets if we want to, whereas it used to say `if` -> keep the target. Therefore there is also no `not`  involved for `__t in _rm_t`

PR #396, opened 2021-03-11
New EventPositions for XENONnT

ftoschi:

The `EventPositions` plugin has been changed to take into account the new FDC maps. 

`EventPositions` checks if the `fdc_map` option is a tuple or a string: in the first case this is handled by CMT and therefore it is assumed to be XENONnT (aka we need to pick an algorithm), otherwise it is the path to a map. If it is XENONnT data, the FDC map for the default algorithm is loaded. The maps are now given as (x, y, drift time), therefore the z coordinate needs to be scaled. The corrections are calculated as previously.

Minor changes:
- default drift time for XENONnT changed to 0.632 mm/µs;
- added function to scale InterpolationMap.

```
import straxen

st = straxen.contexts.xenonnt_online()
st_gcn = st.new_context(config=dict(default_reconstruction_algorithm='gcn'))

run_id = ['000000', '000001']
df = st.get_df(run_id, ['event_positions'])
df_gcn = st_gcn.get_df(run_id, ['event_positions'])
```

where this would reprocess `event_basics` for the gcn context.

### DIFF
--- a/straxen/contexts.py
+++ b/straxen/contexts.py
@@ -46,6 +46,11 @@ xnt_common_config = dict(
     # Clustering/classification parameters
     s1_max_rise_time=100,
     s2_xy_correction_map=("CMT_model", ('s2_xy_map', "ONLINE"), True),
+    default_reconstruction_algorithm='mlp',
+)
+
+xnt_common_config.update(dict(
+    fdc_map=("CMT_model", (f'fdc_map_{xnt_common_config["default_reconstruction_algorithm"]}', "ONLINE"), True))
 )
 
 # Plugins in these files have nT plugins, E.g. in pulse&peak(let)
--- a/straxen/contexts.py
+++ b/straxen/contexts.py
@@ -1,7 +1,6 @@
 from immutabledict import immutabledict
 import strax
 import straxen
-import numpy as np
 
 
 common_opts = dict(
@@ -46,6 +45,7 @@ xnt_common_config = dict(
         nveto_blank=(2999, 2999)),
     # Clustering/classification parameters
     s1_max_rise_time=100,
+    s2_xy_correction_map=("CMT_model", ('s2_xy_map', "ONLINE"), True),
 )
 
 # Plugins in these files have nT plugins, E.g. in pulse&peak(let)
@@ -242,6 +242,7 @@ x1t_common_config = dict(
     left_event_extension=int(1e6),
     right_event_extension=int(1e6),
 )
+
 
 
 def demo():
--- a/straxen/contexts.py
+++ b/straxen/contexts.py
@@ -46,11 +46,8 @@ xnt_common_config = dict(
     # Clustering/classification parameters
     s1_max_rise_time=100,
     s2_xy_correction_map=("CMT_model", ('s2_xy_map', "ONLINE"), True),
-    default_reconstruction_algorithm='mlp',
-)
-
-xnt_common_config.update(dict(
-    fdc_map=("CMT_model", ('fdc_map', "ONLINE"), True))
+    elife_file=("elife_model", "ONLINE",True),
+    fdc_map=("CMT_model", ('fdc_map', "ONLINE"), True),
 )
 
 # Plugins in these files have nT plugins, E.g. in pulse&peak(let)
@@ -156,7 +153,7 @@ def xenonnt_simulation(output_folder='./strax_data'):
     st = strax.Context(
         storage=strax.DataDirectory(output_folder),
         config=dict(detector='XENONnT',
-                    fax_config='fax_config_nt.json',
+                    fax_config='fax_config_nt_design.json',
                     check_raw_record_overlaps=False,
                     **straxen.contexts.xnt_common_config,
                     ),
--- a/straxen/contexts.py
+++ b/straxen/contexts.py
@@ -50,7 +50,7 @@ xnt_common_config = dict(
 )
 
 xnt_common_config.update(dict(
-    fdc_map=("CMT_model", (f'fdc_map_{xnt_common_config["default_reconstruction_algorithm"]}', "ONLINE"), True))
+    fdc_map=("CMT_model", ('fdc_map', "ONLINE"), True))
 )
 
 # Plugins in these files have nT plugins, E.g. in pulse&peak(let)
--- a/straxen/plugins/event_processing.py
+++ b/straxen/plugins/event_processing.py
@@ -251,7 +251,7 @@ class EventBasics(strax.LoopPlugin):
                 if s_i == 2:
                     for name in posrec_save:
                         posrec_result[f's{s_i}_{name}'] = main_s[s_i][name]
-
+                        
             # Store alternate signal properties
             if _alt_i is None:
                 result[f'alt_s{s_i}_index'] = -1
@@ -290,7 +290,7 @@ class EventBasics(strax.LoopPlugin):
         default=0.632e-4
     ),
     strax.Option(
-        'fdc_map',
+        name='fdc_map',
         help='3D field distortion correction map path',
         default_by_run=[
             (0, pax_file('XENON1T_FDC_SR0_data_driven_3d_correction_tf_nn_v0.json.gz')),  # noqa
@@ -305,16 +305,21 @@ class EventBasics(strax.LoopPlugin):
             170925_0622, pax_file('XENON1T_FDC_SR1_data_driven_time_dependent_3d_correction_tf_nn_part4_v1.json.gz'))],
     # noqa
     ),
+    strax.Option(
+        name="default_reconstruction_algorithm_fdc",
+        help="Default reconstruction algorithm used for field distortion corrected position",
+        default="mlp"),
     *DEFAULT_POSREC_ALGO_OPTION
 )
 class EventPositions(strax.Plugin):
     """
     Computes the observed and corrected position for the main S1/S2
     pairs in an event. For XENONnT data, it returns the FDC corrected
-    positions of the default_reconstruction_algorithm.
+    positions of the default_reconstruction_algorithm_fdc, not necessary 
+    identical to the default_reconstruction_algorithm.
     """
 
-    depends_on = ('event_basics', 'event_posrec_many')
+    depends_on = ('event_basics','event_posrec_many')
 
     __version__ = '0.1.3'
 
@@ -339,17 +344,24 @@ class EventPositions(strax.Plugin):
 
     def setup(self):
 
-        is_CMT = isinstance(self.config["field_map"], tuple)
+        if self.config['default_reconstruction_algorithm_fdc'] != self.config['default_reconstruction_algorithm']:
+            warn(f'Default posrec algorithm ({self.config["default_reconstruction_algorithm"].upper()}) '
+                 'used for uncorrected positions is different than the default algorithm used for '
+                 f'the corrected positions ({self.config["default_reconstruction_algorithm_fdc"].upper()}).')
+        is_CMT = isinstance(self.config['fdc_map'], tuple)
 
         if is_CMT:
-            name_map = self.config["field_map"][1][0].lower()
-            if not self.config['default_reconstruction_algorithm'].lower() in name_map: # check the name
-                raise NotImplementedError(('No FDC map found for posrec algorithm ',
-                                           f'{self.config["default_reconstruction_algorithm"].upper()}'))
+            
+            map_algo = list(self.config['fdc_map'])
+            name_map = list(map_algo[1])
+            name_map[0] = '_'.join([name_map[0], self.config['default_reconstruction_algorithm_fdc']])
+            map_algo[1] = tuple(name_map)
+            map_algo = tuple(map_algo)
+            
             self.map = InterpolatingMap(
-                get_resource(get_config_from_cmt(self.run_id, self.config['fdc_map']), fmt='binary'))
+                get_resource(get_config_from_cmt(self.run_id, map_algo), fmt='binary'))
 
-        elif isinstance(self.config["field_map"], str):
+        elif isinstance(self.config['fdc_map'], str):
             self.map = InterpolatingMap(
                 get_resource(self.config['fdc_map'], fmt='binary'))
 
@@ -362,9 +374,11 @@ class EventPositions(strax.Plugin):
 
         result = {'time': events['time'],
                   'endtime': strax.endtime(events)}
+        
+        algo = self.config['default_reconstruction_algorithm_fdc']
 
         z_obs = - self.config['electron_drift_velocity'] * events['drift_time']
-        orig_pos = np.vstack([events['s2_x'], events['s2_y'], z_obs]).T
+        orig_pos = np.vstack([events[f's2_x_{algo}'], events[f's2_y_{algo}'], z_obs]).T
         r_obs = np.linalg.norm(orig_pos[:, :2], axis=1)
         delta_r = self.map(orig_pos)
 
--- a/straxen/plugins/event_processing.py
+++ b/straxen/plugins/event_processing.py
@@ -3,7 +3,7 @@ import strax
 import numpy as np
 
 from straxen.common import pax_file, get_resource, first_sr1_run, aux_repo
-from straxen.get_corrections import get_elife
+from straxen.get_corrections import get_elife, get_config_from_cmt
 from straxen.itp_map import InterpolatingMap
 export, __all__ = strax.exporter()
 
@@ -366,8 +366,10 @@ class EventPositions(strax.Plugin):
             (0, pax_file('XENON1T_s1_xyz_lce_true_kr83m_SR0_pax-680_fdc-3d_v0.json')),  # noqa
             (first_sr1_run, pax_file('XENON1T_s1_xyz_lce_true_kr83m_SR1_pax-680_fdc-3d_v0.json'))]),  # noqa
     strax.Option(
-        's2_relative_lce_map',
-        help="S2 relative LCE(x, y) map",
+        's2_xy_correction_map',
+        help="S2 (x, y) correction map. Correct S2 position dependence "
+             "manly due to bending of anode/gate-grid, PMT quantum efficiency "
+             "and extraction field distribution, as well as other geometric factors.",
         default_by_run=[
             (0, pax_file('XENON1T_s2_xy_ly_SR0_24Feb2017.json')),
             (170118_1327, pax_file('XENON1T_s2_xy_ly_SR1_v2.2.json'))]),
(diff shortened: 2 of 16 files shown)


ftoschi:
@ahiguera-mx the failing test (`test_plugins`) seems to be realted to CMT, it complains about utilix not having rundb attribute. I tried to run the tests locally and it does not fail.

ahiguera-mx:
> @ahiguera-mx the failing test (`test_plugins`) seems to be related to CMT, it complains about utilix not having rundb attribute. I tried to run the tests locally and it does not fail.

Hi @ftoschi I think you need to add a "default" file to `test_plugins` something like this https://github.com/XENONnT/straxen/blob/master/tests/test_plugins.py#L18 
maybe @jorana can point to you to the current repo/file that need to be added

ftoschi:
I added a minianalyses for people who want to compare the performance of position reconstruction algorithms after the FDC is applied: as the `event_positions` plugin depends on the default posrec algorithm, changing it would imply reprocessing the `event_basics` data. As the needed information is already stored in the `event_posrec_many` plugin, if we want to compare different position reconstruction algorithms it is not necessary to reprocess every time the `event_basics`. Given a specific run_id and a specific set of position reconstruction algorithms (together with their version in CMT), this minianalysis fetches the corresponding FDC map and returns a numpy array with the corrected position for each position.

straxen/analyses/posrec_comparson.py
```diff
@@ -0,0 +1,89 @@
...
+                        are required for different posrec algorithms, default 'ONLINE')
+    :param posrec_algos: list of position reconstruction algorithms to use (default ['mlp', 'gcn', 'cnn'])
+    """
+    
+    posrec_algos = (posrec_algos, ) if isinstance(posrec_algos, str) else posrec_algos
+
+    if hasattr(cmt_version, '__len__') and not isinstance(cmt_version, str) and len(cmt_version) != len(posrec_algos):
+        raise TypeError(f"cmt_version is a list but does not match the posrec_algos ({posrec_algos}) length.")
+        
+    cmt_version = [cmt_version] * len(posrec_algos) if isinstance(cmt_version, str) else cmt_version
```
jorana, line 28: This certainly works, you can however also get this information from the context itself, that way people don't end up with the `ONLINE` versions when we are actually using `v1` in the context.

straxen/analyses/posrec_comparson.py
```diff
@@ -0,0 +1,89 @@
...
+                  f'r_field_distortion_correction_{algo}'), np.float32),
+                ((f'Interaction angular position (radians) - {algo.upper()} posrec algorithm', f'theta_{algo}'),
+                 np.float32)]
+        
+    dtype += [(('Interaction z-position using mean drift velocity only (cm)', 'z_naive'), np.float32)]
+    result = np.zeros(len(events), dtype=dtype)
+
+    drift_speed = context.show_config('event_positions', 'electron_drift_velocity').current.values[0]
+    drift_speed = drift_speed if drift_speed != strax.OMITTED \
+        else context.show_config('event_positions', 'electron_drift_velocity').default.values[0]
```
jorana, line 57: For now, this seems like the reasonable approach but I would guess we'd also track this in CMT rather soon, right @ahiguera-mx ?

By the way, you can also do a little more simple command of:
```python
context.config.get('electron_drift_velocity')
```

straxen/analyses/posrec_comparson.py
```diff
@@ -0,0 +1,89 @@
...
+
+import strax
+import straxen
+
+export, __all__ = strax.exporter()
+
+@straxen.mini_analysis(requires=('event_basics', 'event_posrec_many',))
+def load_corrected_positions(context, run_id, events,
+                             cmt_version='ONLINE',
+                             posrec_algos=['mlp', 'gcn', 'cnn']):
```
jorana, line 12: Can you make this a tuple? List as arguments are:
[the source of all evil](https://florimond.dev/blog/articles/2018/08/python-mutable-defaults-are-the-source-of-all-evil/)

straxen/plugins/event_processing.py
```diff
@@ -288,23 +290,29 @@ def compute_loop(self, event, peaks):
...
-            (first_sr1_run, pax_file('XENON1T_FDC_SR1_data_driven_time_dependent_3d_correction_tf_nn_part1_v1.json.gz')),  # noqa
-            (170411_0611, pax_file('XENON1T_FDC_SR1_data_driven_time_dependent_3d_correction_tf_nn_part2_v1.json.gz')),  # noqa
-            (170704_0556, pax_file('XENON1T_FDC_SR1_data_driven_time_dependent_3d_correction_tf_nn_part3_v1.json.gz')),  # noqa
-            (170925_0622, pax_file('XENON1T_FDC_SR1_data_driven_time_dependent_3d_correction_tf_nn_part4_v1.json.gz'))]),  # noqa
+            (first_sr1_run, pax_file('XENON1T_FDC_SR1_data_driven_time_dependent_3d_correction_tf_nn_part1_v1.json.gz')), # noqa
+            (170411_0611, pax_file('XENON1T_FDC_SR1_data_driven_time_dependent_3d_correction_tf_nn_part2_v1.json.gz')), # noqa
+            (170704_0556, pax_file('XENON1T_FDC_SR1_data_driven_time_dependent_3d_correction_tf_nn_part3_v1.json.gz')), # noqa
+            (170925_0622, pax_file('XENON1T_FDC_SR1_data_driven_time_dependent_3d_correction_tf_nn_part4_v1.json.gz'))], # noqa
+    ),
+    *DEFAULT_POSREC_ALGO_OPTION
```
jorana, line 302: Elegant solution 🎉

straxen/plugins/event_processing.py
```diff
@@ -322,38 +330,63 @@ class EventPositions(strax.Plugin):
...
+                get_resource(get_config_from_cmt(self.run_id, map_algo), fmt='binary'))
+
+        elif isinstance(self.config['fdc_map'], str):
+            self.map = InterpolatingMap(
+                get_resource(self.config['fdc_map'], fmt='binary'))
+
+        else:
+            raise NotImplementedError('FDC map format not understood.')
+
+        self.map.scale_coordinates([1., 1., -self.config['electron_drift_velocity']])
```
jorana, line 357: Is scaling always needed? Also for the 1T maps? Otherwise you might add it under the `if is_CMT` block

straxen/plugins/event_processing.py
```diff
@@ -322,38 +330,63 @@ class EventPositions(strax.Plugin):
...
+
+        is_CMT = isinstance(self.config['fdc_map'], tuple)
+
+        if is_CMT:
+            
+            map_algo = list(self.config['fdc_map'])
+            name_map = list(map_algo[1])
+            name_map[0] = '_'.join([name_map[0], self.config['default_reconstruction_algorithm']])
+            map_algo[1] = tuple(name_map)
+            map_algo = tuple(map_algo)
```
jorana, line 345: If you want to have this a little simpler:
```python
cmt, cmt_conf, is_nt = config['fdc_map']
cmt_conf = (f'{cmt_conf[0]}_{self.config["default_reconstruction_algorithm"]]}' , cmt_conf[1])
map_algo = cmt, cmt_conf, is_nt

File: straxen/plugins/position_reconstruction.py [hunk 7]
```diff
@@ -1,204 +1,205 @@
...
-
-    def compute(self, peaks):
-        result = {dtype: peaks[dtype] for dtype in peaks.dtype.names}
-        algorithm = self.config['default_reconstruction_algorithm']
-        if not 'x_' + algorithm in peaks.dtype.names:
-            raise ValueError
-        for xy in ('x', 'y'):
-            result[xy] = peaks[f'{xy}_{algorithm}']
-        return result
+"""Position reconstruction for Xenon-nT"""
```
jorana, line 1: Hmm, looks like git got a bit tricked (it appears as if you replaced the entire file), if we need to track where a specific line was set later, this is a bit more tricky because the git-history gets scrambled, would it be possible to fix this?

straxen/plugins/event_processing.py
```diff
@@ -322,38 +330,63 @@ class EventPositions(strax.Plugin):
...
-                      r_field_distortion_correction=delta_r,
-                      theta=np.arctan2(orig_pos[:, 1], orig_pos[:, 0]))
-
+        # z correction due to longer drift time for distortion
+        # (geometrical reasoning not valid if |delta_r| > |z_obs|,
+        #  as cathetus cannot be longer than hypothenuse)
         with np.errstate(invalid='ignore'):
             z_cor = -(z_obs ** 2 - delta_r ** 2) ** 0.5
-            invalid = np.abs(z_obs) < np.abs(delta_r)        # Why??
+            invalid = np.abs(z_obs) < np.abs(delta_r)
```
jorana, line 379: Thanks for adding the comment!

straxen/contexts.py
```diff
@@ -47,6 +47,8 @@
     s1_max_rise_time=100,
     s2_xy_correction_map=("CMT_model", ('s2_xy_map', "ONLINE"), True),
     elife_file=("elife_model", "ONLINE",True),
+    fdc_map=("CMT_model", ('fdc_map', "ONLINE"), True),
+    electron_drift_velocity=0.632e-4,
```
jorana, line 51: @ahiguera-mx , Would you agree that it's better to track in CMT?

straxen/analyses/posrec_comparson.py
```diff
@@ -0,0 +1,89 @@
...
+    
+    posrec_algos = (posrec_algos, ) if isinstance(posrec_algos, str) else posrec_algos
+
+    if hasattr(cmt_version, '__len__') and not isinstance(cmt_version, str) and len(cmt_version) != len(posrec_algos):
+        raise TypeError(f"cmt_version is a list but does not match the posrec_algos ({posrec_algos}) length.")
+        
+    cmt_version = [cmt_version] * len(posrec_algos) if isinstance(cmt_version, str) else cmt_version
+    
+    dtype = []
+    posrec_dependent_dtype = ['x', 'y', 'z', 'r', 'r_naive',
```
jorana, line 31: This variable is not used?

straxen/analyses/posrec_comparson.py
```diff
@@ -0,0 +1,89 @@
...
+    drift_speed = drift_speed if drift_speed != strax.OMITTED \
+        else context.show_config('event_positions', 'electron_drift_velocity').default.values[0]
+    
+    for algo, v_cmt in zip(posrec_algos, cmt_version):
+        fdc_tmp = ('CMT_model', (f'fdc_map_{algo}', v_cmt), True)
+        map_tmp = straxen.get_config_from_cmt(run_id, fdc_tmp)
+        itp_tmp = straxen.InterpolatingMap(straxen.common.get_resource(map_tmp, fmt='binary'))
+        itp_tmp.scale_coordinates([1., 1., -drift_speed])
+
+        z_obs = - drift_speed * events['drift_time']
```
jorana, line 65: I think you can safely take this one out of the for loop

straxen/analyses/posrec_comparson.py
```diff
@@ -0,0 +1,89 @@
...
+    """
+    Returns the corrected position for each position algorithm available, without the need to reprocess
+    event_basics, as the needed information is already stored in event_posrec_many.
+    
+    :param cmt_version: CMT version to use (it can be a list of same length as posrec_algos, if different versions
+                        are required for different posrec algorithms, default 'ONLINE')
+    :param posrec_algos: list of position reconstruction algorithms to use (default ['mlp', 'gcn', 'cnn'])
+    """
+    
+    posrec_algos = (posrec_algos, ) if isinstance(posrec_algos, str) else posrec_algos
```
jorana, line 23: We use this indeed all the time, there is a function in strax for this
```python
posrec_algos = strax.to_str_tuple(posrec_algos)
```

jorana (changes requested):
Hi Francesco! Awesome work, also really like the mini analysis you added 👍 . I have some technical comments as added below, overall I think this PR is very well prepared and most of the changes would be minor to implement.

straxen/plugins/position_reconstruction.py
jorana: See https://github.com/XENONnT/straxen/pull/396/files#diff-bfd4351fab05e668e9e4f9cf0f7c1318ada3b2c6cf0bd0fd48523511760bb296

straxen/plugins/event_processing.py
ftoschi: This comes from Jelle, indeed :D

straxen/plugins/event_processing.py
ftoschi: Thanks, much better!

straxen/analyses/posrec_comparson.py
ftoschi: I wanted to use it to make the code cleaner when setting results but I ended up not using it. Apparently I forgot it there...

straxen/analyses/posrec_comparson.py
ftoschi: Then for the moment I will leave it like that and as soon as this is implemented in CMT I will change these lines (anyway, also event_positions will have to be changed)

straxen/analyses/posrec_comparson.py
ftoschi: Done 👍

straxen/analyses/posrec_comparson.py
ftoschi: Thanks!

straxen/analyses/posrec_comparson.py
ftoschi: Ok, this sounds like a better solution

straxen/plugins/event_processing.py
ftoschi: Indeed it is not. When I checked if XENON1T data looked fine I only looked at the (x,y) distribution, but I guess the (r², z) would have shown that this is actually something wrong.

straxen/analyses/posrec_comparson.py
ahiguera-mx: @jorana @ftoschi  regarding the `electron_drift_velocity` I thought that the implementation would use `drift_time`, I guess we decided to use `electron_drift_velocity` at the end. In any case we need to coordinate with ACs and the analyst to provide the data to add this into CMT

straxen/contexts.py
ahiguera-mx: yes, see my comment above

straxen/analyses/posrec_comparson.py
ftoschi: The map itself is defined using the drift_time, but then when interpolating the map it is necessary to have identical units in all dimensions and for this reason we should go from drift time to z.

straxen/plugins/position_reconstruction.py
ftoschi: I am not sure what happened here, but fixed it

straxen/analyses/posrec_comparison.py
```diff
@@ -0,0 +1,88 @@
...
+    :param posrec_algos: list of position reconstruction algorithms to use (default ['mlp', 'gcn', 'cnn'])
+    """
+    
+    posrec_algos = strax.to_str_tuple(posrec_algos)
+    
+    if cmt_version is None:
+        cmt_version = context.get_single_plugin(run_id, 'event_positions').config['fdc_map'][1][1]
+    
+    if hasattr(cmt_version, '__len__') and not isinstance(cmt_version, str) and len(cmt_version) != len(posrec_algos):
+        raise TypeError(f"cmt_version is a list but does not match the posrec_algos ({posrec_algos}) length.")
```
jorana, line 29: The first `if` statement might raise a key error when some non-CMT context is used. Not very important though because this entire mini analysis is build around the concept of CMT.

straxen/plugins/event_processing.py
jorana: As far as I can see, you did not change it though 😉

jorana (approved):
Thanks Francesco! Looks good! You checked that nothing changed for 1T right?

Perhaps ACs also want to have a look otherwise it's technically ready for merging.

Two extremely minor comments I still see:

straxen/analyses/posrec_comparison.py
ftoschi: As you say, the idea is that this should be used only in a CMT context, as for the moment this means XENONnT context. In case this will be changed in the future, this minianalysis will be updated

straxen/plugins/event_processing.py
ftoschi: I must be getting senile, I was really sure I implemented it 😅

jorana (approved):
Thanks again Francesco, I've added a small test just to make sure

straxen/contexts.py
ahiguera-mx: @ftoschi since you have one map for every single reco alg, I guess the option should be something like 
`fdc_map=("CMT_model", ('fdc_map', "ONLINE"), True)`

ahiguera-mx (commented):
I have one small comment regarding the `fdc_map` option should it be
`fdc_map=("CMT_model", ('fdc_map_mlp', "ONLINE"), True)`

ahiguera-mx:
Hi @ftoschi Thanks for the update, you already provided the three maps for the different algs, so no change needed to the actual files, right? just checking what we have in CMT is up to date

ftoschi:
> Hi @ftoschi Thanks for the update, you already provided the three maps for the different algs, so no change needed to the actual files, right? just checking what we have in CMT is up to date

Exactly, the maps are fine

straxen/analyses/posrec_comparison.py
```diff
@@ -0,0 +1,97 @@
...
+                ((f'Correction added to r_naive for field distortion (cm) - {algo.upper()} posrec algorithm',
+                  f'r_field_distortion_correction_{algo}'), np.float32),
+                ((f'Interaction angular position (radians) - {algo.upper()} posrec algorithm', f'theta_{algo}'),
+                 np.float32)]
+        
+    dtype += [(('Interaction z-position using mean drift velocity only (cm)', 'z_naive'), np.float32)]
+    result = np.zeros(len(events), dtype=dtype)
+
+    drift_speed = context.get_single_plugin(run_id, 'event_positions').config['electron_drift_velocity']
+    z_obs = - drift_speed * events['drift_time']
```
l-althueser, line 66: We need to add the gate drift time in order to compute the correct z-positions. Either in this or another PR. See PAX example: https://github.com/XENON1T/pax/blob/c6c813800c03e834265f6f8c33360310248cdaf4/pax/plugins/interaction_processing/BuildInteractions.py#L46

jingqiangye:
Hey everyone, I think we should implement the FDC into straxen soon, so that S1 LCE can be implemented and we will have all corrections in place then. As far I as understand, we just need to take the gate drift time into account and then we can merge this PR?

ftoschi:
@jorana @ahiguera-mx @jingqiangye @l-althueser the maps use drift time, so we need to include the drift speed somehow. For the moment, if not specified, the drift speed is still the one from XENON1T: for this reason I think the best would be to implement the drift time from CMT (unless this has been done already and I missed it). In the meantime, people can simply specify the drift speed within their own context. The correction for the gate time can really come in a second moment, anyway this does not affect how the FDC works as the drift times used in the FDC include this time.

PR #399, opened 2021-03-15
Use code style commentator

jorana:
Instead of relying on reviewers to comment on style (which is always tedious) we could ask a bot to do this.

Github actions 

See for example a much [simpler commit ](https://github.com/jorana/action_tests/pull/2), or that it [works with straxen](https://github.com/jorana/straxen/pull/2)


WenzDaniel:
Would this check be optional? I see already myself speeding an infinite amount of time on this, but in principle I like the idea. Which standard would we use?

jorana:
@WenzDaniel , thanks, for sure, this is an entirely optional check. The comments are posted by the bot but not stopping the PR to get the flag "failed". Also some of the pep8 issues are a bit crazy, we should filter out those over time.

ershockley (approved):
Very nice @jorana!

PR #407, opened 2021-03-22
test python 3.9

jorana:
With [numba 0.53](https://numba.readthedocs.io/en/stable/release-notes.html#version-0-53-0-11-march-2021) we can support python3.9 🎉 !


jorana:
:tada: too soon. I always thought that numba was the limiting factor, but it's tensorflow.

jorana:
TensorFlow 2.5 actually should be supported for py3.9. Let's try again!

WenzDaniel (approved):
I do not see any reason why we should not add it to the tests. Just to be sure we are on the same page, now we are testing things without tensorflow, are not we?

jorana:
@WenzDaniel , ah no for sure not. We are testing **with** tensorflow. I just removed it from the list of pinned versions which we grep from https://github.com/XENONnT/base_environment/blob/master/requirements.txt since tensorflow 2.5 is needed for python 3.9. However, that is not the version provided in the base environment. Therefore, we just let pip resolve a version of tensorflow we can use.

jorana:
> 
> 
> I do not see any reason why we should not add it to the tests. Just to be sure we are on the same page, now we are testing things without tensorflow, are not we? Where are we now loading tensorflow 2.5?

@WenzDaniel , this is now done in the setup.py test:
https://github.com/XENONnT/straxen/pull/407/checks?check_run_id=2455414942#step:8:92

PR #410, opened 2021-03-23
upload compressor from metadata to rundoc

jorana:
## Upload compressor from metadata to rundoc
When uploading several details about the compressor used for a given entry in the rundoc, also include the compressor


darrylmasson (approved):
At some point it may be worth doing a review of everything we're storing in the `data` field and making sure it makes sense.

jorana:
Not a bad idea, on a personal note I have to say that I use everything stored there at least semi-regularly.

PR #413, opened 2021-03-26
Check for overlapping raw_records in nT sims

ramirezdiego:

I could be mistaken, but I though the `check_raw_record_overlaps=False` in the default nT sims context was there to prevent some bug now resolved. Could someone confirm?


petergaemers:
check_overlap=True should work with wfsim

ramirezdiego:
Good, and do I understand correctly that we then want `True` as default?

WenzDaniel (approved):
Yes, we should have this check in place. Nothing to comment, thanks.

PR #404, opened 2021-03-16
Add get_correction_from_cmt to corrections_services.py

ahiguera-mx:
Before you submit this PR: make sure to put all operations-related information in a wiki-note, a PR should be about code and is publicly accessible

**What is the problem / what does the code in this PR do**
Add get_correction_from_cmt, this function would return any correction value, it replace get_elife and serves as a general purpose function for single values
**Can you briefly describe how it works?**
See https://github.com/XENONnT/corrections/blob/master/CMT/examples/CMT_user_guide.ipynb block [5]
`elife_file = ('elife','ONLINE',True)`
`straxen.get_correction_from_cmt('010126', elife_file)`
**Can you give a minimal working example (or illustrate with a figure)?**
see above
Please include the following if applicable:
  - Update the docstring(s)
  - Update the documentation
  - Tests to check the (new) code is working as desired.
  - Does it solve one of the open issues on github?

Please make sure that all automated tests have passed before asking for a review (you can save the PR as a draft otherwise).

PR #417, opened 2021-04-02
Make release

WenzDaniel:
Make new release.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,6 +1,6 @@
 0.16.0 / 2021-04-02
 --------------------
-- restore #404, add get_correction_from_cmt to corrections_services.py (#409)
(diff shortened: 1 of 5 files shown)


HISTORY.md
```diff
@@ -1,3 +1,16 @@
+0.16.0 / 2021-04-02
+--------------------
+- restore #404, add get_correction_from_cmt to corrections_services.py (#409)
```
jorana, line 3: I think you can better use the title of #404 and use both #404 and #409 as reference

jorana (approved):

HISTORY.md
WenzDaniel: Okay, I change it

PR #424, opened 2021-04-07
Abbreviated event display restructure #418

jorana:
#418 is a nice PR that expands the usability of the static event display to single figures. With this PR, we slightly restructure the additions such that the setup of the axes etc. is done in another function then where the plotting takes place, this way, we can more easily support the simple event display as nicely drafted by @kdund .

```python
st.event_display_simple
st.event_display
```
both refer to the easily customizable `_event_display`


kdund (commented):
nice putting it in as a default!

PR #430, opened 2021-04-09
Implement nT S1 correction 

jingqiangye:
Use the nT S1 relative xyz map (generated by @jjakob03 and @l-althueser ) to do S1 correction right. The S1 correction is done for the default PosRec algo, but can also specify other PosRec algos. (Thanks @ahiguera-mx for uploading s1 maps to CMT DB).

Simply load `corrected_areas` (or other plugins dependent on that like `event_info`) with the right electron drift velocity, i.e. `df = st.get_df(run_id, "event_info", config=dict(electron_drift_velocity=0.632e-4))`.

I tested it with a run and the performance is as expected.


jorana:
@jingqiangye , in 2b58fb4 I fixed the tests, is the #Todo which is failing code factor now something you want to add to this PR still?

jorana (approved):

jingqiangye:
No the TODO is not necessary, it's only a reminder for S2 map correction. I just removed it.

jorana:
Thanks @jingqiangye !

PR #431, opened 2021-04-10
update apply_function_to_data

jorana:
After https://github.com/AxFoundation/strax/pull/422, we are slightly differently applying the apply function to data. This PR reflects that change.

Most notably, we changed the signature of the function, the function is now required to take the run_id as a positional argument

### DIFF
--- a/straxen/common.py
+++ b/straxen/common.py
@@ -467,13 +467,15 @@ def remap_channels(data, verbose=True, safe_copy=False, _tqdm=False, ):
     return _dat
 
 
-def remap_old(data, targets, works_on_target=''):
+@export
+def remap_old(data, targets, run_id, works_on_target=''):
     """
     If the data is of before the time sectors were re-cabled, apply a software remap
         otherwise just return the data is it is.
     :param data: numpy array of data with at least the field time. It is assumed the data
         is sorted by time
     :param targets: targets in the st.get_array to get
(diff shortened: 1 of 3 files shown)


straxen/common.py
```diff
@@ -488,11 +490,11 @@ def remap_old(data, targets, works_on_target=''):
...
         # select the old data and do the remapping for this
-        warn("Correcting data of runs with mis-cabled PMTs. \nSee: https://"
-             "xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:dsg:daq:sector_swap. "
-             "Don't use '' selection_str='channel == xx' '' (github.com/XENONnT/straxen/issues/239)")
         mask = data['time'] < TSTART_FIRST_CORRECTLY_CABLED_RUN
-        data[mask] = remap_channels(data[mask])
+        res = data.copy()
+        res[mask] = remap_channels(data[mask])
+        data = res.copy()
+        del res
```
WenzDaniel, line 497: Could you quickly explain why you have to copy the data twice? I know you would not have added it if it would not be needed, but still it looks odd to me. Other than that it looks fine.

WenzDaniel (approved):
Just an explanation please.

straxen/common.py
jorana: Actually it could do with once, good catch, it's needed at least once since it is read-only otherwise.

PR #432, opened 2021-04-10
Raise notimplemented error for peak_min_pmts > 2

jorana:
`raise warn` is not really proper syntax, replace with a NotImplementedError


feigaodm (approved):
My understanding is that this will make lone-hits definition always correct.

jorana:
Exactly, until someone takes the effort to re-split the lone-hits to individual PMTs and removes this check.

PR #433, opened 2021-04-10
Set max-run number in rundb

jorana:
With more and more runs, one might not always need to look at high run-numbers to check for data availabilty. Allow an option to don't look for runs beyond a given range.

Add an extra query

```python
import straxen
st = straxen.contexts.xenonnt_online(_minimum_run_number=15000, 
                                     _maximum_run_number=15010)
st.select_runs()
```


WenzDaniel:
Allow me maybe a naive question. Is there any benefit from this, how is the data in our rundb sorted? If it is not by run_number there should not be any difference is not it?

jorana:
>Allow me maybe a naive question. Is there any benefit from this, how is the data in our rundb sorted? If it is not by run_number there should not be any difference is not it?

I might not understand your question but this is the goal of my PR:
 - If you do `st.select_runs()` , the context does query for all the runs, if you have less runs known to the context, this becomes faster.

PR #434, opened 2021-04-10
Update documentation, add 1T, fix #31, compact conf

### LINKED ISSUE #31: Autogenerated docs failing for multi-output plugin config options
For multi-output plugins, currently only the configuration options of their parent plugins are displayed in the straxen docs. This should be an issue with [build_datastructure_doc.py](https://github.com/XENONnT/straxen/blob/master/docs/source/build_datastructure_doc.py).

jorana:
Update documentaion

### Fix #31
E.g. peaklets heaver their own configs instead of their parents:
afbeelding

### Simplify config options
The config list was getting very long, let's limit the list to only the relevant options:
afbeelding

### Add 1T datastructure
This was missing from the docs
afbeelding


WenzDaniel (approved):
I cannot spot any obvious mistakes.

PR #435, opened 2021-04-11
use strax.apply_selection in om

jorana:
Remove duplicated code and use `strax.apply_selection` for online monitoring purposes.

**Requires**
https://github.com/AxFoundation/strax/pull/425


WenzDaniel (approved):
Not much to say, looks fine.

jingqiangye (approved):
Thanks for the change. It is always good to consolidate ;) Looks good to me

PR #436, opened 2021-04-12
Clustering and classification

jorana:
Requires, https://github.com/AxFoundation/strax/pull/403.

For motivation, please refer to the notes linked here:
https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:analysis:waveforms_team:20210408

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 strax>=0.14.0
-utilix==0.4.1
+utilix>=0.5.2
 # tensorflow>=2.3.0  # Optional, to (re)do posrec
 # holoviews          # Optional, to enable wf display
 # datashader         # Optional, to enable wf display
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 strax>=0.14.0
-utilix==0.4.1
+utilix>=0.4.1
 # tensorflow>=2.3.0  # Optional, to (re)do posrec
 # holoviews          # Optional, to enable wf display
 # datashader         # Optional, to enable wf display
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 strax>=0.14.0
-utilix>=0.4.1
+utilix==0.4.1
 # tensorflow>=2.3.0  # Optional, to (re)do posrec
 # holoviews          # Optional, to enable wf display
 # datashader         # Optional, to enable wf display
--- a/straxen/contexts.py
+++ b/straxen/contexts.py
@@ -249,26 +249,34 @@ x1t_common_config = dict(
         tpc=(0, 247),
         diagnostic=(248, 253),
         aqmon=(254, 999)),
+    # Records
     hev_gain_model=('to_pe_per_run',
                     'to_pe.npy'),
-    gain_model=('to_pe_per_run',
-                'to_pe.npy'),
     pmt_pulse_filter=(
         0.012, -0.119,
         2.435, -1.271, 0.357, -0.174, -0., -0.036,
         -0.028, -0.019, -0.025, -0.013, -0.03, -0.039,
         -0.005, -0.019, -0.012, -0.015, -0.029, 0.024,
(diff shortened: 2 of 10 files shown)


requirements.txt
```diff
@@ -1,5 +1,5 @@
 strax>=0.14.0
-utilix>=0.4.1
+utilix==0.4.1
```
jorana, line 2: Temporary until https://github.com/XENONnT/utilix/pull/41 is fixed

straxen/contexts.py
```diff
@@ -255,26 +255,34 @@ def xenonnt_initial_commissioning(*args, **kwargs):
...
         ((0.5, 1), (3.5, 0.25)),
         ((2, 1), (4.5, 0.4))),
+    peak_min_pmts=2,
+    # MergedS2s
+    s2_merge_max_duration=5000.,
+    s2_merge_max_gap=3500,
+    # Peaks
+    # Smaller right extension since we applied the filter
+    peak_right_extension=30,
+    # Events*
```
jorana, line 285: While adding these 1T configs, I noticed that they were very disorganized. Therefore I slightly re-ordered the configs.

skazama:
Hi @jorana, thanks for the PR. I just checked the code and have a question:
It seems that you have updated value for s2_merge_max_area to 35_000, but this should be for s2_merge_max_duration, right?
Maybe we also need to check which value we should use for s2_merge_max_area, but at least 35_000 is for s2_merge_max_duration if I understand correctly.

jorana:
hi @skazama , thanks for the careful check, you are absolutely right. Too many very similar options 🤔 . 

I had checked `s2_merge_max_area` as well in my studies, but did not find evidence of a need to increase this number.

skazama (approved):
Thanks @jorana for the update. From my side, It looks fine now, but maybe let's wait @zhut19 and @jhowl01's responses.

straxen/plugins/peaklet_processing.py
```diff
@@ -26,8 +26,8 @@
                  # for more information
                  default=(
                      None,  # Reserved
-                     ((0.5, 1), (4, 0.4)),
-                     ((2, 1), (4.5, 0.4))),
+                     ((0.5, 1.), (6.0, 0.4)),
```
zhut19, line 29: This is a typo, right?

zhut19: Sorry, I missed that plot.

skazama: @zhut19 do you think this PR can be merged now?

zhut19 (approved):

jhowl01 (approved):
Looks great! Excellent work @jorana !

My only thing is it seems like we're making these changes to the defaults and then overwriting them for outdated contexts. Wouldn't it be better to explicitly set at least the clustering parameters that are likely to change several times inside `contexts.py`? Maybe in `xnt_common_config`? This is probably a matter of taste but I remember ages of checking the git blame on XENON1T.ini in pax, and in that case it was pretty useful to find all the changes in one place.

Approving though because again, I think this is just one guy's opinion

jorana:
> 
> 
> Looks great! Excellent work @jorana !
> 
> My only thing is it seems like we're making these changes to the defaults and then overwriting them for outdated contexts. Wouldn't it be better to explicitly set at least the clustering parameters that are likely to change several times inside `contexts.py`? Maybe in `xnt_common_config`? This is probably a matter of taste but I remember ages of checking the git blame on XENON1T.ini in pax, and in that case it was pretty useful to find all the changes in one place.
> 
> Approving though because again, I think this is just one guy's opinion

Hi Joey, thanks for asking! The convention we have in straxen is to update the default of the plugins-option if it relates to xenon-nt to increase the readability of the plugins. There are several examples where we don't really follow this convention due to https://github.com/AxFoundation/strax/issues/400. Then for xenon1t (which at this point shouldn't change much), we set the values in the config of the context.

PR #440, opened 2021-04-17
allow faster NV/MV by bootstrax

jorana:
With three hosts processing and three systems (TPC, MV and NV) running in parallel, we sometimes have a bit of a gap in the logic where the old ebs are doing nothing but the new ebs are too busy doing e.g. failed runs to do the new runs of the three detectors.

To this end, let's allow the older ebs to take on MV/NV data which is much less intense then TPC and therefore perfectly handable by the old ebs.

If an eb is not considered allowed to process data since there is a short queue, allow it to check if another MV/NV run is available.

### DIFF
--- a/bin/bootstrax
+++ b/bin/bootstrax
@@ -73,7 +73,7 @@ what bootstrax is thinking of at the moment.
   - **disk_used**: used part of the disk whereto this bootstrax instance
    is writing to (in percent).
 """
-__version__ = '1.1.3'
+__version__ = '1.1.4'
 
 import argparse
 from datetime import datetime, timedelta, timezone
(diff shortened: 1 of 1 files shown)


bin/bootstrax
```diff
@@ -445,6 +445,17 @@ def main_loop():
                 new_runs_seen += 1
                 process_run(rd)
                 continue
+        else:
+            # We are on an old eb with not so much to do, perhaps one of
+            # the veto systems needs processing?
+            rd = consider_run(
+                {'detectors': {'$in': ["muon_veto", "neutron_veto"]},
```
darrylmasson, line 452: This will match linked-mode runs, which maybe we don't want. To avoid this the query should look more like `{'detectors': {'$ne': 'tpc'}}`, but `$ne` isn't a particularly efficient query.

darrylmasson (approved):

bin/bootstrax
jorana: @darrylmasson , thanks, good catch. 

On the bootstrax side, inefficiency is not a problem as this is on the old ebs anyway. Problem is of course that this operator is executed on the mongo side of the story.

darrylmasson: From the docs:
> The inequality operator $ne is not very selective since it often matches a large portion of the index. As a result, in many cases, a $ne query with an index may perform no better than a $ne query that must scan all documents in a collection

PR #441, opened 2021-04-18
Change records default processor

jorana:
The (raw-)records processor is set at the daq. However, when re-making records, let's make sure to make it in a compressed format where ever possible.

We do need to keep in mind though that this might (potentially) lead to problems if the buffers are too big to fit into `zstd`. 

This is especially relevant for (re)processing records.


ershockley (approved):
Thanks @jorana! I didn't realize we weren't using the optimal compressor.

jorana:
@ershockley , we were/are on the DAQ, but not as a default elsewhere

PR #444, opened 2021-04-21
Nveto event display

WenzDaniel:
In this PR we add a dynamic nveto event display based on holoviews. In addition to the event display itself I also added some unity tests which check if the plots can be rendered.  

The plot is based on holoviews and the panel library. The plotting itself is defined in a class called nVETOEventDisplay to accommodate the python callbacks which require the definition of "global"-variables. 

I uploaded an [example note book](https://github.com/XENONnT/analysiscode/blob/master/nveto/nVETOEventDisplay.ipynb).


straxen/analyses/holoviews_waveform_display.py
```diff
@@ -187,26 +195,63 @@ def _hvdisp_plot_records_2d(records,
...
 
 
-def rectangle(time=0, channel=0, width=1.1, height=1):
+def _make_polys(df, scaling=1):
+    """
+    Function which converts the hitlet/record points into polygons
+    for the hitlet/record matrix.
+    """
+    res = []
+    for row_i in range(len(df)):
```
jmosbacher, line 237: Seems a bit strange to iterate over rows, not crucial i guess since number of rows should be small but why not have pandas do the iteration for you?

straxen/analyses/holoviews_waveform_display.py
```diff
@@ -187,26 +195,63 @@ def _hvdisp_plot_records_2d(records,
...
-            record_points.data[['time', 'channel', 'area']].values]
-    polys = hv.Polygons(data, vdims='area')
-    polys = polys.opts(color='level', aspect=4, responsive='width', line_width=0, **kwargs, cmap='viridis')
+    df = record_points.data.copy()
+    if not center_time:
+        df.loc[:, 'time'] = df.loc[:, 'time'] + (df.loc[:, 'length'] / 2 * df.loc[:, 'dt'])
+
+    data = _make_polys(df,
+                       scaling=scaling)
+    polys = hv.Polygons(data,
```
jmosbacher, line 219: Not very important but im a bit curious about why you chose to use hv.Polygons and not hv.Segments here? isnt the y axis just the channel? so wouldnt all of the polygons be the same height?

jmosbacher (approved):
Very nice viewer. At some point in the future i think it would be a good idea if plotting tools are better organized though, its becoming quite difficult to follow the function calls as they go through various files. Will be pretty difficult to maintain. But it works and looks good, that is more important for this kind of thing. One technical point regarding the hitlet matrix plots, I think it would be useful to convert the channel to a string before you plot it so that its handled correctly as a categorical index, just if you do dont forget to sort the values before you convert them to strings (sorting the strings will have a different order than their numerical order).

WenzDaniel:
> Very nice viewer. At some point in the future i think it would be a good idea if plotting tools are better organized though, its becoming quite difficult to follow the function calls as they go through various files.

I totally agree. I am think about making it an optional package for straxen, like straxen[plotting] as it is done in other projects. Thanks to your feedback I switched now completly to holoviews and as a next step I will convert the TPC event display from bokeh to holoviews. During this refactoring I will try to unify the tools as much as possible so one can make a joined display for all three detectors, 
The final goal will be to join the plotting library with Jorans online montior tool such that we can get an interactive one. 

> One technical point regarding the hitlet matrix plots, I think it would be useful to convert the channel to a string before you plot it so that its handled correctly as a categorical index, just if you do dont forget to sort the values before you convert them to strings (sorting the strings will have a different order than their numerical order).

Do you mean when using `hv.Segments `?

jmosbacher:
@WenzDaniel whether you use the segments api or polygons api is not very important, I just think that the Y axis (the channel number) would look a bit better if it was a categorical axis. This is done automatically by holoviews for string types so if you just convert the channel number to a string the yaxis will become categorical. This would mean the all plotted segment would be an equal distance from its neighbor instead of it currently being a bit weird since the channel numbers for nveto are relatively large numbers.

straxen/analyses/holoviews_waveform_display.py
WenzDaniel: Nice suggestion, it works like a charm and makes things much easier. Thanks a lot. FYI: the TPC display still uses the old method via hv.Polygons, I will change it when I refactor the TPC event display from bokeh to holoviews.

straxen/analyses/holoviews_waveform_display.py
WenzDaniel: Okay this function will become obsolete once I refactored things.

PR #445, opened 2021-04-21
Require data to be transferred to dali to load

ershockley:
This fixes a small issue in the RunDB frontend when loading data via rucio. We previously did not require the data to be `transferred`, which made it seem like data was available when it wasn't.  This also drops the `protocol=rucio` from the query since this doesn't seem to do anything.


jorana (approved):
Nice catch, indeed `protocol=rucio` was obsolete given that we already had these lines in the query:
https://github.com/XENONnT/straxen/blob/b51e62605251a629edba8195ed6d2f173b1f5314/straxen/rundb.py#L123-L124

PR #447, opened 2021-04-22
Mveto events

WenzDaniel:
In this PR we add muon-veto events as a child of the neutron-veto events. Like we did with the other plugins before.


straxen/plugins/veto_events.py
```diff
@@ -427,3 +427,51 @@ def first_hitlets(hitlets_per_event: np.ndarray,
...
+                 help="Extends events this many ns to the left"),
+    strax.Option('event_resolving_time_mv', default=300,
+                 child_option=True, parent_option_name='event_resolving_time_nv',
+                 help="Resolving time for fixed window coincidence [ns]."),
+    strax.Option('event_min_hits_mv', default=3,
+                 child_option=True, parent_option_name='event_min_hits_nv',
+                 help="Minimum number of fully confined hitlets to define an event."),
+    strax.Option('gain_model_mv',
+                 child_option=True, parent_option_name='gain_model_nv',
+                 help='PMT gain model. Specify as (model_type, model_config)'),
```
jorana, line 444: Can we add a default, now I am not sure what happens?

straxen/plugins/veto_events.py
```diff
@@ -427,3 +427,51 @@ def first_hitlets(hitlets_per_event: np.ndarray,
...
+        self.channel_range = self.config['channel_map']['mv']
+        self.n_channel = (self.channel_range[1] - self.channel_range[0]) + 1
+        return veto_event_dtype(self.name_event_number, self.n_channel)
+
+    def setup(self):
+        self.to_pe = straxen.get_to_pe(self.run_id,
+                                       self.config['gain_model_mv'],
+                                       self.n_channel)
+
+    def get_window_size(self):
```
jorana, line 473: _nv, is that right? if so, can we add a comment as to why?

jorana (commented):
Thanks Daniel, looks clean

straxen/plugins/veto_events.py
WenzDaniel: No this was a mistake, good catch.

straxen/plugins/veto_events.py
WenzDaniel: You will get as in the normal cases the gain settings as specified in the context: 
```
{'events_mv': ('muVETOEvents',
  '0.0.1',
  {'gain_model_mv': ('to_pe_constant', 'adc_mv'),
   'event_left_extension_mv': 0,
   'event_resolving_time_mv': 300,
   'event_min_hits_mv': 3,
   'nVETOEvents': '0.0.2'})
```
But I can also add a default if you like.

jorana: yeah, would be nice, as far as Jelle and I agreed, this was the convention:
https://github.com/XENONnT/straxen/pull/436#issuecomment-820949904

molinario (approved):

straxen/plugins/veto_events.py
WenzDaniel: Okay I was not aware of this, but it is changed now. Also for the other plugins.

jorana: ok, thanks 😄 , I guess that means we can remove it from contexts.py?

jorana (approved):

Issue #450, opened 2021-04-22
st.select_runs doesn't work for Rucio backend


ershockley:
**Describe the bug**
After the recent straxen update I am reprocessing some data on OSG. I have found that checking the data availability does not work as expected. 

The issue is that there is data that I can load using `st.get_array()` but does not appear as available with `st.select_runs()`.  

**To Reproduce**
(assuming this is not a transient issue, here is a run I've been testing)
```
st = straxen.contexts.xenonnt_online()
runinfo = st.select_runs(available='peak_basics')
available_runs = runinfo.number.values.astype(int)
runid = 16241

runid in available_runs
```
This returns `False`, but I can still load the data if I specify directly. 

```
peaks = st.get_array(f'{runid:06d}', 'peak_basics')
Loading peak_basics: |          | 0.00 % [00:00<?]
```

**Expected behavior**
The `select_runs` method should return all data that can be loaded using e.g. `st.get_array`. 

**Versions**
Please add the output of:
```
python	v3.8.8	(default, Apr 13 2021, 19:58:26) [GCC 7.3.0]
strax	v0.15.0	/opt/XENONnT/anaconda/envs/XENONnT_development/lib/python3.8/site-packages/strax
straxen	v0.18.0	/opt/XENONnT/anaconda/envs/XENONnT_development/lib/python3.8/site-packages/straxen

```

jorana:
Nice catch, I can replicate. Let me write a fix, the find_several query does not understand multiple lineages. So far, we have not dealt with them.

```python
>>>rb = st.storage[0]
>>>key = st.key_for(run_id:='016241', 'peak_basics')
>>>print(rb.find_several([key], fuzzy_for=False, fuzzy_for_options=False))
>>>print(rb.find(key))

[False]
('rucio', '016241-peak_basics-elmqm5dlwt')
```

jorana:
fixed in #451

PR #451, opened 2021-04-22
use did for finding several runs

jorana:
#450 

`RunDB.select_runs` uses a different query than `RunDB._find`. The `RunDB._find` is a bit "sloppy" and only uses the lineage-hash (the `did` field) to verify if we have data. The `RunDB.select_runs` uses the entire lineage ancestry. 

Now the issue in #450 showed up because of a slight misconception in outsource. Nevertheless, after discussing with @ershockley , this entire field might be a bit overkill to put in the rundocs ddocs. 

With this PR, we allow the rucio did to be used.

### DIFF
--- a/straxen/rundb.py
+++ b/straxen/rundb.py
@@ -129,7 +129,11 @@ class RunDB(strax.StorageFrontend):
             'data': {
                 '$elemMatch': {
                     'type': key.data_type,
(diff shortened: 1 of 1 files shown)


ershockley (approved):
Thanks @jorana! I just checked and confirmed this works. 

Do you think we could make a new release with this relatively soon?

PR #452, opened 2021-04-23
mveto config passing

zhut19:

The context configurations that end with '_mv' don't seem to be used in the compute function of the mu veto plugins.


zhut19:
No, this issue only exists in my head. See https://github.com/AxFoundation/strax/pull/297

PR #455, opened 2021-04-27
fix keyerror for uploading data in selectruns

jorana:
KeyError for protocol due to a failing query where non-transferred runs showed up.

```python
st.select_runs(available='peaklets')
```

Rather than only using the `transferred` rule only in `find`, we also need it in `find_several`

_this PR also does a very slight refactor_

_the failing codefactor is something that I will address in #456_

### DIFF
--- a/straxen/rundb.py
+++ b/straxen/rundb.py
@@ -6,6 +6,7 @@ from tqdm import tqdm
 from copy import deepcopy
 import strax
 from pprint import pprint
+
 try:
     import utilix
 except (RuntimeError, FileNotFoundError):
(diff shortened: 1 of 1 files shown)


straxen/rundb.py
```diff
@@ -122,7 +124,9 @@ def __init__(self,
             self.backends.append(strax.rucio(self.rucio_path))
             # When querying for rucio, add that it should be dali-userdisk
             self.available_query.append({'host': 'rucio-catalogue',
-                                         'location': 'UC_DALI_USERDISK'})
+                                         'location': 'UC_DALI_USERDISK',
+                                         'status': 'transferred'
+                                         })
```
jorana, line 129: This is the key fix

l-althueser (approved):
I did not test myself but from what I see on the RunDB comparing the problematic and normal runs, this should fix it in a nice way :)

PR #457, opened 2021-04-28
Standardize CMT options as (correction, version, nT=boolean)

ahiguera-mx:

Standardize CMT options as (correction, version, nT=boolean)

Standardize CMT options as (correction, version, nT=boolean)

Please see https://github.com/XENONnT/corrections/blob/master/CMT/examples/CMT_user_guide.ipynb
This include working examples for all corrections in CMT with the standardization 
Note: Special case if WFsim an examples has been included as well

  - Update the docstring(s)
  - Update the documentation


WenzDaniel (commented):
Hi Aaron thanks a lot. I think one essential thing we should add are tests. CMT has become a very important tool which is not tested much to my knowledge (at least I was not able to find anything beside some tests for the function wrapper). Some unity tests similar to the examples in your notebook would be already enough.

WenzDaniel:
I just noticed it, with your proposed changes we get an error for the peaklet processing of XENON1Td data. In the python 3.8 test you can see that it seems to be due to the splitting function. Can you check if the 1T gains work as expected?

straxen/get_corrections.py
```diff
@@ -37,72 +37,56 @@ def correction_options_wrapped(run_id, conf, *arg):
...
-
-        return to_pe
-
-    elif model_type == 'to_pe_per_run':
-        warn("to_pe_per_run will be replaced by CorrectionsManagementSevices",
-             DeprecationWarning, 2)
-        # Load a npy file specifying a run_id -> to_pe array
-        to_pe_file = model_conf
-        x = straxen.get_resource(to_pe_file, fmt='npy')
+def get_to_pe(run_id, conf, n_pmts):
```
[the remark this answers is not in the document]
ahiguera-mx: It is used to return the `to_pe` values for TPC n-veto and m-veto PMTs respectively, not sure where is the confusion

straxen/get_corrections.py
```diff
@@ -153,25 +136,20 @@ def get_config_from_cmt(run_id, conf):
...
-        raise ValueError("conf must have three elements: "
-                         "the model type, its specific configuration "
-                         "and detector (True = nT)")
-    model_type, model_conf, is_nt = conf
-    if model_type == 'CMT_model':
+    elif isinstance(conf, tuple) and len(conf) == 3:
+        is_nt = conf[-1]
         cmt = straxen.CorrectionsManagementServices(is_nt=is_nt)
-        this_file = cmt.get_corrections_config(run_id, model_conf)
+        this_file = cmt.get_corrections_config(run_id, conf[:2])
```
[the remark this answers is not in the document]
ahiguera-mx: Great catch @WenzDaniel! indeed with the standardization we don't need this two function, that is the whole point. I've fixed it

straxen/get_corrections.py
```diff
@@ -153,25 +136,20 @@ def get_config_from_cmt(run_id, conf):
...
-                         "and modify it accordingly")
-
-    return this_file
+        raise ValueError("Wrong configuration. "
+                         "Please use the following format: "
+                         "(model_type->str, model_config->str or number, is_nT->bool) "
+                         f"User specify {conf} please modify")
 
 
+# This should be removed!!!!!!!
```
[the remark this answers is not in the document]
ahiguera-mx: Done

ahiguera-mx:
> I just noticed it, with your proposed changes we get an error for the peaklet processing of XENON1Td data. In the python 3.8 test you can see that it seems to be due to the splitting function. Can you check if the 1T gains work as expected?

Hi Daniel, 
It seems that due to the new changes in clustering dummy values won't work... which is shame, that force me to paste from 1T values as placeholders. Ideally the new changes in clustering should be robust enough that would work with dummy values, but maybe we are tuning too much that is not the case

ahiguera-mx:
> Hi Aaron thanks a lot. I think one essential thing we should add are tests. CMT has become a very important tool which is not tested much to my knowledge (at least I was not able to find anything beside some tests for the function wrapper). Some unity tests similar to the examples in your notebook would be already enough.

Hi Danie,
Not sure what kind of test your a thinking,  the examples that call CMT in the notebook requires access to the correctionsDB, therefore access to usernames and passwords, so I'm not sure how we can avoid that, that is the whole point of re-settings all configs in `test_plugins` so the automated test can pass without having to access to correctionsDB. Maybe you have another idea?

WenzDaniel:
You can add those credentials to github somehow. This was one of the reasons why we switched to github actions. See for example here: https://github.com/XENONnT/straxen/blob/master/.github/scripts/create-utilix-config.sh and https://github.com/XENONnT/straxen/blob/master/tests/test_mongo_interactions.py. I do not know exactly where you have to store your secret credential information as I am not an admin. But @jorana can help with that.

jorana:
@WenzDaniel , concerning the mongo access I think Aaron did get the gist of it right. Since he is pushing from his own fork, he won't have access to the secrets in this PR.

WenzDaniel:
So in that case it might be better to push first two a new branch of this repository. After that you can check if your tests are working. For the very same reason I am also currently moving from my fork to the upstream repo.

straxen/get_corrections.py
```diff
@@ -37,72 +38,55 @@ def correction_options_wrapped(run_id, conf, *arg):
...
-
-        return to_pe
-
-    elif model_type == 'to_pe_per_run':
-        warn("to_pe_per_run will be replaced by CorrectionsManagementSevices",
-             DeprecationWarning, 2)
-        # Load a npy file specifying a run_id -> to_pe array
-        to_pe_file = model_conf
-        x = straxen.get_resource(to_pe_file, fmt='npy')
+def get_to_pe(run_id, conf, n_pmts):
```
jorana, line 41: Making a docstring here would be super helpful we did not have one in the past, which is a shame.

straxen/get_corrections.py
```diff
@@ -37,72 +38,55 @@ def correction_options_wrapped(run_id, conf, *arg):
...
+
+    if isinstance(conf, str) and conf.startswith('https://raw'):
+        # Legacy support for pax files
+        x = straxen.get_resource(elife_conf, fmt='npy')
```
jorana, line 45: Something went wrong copy pasting things here, `elife_conf` is not defined.

straxen/get_corrections.py
```diff
@@ -37,72 +38,55 @@ def correction_options_wrapped(run_id, conf, *arg):
...
+                         f"User specify {conf} please modify")
 
 
+    elif isinstance(conf, tuple) and len(conf) == 2:
+        model_conf = conf[0]
+        if model_conf in FIXED_TO_PE:
+            to_pe = FIXED_TO_PE[model_conf]
+            if len(to_pe) != n_pmts:
+                raise ValueError(
+                    f"Gain model {gain_model} resulted in a to_pe "
```
jorana, line 75: `gain_model` is not defined

straxen/get_corrections.py
```diff
@@ -37,72 +38,55 @@ def correction_options_wrapped(run_id, conf, *arg):
...
+                    f"of length {len(to_pe)}, but n_pmts is {n_pmts}!")
+                
+            return to_pe
+        else:
+            raise ValueError(
+                "User must add his/her own values to FIXED_TO_PE as: "
+                "FIXED_TO_PE ={str(my_constant_gains): np.repeat(0.005, straxen.n_tpc_pmts)}")            
+
+    else:
+        raise NotImplementedError(f"Gain model type {model_conf} not implemented")
```
jorana, line 85: `model_conf` is not defined

straxen/get_corrections.py
```diff
@@ -119,10 +103,9 @@ def get_correction_from_cmt(run_id, conf):
...
     if isinstance(conf, tuple) and len(conf) == 3:
         is_nt = conf[-1]
-
-        model_type, global_version = conf[:2]
+        model_conf, global_version = conf[:2]
         correction = global_version  # in case is a single value
-        if 'constant' in model_type:
+        if 'constant' in model_conf:
             if not isinstance(global_version, (float, int)):
                 raise ValueError(f"User specify a model type {model_type} "
```
jorana, line 110: `model_type` is not defined

straxen/get_corrections.py
```diff
@@ -134,52 +117,18 @@ def get_correction_from_cmt(run_id, conf):
                 raise ValueError(f"Could not find a value for {model_type} "
```
jorana, line 117: also here

straxen/get_corrections.py
```diff
@@ -119,10 +103,9 @@ def get_correction_from_cmt(run_id, conf):
         return conf
     if isinstance(conf, tuple) and len(conf) == 3:
         is_nt = conf[-1]
```
jorana, line 105: also for `get_correction_from_cmt` now would be a good time to add a docstring what inputs are expected and what the function does

straxen/get_corrections.py
```diff
@@ -134,52 +117,18 @@ def get_correction_from_cmt(run_id, conf):
...
-    # 1T support for electron lifetimes from a file
-    # Let's remove these functions and only rely on the CMT in the future
-    x = straxen.get_resource(elife_conf, fmt='npy')
-    run_index = np.where(x['run_id'] == int(run_id))[0]
-    if not len(run_index):
-        # Electron lifetime not known: using placeholders
-        e = 623e3
-    else:
-        e = x[run_index[0]]['e_life']
-    return float(e)
```
jorana, line 185: It's probably easier to leave this as it is right, rather than putting this in `get_to_pe`

straxen/plugins/event_processing.py
```diff
@@ -452,25 +451,20 @@ class CorrectedAreas(strax.Plugin):
...
-            cmt, cmt_conf, is_nt = self.config['s1_xyz_correction_map']
-            cmt_conf = (f'{cmt_conf[0]}_{self.config["default_reconstruction_algorithm"]}', cmt_conf[1])
-            map_algo = cmt, cmt_conf, is_nt
-
-            self.s1_map = InterpolatingMap(get_resource(get_config_from_cmt(self.run_id, map_algo)))
+            is_nt = self.config['s1_xyz_correction_map'][-1]
+            conf = self.config['s1_xyz_correction_map'][:2]
+            map_algo = (f'{conf[0]}_{self.config["default_reconstruction_algorithm"]}')
+            cmt_conf = map_algo, conf[1], is_nt
+            self.s1_map = InterpolatingMap(get_resource(get_correction_from_cmt(self.run_id, cmt_conf)))
```
jorana, line 459:
```suggestion
            config_name, cmt_version, is_nt = self.config['s1_xyz_correction_map']
            map_algo = f'{config_name}_{self.config["default_reconstruction_algorithm"]}'
            cmt_conf = map_algo, cmt_version, is_nt
            self.s1_map = InterpolatingMap(get_resource(get_correction_from_cmt(self.run_id, cmt_conf)))
```

straxen/contexts.py
```diff
@@ -292,7 +290,7 @@ def xenonnt_initial_commissioning(*args, **kwargs):
     # Events*
     left_event_extension=int(1e6),
     right_event_extension=int(1e6),
-    elife_conf=straxen.aux_repo + '3548132b55f81a43654dba5141366041e1daaf01/strax_files/elife.npy',
+    elife_conf=('elife','v1', False),
```
jorana, line 293:
```suggestion
    elife_conf=('elife', 'v1', False),
```

jorana (changes requested):
Hi @ahiguera-mx , thanks for the thorough work! It's nice that these naming conventions are now smoothed out! 
## General / political
One broader question I have is if we get CMT working as a context attribute, would it still make sense to have the  (correction, version, **nT=boolean**), since that *nT=boolean** will easily be captured in the registering of the CMT for a given context. If this is the case, having a reprocessing twice to reflect these changes in the options naming might be a bit much. Since this is a political question rather than the code itself, we might want to move the discussion to slack.

## Physics review
I hope one of the ACs or someone from corrections may provide a review where the values obtained from CMT are tested, e.g. for 1T to see if the results from CMT are the same.

## Code review
I've found a few instances where some some variable is not introduced correctly but I'm sure that will be easy to fix. Also, perhaps now that everything has settled down, it is a good time to add some doc strings to the core functions.

## Testing
I agree with Daniel that there is a merit to also run this PR on a branch of XenonnT as this is one of the PRs that needs to test with a database most. This would also hopefully spot whatever we don't need (e.g., why haven't the errors I point out been caught earlier, most likely because those lines are never used). For things we do need, indeed adding some tests would be great.

jorana:
By the way a941892 fixes the minianalysis from @ftoschi to use this new CMT convention

straxen/get_corrections.py
ahiguera-mx: done

straxen/get_corrections.py
ahiguera-mx: fixed

straxen/get_corrections.py
ahiguera-mx: fixed

straxen/get_corrections.py
ahiguera-mx: fixed

straxen/get_corrections.py
ahiguera-mx: fixed

straxen/get_corrections.py
ahiguera-mx: done

straxen/get_corrections.py
ahiguera-mx: That is not what was done not sure what do you mean, that being said, there is no reason to duplicate functions that do the same, this is something I learn from Daniel

straxen/contexts.py
ahiguera-mx: Done

straxen/plugins/event_processing.py
ahiguera-mx: done

tests/test_cmt.py
```diff
@@ -1,8 +1,89 @@
...
+from straxen.common import aux_repo
 
+def test_connect_to_db():
+    """
+    Test connection to db
+    """
+    if not straxen.utilix_is_configured():
+        warn('Cannot do test becaus  '
+             'no have access to the database.')
+        return
```
WenzDaniel, line 19: This is only the case if you push from one branch of the origin repository to another. Hence when pushing from your fork those tests wont be executed.

tests/test_cmt.py
```diff
@@ -1,8 +1,89 @@
...
+def test_1T_elife():
+    """
+    Test elife from CMT DB against historical data(aux file)
+    """
+    if not straxen.utilix_is_configured():
+        warn('Cannot do test becaus  '
+             'no have access to the database.')
+        return
+
+    elife_conf = ('elife', 'ONLINE', False)
```
WenzDaniel, line 45: I have the impression, that this will become difficult in the long run. Maybe we should consider to add to each new correction a "test" column which contains just a static value.

tests/test_cmt.py
```diff
@@ -1,8 +1,89 @@
...
+    assert elife_cmt == elife, mes
+
+def test_nT_to_pe_values(execute=True):
+    """
+    Test to_pe average values nT
+    """
+    if not straxen.utilix_is_configured() or not execute:
+        warn('Cannot do test becaus  '
+             'no have access to the database.')
+        return
```
WenzDaniel, line 61: Why would you not execute a test? It is there purpose to be always executed no matter what. 
```suggestion
def test_nT_to_pe_values():
    """
    Test to_pe average values nT
    """
    if not straxen.utilix_is_configured():
        warn('Cannot do test becaus  '
             'no have access to the database.')
        return

```

tests/test_cmt.py
```diff
@@ -1,8 +1,89 @@
...
+
+    gain_model = ('to_pe_model', 'ONLINE', True)
+    to_pe_cmt = straxen.get_to_pe(test_run_id_nT, gain_model, straxen.n_tpc_pmts)
+    ave_to_pe_cmt = round(np.average(to_pe_cmt), 3)
+    ave_to_pe = 0.003  # average value for the nT test run
```
WenzDaniel, line 66: See my comment above about the test column. I could imagine that this may not be always true.

tests/test_cmt.py
```diff
@@ -1,8 +1,89 @@
...
+    mes = 'Average values do not match. Please check'
+    assert np.isclose(ave_to_pe_cmt, ave_to_pe), mes
+
+def test_cmt_conf_option(option='mlp_model', version='ONLINE', is_nT=True):
```
WenzDaniel, line 70: In most cases a test does not need any options. They are designed to test just one specific feature as they are executed automatically no one will ever change its arguments.  The story becomes different when you use randomized tests with hypothesis. But this wont be necessary here.

tests/test_cmt.py
```diff
@@ -1,8 +1,89 @@
...
+    """
+    if not straxen.utilix_is_configured():
+        warn('Cannot do test becaus  '
+             'no have access to the database.')
+        return
+
+    conf = option, version, is_nT
+    if option == 'to_pe_model':
+        correction = straxen.get_to_pe(test_run_id_nT, conf, straxen.n_tpc_pmts)
+        assert isinstance(correction, np.ndarray)
```
WenzDaniel, line 83: This part is never tested

straxen/get_corrections.py
```diff
@@ -114,72 +93,44 @@ def get_to_pe(run_id, gain_model, n_pmts):
...
+ 
     if isinstance(conf, str) and conf.startswith('https://raw'):
         # Legacy support for pax files
         return conf
     if isinstance(conf, tuple) and len(conf) == 3:
         is_nt = conf[-1]
-
-        model_type, global_version = conf[:2]
+        model_conf, global_version = conf[:2]
         correction = global_version  # in case is a single value
```
WenzDaniel, line 109: Could we just rename it in `global_version_or_constant_value` ? In the current version it is super hard to understand because it is very inconsistently used.  
In line 111 you check if global_version is a float or int while the setting "global_version" should be a string if specified as version. 
In line 117 you overwrite correction with a correction value etc.

straxen/get_corrections.py
```diff
@@ -37,72 +38,50 @@ def correction_options_wrapped(run_id, conf, *arg):
...
-
-        try:
-            # Uniform gain, specified as a to_pe factor
-            to_pe = np.ones(n_pmts, dtype=np.float32) * model_conf
-        except np.core._exceptions.UFuncTypeError as e:
-            raise (str(e) +
-                   f"\nTried multiplying by {model_conf}. Insert a number instead.")
-    else:
-        raise NotImplementedError(f"Gain model type {model_type} not implemented")
+def get_to_pe(run_id, conf, n_pmts):
```
WenzDaniel, line 41: I think we can join this function with `get_correction_from_cmt`. See my comments below.

straxen/get_corrections.py
```diff
@@ -37,72 +38,50 @@ def correction_options_wrapped(run_id, conf, *arg):
...
+        if 'to_pe_model' in model_conf:
+            corrections = straxen.CorrectionsManagementServices(is_nt=is_nt)
+            to_pe = corrections.get_corrections_config(run_id, conf[:2])
+
+            return to_pe
+        else:
+            raise ValueError("Wrong configuration. "
+                         "Please use the following format: "
+                         "(config->str, model_config->str or number, is_nT->bool) "
+                         f"User specify {conf} please modify")
```
WenzDaniel, line 61: This is more of less identical with first part of `get_correction_from_cmt`

straxen/get_corrections.py
```diff
@@ -37,72 +38,50 @@ def correction_options_wrapped(run_id, conf, *arg):
...
+    elif isinstance(conf, tuple) and len(conf) == 2:
+        model_conf = conf[0]
+        if model_conf in FIXED_TO_PE:
+            to_pe = FIXED_TO_PE[model_conf]
+            if len(to_pe) != n_pmts:
+                raise ValueError(
+                    f"Gain model {model_conf} resulted in a to_pe "
+                    f"of length {len(to_pe)}, but n_pmts is {n_pmts}!")
+                
+            return to_pe
```
WenzDaniel, line 73: I think you can drop n_pmts from the function arguments. As its only purpose is this check which is not really needed in my opinion. I think the argument is more legacy code which was required before to check whether the detector is 1T or nT.

WenzDaniel (commented):
Thanks Aaron, I pulled your branch and the tests seem to work. I think you did already a great job, but I think one can simplify things even further see my comments. In addition I added some comments about the tests.

straxen/get_corrections.py
```diff
@@ -114,72 +93,44 @@ def get_to_pe(run_id, gain_model, n_pmts):
 @export
 @correction_options
 def get_correction_from_cmt(run_id, conf):
+    """
+    Get correction from CMT
+    :param run_id: run id from runDB
+    :param conf: configuration
```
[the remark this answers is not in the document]
ahiguera-mx: You are right, I've added additional information

tests/test_cmt.py
ahiguera-mx: we are using a lot of aux/pax fails in testing so don't understand why this is an issue

tests/test_cmt.py
ahiguera-mx: Here I'm following Joran's conventions, reading 494 values can be slow, therefore we add the option to execute or not the test

tests/test_cmt.py
ahiguera-mx: As it says it is valid only for the nT test run. But we can get rid of this test if we get rid of `get_to_pe()`

tests/test_cmt.py
ahiguera-mx: fixed

tests/test_cmt.py
ahiguera-mx: we remove, see below

straxen/get_corrections.py
ahiguera-mx: fixed

straxen/get_corrections.py
ahiguera-mx: you are right, I remove `get_to_pe()`

straxen/get_corrections.py
ahiguera-mx: fixed

straxen/get_corrections.py
ahiguera-mx: agree

ahiguera-mx:
Hi Daniel,
Thanks a lot for the very useful comments. I've implemented the changes as suggested

straxen/get_corrections.py
```diff
@@ -34,152 +35,76 @@ def correction_options_wrapped(run_id, conf, *arg):
...
-                                 "please check it is implemented in CMT. "
-                                 f"for nT = {is_nt}")
-
-        if 'samples' in model_type:
+        elif 'samples' in model_conf: # int baseline samples, etc
             return int(correction)
+
         else:
-            return float(correction)
+            return float(correction) # float elife, drift velocity, etc
```
WenzDaniel, line 94: We could reduce the number of return statements here to a single one. This makes is slightly more readable. You could do something like 
```
if ....
    correction = ' '.join(map(str, correction))
elif ...... 
    correction = int(correction)
....
return correction 
```
In that case you can also remove if to_pe in ....

WenzDaniel (approved):
Hej Aaron thanks a lot. Your new function is now very nice and clean. Just one minor thing which stroked my eye. In addition we should be cautious that we do not add for each new correction another elif now. In that case we may have to think a bit how this can be avoided.

straxen/get_corrections.py
ahiguera-mx: Good suggestion, by the way we should be fine since we have all types: str, int, float and array

straxen/get_corrections.py
```diff
@@ -34,152 +36,75 @@ def correction_options_wrapped(run_id, conf, *arg):
...
-            f"of length {len(to_pe)}, but n_pmts is {n_pmts}!")
-    return to_pe
+        # special case constant single value
+        elif 'constant' in model_conf:
+            if not isinstance(cte_value, (float, int)):
+                raise ValueError(f"User specify a model type {model_conf} "
+                                 "and should provide a number. Got: "
+                                 f"{type(cte_value)}")
+            correction = cte_value
+            return correction
```
WenzDaniel, line 74: No, comment. I just need to have some thing I can link in https://github.com/XENONnT/mc_chain/pull/14.

WenzDaniel: I apologize, there was not any other way of linking to the code in this PR directly. I was discussion with Diego something about https://github.com/XENONnT/mc_chain/pull/14.

ahiguera-mx: no worries

straxen/get_corrections.py
```diff
@@ -34,152 +36,75 @@ def correction_options_wrapped(run_id, conf, *arg):
...
+    get_correction_from_cmt(run_id, conf[:2])
+    special cases:
+    version can be replaced by consant int, float or array
+    when user specify value(s)
+    :param run_id: run id from runDB
+    :param conf: configuration 
+    :return: correction value(s)
+    """
+ 
+    if isinstance(conf, str) and conf.startswith('https://raw'):
```
zhut19, line 55: Do you think 1T compatibility check can already handle the is_CMT check [here](https://github.com/ahiguera-mx/straxen/blob/master/straxen/plugins/event_processing.py#L452-L460)? While I would also propose moving this making `map_algo` part to [the decorator function](https://github.com/ahiguera-mx/straxen/blob/master/straxen/get_corrections.py#L27-L32)

zhut19: Might also consider in the 'http://raw' case, return the data after get_resource?

straxen/get_corrections.py
```diff
@@ -34,152 +36,75 @@ def correction_options_wrapped(run_id, conf, *arg):
...
+            return correction
 
-    if len(to_pe) != n_pmts:
-        raise ValueError(
-            f"Gain model {gain_model} resulted in a to_pe "
-            f"of length {len(to_pe)}, but n_pmts is {n_pmts}!")
-    return to_pe
+        # special case constant single value
+        elif 'constant' in model_conf:
+            if not isinstance(cte_value, (float, int)):
```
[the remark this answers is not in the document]
zhut19: Right, but it will be nice to pass only the file name instead of the full path, and then the mongo downloader can take care of the rest. The same concept also applies to [the correction files](https://github.com/ahiguera-mx/straxen/blob/ab7b4c485e1270b12302196e8c89ba0e4adbe473/straxen/plugins/event_processing.py#L459-L460), I believe.

straxen/get_corrections.py
ahiguera-mx: You mentioned 1T, that is why I made the comment

straxen/get_corrections.py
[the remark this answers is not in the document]
ahiguera-mx: I see what you mean something like 

```
if isintance(conf, str):
      downloader = straxen.MongoDownloader()
      downloader.download_single(conf)
      return path_to_the_file
```

But at then end you need to pass the path to the file, right?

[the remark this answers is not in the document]
ahiguera-mx: Wait, I think you can do what you one already since `get_resources()` can handle a few options [here](https://github.com/ahiguera-mx/straxen/blob/master/straxen/common.py#L152)

jorana (approved):
Let's start staging this, thanks again Aaron!

PR #458, opened 2021-04-29
do check for overlaps in NV

jorana:
This PR reactivates the NV overlapping pulse checks.

@WenzDaniel, you said this was checked and confirmed right?


WenzDaniel:
Yes, but I checked only like 4 h of data by now. I would like to check a few more first.

PR #459, opened 2021-04-29
Added support of dill for resource files

terliuk:

Adds support for dill files in resources in `get_resource`/`open_resource` functions

Usage is identical to pickle files

  - Update the docstring(s)
  - Update the documentation

### DIFF
--- a/straxen/common.py
+++ b/straxen/common.py
@@ -106,16 +106,16 @@ def open_resource(file_name: str, fmt='text'):
             result = result_slurped
     elif fmt == 'pkl':
         with open(file_name, 'rb') as f:
-            result = pickle.load(f)
+            result = pickle.load(f)  # nosec
     elif fmt == 'pkl.gz':
         with gzip.open(file_name, 'rb') as f:
-            result = pickle.load(f)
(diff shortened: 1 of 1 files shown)


terliuk:
Previously, it was not possible to store some of the scipy objects in pickle. In particular, `RegularGridInterpolator`. However, I checked now and it looks like now it works with standard pickle. I think it can still be useful, but up to you to decide.

jorana:
@terliuk , looks fine to me, only codefactor has some security issues (like it has for pickle), would you say we need to worry about these?

terliuk:
In general, it the future it is better if we use dedicated formats, but i am not sure how realistic it is. My understanding is that dill/pickle does full serialization, so technically can contain malicious binary code/class inside. So given that we use pickle already, it shouldn't lead to new concerns in addition to pickle ones.

jorana:
Right, sure, I think that summarizes it quite nicely. In the file-repos, we highly discourage the usage of such files, however supporting them in principle may not hurt. We should try depending on them as little as possible but e.g. dedicated scripts may benefit from this utility function. Otherwise people just `import dill` and do it manually.

PR #460, opened 2021-04-30
Update veto_hitlets.py

petergaemers:
To avoid problems with "double defaults" options used by multiple plugins shouldn't have defaults set in their class but by the context.


jorana:
@petergaemers , we usually try keeping the code clear by setting defaults at the plugin to keep it simple. Would this issue persist if we merge https://github.com/XENONnT/WFSim/pull/145/files? If not, I'd prefer to keep the default set at the plugin.

petergaemers:
It will not. It will when you add another plugin who wants nveto gains

WenzDaniel:
Based on the discussion in slack I guess we can close?

jorana:
sounds good, @petergaemers we can always re-open if needed

PR #461, opened 2021-04-30
fix 1T sim context to have working dep. trees

jorana:
In https://github.com/XENONnT/WFSim/pull/146 we added a test that should test if the sim contexts are properly constructed. This turns out to be that some plugins are either missing or are depending on datatypes unknown to the context.

In this PR, we either remove or add missing plugins such that all datatypes are properly known to the context.

Run https://github.com/XENONnT/WFSim/pull/146

### DIFF
--- a/straxen/contexts.py
+++ b/straxen/contexts.py
@@ -200,6 +200,10 @@ def xenonnt_simulation(output_folder='./strax_data'):
                     ),
         **straxen.contexts.xnt_common_opts)
     st.register(wfsim.RawRecordsFromFaxNT)
+    # Deregister NV/MV
+    for plugin in list(st._plugin_class_registry.keys()):
+        if plugin.endswith('_nv') or plugin.endswith('_mv'):
+            del st._plugin_class_registry[plugin]
     return st
 
 
@@ -389,5 +393,5 @@ def xenon1t_simulation(output_folder='./strax_data'):
             detector='XENON1T',
             **straxen.contexts.x1t_common_config),
         **straxen.contexts.common_opts)
-    st.register(wfsim.RawRecordsFromFax1T)
+    st.register([wfsim.RawRecordsFromFax1T, straxen.PeakPositions1T])
     return st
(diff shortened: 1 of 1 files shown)


WenzDaniel:
Wait what, why are the veto systems unknow to wfsim?

jorana:
> 
> 
> Wait what, why are the veto systems unknow to wfsim?
There is no NV or MV in the RR plugin here:
https://github.com/XENONnT/WFSim/blob/master/wfsim/strax_interface.py#L461

WenzDaniel:
@Peter @ramirezdiego what are your suggestions here, will we have different contexts for different simulations, or can we unify things? E.g. will we have a full chain context the mc chain https://github.com/XENONnT/WFSim/blob/fe0148408b5b624c6569a929b435f68530583cb6/wfsim/strax_interface.py#L529 and another context for vetos only?

ramirezdiego:
Sorry guys, you must explain this a bit slower to me 😄 
@jorana introduced the context test in https://github.com/XENONnT/WFSim/pull/146, which complains because `RawRecordsFromFaxNT` doesn't expect any n/muVeto plugin. So this fix is to cleanup these datatypes when straxen runs?

In WFSim, we have `RawRecordsFromFaxnVeto`, for when running nVeto sims. And for the full-chain with both we load the context in mc_chain using this trick: https://github.com/XENONnT/mc_chain/blob/2c9b58ea4e0a5ae6b57601fa4d2958e25718ed12/run_wfsim.py#L108 (still under review) @WenzDaniel Does this respond to your question?

jorana:
Alright, after a discussion on slack, let's just leave it at fixing the 1T sim context

WenzDaniel:
> In WFSim, we have RawRecordsFromFaxnVeto, for when running nVeto sims. And for the full-chain with both we load the context in mc_chain using this trick: https://github.com/XENONnT/mc_chain/blob/2c9b58ea4e0a5ae6b57601fa4d2958e25718ed12/run_wfsim.py#L108 (still under review) @WenzDaniel Does this respond to your question?

Okay, thanks. I do not know in detail how you guys envision the simulation context. But the current approach is a bit cumbersome if you want to load any MC  chain simulated data. You must do:
```
st = straxen.contexts.xenonnt_simulation(output_folder='./strax_data')
st.register(wfsim.RawRecordsFromMcChain)
```
Because the lineage-hash of MC chain is different from the default simulation context:
```
1-raw_records-m4hi7bp7ta
1-records-nojcknf3h4
```

PR #463, opened 2021-05-01
Refactor veto plugins

WenzDaniel:
In this PR we refactor the veto plugins. The following things were changed:

1. Removed clean_up_empty_records as it did not help much to reduce the overall data size. 
2. Removed gain option from event plugins since they are not needed any longer.
3. Simplified hitlet plugin by clustering some of its code into a new function which is now part of [strax](https://github.com/AxFoundation/strax/pull/436).

Please note, that the test are going to fail until [#436](https://github.com/AxFoundation/strax/pull/436) is merged.


straxen/plugins/veto_events.py
```diff
@@ -45,11 +43,6 @@ def infer_dtype(self):
         self.n_channel = (self.channel_range[1] - self.channel_range[0]) + 1
         return veto_event_dtype(self.name_event_number, self.n_channel)
 
-    def setup(self):
-        self.to_pe = straxen.get_to_pe(self.run_id,
-                                       self.config['gain_model_nv'],
-                                       self.n_channel)
-
```
jorana, line 52: Right, was this ever needed actually?

jorana (approved):
Thanks Daniel

straxen/plugins/veto_events.py
WenzDaniel: Only in a local version of mine. I forgot to remove it.

PR #465, opened 2021-05-03
Veto plugins

WenzDaniel:
In this PR we are adding a veto plugin for the neutron and muon-veto. The veto plugins define the regions in which S1 peaks would be tagged by the muon- or neutron-veto. The corresponding event plugin for the TPC will be shipped in an additional PR. The corresponding tests had been added too.

To avoid any duplicated code, I had to refactored the function _merge_intervals in the nveto recorder. Due to this modification I also had to slightly adjust the nveto_recorder plugin as well as the function find_coincidence. Since neither merge_intervals nor find_coincidence was tested by unit tests yet. I also added those.


straxen/plugins/veto_vetos.py
```diff
@@ -0,0 +1,160 @@
...
+        help='Veto time in ns left t the start of a vetoing event.'),
+    strax.Option(
+        'veto_right_nv', default=0, type=int, track=True,
+        help='Veto time in ns right to the end of a vetoing event.'),
+)
+class nVETOveto(strax.OverlapWindowPlugin):
+    """
+    Plugin which defines the time intervals in which peaks should be
+    tagged as vetoed. An event must surpass all three criteria to trigger
+    a veto.
```
jorana, line 33: _minor_: I think this answers my question 😉 , is it useful to specify?

straxen/plugins/veto_vetos.py
```diff
@@ -0,0 +1,160 @@
...
+    strax.Option(
+        'min_veto_channel_nv', default=0, type=int, track=True,
+        help='Minimal number PMT channel contributing to the event.'),
+    strax.Option(
+        'veto_left_nv', default=500_000, type=int, track=True,
+        help='Veto time in ns left t the start of a vetoing event.'),
+    strax.Option(
+        'veto_right_nv', default=0, type=int, track=True,
+        help='Veto time in ns right to the end of a vetoing event.'),
+)
```
jorana, line 28: _medium_: perhaps call them extensions like we do elsewhere

tests/test_nveto_recorder.py
```diff
@@ -0,0 +1,75 @@
...
+        intervals[0]['time'] = 2
+        intervals[0]['endtime'] = 4
+        intervals[1]['time'] = 3
+        intervals[1]['endtime'] = 7
+        intervals[2]['time'] = 7
+        intervals[2]['endtime'] = 8
+
+        # Second interval:
+        intervals[3]['time'] = 20
+        intervals[3]['endtime'] = 22
```
jorana, line 28: _minor_: I guess you thought this might be useful for the sake of the comments right? I think it might read simpler if you just used `intervals['time'] = [2, 3, 4, 7, 20]` or even `intervals = np.ndarray([[2, 3, 4, 7, 20], ....., dtype=...)`

Especially since you do this for several tests below.

tests/test_nveto_recorder.py
```diff
@@ -0,0 +1,75 @@
...
+
+        intervals = straxen.merge_intervals(intervals)
+
+        assert len(intervals) == 2, 'Got the wrong number of intervals!'
+        assert intervals[0]['time'] == 2, 'First interval has the wrong time!'
+        assert intervals[0]['endtime'] == 8, 'First interval has the wrong endtime!'
+        assert intervals[1]['time'] == 20, 'Second interval has the wrong time!'
+        assert intervals[1]['endtime'] == 22, 'Second interval has the wrong endtime!'
```
jorana, line 36: _minor_: This can also be a little simpler to read right, if you keep your input you can compare times wrt to the input rather than hardcode it

tests/test_nveto_recorder.py
```diff
@@ -0,0 +1,75 @@
...
+        intervals[2]['endtime'] = 10
+
+        intervals[3]['time'] = 38
+        intervals[3]['endtime'] = 42
+        self.intervals = intervals
+
+    def test_empty_inputs(self):
+        intervals = np.zeros(0, dtype=strax.time_fields)
+        intervals = straxen.find_coincidence(intervals)
+        assert len(intervals) == 0, 'Empty input should return empty intervals!'
```
jorana, line 58: _medium_: same comments as above apply here

tests/test_veto_vetos.py
```diff
@@ -0,0 +1,78 @@
+import straxen
+
+import numpy as np
+import unittest
+
+
+class TestCreateVetoIntervals(unittest.TestCase):
```
jorana, line 7: _minor_: same comments as above apply here

jorana (commented):
Thanks @WenzDaniel I think overall it looks good. Just mostly minor comments below:

straxen/plugins/veto_vetos.py
```diff
@@ -0,0 +1,160 @@
+import strax
+import straxen
+
```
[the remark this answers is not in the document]
WenzDaniel: This is just some personal preference. I like to sort the imports a bit to give it some structure, but I also do not mind to remove them.

straxen/plugins/nveto_recorder.py
```diff
@@ -105,29 +105,37 @@ def compute(self, raw_records_nv, start, end):
                                min_amplitude=self.config['hit_min_amplitude_nv'])
         del temp_records
```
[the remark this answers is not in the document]
WenzDaniel: The resulting raw_records_coin_nv should be identical, but we can also increase the version number for bookkeeping. It will just mean a lot of reprocessing.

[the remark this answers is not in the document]
jorana: if you are sure that the results are the same, I stand corrected and no version bump would be needed

WenzDaniel:
> One general question, you decided against merging this plugin with the plugins from https://github.com/XENONnT/straxen/pull/487/files since we at the moment do not know the correct settings or is there a more fundamental reasoning behind this?

Hm mainly two reasons. Back then it was not fully clear which way to go either tag all S1 Peaks or just the one classified as main/alt S1 in event basics. The second reason, I did not want extend this PR with another feature as it is already quite long.

straxen/plugins/nveto_recorder.py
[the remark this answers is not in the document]
WenzDaniel: I checked with a 6 min run (42 million raw_record fragments and 1.6 million raw_records_coin_nv). Shape and all entries are identical.

jorana (approved):
> 
> 
> > One general question, you decided against merging this plugin with the plugins from https://github.com/XENONnT/straxen/pull/487/files since we at the moment do not know the correct settings or is there a more fundamental reasoning behind this?
> 
> Hm mainly two reasons. Back then it was not fully clear which way to go either tag all S1 Peaks or just the one classified as main/alt S1 in event basics. The second reason, I did not want extend this PR with another feature as it is already quite long.

I mean, if you considered making it a single plugin instead of multiple?

Other than that, I think you looked at all the comments right without any controversies?

WenzDaniel:
> I mean, if you considered making it a single plugin instead of multiple?
We may want to replace and/or extend the way we decide when to veto a certain events. Hence I thought it would be simpler to distinguish between the vetoing and tagging part.

> Other than that, I think you looked at all the comments right without any controversies?
No, no objects here. I implemented them.

jorana:
> 
> 
> > I mean, if you considered making it a single plugin instead of multiple?
> 
> We may want to replace and/or extend the way we decide when to veto a certain events. Hence I thought it would be simpler to distinguish between the vetoing and tagging part.

yes, for the development this is certainly true. I don't know how big the penalty (if there exists one) is for having to communicate between plugins is.

PR #467, opened 2021-05-03
Added exceptions to review do

WenzDaniel:
Adding additional exceptions to review dog. 

- DAR101 does not seem to work, see e.g. https://github.com/XENONnT/straxen/pull/465/files#r625016503
- S101 and WPS218  are annoying in tests. 
- E501 can be discussed. If we say, that this is our limit I adjust my pycharm.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -31,7 +31,6 @@ exclude =
 # DAR101 Missing parameter(s) in Docstring: - min_hits
 # E128 continuation line under-indented for visual indent
 # E251 unexpected spaces around keyword / parameter equals
-# E501 line too long (105 > 100 characters)
 # F403 'from .veto_events import *' used; unable to detect undefined names
 # F541 f-string is missing placeholders
 # I001 isort found an import in the wrong position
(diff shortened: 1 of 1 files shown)


setup.cfg
```diff
@@ -28,18 +28,23 @@ exclude =
 # D400 First line should end with a period
 # D401 First line should be in imperative mood
 # D403 First word of the first line should be properly capitalized
+# DAR101 Missing parameter(s) in Docstring: - min_hits
 # E128 continuation line under-indented for visual indent
 # E251 unexpected spaces around keyword / parameter equals
+# E501 line too long (105 > 100 characters)
```
jorana, line 34: @WenzDaniel , with this one I would disagree, it would be nice if we can try and limit this line length. Otherwise we can increase the value to be a bit higher using the 100 a few lines above.

setup.cfg
```diff
@@ -28,18 +28,23 @@ exclude =
 # D400 First line should end with a period
 # D401 First line should be in imperative mood
 # D403 First word of the first line should be properly capitalized
+# DAR101 Missing parameter(s) in Docstring: - min_hits
```
jorana, line 31: Yeah, it is a good check but if there are multiple, it turns messy quite quickly

setup.cfg
jorana: Nevertheless, I think 100 is a good rule of thumb

jorana (changes requested):
Looks good apart from E501

setup.cfg
jorana: otherwise 120 (that is actually my PyCharm hard-limit)

WenzDaniel: For me it was 110. But I also do not mind 100. Both fits on my small 18 inch screen :D

jorana: Yeah, let's not forget the scope of this bot, it should be a gentle reminder, not a barking dog (which I agree it sometimes is)

jorana:
Thanks Daniel!

PR #468, opened 2021-05-04
Remove zero gain hits

WenzDaniel:
In this PR we add a tiny function which removes all hits (hitlets) for channels which gain was set to zero, if any.

### DIFF
--- a/straxen/plugins/veto_hitlets.py
+++ b/straxen/plugins/veto_hitlets.py
@@ -89,8 +89,7 @@ class nVETOHitlets(strax.Plugin):
 
     def compute(self, records_nv, start, end):
         hits = strax.find_hits(records_nv, min_amplitude=self.config['hit_min_amplitude_nv'])
-        active_channel = np.argwhere(self.to_pe > 0).flatten()
-        hits = hits[np.isin(hits['channel'], active_channel)]
+        hits = remove_switched_off_channels(hits, self.channel_range, self.to_pe)
 
         temp_hitlets = strax.create_hitlets_from_hits(hits,
                                                       self.config['save_outside_hits_nv'],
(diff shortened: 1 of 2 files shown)


straxen/plugins/veto_hitlets.py
```diff
@@ -120,6 +122,23 @@ def compute(self, records_nv, start, end):
...
+        for the neutron-veto.
+    :param to_pe: conversion factor from ADC per sample.
+    :return: Hits
+    """
+    channel_on = np.argwhere(to_pe > 0).flatten()
+    all_channel = np.arange(channel_range[0], channel_range[1] + 1)
+    channel_off = set(all_channel) - set(channel_on)
+    mask_on = ~np.isin(hits['channel'], list(channel_off))
+    hits = hits[mask_on]
+    return hits
```
jorana, line 139:
```suggestion
def remove_switched_off_channels(hits, to_pe):
    """Removes hits which were found in a channel without any gain.
    :param hits: Hits found in records.
    :param to_pe: conversion factor from ADC per sample.
    :return: Hits
    """
    channel_off = np.argwhere(to_pe == 0).flatten()
    mask_off = np.isin(hits['channel'], channel_off)
    hits = hits[~mask_off]
```

jorana (commented):
Can be a bit simpler I believe, I don't see why a channel-range is needed if one uses to-pe indexes

WenzDaniel:
I did it mainly for performance reasons. I started with the example you are showing, but in that case we have to compare each hit ~2000 times with a number. While in the other case we reduce it to something in the order to 1 or 2.

straxen/plugins/veto_hitlets.py
```diff
@@ -120,6 +122,23 @@ def compute(self, records_nv, start, end):
...
+    """Removes hits which were found in a channel without any gain.
+
+    :param hits: Hits found in records.
+    :param channel_range: Tuple of min/max channel. E.g. (2000, 2119)
+        for the neutron-veto.
+    :param to_pe: conversion factor from ADC per sample.
+    :return: Hits
+    """
+    channel_on = np.argwhere(to_pe > 0).flatten()
+    all_channel = np.arange(channel_range[0], channel_range[1] + 1)
```
jorana, line 135: By the way, you have an off by one here, since https://github.com/XENONnT/straxen/blob/0fa353810faab6b7b80a9bc0dbf7d4c8835bb993/straxen/plugins/veto_hitlets.py#L78 is end inclusive right?

straxen/plugins/veto_hitlets.py
jorana, line 139: So if this is really impacting performance, you could of course make it simpler still, but ok, this is semantics
```suggestion
    channel_off = np.argwhere(to_pe[channel_range[0]: channel_range[1]] == 0).flatten() + channel_range[0]
    mask_off = np.isin(hits['channel'], channel_off)
    hits = hits[~mask_on]
    return hits
```

WenzDaniel (commented):
It is not large a factor 20 or so for the function and probably negligible for the whole plugin. But I also did not expect such a strong opinion on that, as the code is just 5 lines. But wen can also go with the simple approach, I really do not mind.

PR #469, opened 2021-05-04
Scada allowed, fix #392

jorana:
Some of the testing settings need to be updated to allow #392 

When testing, we use a tempory secret .xenon_config file to allow access to to-test entities like the database and scada. 

If we e.g. change a URL, we need to update the secret in the config-file accordingly.

We use this on two places:
- coveralls.yml -> for making a coveralls report and uploading the code coverage
- pytest.yml -> for testing with different python versions etc.

**Where is this accessible?**
Only on the XENONnT fork. That provides us with an additional layer of protection, in addition to the already secure usage of [github sectets](https://docs.github.com/en/actions/reference/encrypted-secrets) (I've tried debugging these things, in fact they are so secret that I grew a few grey hairs by the fact that the secrets where so secret I could impossibly understand what was going on.).

For that reason, always tests should also pass as if we did not have access to the database.

### DIFF
--- a/.github/scripts/create_readonly_utilix_config.sh
+++ b/.github/scripts/create_readonly_utilix_config.sh
@@ -22,9 +22,11 @@ pymongo_database = $PYMONGO_DATABASE
 [scada]
 scdata_url = $SCADA_URL
 sclastvalue_url = $SCADA_VALUE_URL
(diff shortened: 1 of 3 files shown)


WenzDaniel (approved):
Thanks a lot

PR #471, opened 2021-05-05
Itp map patch

zhut19:
Adding more (now only RectBivariateSpline and RegularGridInterpolator) to InterpolatingMap class

This pr will add keyword arg `method` for initializing the `InterpolateMap` class
If not specified the original method of the inverse-distance weighted average the nearest neighbors will be used.
And to use the following line to interpolate with the added method.
`straxen.InterpolatingMap(data, method='RectBivariateSpline')`


zhut19:
> I am not sure what is the best solution, but I lean towards making self.data = copy(data) and then always deal with self.data instead of just data, which is pointer to straxen, so modifying the data here actually modifies original copy , resulting in impossibility to use that data twice/making different copy of interpolation object.
> See example here:
> https://github.com/XENONnT/analysiscode/blob/master/SmoothLLHDevelopment/notebooks/MapProblemStraxen.ipynb

I also don't like the idea of compressing map data inside the dictionary, however, some old files use such structure. But future files will be more JSON-like and won't need this decompressing. But for now, I changed the code back to the original order so that once the 'map' field had been decompressed, the 'compressed' field would be deleted.

terliuk:
I quickly checked regular grid iterpolation (solid) vs previous inverse distance one (dashed) and it seems to be much better now. 
S2_eff_standard

jorana:
@terliuk thanks, is that an approval from your side?

@zhut19, thanks! Is there an opportunity add a test for each of these interpolation methods?

terliuk:
I realized that I forgot to check the behavior outside of the last bin. It will fail if it receives a value outside. For example, our maps currently go in bins between [-r_tpc, + r_tpc], so technically the region around TPC edge can rather badly behave. I would probably suggest to add `bounds_error=False, fill_value=None` to RegularGridInterpolation.  @zhut19, dose it make sense?

terliuk (approved):
Looks fine, i have tested a couple of maps. There is still a question about preparation of maps and None/nan values (for example, in regions outside of TPC), in particular with RectBivariateSpline, but I think it is not really related to this PR, but that has to be solved separately during map preparation rather inside of straxen.

jorana:
Thanks Andrii and Tianyu

Issue #442, opened 2021-04-19
S1 aft bigger than 1


GiovanniVolta:
I observed some events with area fraction top slight greater than one.
For instance, consider **event_info**  of run 012126, s1_area_fraction_top 1.0000001192092896 (index: 1463 ).
I guess it is due to the fact that area_fraction_top is **numpy.float32**

Looking at **peaks** level and selecting this event ((peaks['time'] >= 1611682557517630560) & (peaks['time'] <= 1611682557517631090)). We have:

- area_top = 9.05191
- area_tot = 9.051909

A solution can be:
```
r['area_fraction_top'][m] = area_top[m].astype('f2')/check['area'][m].astype('f2')
```
 I am working with:
```
strax 0.14.0 /dali/lgrandi/strax/strax/strax/__init__.py
straxen 0.17.0 /dali/lgrandi/strax/straxen/straxen/__init__.py
```

jingqiangye:
Thanks @GiovanniVolta . Indeed `area_top` is np.float32 (this [line](https://github.com/XENONnT/straxen/blob/master/straxen/plugins/peak_processing.py#L76)) while the total area `area` is more complicated but looks to have better precision (these [lines](https://github.com/XENONnT/straxen/blob/master/straxen/plugins/peaklet_processing.py#L377-L388)). The different precisions of these two lead to some `area_fraction_top` that are slightly larger than 1.

jorana:
@GiovanniVolta @jingqiangye, thanks for spotting this issue! Both `area_per_channel` and `area_fraction_top` are `np.float32` (see reference from Jingqiang and [the peakdtype](https://github.com/AxFoundation/strax/blob/master/strax/dtypes.py#L191-L192). I think the issue is that there might be an every so slightly negative area in the bottom that makes the `area_fraction_top` go to values >1.

My guess the biggest issue with aft >1 is that it does not pass a trivial aft>1 cut?

GiovanniVolta:
> My guess the biggest issue with aft >1 is that it does not pass a trivial aft>1 cut?
The binomial test B(n, k, p), where n is the S1*S1_AFT, k is S1 and p is AFT (from map), wants n<=k. Otherwise, it fails.
So either I change the cut or we modify the aft at peak level.

jorana:
Right, this is valid. So in principle, your solution is a fine one, with the downside that we are throwing away some precision and then casting it back into a float32. 

[...]

PR #477, opened 2021-05-07
Fix AFT close but not quite 1

jorana:
## What does this PR fix
As explained in #442, there are some peaks where the AFT appears slightly larger than 1 due to numerical inaccuracies that are accumulated over the scattered area-calculation steps.

In this PR, we make sure that the AFT is calculated only using the `area_per_channel` thereby not being affected by any inaccuracies in the `area` field.

Additionally, we add a small sanity check to the peak building that the values in `area` and `area_per_channel` are consistent.

### The number of affected peaks
I did a small test on run `011680`. **Before** the change we get `0.40 %` of peaks with an AFT : 1.001 > AFT > 1. **After** this is `0.02 %` most likely due to negative area in the bottom array.

### check_peak_sum_area
I've added a small check to see if the area and the sum of the area-per-channel fields are close to one another, perhaps we should use another setting than the `np.isclose` default to see what ranges are acceptable.

### Testing
I've added a small test to run the compute argument of PeakBasics and make sure the diagnostic error message gets triggered in the correct way.

### Performance
I've checked that (if we pre-compute) peaklets etc. During the test, I set `'check_peak_sum_area'==False`. With the context set to the following config we get the performance as below.
```python
# My context config
st.context_config.update({
    'timeout': 600,
    'allow_lazy': False,
    'max_messages': 8,
})
```

Calculating total computation (and deletion time) of the peak-basics:
```python
%%timeit
!rm -r $temp_dir
st.make(run_id, target, progress_bar=True)
```
Yields:
#### Before 
19.4 s ± 1.12 s per loop (mean ± std. dev. of 7 runs, 1 loop each)
#### After 
20.6 s ± 1.65 s per loop (mean ± std. dev. of 7 runs, 1 loop each)

Which is only a ~5% change which even falls within the accuracy of this test.


setup.cfg
```diff
@@ -19,6 +19,7 @@ exclude =
   .eggs
   *.egg
   dist
+  tests/*
```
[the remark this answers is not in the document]
WenzDaniel: Hmm, if it becomes too annoying sure, but in general I think it does not hurt to also check the test code. For this reason I deactivated those "too many assert"-warning.s

WenzDaniel (commented):
I have a general question about this PR. Does not this PR indicate that that the regular peak area is biased systematically towards lower values? I mean we are starting with the issue that the area_fraction_top is too often > 1 if it is computed as area_top/area. Hence you introduce area_total which is the sum of area_per_channel. Now you compute area_fraction_top as area_top/area_total which reduces the number of peaks with an aft larger one. But in the same go you check that area_total is close to area. Does not lead this to the conclusion, that area is systematically smaller than area_total?

jorana:
@WenzDaniel , I would like to refer to #442, it looks like it's just a numerical issue. If it were systematically wrong, we would hit the test I added.

WenzDaniel:
Ah okay I see. Yes the issue seems to be numerically, but it leads to a systematically lower area value. Then it makes sense, thanks.

setup.cfg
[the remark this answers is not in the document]
jorana: ok sure, I was 55/45% sold on the idea, let me remove it then. 👍

straxen/plugins/peak_processing.py
```diff
@@ -96,11 +104,43 @@ def compute_center_times(peaks):
...
+
+        if not is_close.all():
+            for peak in peaks[positive_area][~is_close]:
+                print('bad area')
+                strax.print_record(peak)
+
+            p_i = np.where(~is_close)[0][0]
+            peak = peaks[positive_area][p_i]
+            area_fraction_off = 1 - area_per_channel_sum[p_i] / peak['area']
+            message = (f'Area not calculated correctly, it\'s '
```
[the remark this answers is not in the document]
jorana: 👎

straxen/plugins/peak_processing.py
```diff
@@ -14,7 +14,11 @@
 @export
 @strax.takes_config(
     strax.Option('n_top_pmts', default=straxen.n_top_pmts,
-                 help="Number of top PMTs"))
+                 help="Number of top PMTs"),
+    strax.Option('check_peak_sum_area', default=True, track=False,
+                 help="Check if the sum area and the sum of area per "
+                      "channel are the same."),
```
WenzDaniel, line 20: Just to be sure, you would like to test during processing too and not only during the testing.

tests/test_peaks.py
```diff
@@ -0,0 +1,71 @@
...
+        peaks = self.peaks_basics_compute(test_data)
+        assert peaks[test_peak_idx]['area_fraction_top'] == 1
+
+    @settings(deadline=None)
+    @given(strategies.floats(min_value=0,
+                             max_value=2,
+                             ),
+           strategies.integers(min_value=0,
+                               max_value=TEST_DATA_LENGTH - 1),
+           )
```
WenzDaniel, line 42: You could also filter the strategies itself. In that way you make sure that you do not "test" a case which is not actually tested in the end. I am refering to 
```
if np.isclose(off_by_factor, 1) or np.isclose(off_by_factor, 0):
            return```

WenzDaniel (approved):
Looks fine too me. Just a minor comment and a question. Thanks a lot.

straxen/plugins/peak_processing.py
jorana: This would indeed be the idea. However, perhaps it is better to run it once now on a couple of runs to not waste computing it for every single run?

tests/test_peaks.py
jorana: Sure, the advantage being that we are not supplying meaningless numbers to the test.

I've added this in .filter(_not_close_to_0_or_1)

straxen/plugins/peak_processing.py
jorana: @WenzDaniel , in da4a0f1 I've set the default to None, and made the `rtol` configurable. I noticed with processing that I needed for a random sample of runs (in range 11000 - 19000) that I needed a rtol between 2e-4 and 1e-3 to let all the runs complete. Peaks (just a few runs) that failed at an rtol ~= 1e-4 where mostly in the order of `10***7` PE.

I checked as a sanity check if the saturation correction could be involved but it was not (at least disabling made no difference).

WenzDaniel (approved):
Thanks for the changes I still approve.

PR #480, opened 2021-05-07
set event_info_double as endpoint for kr

jorana:
If we are ever to take Kr runs, if would be nice if we in that case could automatically set the target to event_info_double such that it's immediately available for analysers.

In bootstrax, we do a couple of checks for what kind of target we are going to process. Here we simply replace event_info -> event_info_double in the case of a Kr mode.


DonNabla (approved):
Thanks Joran for implementing this feature, it will be really useful for analysts in the future.
Everything looks good from my side. (Actually my brain stopped for 1 min in order to understand the _len(targets)_ ^^)

PR #484, opened 2021-05-08
use_per_run_defaults explicitly for 1T

jorana:
# What is the problem / what does the code in this PR do
The `use_per_run_defaults` is a horrible option in strax that is preferably not used. It makes for almost impossible bookkeeping as the lineage depends on a run number.

Additionally, this makes run selection very slow.

To this end in https://github.com/AxFoundation/strax/pull/440 we have added some logic that greatly speeds up run selection if a context does not allow `use_per_run_defaults`.

For XENONnT this is the case, thanks to CMT. For 1T however, we do use this feature. It is also much less of a problem because we simply have much less data for 1T.

Please checkout https://github.com/AxFoundation/strax/pull/440


WenzDaniel (approved):
I checked https://github.com/XENONnT/straxen/pull/484 and it looks fine to me, so I already approve. (the github code collapsing is very misleading here :D hiding the # XENON1T comment behind XENONnT :D)

jorana:
Thanks Daniel! Yes, I also had to take a second look 😂

PR #485, opened 2021-05-08
remove context since they are not used

jorana:
# What is the problem / what does the code in this PR do
There were two contexts that we were bookkeeping but they remain unused (one actually only returned a value error).

Let's remove these context if they are no use.

### DIFF
--- a/straxen/contexts.py
+++ b/straxen/contexts.py
@@ -50,6 +50,7 @@ xnt_common_config = dict(
 # these are placeholders to avoid calling cmt with non integer run_ids. Better solution pending.
 # s1,s2 and fd corrections are still problematic
 xnt_simulation_config = deepcopy(xnt_common_config)
+# This gain model is the average to_pe. For something more fancy use the CMT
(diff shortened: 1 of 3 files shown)


straxen/contexts.py
```diff
@@ -50,6 +50,7 @@
 # these are placeholders to avoid calling cmt with non integer run_ids. Better solution pending.
 # s1,s2 and fd corrections are still problematic
 xnt_simulation_config = deepcopy(xnt_common_config)
+# This gain model is the average to_pe. For something more fancy use the CMT
```
WenzDaniel, line 53: Do we need this comment I mean it does not even fully fit. e.g. MV and elif.

WenzDaniel (approved):

straxen/contexts.py
jorana: the place I moved it from it made even less sense. I think it's Diegos or Peters. I can remove it

ershockley (approved):

PR #488, opened 2021-05-12
Add hit thresholds to CMT

ahiguera-mx:
This will replace `hit_finderthresholds.py` by retrieving the data from CMT
This will replace `hit_finderthresholds.py` by retrieving the data from CMT
See examples on cell 17 from this list of [examples](https://github.com/XENONnT/corrections/blob/master/CMT/examples/CMT_user_guide.ipynb)
See examples on cell 17 from this list of [examples](https://github.com/XENONnT/corrections/blob/master/CMT/examples/CMT_user_guide.ipynb)

  - Update the docstring(s)
  - Update the documentation

Note: There are some tweaks I need to add to handle the fact that mongo only encodes standard python types


jorana:
Thanks guys! Nice that this will be taken over by CMT.

Some early comments already (it's a bit hard to read since it's intertwined with https://github.com/XENONnT/straxen/pull/457). It might be nice (and simple) if we keep some form of backward compatibility. If we merge this currently, it's going to be quite a lot of effort to load older data (with the previous hitfinder thresholds), and given that this is such a low-level change, we might be expecting a lot of people running into this.
  - Keeping the possibility to load the hitfinder thresholds as they were allows one to load the data also from former times if we keep the machinery in straxen. There is nothing particularly sensitive to this information so there should be little "cost" of keeping it around. So I think it would be nice if can leave the `hitfinder_thresholds.py` etc in the code. Would you agree from this point of view?
  - I don't think there is a particularly strong case for updating the version number if we can avoid it, after all, the code stays the same, it's the option that changes.

FaroutYLq:
@jorana Ah so sorry that I completely missed the message here! Now I learned that I need to check GitHub frequently... Regarding your comments:
- For your second point about the version, yes I can modify them back.
- For the first point, I am a bit confused how we can keep the old way. In the cmt based hitfinder thresholds I am trying to make the default of `hit_min_amplitude*` something like: `('hit_thresholds, 'ONLINE', True)`, while in the old way, we set the default to be something like `"nt_commissioning_initial"`. For the users, they will run into the re-computation issue if we update the default to `('hit_thresholds, 'ONLINE', True)`. I am not sure if there is a way we can balance. Any idea? Let me know if you think a zoom chat might be easier.
Thanks a lot for the feedback and again apologize for missing the message here!

jorana:
> For the users, they will run into the re-computation issue if we update the default to ('hit_thresholds, 'ONLINE', True)

Yes, but that is what we want right? The option changes so you need to recompute.

I think what you write down (having the two options both working) sounds like a desirable situation

FaroutYLq:
Thanks @jorana . I was actually confused by a `strax` question. After I change the default to the cmt style, when people want to get the old runs which has the old `hitfinder_threshold.py` style default (e.g. `st.get_array('012345', 'peak_basics')`), will they have to wait a long time to reprocess or will they just immediately load the processed one with old `hitfinder_threshold.py` style default?

ahiguera-mx:
Hi guys 

not sure if it is better to wait until #457[https://github.com/XENONnT/straxen/pull/457] is merged so we all are in the same page, also a lot of conflicts to fix with this branch

WenzDaniel:
Hi Aaron nice to have you back. I would probably follow as you suggested. First merge the refactoring PR and afterwards this one.

jorana:
> 
> 
> Thanks @jorana . I was actually confused by a `strax` question. After I change the default to the cmt style, when people want to get the old runs which has the old `hitfinder_threshold.py` style default (e.g. `st.get_array('012345', 'peak_basics')`), will they have to wait a long time to reprocess or will they just immediately load the processed one with old `hitfinder_threshold.py` style default?

If you change the defaults, it will load the new option by default so unless we reprocess the data, that wouldn't work. However one can set the defaults back if need be (that was why I requested not to remove the associated file). Does that make sense?

ahiguera-mx:
Close PR since there is too many conflicts to resolve. @FaroutYLq will continue the implementation here https://github.com/XENONnT/straxen/pull/543. Thanks to Joran and Daniel for very useful comments

PR #489, opened 2021-05-13
Activated overlaping check for mveto again.

WenzDaniel:
Agreed in yesterday telecon to activate the check again. I also changed number of baseline sample and hit thresholds as agreed on with Andrea.


jorana (approved):
Thanks Daniel and Andrea.

PR #492, opened 2021-05-17
Fix master dependencies

jorana:
Fix master for the dependencies


jorana:
Thanks for the hotfix!

PR #493, opened 2021-05-20
fix veto veto regions

jorana:
Fix some misc from #460 : 
- set as child plugin
- use updated config names
- register new plugins


WenzDaniel (approved):
This is somewhat embarrassing... Thanks a lot.

PR #494, opened 2021-05-25
Allow variable event duration

jorana:
Allow the event-duration to vary by using the max-drift length to determine the left extension.

Add the total drift length of the TPC to the left extension.


jingqiangye (approved):
@jorana Thanks a lot for the improvement! It is good to associate the left extension with the max drift time. The units also work right (max drift time, etc). Looks good to me.

PR #495, opened 2021-05-25
Max-size for rechunkable raw-records

jorana:
Set the max size of the raw-records-aqmon* to be max 50 MB to avoid overflows in the mongo buffer for veto_intervals.

Max-size for rechunkable raw-records

16 MB is the max size of a mongo document. We cannot exceed that without redesigning the mongo-storage frontend. If we split the veto_intervals over more chunks, this size/chunk (i.e. size per document) is reduced. By splitting raw-records-aqmon over more chunks, veto_intervals will also be split more, reducing the size per chunk.


petergaemers (approved):

WenzDaniel:
Thanks, sorry I had been in the car.

PR #499, opened 2021-05-26
S1/S2 event patternfit and S1 AFT test

terliuk:

This code implements the S1 AFT binomial test, as well as S1 and S2 pattern-fit/likelihood value that can be used for data quality cuts. 

Two modules are provided. One stores area per channel for peaks participating in the events, while second one actually computes likelihood and AFT binomial tests. See a drawing at the end of the note here:
https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:terliuk:nt_patternfit_parameters

Please see examples in the notes here:
https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:terliuk:nt_patternfit_parameters
https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:giovo:even_patterfit_s1


straxen/plugins/event_area_per_channel.py
```diff
@@ -0,0 +1,38 @@
...
+import strax
+import straxen
+export, __all__ = strax.exporter()
+
+@export
+@strax.takes_config(
+    strax.Option('n_tpc_pmts', type=int,
+                 help='Number of TPC PMTs'),
+)
+class event_area_per_channel(strax.LoopPlugin):
```
jorana, line 11:
```suggestion
class EventsAreaPerChannel(strax.LoopPlugin):
```

straxen/plugins/event_area_per_channel.py
```diff
@@ -0,0 +1,38 @@
...
+    """
+    Simple plugin that provides area per channel for main 
+    and alternative S1/S2 in the event. 
+    """
+    depends_on = ('events', 'event_basics', 'peaks')
+    parallel = False
```
jorana, line 17: Can you explain why you set this to this?

straxen/plugins/event_area_per_channel.py
```diff
@@ -0,0 +1,38 @@
...
+@strax.takes_config(
+    strax.Option('n_tpc_pmts', type=int,
+                 help='Number of TPC PMTs'),
+)
+class event_area_per_channel(strax.LoopPlugin):
+    """
+    Simple plugin that provides area per channel for main 
+    and alternative S1/S2 in the event. 
+    """
+    depends_on = ('events', 'event_basics', 'peaks')
```
jorana, line 16: Unless I'm mistaken I don't think you need `events`

straxen/plugins/event_area_per_channel.py
```diff
@@ -0,0 +1,38 @@
...
+@export
+@strax.takes_config(
+    strax.Option('n_tpc_pmts', type=int,
+                 help='Number of TPC PMTs'),
+)
+class event_area_per_channel(strax.LoopPlugin):
+    """
+    Simple plugin that provides area per channel for main 
+    and alternative S1/S2 in the event. 
+    """
```
jorana, line 15: Would it make sense to make this a savewhen = strax.SaveWhen.NEVER plugin as it's computation is trivial?

straxen/plugins/event_area_per_channel.py
```diff
@@ -0,0 +1,38 @@
...
+        dtype.append( (("Area per channel for main S2", "s2_area_per_channel"), 
+                       np.float32, (self.config['n_tpc_pmts'],)) )
+        dtype.append( (("Area per channel for alternative S2", "alt_s2_area_per_channel"), 
+                       np.float32, (self.config['n_tpc_pmts'],)) )
+        dtype.append( (("Area per channel for main S1", "s1_area_per_channel"), 
+                       np.float32, (self.config['n_tpc_pmts'],)) )
+        dtype.append( (("Area per channel for alternative S1", "alt_s1_area_per_channel"), 
+                       np.float32, (self.config['n_tpc_pmts'],)) )
+        dtype += strax.time_fields
+        return dtype
```
jorana, line 31:
```suggestion
    def infer_dtype(self):
        dtype = [(("Area per channel for main S2", "s2_area_per_channel"), 
                  np.float32, (self.config['n_tpc_pmts'],)),
                 (("Area per channel for alternative S2", "alt_s2_area_per_channel"),
                  np.float32, (self.config['n_tpc_pmts'],)),
                 (("Area per channel for main S1", "s1_area_per_channel"), 
                  np.float32, (self.config['n_tpc_pmts'],)),
                 (("Area per channel for alternative S1", "alt_s1_area_per_channel"), 
                  np.float32, (self.config['n_tpc_pmts'],)),
                 ]
        dtype += strax.time_fields
        return dtype
```

straxen/plugins/event_area_per_channel.py
```diff
@@ -0,0 +1,38 @@
...
+    
+    def compute_loop(self, event, peaks):
+        result = dict()
+        result['time'], result['endtime'] = event['time'], strax.endtime(event)
+        for type_ in ['s1','s2','alt_s1','alt_s2']:
+            result[type_+'_area_per_channel'] = (peaks['area_per_channel'][event[type_+"_index"]] if event[type_+"_index"]!=-1 else 0.0)
+        return result
```
jorana, line 38:
```suggestion
    def compute_loop(self, event, peaks):
        result = np.zeros(1, self.dtype)
        result['time'] = event['time']
        result['endtime'] = strax.endtime(event)
        
        for type_ in ['s1', 's2', 'alt_s1', 'alt_s2']:
            type_index = event[f'{type_}_index']
            if type_index != -1:
                type_ara_per_channel = peaks['area_per_channel'][type_index]
                result[f'{type_}_area_per_channel'] = type_ara_per_channel
        return result
```

The 0.0 wouldn't have worked right?

straxen/plugins/event_patternfit.py
```diff
@@ -0,0 +1,410 @@
...
+
+@export
+@strax.takes_config(
+    strax.Option('s1_optical_map', help="S1 (x, y, z) optical/pattern map.",
+                 default="XENONnT_s1_xyz_patterns_LCE_corrected_qes_MCva43fa9b_wires.pkl"),
+    strax.Option('s2_optical_map', help="S2 (x, y) optical/pattern map.",
+                 default="XENONnT_s2_xy_patterns_LCE_corrected_qes_MCva43fa9b_wires.pkl"),
+    strax.Option('s1_aft_map', help="Date drive S1 area fraction top map.",
+                 default='s1_aft_dd_xyz_XENONnT_Kr83m_41500eV_25May2021.json'),
+    strax.Option('MeanPEperPhoton', help="Mean of full VUV single photon responce",
```
jorana, line 17: please try to be consistent with capital and lower case / snake_case vs. CamelCase

straxen/plugins/event_patternfit.py
```diff
@@ -0,0 +1,410 @@
...
+    strax.Option('gain_model',
+                 help='PMT gain model. Specify as (model_type, model_config)'),
+    strax.Option('n_tpc_pmts', type=int,
+                 help='Number of TPC PMTs'),
+    strax.Option('n_top_pmts', type=int,
+                 help='Number of top TPC PMTs'),
+    strax.Option('s1_min_reconstruction_area',
+                 help='Skip reconstruction if S1 area (PE) is less than this'),
+    strax.Option('s2_min_reconstruction_area',
+                 help='Skip reconstruction if S2 area (PE) is less than this'),
```
jorana, line 28: Perhaps in comments and helps, it's useful to write out LLH

straxen/plugins/event_patternfit.py
jorana, line 17:
```suggestion
    strax.Option('MeanPEperPhoton', help="Mean of full VUV single photon response",
```

straxen/plugins/event_patternfit.py
```diff
@@ -0,0 +1,410 @@
...
+    strax.Option('StorePerChannel', default=False, type=bool,
+                 help="Store normalized LLH per channel for each peak"),
+    strax.Option("max_r", default = straxen.tpc_r, type=float,
```
jorana, line 31: max_r might be a bit too generic as this could mean many different things

straxen/plugins/event_patternfit.py
```diff
@@ -0,0 +1,410 @@
...
+                       np.float32, (self.config['n_top_pmts'], )))
+            dtype.append((("Pattern for main S1", "s1_pattern"),
+                       np.float32, (self.config['n_tpc_pmts'], )))
+            dtype.append((("2LLH per channel for main S1", "s1_2llh_per_channel"),
+                       np.float32, (self.config['n_tpc_pmts'], )))
+        dtype += strax.time_fields
+        return dtype
+    
+    def setup(self):
+        self.mean_sPhoton = self.config['MeanPEperPhoton']
```
jorana, line 80:
```suggestion
        self.mean_pe_photon = self.config['MeanPEperPhoton']
```

straxen/plugins/event_patternfit.py
```diff
@@ -0,0 +1,410 @@
...
+            s1_area_fraction_top_probability(_aft, _s1, _s1_aft, mode='discrete') if not np.isnan(_aft*_s1*_s1_aft)
+            else np.nan for _aft, _s1, _s1_aft in zip(aft_prob, events['s1_area'], events['s1_area_fraction_top'])]
+        
+        result['s1_photon_fraction_top_continuos_probability'] = [
+            s1_area_fraction_top_probability(_aft, _s1/self.config['MeanPEperPhoton'], _s1_aft) if not np.isnan(_aft*_s1*_s1_aft) 
+            else np.nan for _aft, _s1, _s1_aft in zip(aft_prob, events['s1_area'], events['s1_area_fraction_top'])]
+        
+        result['s1_photon_fraction_top_discrete_probability'] = [
+            s1_area_fraction_top_probability(_aft, _s1/self.config['MeanPEperPhoton'], _s1_aft, mode='discrete') if not np.isnan(_aft*_s1*_s1_aft)
+            else np.nan for _aft, _s1, _s1_aft in zip(aft_prob, events['s1_area'], events['s1_area_fraction_top'])]
```
jorana, line 128: Sorry I cannot follow this, can we break it down?

straxen/plugins/event_patternfit.py
```diff
@@ -0,0 +1,410 @@
...
+            if self.config['StorePerChannel']:
+                result[t_+'_pattern'][:] = 0.0
+                result[t_+'_2llh_per_channel'][:] = 0.0
+                store_patterns = np.zeros((s2_pattern.shape[0], self.config['n_top_pmts']) )
+                store_patterns[:, self.pmtbool_top] = s2_pattern
+                result[t_+'_pattern'][cur_s2_bool] = store_patterns#:s2_pattern[cur_s2_bool]
+                
+                store_2LLH_ch = np.zeros((norm_llh_val.shape[0], self.config['n_top_pmts']) )
+                store_2LLH_ch[:, self.pmtbool_top] = norm_llh_val
+                result[t_+'_2llh_per_channel'][cur_s2_bool] = store_2LLH_ch
```
jorana, line 245: Would it possible to simplify this a bit and re-use the overlapping code between S1s and S2?

straxen/plugins/event_patternfit.py
```diff
@@ -0,0 +1,410 @@
...
+
+        # One side or two side
+        # binomial test
+        if k * j == 0:
+            pval = binom_sf(max(k, j), n, p)
+        else:
+            pval = binom_cdf(min(k, j), n, p) + binom_sf(max(k, j), n, p)
+        pval = min(1.0, pval)
+        
+    return pval
```
jorana, line 377: Can't we use scipy here?

straxen/plugins/event_patternfit.py
```diff
@@ -0,0 +1,410 @@
...
+
+def s1_area_fraction_top_probability(aft_prob, area_tot, area_fraction_top, mode='continuos'):
+    """
+    Wrapper that does the S1 AFT probability calculation for you
+    """
+    
+    size_top = area_tot * area_fraction_top
+    size_tot = area_tot
```
jorana, line 385:
```suggestion
```

straxen/plugins/event_area_per_channel.py
terliuk: Removed "events" from the list

straxen/plugins/event_area_per_channel.py
terliuk: It depends whether we will use it only here or in other plugins as well. My plan is to store it, as I want to use it for event level position re-reconstruction, so having it stored definitely helps. And there are plenty of studies where these areas can be extremely useful.

straxen/plugins/event_area_per_channel.py
terliuk: Copy-paste, i just removed it to use defaults. I think it should work in parallel mode out of the box, but I have not tested it.

straxen/plugins/event_area_per_channel.py
terliuk: I renamed it to EventAreaPerChannel

straxen/plugins/event_area_per_channel.py
```diff
@@ -0,0 +1,38 @@
+import numpy as np
```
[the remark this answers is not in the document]
terliuk: I think we need it as np is used inside this file to define storage data types. (np.float32), right?

straxen/plugins/event_area_per_channel.py
jorana: Perhaps it's good to know what numbers we are talking about, what is the typical size / computation time of this datatype?

The fact that it's used elsewhere too is not a problem, in fact, `Peaks` as also `strax.SaveWhen.NEVER` and that is used on many different places.

straxen/plugins/event_area_per_channel.py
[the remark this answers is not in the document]
jorana: Sorry this comment wasn't about this line but about `straxen/plugins/event_area_per_channel.py`-file (this just happened to be the first line)

straxen/plugins/event_patternfit.py
terliuk: Renamed it as suggested

straxen/plugins/event_patternfit.py
terliuk: I am not entirely sure whether making generalized version will make it much easier, given different shapes (only top, vs total vs bottom arrays) etc. And it cuts off the possibility of treating S1 and S2 differently, if we decide so. Probably it is easier to leave it as it is.

straxen/plugins/event_area_per_channel.py
terliuk: For background data that I tested, it is an order of 40-50 MB per run of 30 minutes, so it should not be too much space. As for peaks - the amount of data that has to be loaded and then not used is very different when one compares peaks vs peaklets and event areas vs peak(lets).

straxen/plugins/event_area_per_channel.py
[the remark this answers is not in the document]
terliuk: Sorry for misunderstanding. We can put it there, but then our source files grow really large as we develop. My personal preference is to keep fewer classes/modules in one .py file and make sub-folder structure, like "event_processing" or similar.

terliuk:
>     * Do we need to register it in the context or do we not need this in the main pipeline
Eventually yes, but see my comment below about some runs. 

>     * Do we have any feeling for the performance of this plugin? I remember in 1T this used to be terribly slow, it would be very nice if this would be no longer the case and optimize otherwise?

Here, we do not do any fits, but only getting the likelihood for already reconstructed position. For my basic benchmarks, it takes about a minute to compute areas and then ~20 seconds to get all the values for events per 30 min bkg data run. Actual likelihood fitter will follow and i expect it to be slower, based on my previous studies. 
 
>     * Can we use scipy for some of the LLH computations?
I need to see whether something like "stretched" Poisson probability is available somewhere. 

>     * This relates to point 1 a bit, but currently, none of this code is tested. Registering the plugins to the context would already help a bit but perhaps some dedicated tests are needed/desirable.
I think it makes sense to register it by default. However, i had a problem with one of the runs, 019612, which creates (or it looks like that) a spike in RAM usage killing the notebook, and i see nothing special about this run, so I am not sure what is the issue.

straxen/plugins/event_patternfit.py
GiovanniVolta: Hi Joran, yes I try to simplify it. Let me tested before push it

straxen/plugins/event_patternfit.py
GiovanniVolta: You mean the [scipy.stats.binom_test](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.binom_test.html) ? No we can't. The package in scipy is defined such that is valid for integer number. Here we need the continuos version, like in [XENON1T](https://github.com/XENON1T/pax/blob/master/pax/plugins/interaction_processing/S1AreaFractionTopProbability.py)

straxen/plugins/event_patternfit.py
GiovanniVolta: You are right. Let me change it

straxen/plugins/event_area_per_channel.py
GiovanniVolta: Right! Like this though, strax/straxen complains:
`AttributeError: Please provide result in compute loop as dict`
I have modified with the last push

GiovanniVolta:
Hi All, I have tried to make the binomial test faster. Unfortunately, the loop for finding _the value j on the other side of the mean that has the same probability as k_ takes time. Although with the vectorization we gain something. On the other hand, I have been asked to include the binomial test for the alternative s. Hence, what we gained we lost.

For background runs it takes between 30 s and 1 min. For krypton runs it need around 2 min.

An issue, for which I don't have an answer yet, is for 2 background runs: 019612 and 019614. The plugin `event_area_per_channel` failed with the following error:
`ValueError: Cannot merge chunks with different number of items: [[019614.event_area_per_channel: 1621319289sec 0 ns - 1621319374sec 899344920 ns, 553 items, 0.1 MB/s], [019614.event_info: 1621319289sec 0 ns - 1621319374sec 899344920 ns, 414 items, 0.0 MB/s]]`

GiovanniVolta:
Thanks, @jorana, for the review. See my reply in line.
>     * Do we need to register it in the context or do we not need this in the main pipeline?
My opinion is that we need these variables in the main pipeline since several data quality cuts are based on them. Especially the s1 binomial test

>     * Do we have any feeling for the performance of this plugin? I remember in 1T this used to be terribly slow, it would be very nice if this would be no longer the case and optimize otherwise?
I can compare to XENON1T but to me seems not incredible slow. In average, it needs ~30 sec for background runs and ~1 min for krypton calibration runs.

>     * Can we use scipy for some of the LLH computations?
I would leave @terliuk replies to this since it knows better the algorithms.

>     * I think you need to add some defaults for the S1/S1 min reconstruction area
With the last push I have set the default `s1_min_reconstruction_area` to 3 PE which sound reasonable given the PMT coincidence requirement.

straxen/plugins/event_patternfit.py
terliuk: changed to `max_r_patternfit`, probably it is a bit more clear this way.

straxen/plugins/event_patternfit.py
terliuk: changed everything to lower case

GiovanniVolta:
Hi @skazama, do you have comments/suggestion for Andrii and me? Is there anything else we can do?

jorana:
Hi Giovanne and Andrii,
Thanks, for the changes, are there any review comments pending?

- @GiovanniVolta https://github.com/XENONnT/straxen/pull/499#issuecomment-852325833 sounds quite serious, does it persists? If so we need to fix it.
- In c75feb8 , I've registered the plugins, lets see if the tests still pass after them
- Thanks for checking the performance!
- Minor but I guess you mean `continuous` instead of `continuos`

straxen/plugins/event_area_per_channel.py
jorana: thanks!

straxen/plugins/event_patternfit.py
jorana: Thanks! Perhaps it's worth considering adding to strax.processing as it's so general.

straxen/plugins/event_patternfit.py
```diff
@@ -0,0 +1,435 @@
...
+        positions = np.vstack([events['x'], events['y'], events['z']]).T
+        aft_prob = self.s1_aft_map(positions)
+        
+        # main s1 events
+        mask_s1 = ~np.isnan(aft_prob)
+        mask_s1 &= ~np.isnan(events['s1_area'])
+        mask_s1 &= ~np.isnan(events['s1_area_fraction_top'])
+        
+        # compute binomial test only if we have events that have valid aft prob, s1 area and s1 aft
+        if len(mask_s1):
```
GiovanniVolta, line 130: This check is to avoid input length equal to 0 for the vectorized version of _s1_area_fraction_top_probability_ function.

GiovanniVolta:
Hi @jorana, please see my reply in line.

> Hi Giovanni and Andrii,
> Thanks, for the changes, are there any review comments pending?
> 
>     * @GiovanniVolta [#499 (comment)](https://github.com/XENONnT/straxen/pull/499#issuecomment-852325833) sounds quite serious, does it persists? If so we need to fix it.

I have not been able to reproduce the error again. I have tested the plugin with several runs, using the all the calibration source and background runs. 

> 
>     * In [c75feb8](https://github.com/XENONnT/straxen/commit/c75feb80bc72f752d951cfe22f7ff60303f894d4) , I've registered the plugins, lets see if the tests still pass after them
> 
>     * Thanks for checking the performance!
> 
>     * Can we also use this plugin for xenon1T. Probably we'd need to use a 1T aft prob. map right?

Yes sure, but I don't know what I have to do for this. The map used in XENON1T is stored [here](https://github.com/XENON1T/pax/blob/master/pax/data/XENON1T_s1_aft_xyz_20170808.json)

> 
>     * Minor but I guess you mean `continuous` instead of `continuos`

Thanks!

> 
>     * Finally, the tests are failing for three reasons:
>       
>       * We cannot open this map on testing, could we upload a dummy map that has the same shape perhaps to strax_auxfiles

See [#20 PR](https://github.com/XENONnT/strax_auxiliary_files/pull/20)

>       * For 1T we are going to have a different shape, can we use the pax-files for that?

I need assistance here. I am not sure how to define the map for XENON1T context.

>       * np.vectorize is not suitable if the length of the input is 0

With the [last update](https://github.com/XENONnT/straxen/pull/499/files#r645993806), this should be avoided.

straxen/plugins/event_patternfit.py
jorana: should it not be np.sum(mask_s1)? Otherwise if all is False, it will still fail right?

GiovanniVolta: You are right, let me update it

jorana:
@l-althueser would you be able to help out using the 1T maps here or do you think that would be too complex?

jorana:
Thanks guys!

jorana (approved):

PR #503, opened 2021-06-01
Fix failing tests

jorana:
Fix failing tests

Dask is horrible

If the tests pass it works otherwise it does not


jorana:
🎉  it's done! It's finally over, what a horrible exercise again after #474 , I deeply started hating dask

jorana:
Nice great to be back

PR #532, opened 2021-06-04
Fix CMT from PR #457

jorana:
Master is failing because there were a few places missed in #457 where the old convention was still used.

Also the gains for MV were still the placeholder ones?

See #457


ahiguera-mx (approved):
Looks good to me, thanks Joran

PR #534, opened 2021-06-04
Drop gcn_model

jorana:
If GCN is not used anymore, we might as well stop processing it.

When we set the config to None, it will not do any processing (thanks to Andrii foreseeing this).

## Alternative approach
Drop the plugin all together from the context by not registering it. I'm not sure if we want to go that drastic already but @reallsx might comment.


reallsx (approved):

reallsx:
New modeling is almost ready and will use the GCN slot. It's also graph based and much faster.

jorana:
Thanks @reallsx

Issue #527, opened 2021-06-03
Follow the FDC updates


l-althueser:
**Describe the bug**
We decided that we want to remove the z-correction for the moment and change the variable naming a bit. This should change the lines around [event_processing.py#L410](https://github.com/XENONnT/straxen/blob/master/straxen/plugins/event_processing.py#L410). 

What we have:

- `z_naive` as observed z
- `z` as corrected z

What we want to have:

- `z` as observed z
- `z_naive` as observed z
- `z_field_distortion_correction` as delta_z

So that we could in future release change the `z` to one of the other parameters.

This is following https://github.com/XENONnT/private_nt_aux_files/pull/66 and https://github.com/XENONnT/corrections/pull/31

PR #535, opened 2021-06-05
z coordinate update

jingqiangye:

* Take `electron_drift_time_gate` into account when calculating `z_naive`
* Fixing issue [#527](https://github.com/XENONnT/straxen/issues/527)

A test is done in 
https://xenonnt.slack.com/archives/C020GRDUNN8/p1622855301162000. Not sure why it fails automatic checks. Is it related to the Travis shutdown?

### DIFF
--- a/straxen/contexts.py
+++ b/straxen/contexts.py
@@ -272,6 +272,7 @@ x1t_common_config = dict(
     electron_drift_velocity=("electron_drift_velocity_constant", 1.3325e-4, False),
     event_info_function='disabled',
     max_drift_length=96.9,
+    electron_drift_time_gate=("electron_drift_time_gate_constant", 1700),
 )
 
 
--- a/straxen/corrections_services.py
+++ b/straxen/corrections_services.py
@@ -16,7 +16,7 @@ corrections_w_file = ['mlp_model', 'gcn_model', 'cnn_model',
                       'fdc_map_cnn']
 
 single_value_corrections = ['elife', 'baseline_samples_nv',
-                            'electron_drift_velocity']
+                            'electron_drift_velocity', 'electron_drift_time_gate']
 
 
 @export
--- a/straxen/plugins/event_processing.py
+++ b/straxen/plugins/event_processing.py
@@ -328,6 +328,11 @@ class EventBasics(strax.LoopPlugin):
         default=("electron_drift_velocity", "ONLINE", True)
     ),
     strax.Option(
+        name='electron_drift_time_gate',
+        help='Electron drift time from the gate in ns',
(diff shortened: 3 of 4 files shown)


l-althueser:
We need to change the description for the z variable. Otherwise it looks good :)

jorana:
Hi @jingqiangye , thanks!
- The tests fail because some tests (py3.6) do not have access to CMT database, if we can upload a placeholder dummy value [here](https://github.com/XENONnT/straxen/blob/gate_drift_time/tests/test_plugins.py#L27), they should pass.
- The tests with database access (coveralls, py3.8 etc) fail because this implementation did not take into account xenon1t. So perhaps its good if we upload the 1T value [similar as we do for the drift velocitiy](https://github.com/XENONnT/straxen/blob/gate_drift_time/straxen/contexts.py#L270)

jingqiangye:
> Hi @jingqiangye , thanks!
> 
> * The tests fail because some tests (py3.6) do not have access to CMT database, if we can upload a placeholder dummy value [here](https://github.com/XENONnT/straxen/blob/gate_drift_time/tests/test_plugins.py#L27), they should pass.
> * The tests with database access (coveralls, py3.8 etc) fail because this implementation did not take into account xenon1t. So perhaps its good if we upload the 1T value [similar as we do for the drift velocitiy](https://github.com/XENONnT/straxen/blob/gate_drift_time/straxen/contexts.py#L270)

Thanks for the explanation and suggestion! I added the placeholder for `electron_drift_time_gate` in 1T and nT config, but it still fails the check. Does one need to add dummy values for many other parameters? (1T config has much fewer params) Or maybe I overlooked smth else?

straxen/contexts.py
```diff
@@ -272,6 +272,7 @@ def xenonnt_simulation(output_folder='./strax_data'):
     electron_drift_velocity=("electron_drift_velocity_constant", 1.3325e-4, False),
     event_info_function='disabled',
     max_drift_length=96.9,
+    electron_drift_time_gate=("electron_drift_time_gate_constant", 1700),
```
jorana, line 275: @jingqiangye do you happen to know the correct value for 1T?

jingqiangye: Yeah I think it is the `electron_drift_time_gate_constant` for 1T, see [Tab. 1](https://xe1t-wiki.lngs.infn.it/doku.php?id=dandrea:diffusionstudywithkrdata)

ershockley (approved):
@jingqiangye @jorana looks good. Can we merge this guy?

jingqiangye:
@ershockley Yeah I think so

jorana:
Looks good 😄

PR #541, opened 2021-06-08
Fixing saturation correction bugs

zhut19:
 - This fixes the bugs in peak saturation correction. Especially now that the baseline can not be directly compared with the samples in records, instead, it should be converted to int16 first.
 - Additionally, we now compare the buffer pulse to the smallest baseline value of all the records of the channel touching the peaklet. To deal with cases when the saturated records having a different baseline value than that of the first record.
 - Last but not least, correct the condition for skipping the correction if there's not enough non-saturated channel to build the pulse model.


straxen/plugins/peaklet_processing.py
```diff
@@ -352,7 +364,7 @@ def _peak_saturation_correction_inner(channel_saturated, records, p,
         if (b_sumwf[ref] > 1).sum() < min_reference_length:
             # the same condition applies to the waveform model
             continue
-        if np.sum(b[ref]) * to_pe[ch] / np.sum(b_sumwf[ref]) < 1:
+        if np.sum(b[ref]) * to_pe[ch] / np.sum(b_sumwf[ref]) > 1:
```
skazama, line 367: Was this simply a bug and just corrected now?

skazama:
Hi @zhut19, thanks for the PR! I just put here the link to your [note](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenon1t:tzhu:straxensatcorrectionfix) and have one comment only;

> Additionally, we now compare the buffer pulse to the smallest baseline value of all the records of the channel touching the peaklet. To deal with cases when the saturated records having a different baseline value than that of the first record.

Do you have some examples of how big the baseline difference is? Is this problem (=baseline difference) only occurring for saturated signals or it could happen also for large S2 signals?

straxen/plugins/peaklet_processing.py
zhut19: I think this really is a bug

straxen/plugins/peaklet_processing.py
```diff
@@ -168,6 +168,12 @@ def compute(self, records, start, end):
         # Saturation correction using non-saturated channels
         # similar method used in pax
         # see https://github.com/XENON1T/pax/pull/712
+        # Cases when records is not writeable for unclear reason
+        # only see this when loading 1T test data
+        # more details on https://numpy.org/doc/stable/reference/generated/numpy.ndarray.flags.html
+        if not records['data'].flags.writeable:
+            r = records.copy()
```
jorana, line 175: is it the same with `r` instead of `records`? Perhaps a bit clearer if we use `r`?

also don't know what the issue is but I have seen such behavior before too

jorana (approved):

zhut19:
> Do you have some examples of how big the baseline difference is? Is this problem (=baseline difference) only occurring for saturated signals or it could happen also for large S2 signals?

It happens to 20% of the population, I think it's normal given that deep S2s peaks can actually be built from multiple pulses in a channel. See plot [here](https://xe1t-wiki.lngs.infn.it/lib/exe/detail.php?id=xenon%3Axenon1t%3Atzhu%3Astraxensatcorrectionfix&media=xenon:xenon1t:tzhu:bl_difference.png)

straxen/plugins/peaklet_processing.py
zhut19: I think yes, here r and records are referencing the same thing, as `r` is here probably just to make the code more succinct

skazama:
> > Do you have some examples of how big the baseline difference is? Is this problem (=baseline difference) only occurring for saturated signals or it could happen also for large S2 signals?
> 
> It happens to 20% of the population, I think it's normal given that deep S2s peaks can actually be built from multiple pulses in a channel. See plot [here](https://xe1t-wiki.lngs.infn.it/lib/exe/detail.php?id=xenon%3Axenon1t%3Atzhu%3Astraxensatcorrectionfix&media=xenon:xenon1t:tzhu:bl_difference.png)

ok thanks for the plot! Maybe we have to think about it for large S2s too, but the way how you defined sounds fine for saturated S2s.

skazama (approved):

straxen/plugins/peaklet_processing.py
zhut19: Sure, I can change it to r. It just `r=r.copy()` looks unsettling to me, but I did some quick memory profiling, this copying line doesn't even use any new memory. Really have no idea what it's doing internally.

PR #542, opened 2021-06-09
Compute width again after saturation correction

zhut19:
The peak width is now being re-computed after the saturation correction. Those peaks previously computed with smaller widths.

This will need https://github.com/AxFoundation/strax/pull/465 merged first.


skazama:
Thanks Tianyu for the PR. It looks fine once the [AxFoundation/strax#465](https://github.com/AxFoundation/strax/pull/465) is merged

straxen/plugins/peaklet_processing.py
```diff
@@ -316,6 +319,7 @@ def peak_saturation_correction(records, peaks, to_pe,
         peaks[peak_i]['dt'] = dt
 
     strax.sum_waveform(peaks, records, to_pe, peak_list)
+    return peak_list
```
jorana, line 322: maybe makes sense to just do `strax.compute_widths(peaklets, select_peaks_indicies=peak_list)` here together with the sum_wf, also alleviates the need to return this variable?

jorana: @zhut19 , did you see my suggestion?

zhut19: Oh, sorry, I missed this. I couldn't do this because that `strax.compute_width` is not a numba function.

WenzDaniel: You could use a numba.typed.List, but then you should check the impact on the performance.

jorana (approved):

zhut19:
@jorana Let's leave it like this now.

Issue #478, opened 2021-05-07
Default corrections (for full-chain simulation)


ramirezdiego:
At the moment, these lines:
https://github.com/XENONnT/straxen/blob/3913577586b374cd26a3c8bd53787094cf38a7ab/straxen/contexts.py#L53-L57
together with these:
https://github.com/XENONnT/straxen/blob/3913577586b374cd26a3c8bd53787094cf38a7ab/straxen/contexts.py#L192-L203
aim at providing with a simulations context that doesn't rely on CMT defaults.

These only correct for electron lifetime and gain model defaults, therefore a full-chain run beyond `event_basics` would still fetch the straxen defaults for each plugin. In case we want to point CMT to a particular run for our corrections, we have now the tools to do it (see https://github.com/XENONnT/straxen/pull/443 and [these lines](https://github.com/XENONnT/mc_chain/blob/82cb5300f24ec9685a7b420233454b4be18c8d65/run_wfsim.py#L113-L126)).

But as a default we (probably) don't want to use CMT at all and apply custom correction instead. For this we/I should update `xnt_simulation_config` to the point in which we define a non-CMT default for each single correction. Thoughts?

jorana:
Hi Diego, 
So breaking down your proposal, you would:
1. Remove all options where CMT is the defaults and set it to some "reasonable number"
2. Combine all the CMT options into one immutable, which we use for xenonnt_online but not for simulation.

This would be slightly different from what we are currently doing (as much as possible):
- Set the nT defaults as much as possible at the plugin-option level.
- Set 1T/simulations options, different from nT in the context using the appropriate dict.

In principle, I'm fine with your proposal if I understood it right. In fact, there might be two additional advantages of this: 
1.  Make the [testing config used here](https://github.com/XENONnT/straxen/blob/master/tests/test_plugins.py#L17-L36) a bit simpler. Now, for each option that would be pulled from the database (e.g. CMT), we need to specify some offline counterpart. Of course, such a change would not affect e.g. neural nets that are only available in our database.
2. Make the way for a CMT global version v1-context slightly simpler (CMT guys would call this "an offline context"). One would just pull the global version dict (mentioned at 2. above) from CMT instead of using the one as written down in the context.py file for the online-context.

There are two downsides however. There would be a political debate of what numbers to put then as defaults for nT (simulation) which we were avoiding using CMT. Additionally, we were trying to keep the nT defaults next to the plugins as much as possible for readability of the code (was not always possible).

[...]

PR #544, opened 2021-06-10
Download contexts from the database

ershockley:
This PR adds a new function `straxen.contexts.load_context`. This function uses utilix to download additional context files from the database, so that we can define production contexts outside of the straxen installation. This might be useful if e.g. we need to update corrections but we don't want to update to a new straxen version. It also will help resolve some issues we've had in the MC group about context definitions. 

It takes 2 arguments: `context_name` and `file`. The `context_name` is the name of the context function that will be loaded, and `file` is the name of the file in the files database. The default filename is `xenonnt.py`, a version which I have already uploaded. 

First, the `file` is downloaded using utilix, which requires a utilix config file with the proper credentials. This file is downloaded to your current directory so that it is automatically in your python path. We then use `eval`, `exec` and `getattr` to import the `context_name` from the file, and initialize it. You can pass normal context kwargs to `load_context` which then get propagated to the initialization of the context. The initialized context is returned. 
 

I have uploaded a file which is essentially a copy of contexts.py, but with xenonnt_online renamed to just `xenonnt`. You can load it as follows
```
st = straxen.contexts.load_context('xenonnt')
print(st.provided_dtypes()['peaklets']['hash'])

# returns 2as4vw56ce
```

### DIFF
--- a/straxen/contexts.py
+++ b/straxen/contexts.py
@@ -3,6 +3,25 @@ import strax
 import straxen
 from copy import deepcopy
 
+# initialize runDB API if utilix is configured.
+if straxen.utilix_is_configured():
+    from utilix.mongo_files import APIDownloader
+    downloader = APIDownloader(store_files_at='./')
+
+
+def load_context(context_name, file='xenonnt.py', **kwargs):
+    """Downloads a context file from the files database, imports it, and initializes it.
+    :param context_name: name of context to be imported
+    :param file: the name of the file that will be downloaded."""
+    # download file
+    assert file.endswith('.py'), f"The contexts file must have a .py suffix. You passed {file}"
+    path = downloader.download_single(file, human_readable_file_name=True)
+    module = file.split('.')[0]
+    exec(f'import {module}')
+    st = getattr(eval(module), context_name)(**kwargs)
+    return st
+
+
 common_opts = dict(
     register_all=[
         straxen.event_processing,
--- a/straxen/misc.py
+++ b/straxen/misc.py
@@ -73,7 +73,7 @@ def print_versions(modules=('strax', 'straxen'), return_string=False):
 
 
 @export
-def utilix_is_configured(header='RunDB', section='pymongo_database') -> bool:
+def utilix_is_configured(header='RunDB', section='xent_database') -> bool:
     """
     Check if we have the right connection to
     :return: bool, can we connect to the Mongo database?


ershockley:
I should add that I created this branch off of the cmt branch, so I would prefer that this PR get merged before #531, if everyone agrees.

straxen/contexts.py
```diff
@@ -3,6 +3,25 @@
...
+def load_context(context_name, file='xenonnt.py', **kwargs):
+    """Downloads a context file from the files database, imports it, and initializes it.
+    :param context_name: name of context to be imported
+    :param file: the name of the file that will be downloaded."""
+    # download file
+    assert file.endswith('.py'), f"The contexts file must have a .py suffix. You passed {file}"
+    path = downloader.download_single(file, human_readable_file_name=True)
+    module = file.split('.')[0]
+    exec(f'import {module}')
+    st = getattr(eval(module), context_name)(**kwargs)
```
WenzDaniel, line 21: These lines look a bit hacky to me and I think one should try to avoid eval whenever possible. 
Would not for your purpose something like the following work? 
```
import importlib
spec = importlib.util.spec_from_file_location("straxen_context", 
                                              "/home/dwenz/mymodules/straxen/straxen/contexts.py")
context = importlib.util.module_from_spec(spec)
print(context)
spec.loader.exec_module(context)
straxen.contexts = context.xenonnt_online()
```
credits to https://stackoverflow.com/questions/67631/how-to-import-a-module-given-the-full-path

WenzDaniel (commented):
Hej Evan thanks a lot. I have a few questions though, please correct me if I am wrong.

1. Could we make the path a parameter such that users can specify its location? Could the default location be our resource_cache? I see already now people having thousands of those files lying around in different directories. This will become a mess over time. 
2. Is there any version control of the file? If I understand you correctly one of the goals should be that we can exchange our context in an easier way if only some settings change. However, if I understand [these lines](https://github.com/XENONnT/utilix/blob/9743a731ad951ac4d0c2e3441fb3fbf6e8d1b9e8/utilix/mongo_files.py#L535-L539) correctly your downloader will keep the old version of the context file if it already exists.

jorana (changes requested):
As you know, I'm not a particularly big fan of this proposal 😉. I foresee the following (potential) problems:
- Updating a "database" contexts will be a mess due to the human-readable format, if there is an update to a given context at any given time, it would not be reflected in the file that someone once downloaded. Who knows what context you would be loading as the files stay around.
- Documentation will not exist for these contexts and the straxen documentation will be incomplete. Accessing the docstrings might not be trivial either.
- Bookkeeping is tricky as not all database contexts will be compatible with all straxen versions. If we change change where we for example store some plugin, we also need to update the database context making bookkeeping more complex. I imagine this will become very tricky very quickly if we need to keep making contexts for the next 5 years.
- There would be yet another place where configs can be set, making their implementation less transparent to average users.
- Bugs in database contexts will remain untested as CI is very ambiguous for such contexts, would we want to run them for all contexts we put in the database?

I also see some advantages:
 - There is no good way of doing an offline context for CMT currently, this would help.
 - Configuration changes can be made in the past for a given straxen version (although in my opinion, adding a function that just greps a dictionary from the database is easier to maintain and arguably more transparent).
 - Any changes in storage could be helpful although this should be easy to just pass on as a kwarg to the contexts.

What would perhaps already help is:
 - Putting the .py files in the private repo.
 - Basing any database context functions on e.g. the xenonnt_online context, and just overwrite if a change is required. My guess is that most of the motivations are going to be updates of configs (and perhaps some storage frontend) but not much more.
 - Making it very clear what is, and is not, the scope of such a database context, as most likely, you are just concerned with updating the straxen-options.

ershockley:
Based on discussion in todays computing meeting we decided to drop this in favor of a python package due to security concerns. I try to address some of the questions above just in case we want to revisit this in the future. 

@WenzDaniel, using the importlib bit you mentioned would make passing configurable paths easier, so I definitely would be in favor of that. Actually the local paths thing is something I would like to address in straxen generally at some point, as I at least have zillions of `strax_data` and `resource_cache` directories everywhere now which is a bit annoying/inefficient, and as you say this would have only added an additional one. And yes it was absolutely hacky :) 

For version control, I envisioned the following:
- We have a repo (xenon_contexts) that is of course under version control. The files themselves would be versioned as well. 
- When we make a new release of the contexts repo, we automatically push the files to the database using Actions. The files database already has version control implemented, and it would be relatively straightforward (but a small change) to include the version of the repo/file in the database as well for better querying.
- As a bonus, we could also update the contexts collection in this new repo instead of straxen; this would allow us to track hashes of data with various CMT versions better which is going to be very important down the road. 
- We would need to implement the actual download side to include the version control and update the current file if needed. This is already essentially implemented in the MongoFiles classes Joran added but I didn't include here due to the quick/hacky nature. 


@jorana 

> Updating a "database" contexts will be a mess due to the human-readable format, if there is an update to a given context at any given time, it would not be reflected in the file that someone once downloaded. Who knows what context you would be loading as the files stay around.

The context would be defined in a repo which is where people would be able to check things. I don't see it as being all that different from how xenonnt_online is defined now; people need to check the source code itself to see how the context is setup. For files staying around, we could consider deleting the file immediately after importing it, or we implement the version control that Daniel mentioned above.

> Documentation will not exist for these contexts and the straxen documentation will be incomplete. Accessing the docstrings might not be trivial either.

The documentation would be in the other repo, probably just a README that summarizes what the different contexts are and what they should be used for. I don't think analysts need much more beyond that. 

> Bookkeeping is tricky as not all database contexts will be compatible with all straxen versions. If we change change where we for example store some plugin, we also need to update the database context making bookkeeping more complex. I imagine this will become very tricky very quickly if we need to keep making contexts for the next 5 years.

The strax(en) version compatibility is definitely a concern with doing it this way, I agree with that. We could maybe specify in the contexts themselves what strax(en) versions they are compatible with, as we did in the past with lax/hax versioning. 

However I think the general data/context bookkeeping would be much improved if we did something like this, as we can then track hashes of different production contexts in a much better way. We can also name contexts more precisely which was always the plan from when strax was first discussed but we have not really followed through with that. Right now we only update the contexts collection for xenonnt_online, xenonnt_led, and xenonnt_simulation, but that will surely not be the only contexts we use for production. We absolutely 100% must track hashes when we e.g. use a different CMT version, etc. and it needs to be done in a careful/automatic way.  

>There would be yet another place where configs can be set, making their implementation less transparent to average users.
Bugs in database contexts will remain untested as CI is very ambiguous for such contexts, would we want to run them for all contexts we put in the database?
Again I don't think this is really all that different from how it currently is -- the context would be easy to see by looking at the xenon_contexts repo where the files are (or by checking the downloaded file directly). For the xenonnt_online context you need to check similar things, or do st.show_config or whatever it is, which also would work here of course.

For CI we could add it to the xenon_contexts repo, though to be honest I'm not too worried about it. Such contexts would be thoroughly tested by analysts before being put into production anyway. 

For the benefits you mention: 

>There is no good way of doing an offline context for CMT currently, this would help.
Yes I think this or something like it is the best way to handle offline contexts. It also would help enormously with the simulation contexts as stressed by @ramirezdiego several times over the last months and discussed in #478. 

>Configuration changes can be made in the past for a given straxen version (although in my opinion, adding a function that just greps a dictionary from the database is easier to maintain and arguably more transparent).
This could be an option as well, but I think it is much less transparent than just writing out the python code directly. 

>Any changes in storage could be helpful although this should be easy to just pass on as a kwarg to the contexts.
Yes from the data processing side I think it would be amazing to define an OSG context that inherits from the production contexts but sets up the correct storage frontends etc. Right now I have to do it manually in several places which is a bit of a headache. 

I totally agree with your suggestions -- I think I wasn't clear before about the other repo that holds the files. That is some key info. 

Anyway I will close this for now, though I welcome more comments if you have any. Thanks @WenzDaniel and @jorana for the fast feedback!

PR #547, opened 2021-06-10
Fix example command

mzks:
The example command didn't work on dali.
I fixed it.
Excuse me for very tiny PR, but I believe it saves analyzer's time.


WenzDaniel (approved):
Thanks a lot.

Issue #526, opened 2021-06-03
MergedS2s max-duration leads to many S2s


jorana:
Perhaps we need to rethink if this line makes sense:
https://github.com/XENONnT/straxen/blob/a92dcbec6d8b76654bc5d15723a955bf54d5a505/straxen/plugins/peaklet_processing.py#L574

Since it prevents merges happening at all if a S2-peak cluster is longer than the max length. Or, we should increase the max length by a huge ammount.

jorana:
Fixed in #548

PR #548, opened 2021-06-12
Rewrite merge s2

zhut19:
  - This would solve #526, the idea is to loop over gaps from small ones to large ones and would merge as many as possible.
  - Now we can put a varying gap threshold for merging based on the area of the merged peak as suggested based on [the plot here](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenon1t:tzhu:straxencalibrateclustering#current_status).

Since the idea centers around looping over gaps instead of time, it's important to keep track of the relation of the gaps. We do this via two lists, one marking the previous gap that's not yet merged, and one marking the next gap that's not yet merged. As show below
```
0 1 2 3 4 5 6 7 8 9  gap i

- 0 1 2 3 4 5 6 7 8  previous
1 2 3 4 5 6 7 8 9 10 next
              x
- 0 1 2 3 4 5 . 6 8  previous
1 2 3 4 5 6 8 . 9 10 next
      x
- 0 1 . 2 4 5 . 6 8  previous
1 2 4 . 5 6 8 . 9 10 next
          x
- 0 1 . 2 . 4 . 6 8  previous
1 2 4 . 6 . 8 . 9 10 next
        x
- 0 1 . . . 2 . 6 8  previous
1 2 6 . . . 8 . 9 10 next
  x
- . 0 . . . 2 . 6 8  previous
2 . 6 . . . 8 . 9 10 next
x
. . 0 . . . 2 . 6 8  previous
. . 6 . . . 8 . 9 10 next
```


straxen/plugins/peaklet_processing.py
```diff
@@ -557,35 +558,37 @@ class MergedS2s(strax.OverlapWindowPlugin):
...
                     + self.config['s2_merge_max_duration'])
 
     def compute(self, peaklets):
-        if not len(peaklets):
+        if len(peaklets) <= 1:
             return peaklets[:0]
 
-        if self.config['s2_merge_max_gap'] < 0:
+        gap_thresholds = self.config['s2_merge_gap_thresholds']
+        if gap_thresholds[0][1] < 0:
```
jorana, line 575: perhaps we should also add this to the help of the option

straxen/plugins/peaklet_processing.py
```diff
@@ -557,35 +558,37 @@ class MergedS2s(strax.OverlapWindowPlugin):
...
         return strax.unpack_dtype(self.deps['peaklets'].dtype_for('peaklets'))
 
     def get_window_size(self):
-        return 5 * (self.config['s2_merge_max_gap']
+        return 5 * (int(self.config['s2_merge_gap_thresholds'][0][1])
                     + self.config['s2_merge_max_duration'])
 
     def compute(self, peaklets):
-        if not len(peaklets):
+        if len(peaklets) <= 1:
```
jorana, line 571: out of curiosity, for 1 peaklet, we don't have to return a mergedS2 right as the data is from the peaklets already?

straxen/plugins/peaklet_processing.py
```diff
@@ -601,47 +604,80 @@ def compute(self, peaklets):
...
+
+        # Check gaps from small to large
+        for gap_i in np.argsort(peaklet_gaps):
+            left_i = previous_valid_gap[gap_i] + 1
+            # the right is inclusive
+            right_i = next_valid_gap[gap_i]
+
+            sum_area = areas[left_i:right_i+1].sum()
+
+            if peaklet_gaps[gap_i] > max_gap:
```
jorana, line 629: smart iteration 😄

straxen/plugins/peaklet_processing.py
```diff
@@ -601,47 +604,80 @@ def compute(self, peaklets):
...
-            start_merge_at[n_to_merge] = left_i
-            end_merge_at[n_to_merge] = right_i
-            n_to_merge += 1
+        i_gaps_to_merge = np.where(gaps_to_merge)[0]
+        start_merge_at = i_gaps_to_merge[~gaps_to_merge[i_gaps_to_merge - 1]] - 1
+        # end_merge_at is now exclusive
+        end_merge_at = i_gaps_to_merge[~gaps_to_merge[i_gaps_to_merge + 1]] + 1
+
+        # Doesn't start with S2: do not merge
+        start_merge_with_s2 = types[start_merge_at] == 2
```
jorana, line 660: I hope I understood this right. But, would it make sense to do this earlier, because now, if this is true, nothing is merged at all?

straxen/plugins/peaklet_processing.py
```diff
@@ -601,47 +604,80 @@ def compute(self, peaklets):
...
-            cluster_stop = cluster_stops[cluster_i]
+            gap_thresholds, max_duration, max_gap, max_area):
+        """Finding the group of peaklets to merge
+        returns: list of the first index of peaklet to be merged and
+        list of the exclusive last index of peaklet to be merged
+        """
+        peaklet_gaps = peaklet_starts[1:] - peaklet_ends[:-1]
+        n_gaps = len(peaklet_gaps)
+
+        previous_valid_gap = np.arange(n_gaps) - 1
```
jorana, line 615: Thanks Tianyu, I really had to make a sketch to keep track of all the shifts by +1 or -1 and go over it. I think it looks good but it's a bit hard to fully keep track of all those indices.

straxen/plugins/peaklet_processing.py
```diff
@@ -601,47 +604,80 @@ def compute(self, peaklets):
...
-            max_duration, max_area):
-        start_merge_at = np.zeros(len(cluster_starts), dtype=np.int32)
-        end_merge_at = np.zeros(len(cluster_starts), dtype=np.int32)
-        n_to_merge = 0
-        left_i = 0
-
-        for cluster_i, cluster_start in enumerate(cluster_starts):
-            cluster_stop = cluster_stops[cluster_i]
+            gap_thresholds, max_duration, max_gap, max_area):
+        """Finding the group of peaklets to merge
```
jorana, line 608:
```suggestion
        """Finding the group of peaklets to merge. To do this start with the 
        smallest gaps and keep merging until the new, merged S2 has such a 
        large area or gap to adjacent peaks that merging is not required 
        anymore.
```

jorana (commented):
Thanks Tianyu, it took me some time to understand but I think the idea is brilliant.

Based on simulation, it looks very good, some data-cross checks would be good to validate too.

The code is a bit tarse with all the indices flowing around, an off by one error could easily be made. I'm not sure if there are places we can simplify to prevent this, what do you think?

straxen/plugins/peaklet_processing.py
zhut19: Yes, I have a sketch that I should have put in the pr first. But not sure how much commenting should be in the code, or just explain this in the pr is enough.

straxen/plugins/peaklet_processing.py
zhut19: Right, I don't see an obvious place to do this during the loop as the whole loop is by size so the time info is abandoned.
One thing we can do is loop over the non-s2 merge starts and find the first that's an S2.

straxen/plugins/peaklet_processing.py
zhut19: Right, not sure if there's a difference between return empty list and None.

jorana: Giving the instruction to overwrite a single peak with the same single peak is not really needed unless I misunderstand

straxen/plugins/peaklet_processing.py
jorana: Yes, I also think that would be basically the only way to solve it if this is really a problem. Did you see any such worrisome topology in the data?

straxen/plugins/peaklet_processing.py
```diff
@@ -535,20 +535,23 @@ class PeakletClassificationHighEnergy(PeakletClassification):
...
 @strax.takes_config(
-    strax.Option('s2_merge_max_area', default=5000.,
-                 help="Merge peaklet cluster only if area < this [PE]"),
-    strax.Option('s2_merge_max_gap', default=10_000,
-                 help="Maximum separation between peaklets to allow merging [ns]"),
     strax.Option('s2_merge_max_duration', default=50_000,
                  help="Do not merge peaklets at all if the result would be a peak "
-                      "longer than this [ns]"))
+                      "longer than this [ns]"),
+    strax.Option('s2_merge_gap_thresholds', default=((1.7, 2.65e4), (4.0, 2.6e3), (5.0, 0.)),
```
skazama, line 548: Hi Tianyu, thanks for the nice PR. I think it's good to merge this for the next reprocessing, but can you explain how you decided these values? From the plots [here](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenon1t:tzhu:straxencalibrateclustering#current_status), it looks bit different.

straxen/plugins/peaklet_processing.py
zhut19: Now I do see some strange topology that might be like this (starting with an unknown). This will increase the n_competing values around it, which is the other thing we need to be a little careful about with this update (or any clustering updates), see [the second strange topology example](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenon1t:tzhu:straxenmerges2s#events_merged_split_lost_and_found)

straxen/plugins/peaklet_processing.py
zhut19: Hi Shingo, I did redo [that plot](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenon1t:tzhu:straxenmerges2s#merge_thresholds) with simulation, and think that we can tolerate 1% loss of some long and small S2 peaks near the cathode.

skazama: Thanks Tianyu for the updated note. It looks really promising!! I agree that 1% loss should be fine (around 3000 pe, it might be ~5% loss?).
I think you are still investigating, but it's interesting to see where the population " XX events found in default is now missing" went to.

skazama: As you directly explained to me, this requirement is conservative considering that parameters are estimated using cathode events. Then, I'm fine with these numbers

jorana:
@zhut19 , I retested it for simulation https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:j_angevaare:cluster_and_classification_finetuning#pre-release_performance_and_2_fold_example and the result still looks stellar. Thanks!

PR #549, opened 2021-06-14
Plugin for afterpulse processing

hoetzsch:
This PR adds the afterpulse_processing plugin used to process LED afterpulse calibration data. 
Added file: [straxen/plugins/afterpulse_processing.py](https://github.com/XENONnT/straxen/blob/ae162d4da034e283766dec53e4608444901d79ec/straxen/plugins/afterpulse_processing.py)
Edited file: [straxen/contexts.py](https://github.com/XENONnT/straxen/blob/ae162d4da034e283766dec53e4608444901d79ec/straxen/contexts.py#L139) (register plugin in xenonnt_online context)

The plugin is based on the [pulse_processing plugin](straxen/plugins/pulse_processing.py) and the [strax function find_hits](https://github.com/AxFoundation/strax/blob/1d7b86a493d6ed1bc3d4137e8e684f0443c5dc4f/strax/processing/pulse_processing.py), with the addition of a predefined LED window in which the LED hits are expected. A field 'tdelay' is added that gives the delay time (in samples) of each hit in the same WF w.r.t. the LED hit. Further additional fields give the height in PE for all hits (allows the application of a height cut for noise reduction), and several hit timing parameters (e.g. 10 and 50 percent area quantiles) for hits on a per-channel basis.

Example: Histogram of all hits in (tdelay, area_pe) for channel 131. All LED hits lie at tdelay=0. Theoretical (calculated) afterpulse delay times are indicated by red lines.

apspectrum_pmt131_20210610-2


WenzDaniel (commented):
Hej Luisa I have already some minor comments. But before I review your hit finding method I have two questions: 
1. Is the LED calibration data divided into fragments? E.g. your plot lasts for 700ish samples. Are those 700ish samples long pulses divided into fragments of a length of 110 samples each? 
2. I have to admit I am not a friend of your copy pasted hitfinder method as it means maintaining the same thing twice. Have you tried a different approach too? Are there any fundamental issues which need to be solved? E.g. could not we just apply the regular hit finding and apply a different function to estimate the delay?

hoetzsch:
Hey Daniel,

> 1. Is the LED calibration data divided into fragments? E.g. your plot lasts for 700ish samples. Are those 700ish samples long pulses divided into fragments of a length of 110 samples each?

The afterpulse LED data is taken with an external trigger that triggers both the LED pulser and tells the DAQ to start recording everything for 800 samples with all channels. That means that each 'fragment' (if that is the correct terminology in this case) is always exactly 800 samples long and contains the LED hits and all afterpulses (the latest one being Xe+ at ~3µs delay time).

> 2. I have to admit I am not a friend of your copy pasted hitfinder method as it means maintaining the same thing twice. Have you tried a different approach too? Are there any fundamental issues which need to be solved? E.g. could not we just apply the regular hit finding and apply a different function to estimate the delay?

I think the afterpulse data is fundamentally different from the usual TPC data, because of the fact that it is triggered and always has a fixed 'waveform' length of 800 samples.
The afterpulse hitfinder is therefore different from the usual hitfinder, in the definition of an LED window that allows to reliably identify which hit is the LED hit, and allows for a per-waveform determination of the delay time of afterpulses.
The latter is the main advantage in the afterpulse hitfinder. The afterpulse data is in the format I explained above exactly to make this possible: Because all the afterpulse hits are in the same 'fragment' as the LED hit, the afterpulse hitfinder can give the delay time of an afterpulse w.r.t. the actual LED hit that produced it, which increases the accuracy in the afterpulse timing.
It could be difficult to find a way to implement this feature with the fixed LED window definition outside of the loop over the samples in the hitfinder (where it's currently implemented). But I am in no way a strax/en expert, so please let me know if you know a nice solution that includes the regular hit finder.
I also see your point regarding maintenance, however I've been working with the plugin as it is for about a year now, with very little effort needed on that side.

WenzDaniel:
Hi Luisa thanks for the clarification. 
About point 1. just to be 100% sure.
>  That means that each 'fragment' (if that is the correct terminology in this case) is always exactly 800 samples long and contains the LED hits and all afterpulses (the latest one being Xe+ at ~3µs delay time).

You mean all your raw_records have a `length == pulse_length == 800` samples? 

> About point 2. 

I understand the difference but I think we can achieve the same without duplicating the hitfinder. I am wondering if you have ever tried to do something like (assuming my assumption in point 1 is correct):

1. Find hits
2. Group hits by record_i e.g. simply sort first by record_i and then time (np.sort can do that). 
3. Loop over hits and:
   1. Identify LED hit for hit group  in record_i
   2. Compute the desired properties.
   3. If record_i changes push results and compute new properties

straxen/plugins/afterpulse_processing.py
```diff
@@ -0,0 +1,300 @@
...
+
+import numpy as np
+
+import numba
+export, __all__ = strax.exporter()
+
+# define new data type for afterpulse data
+dtype_ap = [(('Channel/PMT number', 'channel'),
+              '<i2'),
+              # = PMT number
```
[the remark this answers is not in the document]
kdund: I think we have different desiderata for terseness here? I would always encourage leaving sensible comments everywhere-- code degrades faster than the intention behind.

hoetzsch:
Hi everyone,

finally I've finished the new plugin version. It basically follows now the steps that @WenzDaniel suggested above: (thanks for all the help btw :))

> 1. Find hits
> 2. Group hits by record_i e.g. simply sort first by record_i and then time (np.sort can do that).
> 3. Loop over hits and:
>    1. Identify LED hit for hit group in record_i
>    2. Compute the desired properties.
>    3. If record_i changes push results and compute new properties

In addition I've added extensions of the integration windows for the PMT pulses (strax options `hit_left_extension` and `hit_right_extension`), similar to what is done in `straxen/plugins/peaklet_processing.py`.

Here's an example of a resulting afterpulse spectrum (same PMT as above):
apspectrum_pmt131_20210917

straxen/plugins/afterpulse_processing.py
```diff
@@ -0,0 +1,345 @@
...
+                                      'Specify as a tuple of length n_tpc_pmts, or a number,'
+                                      'or a string like "pmt_commissioning_initial" which means calling'
+                                      'hitfinder_thresholds.py'
+                                      'or a tuple like (correction=str, version=str, nT=boolean),'
+                                      'which means we are using cmt.',
+                                ),
+                    strax.Option('hit_min_height_over_noise',
+                                 default=4,
+                                 help='Minimum hit amplitude in numbers of baseline_rms above baseline.'
+                                      'Actual threshold used is max(hit_min_amplitude, hit_min_height_over_noise * baseline_rms).',
```
WenzDaniel, line 89: So for regular data this is not really used actually. And I would think as you are using triggered data this should be needed even less. Or are you suspecting that the LED pulser makes some noise?

straxen/plugins/afterpulse_processing.py
```diff
@@ -0,0 +1,345 @@
...
+    buffer = np.zeros(len(hits), dtype=dtype_ap)
+    if not len(hits):
+        return buffer
+    res = _find_ap(hits, records, LED_window_left, LED_window_right, hit_left_extension, hit_right_extension, buffer=buffer)
+    return res
+            
+
+@numba.jit(nopython=True, nogil=True, cache=True)
+def _find_ap(hits, records, LED_window_left, LED_window_right, hit_left_extension, hit_right_extension, buffer=None):
+    # hits need to be sorted by record_i, then time!
```
WenzDaniel, line 179: You could add a test in `find_ap`for this condition. Then you do not have to leave a comment here. I guess something like np.diff(hits['record_i'] >= 0) would be probably already sufficient. 
Option 2. maybe even more elegant you do the sorting in `find_ap`.

straxen/plugins/afterpulse_processing.py
```diff
@@ -0,0 +1,345 @@
...
+            baseline_fpart = records[prev_record_i]['baseline'] % 1
+        
+        res = buffer[offset]
+
+            
+        if h['left'] < LED_window_left:
+            # if hit is before LED window: discard
+            continue
+            
+        if (h['left'] >= LED_window_left) & (h['left'] < LED_window_right):
```
WenzDaniel, line 210: I think `(h['left'] >= LED_window_left)` is already always true here since otherwise  we would have called:

```
if h['left'] < LED_window_left:
            # if hit is before LED window: discard
            continue
```

Maybe check and if I am not wrong you can reduce it to `h['left'] < LED_window_right`.

straxen/plugins/afterpulse_processing.py
```diff
@@ -0,0 +1,345 @@
...
+    
+    prev_record_i = hits[0]['record_i']
+    record_data = records[prev_record_i]['data']
+    record_len = records[prev_record_i]['length']
+    baseline_fpart = records[prev_record_i]['baseline'] % 1
+    
+    for h_i, h in enumerate(hits):
+        
+        if h['record_i'] > prev_record_i:
+            # start of a new record
```
WenzDaniel, line 194: Please remind me here it has been sometime since we spoke. In your case a single record is equal to a single pulse is not it? E.g. pulse_length of records is equal to length of records?

straxen/plugins/afterpulse_processing.py
```diff
@@ -0,0 +1,345 @@
...
+            
+        if h['left'] < LED_window_left:
+            # if hit is before LED window: discard
+            continue
+            
+        if (h['left'] >= LED_window_left) & (h['left'] < LED_window_right):
+            # hit is in LED window
+            if not is_LED:
+                # this is the first hit in the LED window
+                res['time'] = h['time']
```
WenzDaniel, line 214: You have to be care full here. Since you specify `length` to be `res['right_integration'] - res['left_integration']` you have to updated time by `h['time'] - hit_left_extension*h['dt'] `. Otherwise time + length * dt will not cover the correct interval.

straxen/plugins/afterpulse_processing.py
```diff
@@ -0,0 +1,345 @@
...
+            
+            hit_data = record_data[res['left_integration']:res['right_integration']]
+            res['area'] = hit_data.sum() + res['length'] * baseline_fpart
+
+            res['sample_10pc_area'] = res['left_integration'] + get_sample_area_quantile(hit_data, 0.1, baseline_fpart)
+            res['sample_50pc_area'] = res['left_integration'] + get_sample_area_quantile(hit_data, 0.5, baseline_fpart)
+
+            res['max'] = res['left_integration'] + hit_data.argmax()
+
+            t_LED = res['sample_10pc_area']
```
WenzDaniel, line 259: I am not sure if this is possible. But to me it looks like that this fraction of the code is required several times. Maybe it is worth to be put in a small extra function.

WenzDaniel (commented):
Looks mostly fine to me a few small comments. Would you like to share with us your results with this new plugin in an upcoming Team B meeting? Also tagging @zhut19

straxen/plugins/afterpulse_processing.py
```diff
@@ -0,0 +1,345 @@
...
+        # if cmt config
+        if is_cmt_option(self.config['hit_min_amplitude']):
+            self.hit_thresholds = straxen.get_correction_from_cmt(self.run_id,
+                self.config['hit_min_amplitude'])
+        # if hitfinder_thresholds config
+        elif isinstance(self.config['hit_min_amplitude'], str):
+            self.hit_thresholds = straxen.hit_min_amplitude(
+                self.config['hit_min_amplitude'])
+        else: # int or array
+            self.hit_thresholds = self.config['hit_min_amplitude']
```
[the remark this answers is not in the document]
FaroutYLq: I agree that it looks quite annoying, but at this moment I have no idea how it can be done more elegantly... Any suggestion?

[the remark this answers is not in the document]
ahiguera-mx: Hi guys, 
I think the issue is because we want get the `hit_min_amplitude` from three different places, 1) CMT, 2) `hitfinder_thresholds.py` and 3) passing an int or array. Do we really need to support these three approaches? I would suggest we don't need to support `hitfinder_thresholds.py` anymore, we can use CMT for analysis and option 3) for testing.  Or did I missed something here @WenzDaniel

straxen/plugins/afterpulse_processing.py
```diff
@@ -0,0 +1,345 @@
...
+                                      'Actual threshold used is max(hit_min_amplitude, hit_min_height_over_noise * baseline_rms).',
+                                ),
+                    strax.Option('hit_left_extension',
+                                 default=2,
+                                 help='Extend hits by this many samples left',
+                                ),
+                    strax.Option('hit_right_extension',
+                                 default=20,
+                                 help='Extend hits by this many samples right',
+                                ),
```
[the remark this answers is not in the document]
hoetzsch: No, no particular reason (I wasn't aware this existed on pulse-level already). I'll change it to use the same one. Do I do this by just replacing my options with the four lines from pulse_processing.py or is there a cleverer way to use an already existing option?

straxen/plugins/afterpulse_processing.py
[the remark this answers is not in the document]
hoetzsch: Is there a decision on this yet? Then I can adjust accordingly

straxen/plugins/afterpulse_processing.py
hoetzsch: I think yes. I'm not sure what the exact definition of a 'pulse' is in strax/straxen, but in the triggered AP data every recorded single-channel WF (=record?) is always 800 samples long and contains the physical PMT pulses from LED and afterpulses. So every record corresponds to one LED trigger.

WenzDaniel: Sounds good. Yes so a pulse in DAQ and strax/straxen jargon is one entire waveform independently if it is fragment (normal data) or not (your case.)

hoetzsch:
I've implemented all the comments now, should be ready to merge now (hopefully) :)

WenzDaniel (approved):
Thanks a lot for your hard work. It has become a really nice plugin.

PR #550, opened 2021-06-16
Restart GCN processing

reallsx:

Start [new GCN](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:analysis:reconstruction_team:graph_network_porec2) processing.

removed the line that disable GCN in `contexts.py`

### DIFF
--- a/straxen/contexts.py
+++ b/straxen/contexts.py
@@ -43,7 +43,6 @@ xnt_common_config = dict(
         nveto=(2000, 2119),
         nveto_blank=(2999, 2999)),
     # Clustering/classification parameters
-    gcn_model=None,
     # Event level parameters
     s2_xy_correction_map=('s2_xy_map', "ONLINE", True),
     fdc_map=('fdc_map', "ONLINE", True),


jorana:
@reallsx @ahiguera-mx thanks for reactivating this!

Would it make sense to increment the version of the GCN plugin and change the name of the correction in CMT to only point to this new version of the NN? That way we would prevent people from comparing the different GCN-NN with the updated one?

Otherwise run 12000 and run 23000 are might yield entirely different patterns with this updated version of the NN

l-althueser:
Regarding the GCN plugin and CMT updates. I do see that this might happen for ONLINE processed data but this cannot happen for any offline context since the GCN was changed for all conditions. So latest local version only has new GCN-NN. Comparing ONLINE is anyway not a good idea ..
I guess that we want to avoid using different name due to all the changes we would need to do in CMT and other places. Does this kind of helps with the questions you asked @jorana ?

ahiguera-mx:
Hi Joran
I fully agree with what Lutz said, not sure if I can add more to that

jorana:
@l-althueser , alright, thanks if you, as team C leader, think the risk of potential issues is small, I do see your arguments.

l-althueser:
I would like to use my approval powers here. In the end I will anyway be the person that has to catch these kind of issues. I, as Team leader, won't trust any ONLINE analysis and approve using the same name.

Regarding the technical aspect of incrementing plugin versions - I would rely on your expertise.

reallsx:
Looks like we are already getting result from GCN while this is not merged yet?

jorana:
actually, the disabling of GCN never made it to an official release, so yeah, we are processing with GCN

Issue #530, opened 2021-06-03
Importing straxen is slow, smarter imports?


jorana:
_daft_
On my machine importing straxen takes 22 seconds, which is much more than we'd like to.
```bash
python -X importtime -c 'import straxen'
```

Main culprits (random order):

- holoviews
- utilix
- numba
- scipy
- ipywidgets
- bokeh
- strax

jorana:
I re-profiled using [tuna](https://github.com/nschloe/tuna):

```
python -X importtime -c 'import straxen' &> imports.log
tuna imports.log
``` 

![afbeelding](https://user-images.githubusercontent.com/22295914/122199916-d3e1ad00-ce9a-11eb-9245-99a3024615a9.png)

or `python3 -m import_profile strax bokeh holoviews utilix numba scipy straxen`
both pointing to holoviews for a considerable amount of import time

PR #551, opened 2021-06-16
Don't import holoviews and ipywidgets

jorana:
As in #530, importing holoviews is slow and not essential to do when importing straxen. Let's only import it when needed.


jorana:
No idea at all why all of a sudden codefactor is complaining about something unrelated 😕

WenzDaniel (approved):
Okay things are still working. Thanks a lot.

jorana:
Thanks Daniel for checking it out 😃

PR #555, opened 2021-06-18
New nt sim context & update get correction from CMT implementation

WenzDaniel:
In this PR we add a new "offline" simulation context to straxen. ~The context queries correction values from CMT and pass them to straxen as constant values~ The context checks all configurations and adds CMT run id tag and the id in the option tuple. This allows to freeze-in a certain setting given by a cmt_run_id for simulations.
We also update the get correction from CMT functions so it can recognize the cmt_run_id tag as well as the option to add prefix and affix to the correction name. Which will solve the issue of adding default pos rec tag for fdc and s1, s2 correction maps.

First, we create the context and register all plugins. Then we check all Plugins for CMT options, ~query the corresponding correction values, and overwrite the CMT options with "constant" options~ add CMT run id tag and the id in the option tuple. ~There was one issue though. Some corrections also need in addition the name of our default reconstruction algorithm. I added therefore the `_depends_on_pos_algo` function but would prefer to replace it with something less hardcoded. I am open for suggestions.~ The simulation configs and processing configs are now fully decoupled, that one can specify different CMT run ids.


ramirezdiego:
Thank you @WenzDaniel, for the PR and for the private written and ZOOM discussions on it :)

Just a few conclusive points if we move forward, following up on the discussion we had on Tuesday on this topic (cc @ershockley @terliuk @petergaemers @zhut19):
- We conclude that we always need a cmt_run_id for this simulations context, so no option for cmt_run_id=0 to ignore CMT configs.
- Instead, we now have the flag `overwrite_from_fax_file`, which points to the `fax_config*` file variables that match CMT parameter name. This requires renaming the relevant `fax_config*` parameters and propagating this renaming to WFSim.
- In this case, we risk to have, e.g., an `elife` definition in `fax_config*` different than that from CMT. Regardless of the `overwrite_from_fax_file` value, WFSim will always use the one defined in `fax_config*`. We can allow this only when setting `_allow_diverging_settings=False`, so there's a check for it 👌 
- In order to be able to load raw_records with any context (we all agreed on this), we could either
  - hardcode the gain model via context when producing them in WFSim; or
  - untrack the gain model from raw_records, although this requires to properly document how these were produced (we will store the context in mc_chain). -> I prefer this option. Do you all agree?

straxen/contexts.py
```diff
@@ -218,17 +219,100 @@ def xenonnt_led(**kwargs):
...
+
+    st.apply_cmt_version(f'global_{cmt_version}')
+    if not cmt_run_id:
+        raise ValueError('You have to specify a run_id which should be used to initialize '
+                         'the corrections.')
+
+    cmt_options = straxen.get_corrections.get_cmt_options(st)
+
+    p = st.get_single_plugin(cmt_run_id, 'peak_positions')
+    default_reconstruction_algorithm = p.config['default_reconstruction_algorithm']
```
jorana, line 268: I always thought this was a quite slow operation, as you fully init the plugin but I might be remembering it wrong.

jorana (commented):
Thanks Daniel, one consideration which might be worth changing is putting all the CMT work arounds in the CMT module? Just to prevent the already large context.py file from getting even larger.

straxen/contexts.py
WenzDaniel: Initializing the context took me about 5s. So I think it is fine.

WenzDaniel:
> one consideration which might be worth changing is putting all the CMT work arounds in the CMT module?

Sure I can move the corresponding code.

straxen/contexts.py
jorana: I see, perhaps it was only if you are actually loading things like gains that it makes it take a long time.

10.000x faster but completely irrelevant (19.4 ms vs 1.13 µs) would be:
st._plugin_class_registry['peak_positions'].takes_config['default_reconstruction_algorithm'].default

straxen/contexts.py
```diff
@@ -218,17 +219,100 @@ def xenonnt_led(**kwargs):
...
+    for option_key, cmt_setting in cmt_options.items():
+        cmt_setting = _depends_on_pos_algo(cmt_setting)
+        config_name, _, _ = _depends_on_pos_algo(cmt_setting)
+        config_value = straxen.get_correction_from_cmt(cmt_run_id, cmt_setting)
+        if isinstance(config_value, np.ndarray):
+            # We have to convert the arrays into a list to write the metadata
+            # numpy arrays are not json serializable.
+            config_value = config_value.tolist()
+
+        cmt_options[option_key] = (config_name + '_constant',
```
zhut19, line 289: Hi @WenzDaniel , sorry about the late comment. I still don't really get why we can not just change everything in cmt_setting to `('MC', cmt_runid, cmt_setting)`? Isn't it still a valid value to convert to a hash?

WenzDaniel: I understood from Diego that this cause an issue with the lineage. If this still works there would not be a need for a new context.

ramirezdiego (approved):
I think we can merge and start testing.

We don't need to resolve this before merging, but better sooner than later: do we all agree with what I wrote in the bullet point above? This is, can we untrack gain model from the raw_records lineage from now on?

zhut19 (approved):

jorana:
> 
> 
> I think we can merge and start testing.
> 
> We don't need to resolve this before merging, but better sooner than later: do we all agree with what I wrote in the bullet point above? This is, can we untrack gain model from the raw_records lineage from now on?

I would try and not do this whenever possible. You want to know if channels aren't producing data because of gains set to 0. So we should track this in my opinion.

ramirezdiego:
> 
> 
> > I think we can merge and start testing.
> > We don't need to resolve this before merging, but better sooner than later: do we all agree with what I wrote in the bullet point above? This is, can we untrack gain model from the raw_records lineage from now on?
> 
> I would try and not do this whenever possible. You want to know if channels aren't producing data because of gains set to 0. So we should track this in my opinion.

Hi @jorana. I think this was the compromised we reached in the dedicated meeting with @WenzDaniel @ershockley @zhut19 @petergaemers et al. We will document the gains used for that in the mc_chain, for example (context and configs will be tracked in the corresponding database entry), but we want to be able to load raw_records with any context, and having the gains define the lineage would not allow us to do that. I would appreciate if tagged people can share their thoughts here.

terliuk:
> I would try and not do this whenever possible. You want to know if channels aren't producing data because of gains set to 0. So we should track this in my opinion.
It can be saved, but strax(en) should not know/care about generation gains, similarly as it doesn't know about "true" gains in data. We can keep it for information somwhere, but it should not affect behavior of further steps of processing in any way. Ideally, we want to ensure that there is one copy of raw_records, as different generation gains should always mean different MC run number/ID, as it is easier to track it such way.

jorana:
> 
> 
> > I would try and not do this whenever possible. You want to know if channels aren't producing data because of gains set to 0. So we should track this in my opinion.
> 
> It can be saved, but strax(en) should not know/care about generation gains, similarly as it doesn't know about "true" gains in data. We can keep it for information somwhere, but it should not affect behavior of further steps of processing in any way. Ideally, we want to ensure that there is only one copy of raw_records, as different generation gains should always mean different MC run number/ID, as it is easier to track it such way.

We absolteu

jorana:
whoops wrong button

> It can be saved, but ~strax(en)~ WFSim should not know/care about generation gains, similarly as it doesn't know about "true" gains in data. 

This isn't strictly true, unless you mean to run gain calibrations on simulated data 😉 and use those gains. We at least know that the gains we are processing real data with are the best representation of the gains we currently have for that channel. If you untrack the gains at raw-records level and just start doing something completely different later in raw records, you would end up with very strange results and an uncalibrated simulation context right? If the raw-records instruct gains of 1e6  everywhere and you later process some channels with 5e6 and other with 1e5, doesn't that create problems?

Anyway, I don't want to interfere too much if this has been al decided upon before. If you untrack the gains, it only makes sense to process data with the exact same gains as the raw-records were generated (perhaps only dropping x channels, I see the benefit there). So either you track this in code or rely on the user not to do something by accident.

terliuk:
> This isn't strictly true, unless you mean to run gain calibrations on simulated data wink and use those gains. We at least know that the gains we are processing real data with are the best representation of the gains we currently have for that channel. If you untrack the gains at raw-records level and just start doing something completely different later in raw records, you would end up with very strange results and an uncalibrated simulation context right? If the raw-records instruct gains of 1e6 everywhere and you later process some channels with 5e6 and other with 1e5, doesn't that create problems?
>
> Anyway, I don't want to interfere too much if this has been al decided upon before. If you untrack the gains, it only makes sense to process data with the exact same gains as the raw-records were generated (perhaps only dropping x channels, I see the benefit there). So either you track this in code or rely on the user not to do something by accident.

The main reason why we want to have different gains at simulation and later analysis/peaklet level is exactly to check the effect of gain mismatch in data/truth and our processing. We might have problem if we have very different gains at wfsim simulation level and then straxen processing, but it will be the same as we have in data, as we know gains only with finite precision. And using the same gains for both makes assumption that we know gains precisely, so limiting possibility to study effects of gain uncertainties. If we do not track it, we can introduce some fluctuations to our "simulation" gain model (or any other simulation parameter) and the process it with mismatched one to evaluate the impact.

jorana:
Right, for such dedicated studies I can see the point. But I think for such very specific checks, there are other options, like just ignoring the gains when you are performing this detailed check using `fuzzy for`: https://strax.readthedocs.io/en/latest/advanced/fuzzy_for.html

Again, I don't want to stall progress on things already agreed on.

ershockley:
Based on previous discussions we said we would untrack the gains in raw_records. The main point is what Diego said: like we have for real TPC data, any simulated raw data should be able to be loaded using any context (even if the gain model is wrong). I guess using 'fuzzy' is an option too. I don't really know which one would be better

straxen/contexts.py
```diff
@@ -218,17 +219,100 @@ def xenonnt_led(**kwargs):
...
+                                   )
+
+    fax_config = straxen.get_resource(fax_config, fmt='json')
+    for fax_option_key, fax_option_value in fax_config:
+        if fax_option_key not in cmt_options.keys():
+            continue
+
+        if overwrite_from_fax_file:
+            cmt_option_name = cmt_options[fax_option_key][0]
+            cmt_options[fax_option_key] = (cmt_option_name, fax_option_value)
```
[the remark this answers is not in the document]
ahiguera-mx: Small comment here `cmt_version='v4'` should be `cmt_version='v3'` there is no `v3` version yet

straxen/contexts.py
```diff
@@ -218,19 +219,98 @@ def xenonnt_led(**kwargs):
...
 
 
-def xenonnt_simulation(output_folder='./strax_data'):
+def xenonnt_simulation(output_folder='./strax_data',
+                       cmt_run_id='020280',
+                       cmt_version='v3',
+                       fax_config='fax_config_nt_design.json',
+                       cmt_option_overwrite=immutabledict(),
+                       overwrite_from_fax_file=False,
+                       _cmt_run_id_proc_only=None,
```
terliuk, line 228: Should it be processing only or simulation only here? My understanding was that if we set `cmt_run_id`, it is used for simulation (inside wfsim etc) and in processing (in peak creation etc). And now when we set divergent `_cmt_run_id_proc_only`, does it invert the logic from previous comment? Or is it just naming?

straxen/contexts.py
```diff
@@ -218,19 +219,98 @@ def xenonnt_led(**kwargs):
     return st
 
 
-def xenonnt_simulation(output_folder='./strax_data'):
+def xenonnt_simulation(output_folder='./strax_data',
+                       cmt_run_id='020280',
+                       cmt_version='v3',
+                       fax_config='fax_config_nt_design.json',
+                       cmt_option_overwrite=immutabledict(),
```
terliuk, line 226: So, if we provide electron lifetime here, will it overwrite it for wfsim, but not during straxen processing, that will use value from `cmt_run_id`?

terliuk (commented):
Great thanks! 
I think we need also very clear examples for a couple of cases. Some examples:
1. Standard mode - simulation and production are exactly the same (e-life, gains, correction/dummy maps etc) 
2. we use everything from some run, but generate divergent gains (from different run or manual setup)
3. similar as above, but have divergent e-life or some other parameter.

(i realized that order of my comments below is inverted)

straxen/contexts.py
zhut19: It's the other way, that the `cmt_option_overwrite` can only overwrite CMT options in processing or `gain_model_mc`.
If you want to use a user-specific electron lifetime in simulation, you should delete `electron_lifetime_liquid='elife_conf'` from `_config_overlap` and set the lifetime in `fax_config_override`

straxen/contexts.py
zhut19: I think it's the same either way. (After the meeting I realized that cmt_run_id is tracked, as I pass the exact options to wfsim e.g. `'gain_model_mc': ('cmt_run_id', '020280', 'to_pe_model', 'v4', True)`. So we need another cmt_id if the processing is using some other run config.

zhut19:
temp_flow_chart

PR #558, opened 2021-06-21
Fix missing info in bootstrax docs, fix #546

### LINKED ISSUE #546: Bootstrax status not updating
**Describe the bug**
The status keeps saying it's idle when it should be busy

jorana:
Fix https://github.com/XENONnT/straxen/issues/546, update the docs.

There was a lingering "," in that made the `to_report` a tuple like `({'some': 'dict'}, )`

then `'some' in to_report` is `False`

### DIFF
--- a/bin/bootstrax
+++ b/bin/bootstrax
@@ -661,7 +661,7 @@ def set_state(state, update_fields=None):
     if update_fields:
         bootstrax_state.update(update_fields)
     if (previous_entry is None or
-            (now() - previous_entry['_id'].generation_time).seconds > timeouts[
+            (now() - previous_entry['time'].replace(tzinfo=pytz.utc).seconds > timeouts[
                 'min_status_interval']):
         bs_coll.insert_one(bootstrax_state)
     else:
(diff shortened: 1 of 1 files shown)


bin/bootstrax
```diff
@@ -660,7 +660,7 @@ def set_state(state, update_fields=None):
     )
     if update_fields:
         bootstrax_state.update(update_fields)
-    if (not previous_entry or
+    if (previous_entry is None or
             (now() - previous_entry['_id'].generation_time).seconds > timeouts[
```
darrylmasson, line 664: `previous_entry` should contain a `time` field, we can maybe improve the readability if we use it
```suggestion
            (now() - previous_entry['time'].replace(tzinfo=pytz.utc).seconds > timeouts[
```

darrylmasson (approved):
I blame the reviewer of #337 for missing the trailing comma.

bin/bootstrax
jorana: Thanks for the suggestion!

jorana:
Thanks Darryl for reviewing

PR #559, opened 2021-06-22
pre_apply_function from $HOME only in pytest

jorana:
In https://github.com/XENONnT/straxen/pull/497 we introduced a pre_apply_function to straxen. This file should always be loaded from a database.

However, for some CI we don't have always have database access, so we still need to be able to load it from somewhere. In this PR we change where such a testing file (only for github!) would be located.

In the `.github/scripts/create_pre_apply_function.sh`we write a file that we can use for testing, this file is now written to a place where it's close to put it by accident and we add a big warning that a normal user should never use this.


straxen/common.py
```diff
@@ -355,13 +355,23 @@ def pre_apply_function(data, run_id, target, function_name='pre_apply_function')
...
+        # Impossible to load from non-existent folder
+        return function_file
+
+    testing_file = os.path.join(home, testing_folder, function_file)
+
+    if os.path.exists(testing_file):
+        # For testing purposes allow loading from 'home/testing_folder'
+        warn(f'Using local function: {function_file} from {testing_file}! '
+             f'If you are not integrated testing on github you should '
+             f'absolutely remove this file. (See #559)')
```
WenzDaniel, line 372: You can also check if you are running in a pytest session: 

https://docs.pytest.org/en/latest/example/simple.html#pytest-current-test-environment-variable

This might be more robust. Could also be useful at different places during our testing.

WenzDaniel (approved):
Looks good. I found a stackoverflow comment which explain how one can safely check if one is in test session. E.g. to use alternative files or dbs for those tests. Might be handy for us too. I linked it below.

straxen/common.py
jorana: Nice! Let me indeed change this! Should be done in 
5b6286b

Issue #491, opened 2021-05-14
Scada calls fail if float times are used due to using len(str(start)) to do log10(start)


kdund:
The line here: 
https://github.com/XENONnT/straxen/blob/42d3f3839e0f771393c6dbbc61d8380320d87d71/straxen/scada.py#L233
uses the length of the string representation of the number to do log10
that works for ints, not for floats. 
If only ints are to be used, I think a separate test should be done for this.

jorana:
Should be fixed in #560

Issue #315, opened 2020-12-15
Documentation Scada and CMT


jorana:
Straxen would benefit from more elaborate documentation. Lately some fantastic features were added.

It would be good if we can provide more elaborate documentation of do's and don't for:
1. the OnlineMonitor (I, @jorana can work on that)
2. the ScadaInterface (perhaps @WenzDaniel )
3. CMT (perhas @ahiguera-mx)

Some of these things might be better suitable for this than others as it has to be written in such a way so as not to expose any xenon-secrets. Nevertheless, explaining structures can be very helpfull for other people to be pointed to.

The threshold for writing documentation should be as low as possible. Here is an example of documentation I just on the gridfs interface: https://github.com/XENONnT/straxen/pull/311/commits/889572f7a05a254e14d33f12e113237d9849ed3f

WenzDaniel:
I think we should still add SCADA and CMT is not it? At least it is still on my agenda.

jorana:
yes, this was one of the automatic closing actions of github which was unintended

ahiguera-mx:
I can work on the CMT documentation

jorana:
Fixed, thanks guys!

PR #560, opened 2021-06-22
Add scada interface to docs

WenzDaniel:
Adds a brief documentation for the SCADAInterface. Also fixes a little bug in some time range checks. Solves: https://github.com/XENONnT/straxen/issues/491 and partially https://github.com/XENONnT/straxen/issues/315.


docs/source/tutorials/ScadaInterfaceExample.ipynb
```diff
@@ -0,0 +1,615 @@
...
+   },
+   "outputs": [],
+   "source": [
+    "parameters = {'para1': 'XE1T.GEN_HEINZVMON.PI',\n",
+    "              'para2': 'XE1T.CTPC.Board06.Chan011.VMon',\n",
+    "              'para2': 'XE1T.CRY_TE101_TCRYOBOTT_AI.PI',\n",
+    "              'para3': 'XE1T.CRY_PT101_PCHAMBER_AI.PI',\n",
+    "              'para4': 'XE1T.GEN_CE911_SLM1_HMON.PI',\n",
+    "              'para5': 'XE1T.CRY_FCV104FMON.PI',\n",
+    "             }"
```
jorana, line 76: Thanks Daniel, can we add a link or reference where to get these parameter conventions from?

docs/source/tutorials/ScadaInterfaceExample.ipynb
```diff
@@ -0,0 +1,615 @@
+{
```
jorana, line 1: Usually we put these notebooks in the notebooks folder and add a link to that notebook in the tutorials, would it make sense to do the same here?

jorana, line 1: As far as I can see, there is no particularly very sensitive information here, but should we double check with AC/SC?

docs/source/reference/straxen.analyses.rst
```diff
@@ -4,6 +4,30 @@ straxen.analyses package
 Submodules
```
jorana, line 4: I always thought this got auto-updated but that appears not. Thanks for adding it!

jorana (approved):
Thanks Daniel!

docs/source/reference/straxen.analyses.rst
WenzDaniel: Yes, I was also a bit confused about it.

docs/source/tutorials/ScadaInterfaceExample.ipynb
WenzDaniel: Ah okay, what is the difference between the notebooks in "tutorials" and "notebooks" then? I was not sure where to put it.

jorana: The notebooks-folder is convenient because also average analysts might realize that one can actually run/download those.

The tutorials subfolder is quite hidden and only used for the documentation.

docs/source/tutorials/ScadaInterfaceExample.ipynb
WenzDaniel: Yes, good idea. I asked SC group.

PR #561, opened 2021-06-23
Rundb should not crash on fuzzy

jorana:
 - Small patch in bootstrax due to #558 
 - Warn, rather than crash for fuzzy-options. If rundb cannot understand fuzzy for, it should not prevent other frontends from doing so
 - ~Use the straxen RucioLocalBackend instead of strax' rucio backend.~ should go in a seperate PR


WenzDaniel (approved):
Looks good.

PR #564, opened 2021-06-23
Change tight_coincidence

JosephJHowlett:
In pax's waveform simulator, the [tight_coincidence_window](https://github.com/XENON1T/pax/blob/871b241f81456de26c8eab4fd74c282cdd5a9394/pax/plugins/peak_processing/BasicProperties.py#L125) was implemented with <= and >=, but in [straxen](https://github.com/XENONnT/straxen/blob/d298b675b06cddb954b4fad893fbb74364fa1d3e/straxen/plugins/peaklet_processing.py) it used < and > -- this amounts to an effectively 20ns smaller total left-right window. In other words, though we set the left and right window to the same values (50ns) as in pax, we end up with a "tighter" coincidence requirement.

This PR modifies straxen so that the windows have the same definition as in the past.

_edit by Joran_
At the request of the ACs, we are also lowering the tight coincidence to 2.

### DIFF
--- a/straxen/contexts.py
+++ b/straxen/contexts.py
@@ -287,6 +287,7 @@ x1t_common_config = dict(
     peak_right_extension=30,
     s1_max_rise_time=60,
     s1_max_rise_time_post100=150,
+    s1_min_coincidence=3,
     # Events*
     left_event_extension=int(0.3e6),
     right_event_extension=int(1e6),
--- a/straxen/plugins/peaklet_processing.py
+++ b/straxen/plugins/peaklet_processing.py
@@ -92,7 +92,7 @@ class Peaklets(strax.Plugin):
     parallel = 'process'
     compressor = 'zstd'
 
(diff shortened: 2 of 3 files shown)


jorana (approved):
Thanks Joey! Looks good to me, the difference would be near 60 ns in these two plots, so would give a nice boost in S1 performance: for [left](https://xe1t-wiki.lngs.infn.it/lib/exe/fetch.php?media=xenon:j_angevaare:acc_tight_coincidence_window_left_lowe.png) and [right](https://xe1t-wiki.lngs.infn.it/lib/exe/fetch.php?media=xenon:j_angevaare:acc_tight_coincidence_window_right_lowe.png)

skazama:
Thanks Joey! Good catch!

skazama (approved):

jorana:
Thanks again guys!

PR #568, opened 2021-06-24
S2 merge further debug

jorana:
# Debug #548

In #548 there was a bug for indices at the boundaries. In the process, we realized that in addition to fixing that, we could also simplify the logic.

To reproduce the error:
```python
import straxen
st = straxen.contexts.xenon1t_dali()
st.storage += [strax.DataDirectory('/dali/lgrandi/xenon1t/strax_converted/temp_data/strax_data', readonly=True)]
st.make('181027_2244', 'merged_s2s', progress_bar=True)
```

# How does it work
Only keep track of the start and end indices of the merges, that is all the information that one should need.


zhut19 (approved):
This looks great, thanks @jorana @WenzDaniel, for fixing the bug I left behind.

PR #569, opened 2021-06-25
Rewrite EventBasics, set event level S1 tight coincidence

jorana:
 - For speed, we might want to be able to increase the S1-tight coincidence requirement at a later stage in the processing. To do so, add a new option to EventBasics that ignores S1s that are below the tight coincidence requirement. This was a nice idea by @JosephJHowlett 
 - No more loop plugin as it's not needed to do a slow loop (my aim was to get it fully numba, I almost succeeded but in the end I did not manage due to the many fields that need passed on as arguments `copy_largest_peaks_in_events` 😭 ).
 - Set some Nan defaults to parameters for fields. If a parameter is stored as an int, default to -1. This was a request by @jpienaar13 .
 - Split the code a lot make the overall workflow clearer (hopefully), I might have overdone it a bit in the dtype setup, if so, please let me know.
 

To change the S1 coincidence level at event level, you can do:
```python
st_new.set_config(dict(event_s1_min_coincidence=3))
st_new.get_df(run_id, target)
```

## Can you give an example:
afbeelding

### DIFF
--- a/straxen/plugins/event_processing.py
+++ b/straxen/plugins/event_processing.py
@@ -311,8 +311,7 @@ class EventBasics(strax.Plugin):
         self.set_event_properties(event, largest_s1s, largest_s2s, peaks)
 
         # Loop over S1s and S2s and over main / alt.
-        for i, largest_s_i in enumerate([largest_s1s, largest_s2s]):
-            s_i = i + 1
+        for s_i, largest_s_i in enumerate([largest_s1s, largest_s2s], 1):
             # Largest index 0 -> main sx, 1 -> alt sx
             for largest_index, main_or_alt in enumerate(['s', 'alt_s']):
                 peak_properties_to_save = [name for name, _, _ in self.peak_properties]
--- a/straxen/plugins/event_processing.py
+++ b/straxen/plugins/event_processing.py
@@ -135,7 +135,7 @@ class EventBasics(strax.Plugin):
     The main S2 and alternative S2 are given by the largest two S2-Peaks
     within the event. By default this is also true for S1.
     """
-    __version__ = '1.0.0'
+    __version__ = 'bla' #1.0.0'
 
     depends_on = ('events',
                   'peak_basics',
@@ -173,7 +173,6 @@ class EventBasics(strax.Plugin):
         ]
 
         dtype += self._get_posrec_dtypes()
-
         return dtype
 
     def set_dtype_requirements(self):
@@ -260,11 +259,16 @@ class EventBasics(strax.Plugin):
 
         return posrec_dtpye
 
-    def set_nan_defaults(self, result):
-        # TODO don't remember which where the important ones
-        nan_types = 's2_x', 's2_y'
+    @staticmethod
+    def set_nan_defaults(result):
+        """
+        When constructing the dtype, take extra care to set values to
+        np.Nan / -1 (for ints) as 0 might have a meaning
+        """
         for field in result.dtype.names:
-            if any([field in n for n in nan_types]):
+            if np.issubdtype(result.dtype[field], np.integer):
+                result[field][:] = -1
+            else:
                 result[field][:] = np.nan
 
     def compute(self, events, peaks):
@@ -274,10 +278,16 @@ class EventBasics(strax.Plugin):
         fully_contained_in = strax.fully_contained_in(peaks, events)
         for event_i, event in enumerate(events):
             peak_mask = fully_contained_in == event_i
-            result_i = self.get_results_for_event(peaks[peak_mask])
+            n_peaks = np.sum(peak_mask)
 
             result[event_i]['time'] = event['time']
             result[event_i]['endtime'] = event['endtime']
+            result[event_i]['n_peaks'] = n_peaks
+
+            if not n_peaks:
+                continue
+
+            result_i = self.get_results_for_event(peaks[peak_mask])
             for field, value in result_i.items():
                 result[event_i][field] = value
         return result
@@ -327,7 +337,7 @@ class EventBasics(strax.Plugin):
                 result['large_s2_before_main_s2'] = max(s2peaks_before_ms2['area'])
         return result
 
-    # todo numbaft
+    # todo numbafy
     def get_result_for_si(self, result, peaks, s_i, main_si, secondary_si):
         """Get the extracted S1/S2 properties"""
         # Which properties do we need?
--- a/straxen/plugins/event_processing.py
+++ b/straxen/plugins/event_processing.py
@@ -108,9 +108,6 @@ class Events(strax.OverlapWindowPlugin):
         self.events_seen += len(result)
 
         return result
-        # TODO: someday investigate if/why loopplugin doesn't give
-        # anything if events do not contain peaks..
-        # Likely this has been resolved in 6a2cc6c
 
 
 @export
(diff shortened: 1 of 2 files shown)


straxen/plugins/event_processing.py
```diff
@@ -193,26 +164,81 @@ def infer_dtype(self):
...
+
+    @staticmethod
+    def _get_si_dtypes(peak_properties):
+        """Get properties for S1/S2 from peaks directly"""
+        si_dtype = []
+        for s_i in [1, 2]:
+            # Peak indices
+            si_dtype += [
+                (f's{s_i}_index', np.int32, f'Main S{s_i} peak index in event'),
+                (f'alt_s{s_i}_index', np.int32, f'Alternate S{s_i} peak index in event')]
```
WenzDaniel, line 204: Just noticed it, ,aybe we can change this to int16 or an uint16?

straxen/plugins/event_processing.py
```diff
@@ -222,103 +248,159 @@ def infer_dtype(self):
...
 
-    def compute_loop(self, event, peaks):
-        result = dict(n_peaks=len(peaks),
-                      time=event['time'],
-                      endtime=strax.endtime(event))
-        posrec_save = [(xy + algo)
-                       for xy in ['x_', 'y_']
-                       for algo in self.pos_rec_labels]
+            if not n_peaks:
+                continue
```
WenzDaniel, line 285: Should not we raise an error here? An event without peaks should not be possible unless it is a bug is not it?

straxen/plugins/event_processing.py
```diff
@@ -222,103 +248,159 @@ def infer_dtype(self):
...
+        """Get the largest S1/S2. For S1s allow a min coincidence and max time"""
+        # Find all peaks of this type (S1 or S2)
+        s_mask = peaks['type'] == s_i
+        if s_i == 1:
+            s_mask &= peaks['time'] < s1_before_time
+            s_mask &= peaks['tight_coincidence'] >= s1_min_coincidence
+
+        selected_peaks = peaks[s_mask]
+        s_index = np.arange(len(peaks))[s_mask]
+        largest_peaks = np.argsort(selected_peaks['area'])[-number_of_peaks:][::-1]
```
WenzDaniel, line 369:
```suggestion
        largest_peaks = np.argsort(selected_peaks['area'])[::-1][:number_of_peaks]
```
Does not this work too? Minor change but, the other way around it took me a moment to understand :D

straxen/plugins/event_processing.py
```diff
@@ -222,103 +248,159 @@ def infer_dtype(self):
...
-                    main_s[2]['center_time'] - secondary_s[1]['center_time']
-            if 2 in secondary_s:
-                result['alt_s2_interaction_drift_time'] = \
-                    secondary_s[2]['center_time'] - main_s[1]['center_time']
+        if len(largest_s1s) > 0 and len(largest_s2s) > 0:
+            result['drift_time'] = largest_s2s[0]['center_time'] - largest_s1s[0]['center_time']
+            if len(largest_s1s) > 1:
+                result['alt_s1_interaction_drift_time'] = largest_s2s[0]['center_time'] - largest_s1s[1]['center_time']
+            if len(largest_s2s) > 1:
+                result['alt_s2_interaction_drift_time'] = largest_s2s[1]['center_time'] - largest_s1s[0]['center_time']
```
WenzDaniel, line 339: I am wondering why we never added alt_S2 - alt_S1. Could be interesting for multi-scatter studies is not it?

straxen/plugins/event_processing.py
```diff
@@ -222,103 +248,159 @@ def infer_dtype(self):
...
+            s_i=1,
+            s1_before_time=s1_latest_time,
+            s1_min_coincidence=self.config['event_s1_min_coincidence'])
+
+        self.set_sx_index(event, s1_idx, s2_idx)
+        self.set_event_properties(event, largest_s1s, largest_s2s, peaks)
+
+        # Loop over S1s and S2s and over main / alt.
+        for i, largest_s_i in enumerate([largest_s1s, largest_s2s]):
+            s_i = i + 1
```
WenzDaniel, line 315:
```suggestion
        for s_i, largest_s_i in enumerate([largest_s1s, largest_s2s], 1):
```

straxen/plugins/event_processing.py
```diff
@@ -222,103 +248,159 @@ def infer_dtype(self):
...
+    @staticmethod
+    def copy_largest_peaks_in_events(result,
+                                     largest_s_i,
+                                     l_index,
+                                     result_fields,
+                                     peak_fields,
+                                     ):
+        """Fill the event result with the main/alt S1/S2"""
+        if len(largest_s_i) <= l_index:
+            return
```
WenzDaniel, line 383: The sense of this line is a bit hard to understand. From my understanding `len(largest_s_i)` can be either 0,1 or 2 and `l_index` will be 0 and 1. Hence you skip if the main/alt sx does not exists. But is it maybe possible to either make it more explicit or to put a comment?

straxen/plugins/event_processing.py
```diff
@@ -222,103 +248,159 @@ def infer_dtype(self):
...
+
+        if not len(result_fields) == len(peak_fields):
+            raise ValueError("result_fields and peak_names should map 1 to 1")
```
WenzDaniel, line 386: Not sure if you need this, but also no objection in keeping it. I think we want to check that you should check if the order is correct e.g. "if peak_field in event_field".

WenzDaniel (approved):
Looks fine, it almost reads like a well written prose. Mostly minor comments and some stuff we can think about. At two points it took me a bit longer to understand the purpose of the code. Maybe put a comment or so.

straxen/plugins/event_processing.py
jorana: Int16 could do yeah! Its not very heavy on the memort but yeah if we need an int32 for the number of peaks in one event, something is wrong

straxen/plugins/event_processing.py
jorana: Valid point 👍

straxen/plugins/event_processing.py
jorana: Right thats a good suggestion!

straxen/plugins/event_processing.py
jorana: Main reason is most likely that it's not used anywhere. One can of course just subtracy the alt s1 s2 centertimes. I'm actually not entirely convinced we need the alt_*_drift time for this exact reason

straxen/plugins/event_processing.py
jorana: Thanks, it's fully correct what you are saying but let me add a comment

jorana:
Thanks Daniel for the useful review, I don't see any reason to not follow your suggestions so I will gladly implement them 😄

straxen/plugins/event_processing.py
jorana: okay, I thought about it a little more, but if we have a lot of noise e.g. due to a 1 PMT requirement, we don't want to fail because there are 32k peaks in one event. It's unlikely but not excluded.

jorana: I did find some other funky choices that I will push

jorana:
```python
-            ('n_channels',        np.int32,   'count of contributing PMTs'),
-            ('n_competing',       np.float32, 'number of competing PMTs'),
+            ('n_channels',        np.int16,   'count of contributing PMTs'),
+            ('n_competing',       np.int32,   'number of competing peaks'),
```
this is the change I did make

PR #572, opened 2021-06-28
Allow sub-mbs datarates and old runs

jorana:

- The datarate can be below 1 MB/s, we can use a lot of resources then, so let's do that. Otherwise we are defaulting to the args.cores and args.max_messages like we see on the website (see below)
- If a run is finished, use the rate-field instead of a query since thats easier to parse and does not require an extra query
- Also remove some scaffolding eec95c1

afbeelding


darrylmasson (approved):
Looks good to me.

jorana:
Thanks Darryl!

PR #573, opened 2021-06-29
Remove xenonnt_simulation test

ershockley:
Removes a buggy xenonnt_simulation test that's failing over in base_env.


ershockley:
While I beat Joran by about 5 seconds, I'll close mine since his is better documented :)

PR #587, opened 2021-07-06
Fixing veto intervals time

AlexElykov:

Fixing a small bug in the Veto Intervals final dtype, previously it was outputting "endtime" twice. This was pointed to me by several people already. 

Also, I guess it's kinda worth keeping the np.clip, just in case something bugs out and we have at least the start/end times to rely upon.

PS. Plus updated one of the comments.


jorana (approved):
Thanks Alex, nice catch. I incremented the version just to be sure.

PR #588, opened 2021-07-08
Patch scada interface

WenzDaniel:

Fix a bug in scada interface noticed by @HaochenYanICRR. In the current version querying data with a higher precision than seconds will raise an error. In this PR we simply chop any ms/µs/ns entries in the time stamps. I also updated the test accordingly.

In addition I added  piggybacked a small change which removes the following warning when running tests:

```
straxen/plugins/x1t_cuts.py:201
  /home/dwenz/mymodules/straxen/straxen/plugins/x1t_cuts.py:201: DeprecationWarning: `np.bool` is a deprecated alias for the builtin `bool`. To silence this warning, use `bool` by itself. Doing this will not modify any behavior and is safe. If you specifically wanted the numpy scalar type, use `np.bool_` here.
  Deprecated in NumPy 1.20; for more details and guidance: https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations
    dtype = [('cut_s1_max_pmt', np.bool, 'S1 max PMT cut')] + strax.time_field
```

### DIFF
--- a/straxen/plugins/x1t_cuts.py
+++ b/straxen/plugins/x1t_cuts.py
@@ -198,7 +198,7 @@ class S1MaxPMT(strax.LoopPlugin):
     """Removes events where the largest hit in S1 is too large
     port from lax.sciencerun0.py"""
     depends_on = ('events', 'event_basics', 'peak_basics')
-    dtype = [('cut_s1_max_pmt', np.bool, 'S1 max PMT cut')] + strax.time_fields
+    dtype = [('cut_s1_max_pmt', np.bool_, 'S1 max PMT cut')] + strax.time_fields
     provides = 'cut_s1_max_pmt'
     __version__ = 1.2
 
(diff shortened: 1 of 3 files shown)


straxen/plugins/x1t_cuts.py
```diff
@@ -198,7 +198,7 @@ class S1MaxPMT(strax.LoopPlugin):
     """Removes events where the largest hit in S1 is too large
     port from lax.sciencerun0.py"""
     depends_on = ('events', 'event_basics', 'peak_basics')
-    dtype = [('cut_s1_max_pmt', np.bool, 'S1 max PMT cut')] + strax.time_fields
+    dtype = [('cut_s1_max_pmt', np.bool_, 'S1 max PMT cut')] + strax.time_fields
```
jorana, line 201: Please don't forget to bump the version

jorana (approved):
Thanks looks good, just one minor comment

straxen/plugins/x1t_cuts.py
WenzDaniel: Uh that is something I have not thought about. Are you afraid that decoding the data does not work any longer because we changed the dtype? I would have expected that np.bool and np.bool_ behaves the same. But I will do.

PR #591, opened 2021-07-14
use rucio_did in rundb frontend

JoranAngevaare:
As requested by computing, replace the lineage field by the `(rucio-)did` field. Two places are affected:
 - making data
 - looking for data

The lineage was a very long and tedious way of trying to go through the entire lineage and complicates things. Instead computing would like to use one unified variable, the rucio-did.


bin/bootstrax
```diff
@@ -1069,7 +1069,6 @@ def upload_file_metadata(rd):
                     chunk_mb = [chunk['nbytes'] / (1e6) for chunk in md['chunks']]
                     data_size_mb = int(np.sum(chunk_mb))
                     avg_data_size_mb = int(np.average(chunk_mb))
-                    lineage_hash = md['lineage_hash']
```
ershockley, line 1072: I think we want to leave this in right? Unless it is already included somehow?

bin/bootstrax
```diff
@@ -1088,7 +1087,6 @@ def upload_file_metadata(rd):
                       'data.$.meta.straxen_version': straxen.__version__,
                       'data.$.meta.size_mb': data_size_mb,
                       'data.$.meta.avg_chunk_mb': avg_data_size_mb,
-                      'data.$.meta.lineage_hash': lineage_hash,
                       'data.$.meta.compressor': compressor,
                       }
```
ershockley, line 1091:
```suggestion
                      'data.$.lineage_hash': lineage_hash
                      }
```

straxen/rundb.py
```diff
@@ -136,16 +134,12 @@ def _data_query(self, key):
             'data': {
                 '$elemMatch': {
                     'type': key.data_type,
-                    # TODO remove the meta.lineage since this doc
-                    #  entry is deprecated.
-                    '$and': [{'$or': [
-                        {'meta.lineage': key.lineage},
+                    '$and': [
```
ershockley, line 137: If we include the lineage hash and the data type in the top level of the data array, then I think we make this query a bit more clear: 

```
{'data': {'$elemMatch': {'type': key.data_type, 'lineage_hash': key.lineage_hash, 'location': 'UC_DALI_USERDISK', 'status': 'transferred'}}}
```

I see this is part of the `available_query` -- not sure exactly best way to implement.

straxen/rundb.py
```diff
@@ -206,16 +200,16 @@ def _find(self, key: strax.DataKey,
                         'host': self.hostname,
                         'type': key.data_type,
                         'protocol': strax.FileSytemBackend.__name__,
```
ershockley, line 202: Is this field used anywhere else? Since we are updating things maybe we could consider removing this as well?

straxen/rundb.py
```diff
@@ -206,16 +200,16 @@ def _find(self, key: strax.DataKey,
...
-                        # TODO: duplication with metadata stuff elsewhere?
-                        'meta': {'lineage': key.lineage}
+                        'did': key_to_rucio_did(key),
+                        'lineage_hash': key.lineage_hash,
```
ershockley, line 204: Ah I think this answers my question above. So bootstrax is using this frontend and thus this line sets the `lineage_hash` to be a top level field. Is that right?

ershockley (commented):
Thanks @jorana! From my side I think it looks good. If you can clarify a few questions that would be great. After that we will then need to update the runDB for all entries which I am happy to do.

lucascottolavina:
We should think more carefully about those changes in order to not have to change them again and again in the future.

Having the did directly available in the data is a great idea but if we want to have the hash as well, then the presence of two redundant information is a weakness of the robustness of the DB. We should try to avoid to double the information. If having the hash is important to simplify the data field, then I would not have the did and let the software to build, when needed, the did with the information already present (run number, type, hash).

There is also another point. If we have the did, then for the first time we will have (even if not explicitly) also the run number information inside the data field. This could indeed be an advantage for some data management tools, in particular when you extract a data element and you want to remember at which run number it belongs to. Then, why not having directly the run number? That would be much a clean way to store the information. Basically we could have, inside data:

'number'
'type'
'lineage_hash' (that we should simply call 'hash')

and have the did information extracted from number, type and hash with the usual tools.

straxen/rundb.py
jorana: yes, so instead of doing it at bootstrax do this already at the start in the writing part of the frontend. In contrast to the other fields, you already know this up front so bootstrax does not have to upload it later after we finished writing

straxen/rundb.py
jorana: Yes, we use it later to say that it's the `FileSytemBackend` here:
https://github.com/XENONnT/straxen/blob/5d20f66aab548657e39bad57cb620a2ea7c9594a/straxen/rundb.py#L255
and here
https://github.com/XENONnT/straxen/blob/5d20f66aab548657e39bad57cb620a2ea7c9594a/straxen/rundb.py#L255

We could hardcode `'FileSytemBackend'` if needed but I would argue to leave it as it is

straxen/rundb.py
jorana: Right, that is at least the $or field which is not per se needed here but we could change:
```
{'did': {'$regex': f'/*{key.data_type}-{key.lineage_hash}'}}
```
to
```
'type': key.data_type, 'lineage_hash': key.lineage_hash
```

jorana:
>  We should try to avoid to double the information. If having the hash is important to simplify the data field, then I would not have the did and let the software to build, when needed, the did with the information already present (run number, type, hash).
 
This is true

> There is also another point. If we have the did, then for the first time we will have (even if not explicitly) also the run number information inside the data field. This could indeed be an advantage for some data management tools, in particular when you extract a data element and you want to remember at which run number it belongs to. Then, why not having directly the run number? That would be much a clean way to store the information. Basically we could have, inside data:
> 
> 'number'
> 'type'
> 'lineage_hash' (that we should simply call 'hash')
> 
> and have the did information extracted from number, type and hash with the usual tools.

Right, in principle I'm not opposed to this, it would be a much more breaking change and would disable older straxen versions loading data. I was always told that all that mattered to computing is the `did` so I did not know this was up for discussion. If we are going all the way, might as well go for the separate database discussed earlier but I don't know if any of us has currently time to do such a complete overhaul. 

I by the way don't see any reason for adding the run number in the document which is already group by the run-doc which is indexed by run-number 😉.

lucascottolavina (approved):
Fine like that. Other improvements I proposed are not that important compared to the risk to break with old versions.

bin/bootstrax
jorana: I think this was resolved 👍

JoranAngevaare:
will close this PR and keep the reference on my own fork if we ever want to get back to this

PR #592, opened 2021-07-14
Use rucio from straxen & nest RucioRemote imports

jorana:
- Reduce import verbosity
- Nest RucioRemote - imports (do them only when needed, but fail hard)
- Use straxens backend, instead of strax such that we can one day [remove that from strax](https://github.com/AxFoundation/strax/pull/451)

### DIFF
--- a/straxen/rucio.py
+++ b/straxen/rucio.py
@@ -1,6 +1,5 @@
 import socket
 import re
-from tqdm import tqdm
 import json
 from bson import json_util
 import os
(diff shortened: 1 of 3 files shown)


straxen/rucio.py
```diff
@@ -93,6 +77,7 @@ def __init__(self,
             self.local_rucio_path = rucio_prefix
 
         if include_remote:
+            self._set_remote_imports()
```
ershockley, line 80: The local frontend also uses the rucio client to get the local path (see [here](https://github.com/XENONnT/straxen/blob/use_rucio_from_straxen/straxen/rucio.py#L145)). 

In my view people should be using the main envs (with rucio installed) anyway, but I guess it is probably better to just hardcode the rucio prefix rather than fetching it? Then the local backend could be used without rucio, like in the RunDB frontend.

jorana: Right, this is one of the things I should have checked. We did this previously with the `except NameError as e:`, but now we can do the exact same if `self._rse_client is None`

jorana: Should be fixed in cff63ae

PR #593, opened 2021-07-14
increase input_timeout buffer daq reader

jorana:
We are experimenting with longer chunks, let's see how this works out.

Let's for the time being set this to a much longer value (current is 80s). If chunks are ~20 long and a few are in redax' memory 80s might not be enough.


darrylmasson (approved):
I'm still not 100% sure why 80s doesn't work, because with 21 second chunks it shouldn't be more than 42 seconds before data shows up. 300s might be ambitious because there are also other timers in the mailbox system we might start running into as well.

PR #595, opened 2021-07-16
Straxer for other packages

jorana:
Allow straxer to:
- be run for other packages (like cutax)
   ``` --package cutax```
- allow multi-target processing
   ```--target event_info cut_daq_veto```
- allow context and config kwargs (NB: should be JSON)
  ```--context_kwargs '{"use_rucio": 1}'```
  ```--config_kwargs '{"s1_min_tight_coincidence": 10}'```

## Can you give a minimal working example (or illustrate with a figure)
Use all these features in one example:
```
straxer 017918 --target event_info cut_daq_veto \
               --package cutax \
               --context xenonnt_v3 \
               --notlazy --timeout 600 \
               --context_kwargs '{"use_rucio": 1}'
```

### DIFF
--- a/bin/straxer
+++ b/bin/straxer
@@ -191,7 +191,7 @@ def main(args):
     # to do them after argparsing (so --help is fast)
     import strax
     import straxen
-    straxen.print_versions({'strax', 'straxen', args.package})
+    straxen.print_versions(tuple({'strax', 'straxen', args.package}))
 
     st = setup_context(args)
 
--- a/bin/straxer
+++ b/bin/straxer
@@ -113,6 +113,7 @@ def parse_args():
         help='Also add folder to st.storage')
     return parser.parse_args()
 
+
 def setup_context(args):
     # reimport to be safe
     import strax
@@ -120,6 +121,7 @@ def setup_context(args):
 
     context_module = importlib.import_module(f'{args.package}.contexts')
     st = getattr(context_module, args.context)()
+
     if args.context_kwargs:
         logging.info(f'set context kwargs {args.context_kwargs}')
         st = getattr(context_module, args.context)(**args.context_kwargs)
--- a/bin/straxer
+++ b/bin/straxer
@@ -263,7 +263,7 @@ def main(args):
                  f"peak RAM usage was around {peak_ram:.1f} MB.")
 
 
-def register_to_context(st, module:str):
+def register_to_context(st, module: str):
     if not os.path.exists(module):
         raise FileNotFoundError(f'No such file {module}')
     assert module.endswith('.py'), "only py files please!"
@@ -294,4 +294,3 @@ if __name__ == '__main__':
         print(f"Memory profiler says peak RAM usage was: {max(mem):.1f} MB")
     else:
         sys.exit(main(args))
-
--- a/bin/straxer
+++ b/bin/straxer
@@ -9,10 +9,10 @@ import os
 import os.path as osp
 import platform
 import psutil
-import sys
 import json
 import importlib
-
+import sys
+from pprint import pprint
 
 def parse_args():
     parser = argparse.ArgumentParser(
@@ -40,6 +40,11 @@ def parse_args():
         '--context_kwargs',
         type=json.loads,
         help='Use a json-file to load the context with')
+    parser.add_argument(
+        '--register_from_file',
+        type=str,
+        help='do st.register_all from a specified file'
+    )
     parser.add_argument(
         '--config_kwargs',
         type=json.loads,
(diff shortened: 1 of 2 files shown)


bin/straxer
```diff
@@ -112,13 +122,20 @@ def main(args):
...
+    module = getattr(locals()[args.package], 'contexts')
+    st = getattr(module, args.context)()
     if args.diagnose_sorting:
         st.set_config(dict(diagnose_sorting=True))
-    if args.config_from_json != '':
-        conf = json_to_dict(args.config_from_json)
-        st.set_config(conf)
+    if args.context_kwargs:
+        logging.info(f'set context kwargs {args.context_kwargs}')
+        st = getattr(module, args.context)(**args.context_kwargs)
```
WenzDaniel, line 134: Are there use cases in which one has to explicitly overwrite certain context setting. Would not it be sufficient to use `st.new_context` here?

WenzDaniel (commented):
Thanks, could you add one more option. I think it would also be handy to have an option "register plugins from file".

bin/straxer
jorana: No this is different as it also allows context (function) init kwargs like `use_rucio` which is not a context config

bin/straxer
```diff
@@ -112,13 +122,20 @@ def main(args):
     print(f"\tstrax {strax.__version__} at {osp.dirname(strax.__file__)}")
     import straxen
     print(f"\tstraxen {straxen.__version__} at {osp.dirname(straxen.__file__)}")
-
-    st = getattr(straxen.contexts, args.context)()
+    if not args.package == 'straxen':
+        # pylint: disable=exec-used
+        exec(f'import {args.package}')
```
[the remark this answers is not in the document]
jorana: Thanks!

jorana:
@WenzDaniel , not a bad idea at all, I haven't tested it but I think this should work

WenzDaniel:
Okay let me quickly check.

straxen/misc.py
```diff
@@ -57,10 +59,12 @@ def print_versions(modules=('strax', 'straxen', 'cutax'), return_string=False):
...
-            # pylint: disable=exec-used
-            exec(f'import {m}')
-            # pylint: disable=eval-used
-            message += f'\n{m}\tv{eval(m).__version__}\t{eval(m).__path__[0]}'
+            mod = import_module(m)
+            message += f'\n{m}'
+            if hasattr(mod, '__version__'):
+                message += f'\tv{mod.__version__}'
+            if hasattr(mod, '__path__'):
+                message += '\t{mod.__path__[0]}'
```
WenzDaniel:
```suggestion
            message += f'\n{m}'
            if hasattr(mod, '__version__'):
                message += f'\tv{mod.__version__}'
            if hasattr(mod, '__path__'):
                message += f'\t{mod.__path__[0]}'
```

jorana: Yeah I also noticed just before you submitted this: b017abc

bin/straxer
```diff
@@ -155,6 +177,26 @@ def main(args):
...
+        level=logging.DEBUG if args.debug else logging.INFO,
+        format='%(asctime)s - %(threadName)s - %(name)s - %(levelname)s - %(message)s')
+
+    logging.info(f"Starting processing of run {args.run_id} until {args.target}")
+
+    # These imports take a bit longer, so it's nicer
+    # to do them after argparsing (so --help is fast)
+    import strax
+    import straxen
+    straxen.print_versions({'strax', 'straxen', args.package})
```
WenzDaniel, line 194:
```suggestion
    straxen.print_versions(tuple({'strax', 'straxen', args.package}))
```
Otherwise straxer complains

WenzDaniel: Sorry, I just committed but I guess it is fine.

jorana: I'm absolutely furious that you interfere with my PR!

bin/straxer
```diff
@@ -214,15 +259,25 @@ def main(args):
...
+
+def register_to_context(st, module: str):
+    if not os.path.exists(module):
+        raise FileNotFoundError(f'No such file {module}')
+    assert module.endswith('.py'), "only py files please!"
+    folder, file = os.path.split(module)
+    sys.path.append(folder)
+    to_register = importlib.import_module(os.path.splitext(file)[0])
+    st.register_all(to_register)
+    logging.info(f'Successfully registered {file}. Printing plugins')
```
WenzDaniel, line 274: This is a pretty long print-out in the beginning we could reduce to just show the in addition registered once.

WenzDaniel (approved):
Okay it works, but one question though. Have you also modified the print out you get per chunk processed? The progress bar is gone now (but since it was not working nicely anyhow no important loss I would say).

bin/straxer
jorana: Yeah I thought about that but it would require more work since you might also overwrite some plugins. I would rather be a bit more verbose than not know which plugins I use.

jorana:
> 
> 
> Okay it works, but one question though. Have you also modified the print out you get per chunk processed? The progress bar is gone now (but since it was not working nicely anyhow no important loss I would say).

Yes, for me its either the progress bar or the prints, having both is too much I think and also fills up your logs.

The print statements are a de facto progress-bar anyway

PR #601, opened 2021-07-22
Fix peaklet area bias

WenzDaniel:

Complementary PR to https://github.com/AxFoundation/strax/pull/486. 

In addition to the changes in strax we have to set the left and right integration bounds for our hits like we do in case of lone hits. For this purpose `_find_hit_integration_bounds` is used. With `outside_peaks = self.create_outside_peaks_region(peaklets, start, end)` we ensure that hits are not extended beyond peaklet boundaries.


WenzDaniel:
Tests will fail without https://github.com/AxFoundation/strax/pull/486 being merged.

zhut19 (approved):
All looks good to me.

tests/test_peaklet_processing.py
```diff
@@ -0,0 +1,26 @@
+import numpy as np
+from hypothesis import given, settings
+import hypothesis.strategies as st
```
jorana, line 3: st is both context and strategies which can be confusing

straxen/plugins/peaklet_processing.py
```diff
@@ -153,22 +153,44 @@ def compute(self, records, start, end):
...
+        # is not extended beyond a peak.
+        outside_peaks = self.create_outside_peaks_region(peaklets, start, end)
+        strax.find_hit_integration_bounds(hits,
+                                          outside_peaks,
+                                          records,
+                                          (self.config['peak_left_extension'],
+                                           self.config['peak_right_extension']),
+                                           len(self.to_pe),
+                                          allow_bounds_beyond_records=True,
+                                          )
```
jorana, line 176:
```suggestion
        strax.find_hit_integration_bounds(
            hits, outside_peaks, records,
            save_outside_hits=(self.config['peak_left_extension'],
                               self.config['peak_right_extension']),
            n_channels=len(self.to_pe),
            allow_bounds_beyond_records=True,
        )
```

straxen/plugins/peaklet_processing.py
```diff
@@ -153,22 +153,44 @@ def compute(self, records, start, end):
...
+
+        # Updated time and length of hits and sort again:
+        hits_time_shift = (hits['left'] - hits['left_integration']) * hits['dt']
+        hits['time'] = hits['time'] - hits_time_shift
+        hits['length'] = (hits['right_integration'] - hits['left_integration'])
+        hits = strax.sort_by_time(hits)
+        rlinks = strax.record_links(records)
```
jorana, line 183: Not sure if there is much we can do about it. But like 0bbad6c this feels like we could hit issues with this changing definition. Maybe we can leave it for now unless you see some easy way of making this more explicit.

jorana (approved):
Thanks Daniel, as we discussed privately, all the check I did after 0bbad6c look good.

straxen/plugins/peaklet_processing.py
WenzDaniel: We could starting calling these objects `hitlets` instead of hits? Then it would be consistent and the difference should be clear. Hit refers to the central part above threshold a `hitlet` to the entire signal including the left and right extension.

jorana: sounds good 👍 , adding `Hit refers to the central part above threshold a hitlet to the entire signal including the left and right extension.` as a comment will help reminding the people less used to veto-processing

WenzDaniel: Do you thin kit is fine: https://github.com/XENONnT/straxen/pull/601/commits/bc7df5e15275fc0e2a908951de578d2260055b6e ?

Issue #596, opened 2021-07-17
hitlet-processing is bugged (or flaky)


JoranAngevaare:
**Describe the bug**
Processing on the event-builder fails due to hitlet-processing. It remains unclear where this bug is (either MV or NV processing)

**To Reproduce**
Process 25576 

**Expected behavior**
Don't fail

**Versions**
Please add the output of:
```
(py38) xedaq@eb4:~$ python -c "import straxen ; straxen.print_versions()"
Working on eb4.xenon.local with the following versions and installation paths:
python  v3.8.5  (default, Sep  4 2020, 07:30:14) [GCC 7.3.0]
strax   v0.16.1 /home/xedaq/software/strax/strax
straxen v0.19.3 /home/xedaq/software/straxen/straxen
```

**Traceback**
```
Exception info: concurrent.futures.process._RemoteTraceback: 
"""
Traceback (most recent call last):
  File "/home/xedaq/miniconda/envs/py38/lib/python3.8/concurrent/futures/process.py", line 239, in _process_worker
    r = call_item.fn(*call_item.args, **call_item.kwargs)
  File "/home/xedaq/software/npshmex/npshmex.py", line 193, in shm_wrap_f
    result = f(*args, **kwargs)
  File "/home/xedaq/software/strax/strax/plugin.py", line 1032, in do_compute
    r = p.do_compute(**compute_kwargs)
  File "/home/xedaq/software/strax/strax/plugin.py", line 525, in do_compute
    result = self.compute(**kwargs)
  File "/home/xedaq/software/straxen/straxen/plugins/veto_hitlets.py", line 134, in compute
    strax.hitlet_properties(temp_hitlets)
ZeroDivisionError: division by zero
"""

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/xedaq/software/straxen/bin/bootstrax", line 1293, in run_strax
    st_make()
  File "/home/xedaq/software/straxen/bin/bootstrax", line 1271, in st_make
    st.make(run_id, targets,
  File "/home/xedaq/software/strax/strax/context.py", line 1215, in make
    for _ in self.get_iter(run_ids[0], targets,
  File "/home/xedaq/software/strax/strax/context.py", line 1099, in get_iter
    generator.throw(e)
  File "/home/xedaq/software/strax/strax/context.py", line 1072, in get_iter
    for n_chunks, result in enumerate(strax.continuity_check(generator), 1):
[...]

PR #604, opened 2021-07-29
Increase buffer size

WenzDaniel:

Hopefully solves: https://github.com/XENONnT/straxen/issues/596

I processed test wise a few runs with straxer and `NUMBA_DEBUG_CACHE 1` to see how often we trigger recompiling due to a changed buffer size for nv/mv hitlets and it is almost every third chunk. I tripled now the buffer size which reduced the number of recompiling to almost zero. 

I also checked the peak memory usage and the buffer size does not have a big impact. The peak usage increased by about 200 MB to 1700MB for hitlets_nv and 1000MB for hitlets_mv. 

As a nice side effect the plugin is now faster as we are not recompiling everything all the time.

### DIFF
--- a/straxen/plugins/veto_hitlets.py
+++ b/straxen/plugins/veto_hitlets.py
@@ -120,7 +120,10 @@ class nVETOHitlets(strax.Plugin):
         del hits
 
         # Get hitlet data and split hitlets:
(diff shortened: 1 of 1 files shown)


jorana:
Thanks Daniel!

PR #620, opened 2021-08-03
Remove online event monitor

jorana:
Remove the online_event_monitor plugin, it is not used. In case of need, one can always go back in git to implement it again


skazama (approved):
Thanks @jorana, I agree we can remove online event_monitor since it is not used.

Issue #619, opened 2021-08-03
Stop making `events` datatype


jorana:
## Proposal

- Stop saving the `events` datatype 
- include the `event_number` field in `event_basics`.
- Don't include `events` in `event_info`

![afbeelding](https://user-images.githubusercontent.com/22295914/127978991-400b4a3c-ee64-444f-9c3c-bda3cd6e9b29.png)

## Why
`events` is currently almost an empty datatype (`time`, `endtime`, `event_number`). For this one extra field, we might as well add it to the `event_basics`. An additional advantage would be that we could build `event_info` from `event_basics` only. It would require less files/dependencies to build `event_info` so that would also be slightly quicker.

WenzDaniel:
Sounds good. I cannot spot any possible show stoppers.

PR #622, opened 2021-08-03
Disable event saving

jorana:
Disable event saving and build event_info on basis of event_basics

See #619


WenzDaniel (approved):
Looks fine to me. Have you checked if any mini-analysis depends on events? We should change it too,

jorana:
> 
> 
> Looks fine to me. Have you checked if any mini-analysis depends on events? We should change it too,

Thanks @WenzDaniel , good catch, there was only the ev-display now fixed

PR #623, opened 2021-08-03
Add lone hit information to merged S2s.

WenzDaniel:
Straxen counter part for https://github.com/AxFoundation/strax/pull/506. More information can be found also in this [note](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:wenz:comissioning:tpc:sr0:s1_bias#updated).

### DIFF
--- a/straxen/plugins/peaklet_processing.py
+++ b/straxen/plugins/peaklet_processing.py
@@ -547,16 +547,24 @@ class PeakletClassificationHighEnergy(PeakletClassification):
                       "where the gap size of the first point is the maximum gap to allow merging"
                       "and the area of the last point is the maximum area to allow merging. "
                       "The format is ((log10(area), max_gap), (..., ...), (..., ...))"
-                 ))
+                 ),
+    strax.Option('gain_model',
+                 help='PMT gain model. Specify as '
+                      '(str(model_config), str(version), nT-->boolean'),
+)
 class MergedS2s(strax.OverlapWindowPlugin):
     """
     Merge together peaklets if peak finding favours that they would
     form a single peak instead.
     """
-    depends_on = ('peaklets', 'peaklet_classification')
+    depends_on = ('peaklets', 'peaklet_classification', 'lone_hits')
     data_kind = 'merged_s2s'
     provides = 'merged_s2s'
-    __version__ = '0.2.0'
+    __version__ = '0.3.0'
+
+    def setup(self):
+        self.to_pe = straxen.get_correction_from_cmt(self.run_id,
+                                                     self.config['gain_model'])
 
     def infer_dtype(self):
         return strax.unpack_dtype(self.deps['peaklets'].dtype_for('peaklets'))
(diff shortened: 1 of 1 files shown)


jmosbacher (approved):
looks like a pretty small and safe change.

jmosbacher (changes requested):
I think it would be good to add to this PR a change to the requirements.txt file requiring a version of strax that has the add_lone_hits function (after its merged and released).

jorana:
@WenzDaniel , do the tests work well for you, for some reason it appears stuck.

WenzDaniel:
Hmm this is strange. I observe the same on machine now. Let me check.

WenzDaniel:
Okay two little mistakes with made the test stuck. 

1. lone_hits were not write able. 
2. I forgot the @export in strax: https://github.com/AxFoundation/strax/pull/508

WenzDaniel:
With https://github.com/AxFoundation/strax/pull/512 the PR should finally work.

straxen/plugins/peaklet_processing.py
```diff
@@ -702,7 +721,10 @@ def infer_dtype(self):
         return strax.unpack_dtype(self.deps['peaklets_he'].dtype_for('peaklets_he'))
 
     def compute(self, peaklets_he):
-        return super().compute(peaklets_he)
+        # There are not any lone hits for the high energy channel, 
+        #  so create a dummy for the compute method.
+        lone_hits = np.zeros(0, dtype=strax.hit_dtype)
```
jorana, line 726: we could easily make lone_hits_he if needed, but since no one is looking at it anyway. I guess this is fine

jorana (approved):

straxen/plugins/peaklet_processing.py
```diff
@@ -650,6 +658,17 @@ def compute(self, peaklets):
                 max_buffer=int(self.config['s2_merge_max_duration']
                                // peaklets['dt'].min()))
             merged_s2s['type'] = 2
+            
+            # Updated time and length of lone_hits and sort again:
+            lh = np.copy(lone_hits)
+            del lone_hits
+            lh['time'] = (lh['time']
+                                 - (lh['left']-lh['left_integration'])*lh['dt']
+                                 )
```
jorana, line 667:
```suggestion
            lh_time_shift = (lh['left'] - lh['left_integration']) *lh['dt']
            lh['time'] = lh['time'] - lh_time_shift
```

straxen/plugins/peaklet_processing.py
WenzDaniel: I am not sure if it would be worth adding. I would not really expect to see some for the high energy channels. Though I never tested.

jorana: Yeah, I don't think it's worth it

If the threshold for seeing something something is higher, the chances are that at least similar amount of single PMTs are seeing something right for the HE channels (also because it's only half of the TPC)?

jorana (approved):

PR #624, opened 2021-08-05
Add a check for NaN values when reading corrections

ahiguera-mx:

Based on the discussion regarding ONLINE and OFFLINE values, this PR will check for NaNs and the code will rise an error, this basically means that the user want to process some data where the version is not valid

See this notebook where an example is shown. https://github.com/XENONnT/corrections/blob/master/CMT/examples/Check_NaNs.ipynb

### _Notes on testing_


straxen/corrections_services.py
```diff
@@ -241,6 +249,24 @@ def get_config_from_cmt(self, run_id, model_type, version='ONLINE'):
                              f"Please contact CMT manager and yell at him")
         return file_name
 
+    def check_for_nans(self, when, df, version):
```
jmosbacher, line 252: Please use pandas builtin functions so the code is more readable. eg: `df.loc[:when,version].isnull().any()`

jmosbacher (commented):
Are there any tests written for this?

straxen/corrections_services.py
ahiguera-mx: Hi Yossi,
Great suggestion, thanks! 
(I posted a notebook where I test this, see [here](https://github.com/XENONnT/corrections/blob/master/CMT/examples/Check_NaNs.ipynb) )

WenzDaniel:
Hej guys are you planning to add those tests mentioned by Yossi? What is the status here?

ahiguera-mx:
Not sure what kind of test you guys have in mind, can you elaborate on that

jmosbacher:
@ahiguera-mx I was asking about automated tests that people can run when they make changes to the code to verify they haven't broken anything. In this kind of thing I think you would not need a database, just some pandas dataframes to test the logic under various scenarios. The fidelity of the database retrieval should be tested separately.

straxen/corrections_services.py
```diff
@@ -131,13 +132,21 @@ def _get_correction(self, run_id, correction, version):
                 for it_correction in pmts: # loop over all PMTs
                     if correction in it_correction:
                         df = self.interface.read(it_correction)
+                        if self.check_for_nans(when, df, version):
```
jmosbacher, line 135: Im wondering if this is the correct place to check for nans. This essentially checks whether any nans exist before the requests time so I would get an error even if there are nans far back in the beginning of the science run that dont effect the run I requested right? maybe only check the period in time that effects the run? e.g after interpolation and selection

ahiguera-mx: I'm not sure we want a "version" with discontinuities(nans). 
If that is the case we will need to define a period for where to check for nans which is totally arbitrary. 
Doing any check after the interpolation does not make sense since DataFrame.Interpolate() fill nans.

jmosbacher: > If that is the case we will need to define a period for where to check for nans which is totally arbitrary.
Im not sure I follow, in the current API a correction is always requested for a specific run and a run defines a very specific time range so why is that arbitrary?

jmosbacher: Why not just select the datapoints closest to the run time range from each side? or even better only pull the relevant data from the database? e.g something like this:
```python
def before(dt):
    return [
    {
    "$match": {
            "time": {
                    "$lte": pd.to_datetime(dt),
                }
        }
    },
    
    {
    "$sort": {"time": 1}
    },
    {
    "$limit": 1
    }
]

def after(dt):
    return [
    {
    "$match": {
            "time": {
                    "$gte": pd.to_datetime(dt), 
                }
        }
    },
    
    {
    "$sort": {"time": -1}
    },
    {
    "$limit": 1
    }
]
def get_correction(db, correction_name, dt):
    before_df = pdm.read_mongo(correction_name, before(dt), db)
    after_df = pdm.read_mongo(correction_name, after(dt), db)
    return pd.concat([before_df, after_df])

```

ahiguera-mx: This is a great idea, I will implemented although this needs to be done at strax. I'm going to close this PR

PR #625, opened 2021-08-05
Speed up event pattern fit

WenzDaniel:
There was a request to improve the performance of our event_pattern_fit-plugin. With the proposed changes the speed increases by a factor of three. See also attached images.

I profiled the performance of the compute method of the plugin and found binomial_test to be the bottleneck. To increase the performance I numbafied the corresponding functions and created some numbafiable scipy-functions based on their cython implementations.  The new bottleneck is now given by `neg2llh_modpoisson`. However, any attempt to increase its performance failed.  

kr83m_run_event_pattern_fit
binom_test_imporved

### DIFF
--- a/straxen/numbafied_scipy.py
+++ b/straxen/numbafied_scipy.py
@@ -1,0 +1,36 @@
+import numba
+import ctypes
+import strax
+from scipy.special.cython_special import loggamma, betainc, gammaln
+
+export, __all__ = strax.exporter()
+# Numba versions of scipy functions:
+# Based on http://numba.pydata.org/numba-doc/latest/extending/high-level.html but
+# omitting the address step.
+# Unfortunately this approach does not allow to cache the resulting numba functions.
+#  Any ideas in this regard would be welcome.
+# All subsequent function will treat their inputs as 64bit floats:
+double = ctypes.c_double
+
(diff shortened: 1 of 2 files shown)


terliuk:
Do we have a check of pattern fit before and after this change? as we do not expect any changes and outputs should be exactly the same, right?

WenzDaniel:
> Do we have a check of event pattern fit values before and after this change? as we do not expect any changes and outputs should be exactly the same, right?

I compared results before and after the speed up and the out put was identical. I have not tested Giovanni's fixes though. (But I thing here the results do change, do not they?)

jorana:
Thanks all involved, is this ready for merging?

WenzDaniel:
From my side yes. But I think @GiovanniVolta wanted to make some simplifications.

GiovanniVolta:
Sorry guys, let me do one last check. I want to be sure everything is fine for the binomial test.

GiovanniVolta:
It looks like everything is fine. The accepted fraction is higher because we fixed [few bugs](https://github.com/XENONnT/straxen/issues/629) in the p-value calculation.

PR_S1AFT

For me, we can proceed with the merge.

Issue #605, opened 2021-07-30
Superruns don't work with event_info_double


jjakob03:
**Describe the bug**
After creating a superrun in straxen, loading it with event_info works fine. Loading it with event_info_double as a target gives an error: `RuntimeError: Could not load event_info_double for subrun 024418 even though we made it??`

**To Reproduce**
Insert the MWE of how to reproduce the error
```python
import cutax
st = cutax.xenonnt_v3()
superrun_name = '_some_KrRuns'
subrun_ids = ['024425', '024423', '024421', '024418']
st.define_run(superrun_name, data=subrun_ids)

df1 = st.get_df('_some_KrRuns', targets='event_info')

df2 = st.get_df('_some_KrRuns', targets='event_info_double')
```

**Expected behavior**
I would expect two dataframes containing the runs in the two different plugins.

**Versions**
```
Working on dali001.rcc.local with the following versions and installation paths:
python	v3.8.10	(default, Jun  4 2021, 15:09:15) [GCC 7.5.0]
strax	v0.16.1	/opt/XENONnT/anaconda/envs/XENONnT_development/lib/python3.8/site-packages/strax
straxen	v0.19.3	/opt/XENONnT/anaconda/envs/XENONnT_development/lib/python3.8/site-packages/straxen
cutax	v0.1.1	/home/jjakob/.local/lib/python3.8/site-packages/cutax
```

jorana:
Looks like `strax.SaveWhen.NEVER` plugins aren't changed to saving when defining the super-run.

It would indeed be intuitive if this is done behind the scenes. To see that this is the underlying issue you can see that this small adaptation would work):

```python
import cutax
import strax
st = cutax.xenonnt_v3()
superrun_name = '_some_KrRuns_saved'
subrun_ids = ['024425', '024423', '024421', '024418']
st.define_run(superrun_name, data=subrun_ids)

# Temp change the saving behavoir
st._plugin_class_registry['event_info_double'].save_when = strax.SaveWhen.ALWAYS

df1 = st.get_df(superrun_name, targets='event_info')

df2 = st.get_df(superrun_name, targets='event_info_double')
```

WenzDaniel:
Solved with https://github.com/AxFoundation/strax/pull/509 and https://github.com/XENONnT/straxen/pull/628

PR #628, opened 2021-08-06
Change to save when explicit

WenzDaniel:
Change plugin save type to address: https://github.com/XENONnT/straxen/issues/605 in combination with https://github.com/AxFoundation/strax/pull/509


jjakob03 (approved):
Works fine, thanks!

PR #630, opened 2021-08-10
Add a plugin for external trigger run on nVeto calibration

mzks:
Time difference plugin from trigger timing for optical calibration of nVeto

One of the optical calibration devices in nVeto, reflectivity monitor is lasers, which monitor the time constant of observed timing from external trigger. The initial `time` of `raw_records_nv` are available as the trigger timing, however, it has been lost in `hitlets_nv`.
I would like to match them without loading again the entire `raw_records_nv` on the notebook. I propose the small plugin to provide the delta time from the trigger timing for `hitlets_nv`.

Use `strax.fully_contained_in()` to find `raw_records_nv` for each `hitlets_nv`.

https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:dsg:nveto:external_trigger_dt5


straxen/plugins/nveto_ext_timings.py
```diff
@@ -0,0 +1,69 @@
...
+    Note:
+        Channel dependence for trigger timing has been ignored. It should be the same,
+        But unfortunetely not.
+    """
+    depends_on = ('raw_records_nv', 'hitlets_nv')
+    provides = 'ext_timings_nv'
+    data_kind = 'hitlets_nv'
+
+    compressor = 'zstd'
+    ends_with = '_nv'
```
WenzDaniel, line 30:
```suggestion
```
You can remove this line. As long as we are not inheriting this plugin it is not needed.

straxen/plugins/nveto_ext_timings.py
```diff
@@ -0,0 +1,69 @@
...
+    __version__ = '0.0.1'
+
+    def infer_dtype(self):
+        dtype = copy(strax.time_dt_fields)
```
WenzDaniel, line 34:
```suggestion
        dtype = []
        dtype += strax.time_dt_fields
```
Like this you do not need the additional import.

straxen/plugins/nveto_ext_timings.py
```diff
@@ -0,0 +1,69 @@
...
+        dtype += [(('Delta time from trigger timing [ns]', 'delta_time'), np.int16),]
+        return dtype
+
+    def setup(self):
+        self.channel_range = self.config['channel_map']['nveto']
+
+    def compute(self, hitlets_nv, raw_records_nv):
+
+        time_range_dtype = copy(strax.time_fields)
```
WenzDaniel, line 43: Same here.
```suggestion
        time_range_dtype = []
        time_range_dtype  += strax.time_fields
```

straxen/plugins/nveto_ext_timings.py
```diff
@@ -0,0 +1,69 @@
...
+        time_range_dtype += [(('PMT channel', 'channel'), np.int16)]
+        rr_time_ranges = np.zeros(len(raw_records_nv), dtype=time_range_dtype)
```
WenzDaniel, line 45: Maybe call it "pulse". Then it is consistent with our naming conventions.

straxen/plugins/nveto_ext_timings.py
WenzDaniel: We could also move the `time_range_dtype` and `rr_time_ranges` in a separate function like `get_calibration_pulses`.  This makes things more readable as the user can decide if he/she want to know how he/she "gets calibration pulses". But is not strictly needed.

straxen/plugins/nveto_ext_timings.py
```diff
@@ -0,0 +1,69 @@
...
+        for ch in np.arange(self.channel_range[0], self.channel_range[1] + 1):
+            fancy_i_ch = hitlets_nv['channel']==ch
+            fancy_i_ch = np.arange(len(fancy_i_ch))[fancy_i_ch]
+            _rr_time_ranges = rr_time_ranges[rr_time_ranges['channel']==ch]
+            _hitlets_nv = hitlets_nv[fancy_i_ch]
+            _rr_index = strax.fully_contained_in(_hitlets_nv, _rr_time_ranges)
+            _t_delta = _hitlets_nv['time'] - _rr_time_ranges[_rr_index]['time']
+
+            for res_i, t_del in zip(fancy_i_ch, _t_delta):
+                ext_timings_nv['delta_time'][res_i] = t_del
```
WenzDaniel, line 67: I would put this code in a separate function such that you can use numba. This should make things noticeable faster.

WenzDaniel (commented):
Hi Keita, thanks a lot I put some comments and change suggestions which you can commit directly if you like. One general request. Could you move this plugin into https://github.com/XENONnT/straxen/blob/master/straxen/plugins/led_calibration.py. In general this plugin could also be used for LED calibrations and in this case we cluster similar things a bit more.

mzks:
Test notebook and warnings by numba

https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:dsg:nveto:external_trigger_dt5

mzks:
Comments: `straxen.contexts.xenonnt_led()` doesn't include `hitlets_nv` which is required to make `ext_timings_nv`.

straxen/contexts.py
```diff
@@ -216,7 +216,7 @@ def xenonnt_led(**kwargs):
         config=st.config,
         storage=st.storage,
         **st.context_config)
-    st.register([straxen.DAQReader, straxen.LEDCalibration])
+    st.register([straxen.DAQReader, straxen.LEDCalibration, straxen.nVetoExtTimings])
```
WenzDaniel, line 219:
```suggestion
    st.register([straxen.DAQReader, straxen.LEDCalibration, straxen.nVetoExtTimings, straxen.nVETOHitlets])
    st.set_config({"coincidence_level_recorder_nv": 1})
```

straxen/contexts.py
```diff
@@ -216,7 +216,8 @@ def xenonnt_led(**kwargs):
         config=st.config,
         storage=st.storage,
         **st.context_config)
-    st.register([straxen.DAQReader, straxen.LEDCalibration])
+    st.register([straxen.DAQReader, straxen.LEDCalibration, straxen.nVetoExtTimings, straxen.nVETOHitlets])
```
WenzDaniel, line 219:
```suggestion
    st.register([straxen.DAQReader, 
                     straxen.LEDCalibration, 
                     straxen.nVETORecorder,
                     straxen.nVETOPulseProcessing,
                     straxen.nVETOHitlets,
                     straxen.nVetoExtTimings, ])
```

straxen/plugins/led_calibration.py
```diff
@@ -141,3 +142,80 @@ def get_area(records, led_window):
...
+@strax.takes_config(
+    strax.Option('channel_map', track=False, type=immutabledict,
+                 help="immutabledict mapping subdetector to (min, max) "
+                      "channel number."),
+)
+class nVetoExtTimings(strax.Plugin):
+    """
+    Plugin which computes the time differnce from external trigger timing to hitlets_nv.
+    The nearlest `time` of `raw_records_nv` before `hitlets_nv` time are used as the
+    external trigger timings.
```
WenzDaniel, line 157: Actually this is not true.

 `Plugin which computes the time differnce from external trigger timing to hitlets_nv.` is wrong. You compute the time difference between a hitlet and the pulse start it was recorded in. But since our digitizer are connected in a daisy chain scheme the pulse start does not correspond to the trigger start. In fact there can be an offset of several 10ish ns.

Strictly speaking this is not true either `The nearlest `time` of `raw_records_nv` before `hitlets_nv` time are used as the
    external trigger timings.` You are using the `time` of the recorded pulse. `raw_records_nv` also includes the higher fragments of a pulse which could be in principle closer to the hitlet.

straxen/plugins/led_calibration.py
```diff
@@ -141,3 +142,80 @@ def get_area(records, led_window):
...
+        self.nv_pmt_stop = self.config['channel_map']['nveto'][1] + 1
+
+
+    def compute(self, hitlets_nv, raw_records_nv):
+
+        rr_nv = raw_records_nv[raw_records_nv['record_i']==0]
+        pulses = np.zeros(len(rr_nv), dtype=self.pulse_dtype())
+        pulses['time'] = rr_nv['time']
+        pulses['endtime'] = rr_nv['time'] + \
+                                    rr_nv['pulse_length'] * rr_nv['dt']
```
WenzDaniel, line 187:
```suggestion
        pulses['endtime'] = rr_nv['time'] + rr_nv['pulse_length'] * rr_nv['dt']
```
I think it is short enough to be put on a single line.

straxen/plugins/led_calibration.py
```diff
@@ -141,3 +142,80 @@ def get_area(records, led_window):
...
+        pulses['channel'] = rr_nv['channel']
+
+        ext_timings_nv = np.zeros_like(hitlets_nv, dtype=self.infer_dtype())
```
WenzDaniel, line 190:
```suggestion
        ext_timings_nv = np.zeros_like(hitlets_nv, dtype=self.dtype)
```
infer_dtype will store your dtype as a property of the plugin hence you can simply call `self.dtype`

straxen/plugins/led_calibration.py
```diff
@@ -141,3 +142,80 @@ def get_area(records, led_window):
...
+        pulse_dtype += strax.time_fields
+        pulse_dtype += [(('PMT channel', 'channel'), np.int16)]
+        return pulse_dtype
+
+
+    @staticmethod
+    @numba.jit
+    def calc_delta_time(ext_timings_nv_delta_time, pulses, hitlets_nv, nv_pmt_start, nv_pmt_stop):
+        # numpy access with fancy index returns copy, not view
+        # This for-loop is required to substitute in one by one
```
WenzDaniel, line 211: Can you convert this comment in a doc string?

straxen/plugins/led_calibration.py
```diff
@@ -141,3 +142,80 @@ def get_area(records, led_window):
...
+        for ch in range(nv_pmt_start, nv_pmt_stop):
+            fancy_i_ch = hitlets_nv['channel']==ch
+            fancy_i_ch = np.arange(len(fancy_i_ch))[fancy_i_ch]
+            _pulses = pulses[pulses['channel']==ch]
+            _hitlets_nv = hitlets_nv[fancy_i_ch]
```
WenzDaniel, line 216:
```suggestion
           mask_hitlets_in_channel = hitlets_nv['channel']==ch
           _hitlets_nv = hitlets_nv[mask_hitlets_in_channel]
            _pulses = pulses[pulses['channel']==ch]
```
Why not doing both in the same way? See also the suggestion below.

straxen/plugins/led_calibration.py
```diff
@@ -141,3 +142,80 @@ def get_area(records, led_window):
...
+            _rr_index = strax.fully_contained_in(_hitlets_nv, _pulses)
+            _t_delta = _hitlets_nv['time'] - _pulses[_rr_index]['time']
+
+            for res_i, t_del in zip(fancy_i_ch, _t_delta):
+                ext_timings_nv_delta_time[res_i] = t_del
```
WenzDaniel, line 221:
```suggestion
            ext_timings_nv_delta_time[mask_hitlets_in_channel] = t_del
```

WenzDaniel (approved):
Hi Keita thanks a lot looks fine to me. Just a few suggestions.

straxen/plugins/led_calibration.py
mzks: > Why not doing both in the same way? See also the suggestion below.
I would like to index array [0, 4, 5] not boolean index [True, False, False, False, True, True] for the later.

straxen/plugins/led_calibration.py
mzks: Thank you. I fixed the comments. I drop my opinion here.
I reported the issue of the DAQ on telecon, but it has not been fixed. In my general thought, it should be connected to one clock with the same length cable when we need synchronisation. In addition, the trigger wasn't took "AND" on the dizitizer clock, it will make +/-8 ns or +/-16 ns offset in each channels. Moreover, it will occured among each mezzanines, actually, it has clearly been observed.
Without additional hardware module, CAEN digitizer has hardware delay offset thus I hope that the pulse start should correspond to the trigger start for future.
In some nVeto telecon, we were agreed to connect the trigger pulse to one available channel, but it hasn't. 

Thus, this is a one possible solution now.

WenzDaniel: We can discuss this in todays telecon.

WenzDaniel:
@ershockley @jorana any idea why https://github.com/XENONnT/straxen/pull/630/checks?check_run_id=3380108607#step:7:74 fails?

jorana:
> 
> 
> @ershockley @jorana any idea why https://github.com/XENONnT/straxen/pull/630/checks?check_run_id=3380108607#step:7:74 fails?

@WenzDaniel you are adding a test that does not check if we have access to utilix secrets that are needed for hitlets nv since it depends on CMT

mzks:
Thank you very much. Should I do something more?

WenzDaniel:
Thanks @mzks and @jorana

mzks:
Thank you very much!

PR #637, opened 2021-08-17
new plot_pmts

GiovanniVolta:
## What does the code in this PR do / what does it improve?
I am trying to include the dead_pmt in ` straxen.plot_pmts` for the [PMT blessing plots](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:analysis:plots:ongoing_blessing#pmt_performances), but the option already present changes the colormap (see attached picture). Furthermore, I didn't find any way to include the x and y labels. 

```
cmap = plt.cm.viridis  
cmaplist = [cmap(i) for i in range(cmap.N)]
cmap = mpl.colors.LinearSegmentedColormap.from_list('Custom cmap', cmaplist, cmap.N)

array = np.random.uniform(low=50, high=60, size=494)
array[0]   = np.nan
array[10]  = np.nan
array[50]  = np.nan
array[100] = np.nan
array[150] = np.nan
array[170] = np.nan
array[200] = np.nan
array[280] = np.nan
array[310] = np.nan
array[350] = np.nan
array[400] = np.nan
array[450] = np.nan

bounds = np.linspace(np.nanmin(array), np.nanmax(array), 10, dtype=int)
norm = mpl.colors.BoundaryNorm(bounds, cmap.N)

straxen.plot_pmts(array,
                  cmap=cmap, norm=norm, dead_pmts=[0, 10, 50, 100, 150, 170, 200, 280, 310, 350, 400, 450])
```
SPE_old_plot_pmts

## Can you briefly describe how it works?
I have added the [option](https://github.com/XENONnT/straxen/blob/24e23873b2ba0bec0fc0d09ff28a92986ab4fec5/straxen/matplotlib_utils.py#L19) for showing the x and y labels, and I have [changed](https://github.com/XENONnT/straxen/blob/24e23873b2ba0bec0fc0d09ff28a92986ab4fec5/straxen/matplotlib_utils.py#L139) how the dead pmts are shown. 
```
# With new plot_pmts map
straxen.plot_pmts(array, show_axis_labels=True,
                  cmap=cmap, norm=norm, dead_pmts=[0, 10, 50, 100, 150, 170, 200, 280, 310, 350, 400, 450])
```
SPE_new_plot_pmts

_Please include the following if applicable:_
  - [x] _Update the docstring(s)_
  - [x] _Update the documentation_
  - [x] _Tests to check the (new) code is working as desired._
  - [ ] _Does it solve one of the open issues on github?_

PR #638, opened 2021-08-19
pass on mappable

jorana:
Very small tweak to https://github.com/XENONnT/straxen/pull/637 to use the mapable instead of plt.text to fix the issue with the dead-pmts described there.


GiovanniVolta:
I have tested and it works, thank you Joran!

PR #640, opened 2021-08-20
Classify peak built from merged S2s

zhut19:
This pr aims to mitigate the effect of misclassification when the merge S2 plugin merges a normal S1 with a small S2 and labels the new peak S2 regardless of the peak rise time. Find details in [this note](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:yue:classification_parameters#performance_on_regular_radon_run) 

We now add a new plugin that redoes the classification for merged S2s.

### DIFF
--- a/straxen/plugins/peaklet_processing.py
+++ b/straxen/plugins/peaklet_processing.py
@@ -537,7 +537,7 @@ class PeakletClassificationHighEnergy(PeakletClassification):
 
 
 @export
-class MergedS2sClassification(straxen.PeakletClassification):
+class MergedS2sClassification(PeakletClassification):
     """Classify merged_s2s as unknown, S1, or S2."""
     provides = 'merged_s2s_classification'
     depends_on = ('merged_s2s',)
--- a/straxen/plugins/peaklet_processing.py
+++ b/straxen/plugins/peaklet_processing.py
@@ -537,6 +537,18 @@ class PeakletClassificationHighEnergy(PeakletClassification):
 
 
 @export
+class MergedS2sClassification(straxen.PeakletClassification):
+    """Classify merged_s2s as unknown, S1, or S2."""
+    provides = 'merged_s2s_classification'
+    depends_on = ('merged_s2s',)
+    __version__ = '0.0.1'
+    child_plugin = True
+
+    def compute(self, merged_s2s):
+        return super().compute(merged_s2s)
+
+
+@export
 @strax.takes_config(
     strax.Option('s2_merge_max_duration', default=50_000,
                  help="Do not merge peaklets at all if the result would be a peak "
@@ -715,7 +727,7 @@ class Peaks(strax.Plugin):
     (replacing all peaklets that were later re-merged as S2s). As this
     step is computationally trivial, never save this plugin.
     """
-    depends_on = ('peaklets', 'peaklet_classification', 'merged_s2s')
+    depends_on = ('peaklets', 'peaklet_classification', 'merged_s2s', 'merged_s2s_classification')
     data_kind = 'peaks'
     provides = 'peaks'
     parallel = True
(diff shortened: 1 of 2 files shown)


straxen/plugins/peaklet_processing.py
```diff
@@ -536,6 +536,18 @@ def compute(self, peaklets_he):
         return super().compute(peaklets_he)
 
 
+@export
+class MergedS2sClassification(PeakletClassification):
+    """Classify merged_s2s as unknown, S1, or S2."""
+    provides = 'merged_s2s_classification'
+    depends_on = ('merged_s2s',)
+    __version__ = '0.0.1'
+    child_plugin = True
```
WenzDaniel, line 545:
```suggestion
    child_plugin = True
    save_when = strax.SaveWhen.EXPLICIT
```
I guess this data_type is computed quickly, is not it? Maybe we can make computing's life a bit easier if we make it  a SaveWhen Explicit data_type. @ershockley would this help, or do you have to track the lineage of plugins which are not returning data by default, too? 
I am wondering if we should not do the same for PeakletClassification too. Only a few people will look at peaklet level data anyhow.

WenzDaniel (approved):
Looks fine. Beside my comment in the code I have one more commen:

Comment:

Do not forget to add the plugin to the nT and 1T context. This will trigger reprocessing for both detectors. Further, will we lose backwards compatibility to previous data using fuzzy_for.

jorana:
@zhut19 , I'm wondering if this would be better than not merging these non-S2s in the first place? Would e.g. [this wf](https://xe1t-wiki.lngs.infn.it/lib/exe/detail.php?id=xenon%3Axenonnt%3Ayue%3Aclassification_parameters&media=xenon:xenonnt:yue:merge_s2s1.png) be how we'd want to be doing our clustering for such an S1, including great deal of hits into it? 

An alternative approach would be to prevent adding all kinds of S2-peaklets to an S1-peaklet in the mergedS2s plugin as, besides the classification (what you are doing now) also the clustering is off in [this wf](https://xe1t-wiki.lngs.infn.it/lib/exe/detail.php?id=xenon%3Axenonnt%3Ayue%3Aclassification_parameters&media=xenon:xenonnt:yue:merge_s2s1.png). I think we did not want to reclassify after the MergedS2s re-clustering, otherwise it's kind of defeating the purpose of that plugin, right?

WenzDaniel:
>  I'm wondering if this would be better than not merging these non-S2s in the first place? Would e.g. this wf be how we'd want to be doing our clustering for such an S1, including great deal of hits into it?

I think the effect on the S1 is negligible is not it? Once https://github.com/XENONnT/straxen/pull/601 and https://github.com/XENONnT/straxen/pull/623 are merged the bias on area is below half a percent and the additional hits wont contribute to the tight coincidence level. And I also hope that re-classification tags the peak as a S1 peak. 

I think from a physics point of view it can be that a small sized S1 belongs actually to an S2 signal as there is almost no difference in the scintillation process. Maybe we can add here some sort of threshold? To include only S1 signals up to a certain fraction of SE's?

WenzDaniel:
> I think we did not want to reclassify after the MergedS2s re-clustering, otherwise it's kind of defeating the purpose of that plugin, right?

But I do not think that the current approach is correct either and after seeing your example I am also not convinced that simply hard coding everything to be an S2 is correct either.

straxen/plugins/peaklet_processing.py
zhut19: @WenzDaniel, now the test fails at plot waveform when it asks for peak data. Maybe go back to savewhen always if it's not too much a fuss?

ershockley: We track the lineage of all data types, but it definitely would be nice to not have so many datasets running around. I would much prefer save_when to be target or explicit if possible

PR #641, opened 2021-08-24
Fix veto event positions

WenzDaniel:
Tiny fix for a floating point issue in `compute_positions` moved previous line 320 before if-statement. Additionally, I updated the corresponding test to allow for negative hitlet areas and did some refactoring.


ershockley (approved):
Thanks @WenzDaniel!

PR #642, opened 2021-08-25
Break up merge failing natural break

zhut19:
This is one potential way of adding the natural break criteria to S2 merging so that we might improve the splitting of multiple scatters S2s.

This is a rather simple solution to adding the nb, as we can run the merged S2s through the splitting. And use the time information of the split peak to find the overlapping peaklets and build the merged peaks again from those peaklets. The `resolve_merge_split_conflict` function also checks if any peaklet is shared between two merged S2s, in which case the peaklet will be merged to the side containg more area. This is to avoid any split of peaklet, otherwise, we need to compute values like area per channel.


JoranAngevaare:
@zhut19 I'm closing this stale PR, please feel free to reopen if needed

PR #645, opened 2021-08-26
Merge s2 without s1

zhut19:
The current merge s2 plugin includes S1 during the merging, which leads to lower efficiency in reconstructing S1s. Following the discussion [on wiki](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenon1t:tzhu:fixs2merge), we decide to add the option of bypassing S1 during the merging.


WenzDaniel (approved):
Okay I checked everything and I did not spot any obvious mistakes or issues.

PR #646, opened 2021-08-30
Fix n hits

WenzDaniel:

Remove +=1 in n_hits counter introduced in https://github.com/XENONnT/straxen/pull/126 as it is obviously wrong. I cannot reconstruct why we thought back then that it must be added. Also added a small unity test for the future. 

Thanks to @FaroutYLq for spotting.


FaroutYLq (approved):
Seems good to me now. Thanks Daniel for fixing this!

PR #664, opened 2021-09-14
Previous S2 Shadow Plugin draft

chnlkx:

This plug in could compute the previous s2 shadow variable, which could be used to reduce the isolated S1 and S2 events.

we define the shadow parameter as [previous s2 area/ time difference].
To calculate this, we loop over the previous 20 events (about 3 seconds given the event rate now)
 and take the maximum as the shadow value.

here is an example of how to use the variable

https://github.com/XENONnT/analysiscode/blob/master/AC_BG/shadow_plugin_test.ipynb

### _Notes on testing_


skazama:
@chnlkx Hi Kexin, thanks for starting this PR for the S2Shadow cut. I have a question on the definition of previous S2. It seems that previous S2 is based on "event", not "peaks". Do you think it should be changed baed on peaks In case if we miss S1s for some reason? maybe @jingqiangye or @l-althueser have a suggestion?

WenzDaniel:
Hi @chnlkx I also agree with Shingo that the plugin should depend on peaks and events. In addition instead of always taking the last 20 events I think we should make things a bit more reproducible by taking a fixed time window. If you need any help by modifying your plugin let me know via slack.

skazama:
@chnlkx 
Hi Kexin, thanks for the update. I just checked your code and modified some typos. [Here](https://github.com/XENONnT/analysiscode/blob/master/AC_BG/check_plugin.ipynb) you can see my notebook for your plugin test. I managed to create event_shadow plugin and showed s2_area vs s2_shadow in the notebook.

Can you also add some variables like position difference between current s2 and the other s2 used for S2shadow cut calculation?

straxen/plugins/event_shadow.py
```diff
@@ -0,0 +1,80 @@
...
+                ]
+        dtype += strax.time_fields
+        return dtype
+
+    def compute(self, events, peaks):
+        #set a time window to look for peaks
+        roi_dt = np.dtype([(('back in time', 'time'), int),
+                                         (('till it begin','endtime'), int)])
+        roi = np.zeros(len(events), dtype=roi_dt)   
+        n_seconds = int(3e9)
```
skazama, line 43: Do you have any study on this time window (3 seconds)?

straxen/plugins/event_shadow.py
```diff
@@ -0,0 +1,80 @@
...
+        pre_s2_x = np.zeros(len(events))
+        pre_s2_y = np.zeros(len(events))
+        shadow = np.zeros(len(events)) 
+        shadow_distance = np.zeros(len(events))
+        
+        #loop over interested peaks for each event
+        for event_i, event_a in enumerate(events):
+            for peak_i, peak_a in enumerate(split_try[event_i]):
+                new_shadow = 0
+                if  (peak_a['area']>1000)&(peak_a['type']==2):
```
skazama, line 61: Do you also have any studies on this area threshold (1000 pe)?

straxen/plugins/event_shadow.py
chnlkx: Yes,  I check the shadow_dt parameter(time to the previous peak), 3 seconds is a conservative choice.
I check this by setting the time window as 5 seconds and find 95% of 'shadow_dt' within 2 seconds, and 98% of 'shadow_dt' within 3 seconds.

chnlkx: Yes,  I check the shadow_dt parameter(time to the previous peak), 3 seconds is a conservative choice.
I check this by setting the time window as 5 seconds and find 95% of 'shadow_dt' within 2 seconds, and 98% of 'shadow_dt' within 3 seconds.

straxen/plugins/event_shadow.py
chnlkx: We set this area threshold mainly to prevent SE and lone S2s near the event from causing strange behavior. And most of these types of lone events are below 1000 PE. But this 1000 threshold is not a strict one, we might need to tune it as we have a further understanding of lone S2s.

straxen/plugins/event_shadow.py
```diff
@@ -0,0 +1,80 @@
...
+import straxen
+
+
+straxen.print_versions(('strax', 'straxen', 'cutax'))
+import warnings
+export, __all__ = strax.exporter()
+
+import warnings
+export, __all__ = strax.exporter()
+
```
skazama, line 14: You can remove these lines since it is defined above too.

straxen/plugins/event_shadow.py
skazama: Do you have some notes related to this?

skazama:
Can you register your plugin around here in the context?:
https://github.com/XENONnT/straxen/blob/master/straxen/contexts.py#L97

straxen/plugins/event_shadow.py
WenzDaniel: > But this 1000 threshold is not a strict one,

Any threshold window length etc. which might be subject of change (which is basically always true), should be a strax option. See for example one of the other [plugins](https://github.com/XENONnT/straxen/blob/7f66f660326925bd7398137fbb2cf78ab07d9c8b/straxen/plugins/peak_processing.py#L255-L265).

straxen/plugins/event_shadow.py
```diff
@@ -0,0 +1,80 @@
+import numpy as np
+import pandas as pd
+
+import strax
+import straxen
+
+
+straxen.print_versions(('strax', 'straxen', 'cutax'))
```
WenzDaniel, line 8: This line should be deleted.

straxen/plugins/event_shadow.py
```diff
@@ -0,0 +1,80 @@
...
+import warnings
+export, __all__ = strax.exporter()
+
+import warnings
+export, __all__ = strax.exporter()
+
+@export
+class Shadow(strax.Plugin):
+    """Compute several new parameters to describe the noise level.
+        We can use these to reduce AC BG. """
```
WenzDaniel, line 18: You should also updated the doc string here.

straxen/plugins/event_shadow.py
```diff
@@ -0,0 +1,80 @@
...
+
+straxen.print_versions(('strax', 'straxen', 'cutax'))
+import warnings
+export, __all__ = strax.exporter()
+
+import warnings
+export, __all__ = strax.exporter()
+
+@export
+class Shadow(strax.Plugin):
```
WenzDaniel, line 16: Can we call it EventShadow, then it is more clear at which data-level we are.

straxen/plugins/event_shadow.py
WenzDaniel: This should be an option too.

straxen/plugins/event_shadow.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def compute(self, events, peaks):
+        #set a time window to look for peaks
+        roi_dt = np.dtype([(('back in time', 'time'), int),
+                                         (('till it begin','endtime'), int)])
+        roi = np.zeros(len(events), dtype=roi_dt)   
+        n_seconds = int(3e9)
+        roi['time'] = events['time'] - n_seconds
+        roi['endtime'] = events['time']
+        split_try = strax.split_touching_windows(peaks, roi, window=0)
```
WenzDaniel, line 46: I think it is already by default `window=0` so you can remove the keyword argument.

straxen/plugins/event_shadow.py
```diff
@@ -0,0 +1,80 @@
...
+                    new_shadow = peak_a['area']/(event_a['time']-peak_a['center_time'])
+
+                if new_shadow > shadow[event_i]:
+                    pre_s2_area[event_i] = peak_a['area']
+                    shadow_dt[event_i] = event_a['time']-peak_a['center_time']
+                    pre_s2_x[event_i] = peak_a['x']
+                    pre_s2_y[event_i] = peak_a['y']
+                    shadow[event_i] = new_shadow
+                    
+        shadow_distance = ((pre_s2_x - events['s2_x'])**2+(pre_s2_y - events['s2_y'])**2)**0.5
```
WenzDaniel, line 71: This part we should numbafy. If you have time either later today or tomorrow. We can have a chat via zoom and I show you how to do it.

straxen/plugins/event_shadow.py
```diff
@@ -0,0 +1,80 @@
...
+
+        return dict(time=events['time'],
+            endtime=strax.endtime(events),
+                    shadow = shadow,
+                    pre_s2_area = pre_s2_area,
+                    shadow_dt = shadow_dt,
+                    pre_s2_x = pre_s2_x,
+                    pre_s2_y = pre_s2_y,
+                    shadow_distance = shadow_distance)
```
WenzDaniel, line 80: Here we need an additional empty line at the end of the file.

straxen/plugins/event_shadow_test.py
```diff
@@ -0,0 +1,43 @@
+import numpy as np
```
WenzDaniel, line 1: Why do we need this plugin?

WenzDaniel (commented):
Here are some comments. I see why it is so slow, but I think we can seed it up easily.

straxen/plugins/event_shadow_test.py
chnlkx: I should have deleted this file. This is a very early version.

straxen/plugins/event_shadow.py
chnlkx: That would be great! I'm just worried about how to do that. Let's have a chat tomorrow when you are available. Thanks a lot😊

straxen/plugins/event_shadow.py
```diff
@@ -0,0 +1,70 @@
...
+
+        return res
+    
+
+@numba.njit(cache=True)
+def compute_shadow(events, split_peaks, res):
+    for event_i, event_a in enumerate(events):
+        new_shadow = 0
+        for peak_i, peak_a in enumerate(split_peaks[event_i]):
+            new_shadow = peak_a['area']/(event_a['time']-peak_a['center_time'])
```
skazama, line 64: maybe `event_a['time']` should be `event_a['s2_center_time']`?

skazama:
Can you register your plugin around in the context.py?:
https://github.com/XENONnT/straxen/blob/master/straxen/contexts.py#L97

straxen/plugins/event_shadow.py
```diff
@@ -0,0 +1,70 @@
...
+    def compute(self, events, peaks):
+        #set a time window to look for peaks
+        roi_dt = np.dtype([(('back in time', 'time'), int),
+                                         (('till it begin','endtime'), int)])
+        roi = np.zeros(len(events), dtype=roi_dt)   
+        n_seconds = self.config['time_window_backward']
+        roi['time'] = events['time'] - n_seconds
+        roi['endtime'] = events['time']
+        #set interested S2
+        mask_s2 = peaks['type']==2
```
WenzDaniel, line 44: Also here you usually have space between operator `==` and parameters. You only do not use any spaces if it is the argument in a function.

straxen/plugins/event_shadow.py
```diff
@@ -0,0 +1,70 @@
...
+        mask_pre_s2 = peaks['area'] > self.config['pre_s2_area_threshold']   
+        split_peaks = strax.split_touching_windows(peaks[mask_s2 & mask_pre_s2], roi)
```
WenzDaniel, line 46: You can simplify this a bit by using the same mask for both e.g. something like: 

```
mask_s2 = ....
mask_s2 &= peaks['area'] ...
strax.split..... (peaks[mask_s2], ....)
```

straxen/plugins/event_shadow.py
```diff
@@ -0,0 +1,70 @@
...
+    
+        #define the variables we want to calculate
+        res = np.zeros(len(events), self.dtype)       
+        #calculate all the variables
```
WenzDaniel, line 50: I think you can remove the comments they do not add much information.

straxen/plugins/event_shadow.py
WenzDaniel: Yes from reading the code I would suggest the same. If you change it here you also want to probably change also later your `shadow_dt`.

straxen/plugins/event_shadow.py
```diff
@@ -0,0 +1,70 @@
...
+    strax.Option('time_window_backward', default=int(3e9),
+                 help='Search for S2s causing shadow in this time window'))
+class EventShadow(strax.Plugin):
+    """
+    This plugin can find and calculate the previous S2 shadow at event level,
+    with time window backward and previous S2 area as options.
+    It also gives the area and position infomation of these previous S2s.
+    """
+    depends_on = ('event_basics','peak_basics','peak_positions')
+    parallel = False
```
WenzDaniel, line 19: This line you can remove.

straxen/plugins/event_shadow.py
```diff
@@ -0,0 +1,70 @@
...
+    provides = 'event_shadow'
+    __version__ = '0.0.6'
+
```
WenzDaniel, line 22: One question. I guess this plugin will mainly be needed to optimize the corresponding cut is not it? Since the plugin is now reasonably fast I would suggest that we make it a `save_when = strax.SaveWhen.EXPLICIT`-plugin. In that case the output is only stored if asked explicitly.

WenzDaniel (approved):
Looks fine to me. Just a few smaller comments. Please do not forget to register your plugin in the context as Shingo suggested. Just let me know if you need any help.

skazama:
@chnlkx sorry, I forgot to ask you to update [this](https://github.com/chnlkx/straxen/blob/new_kx/straxen/plugins/__init__.py) too. Can you please add your plugin there?

PR #667, opened 2021-09-16
fix some warnings from testing

JoranAngevaare:
Fix some warnings from testing. Most importantly, fix a typo for multiprocessing testing.

### DIFF
--- a/tests/test_cmt.py
+++ b/tests/test_cmt.py
@@ -9,12 +9,13 @@ from .test_basics import test_run_id_1T
 from .test_plugins import test_run_id_nT
 from straxen.common import aux_repo
 
+
 def test_connect_to_db():
     """
     Test connection to db
     """
     if not straxen.utilix_is_configured():
-        warn('Cannot do test becaus  '
+        warn('Cannot do test because  '
              'no have access to the database.')
         return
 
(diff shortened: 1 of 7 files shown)


tests/test_cmt.py
```diff
@@ -9,12 +9,13 @@
...
+
 def test_connect_to_db():
     """
     Test connection to db
     """
     if not straxen.utilix_is_configured():
-        warn('Cannot do test becaus  '
+        warn('Cannot do test because  '
              'no have access to the database.')
         return
```
WenzDaniel, line 20: I am wondering if we should not make this whole thing a utility function. We require it for quite a many tests.

tests/test_plugins.py
```diff
@@ -279,7 +279,7 @@ def test_nT(ncores=1):
     offline_gain_model = ("to_pe_placeholder", True)
     _update_context(st, ncores, fallback_gains=offline_gain_model, nt=True)
     # Lets take an abandoned run where we actually have gains for in the CMT
-    _run_plugins(st, make_all=True, max_wokers=ncores, run_id=test_run_id_nT)
+    _run_plugins(st, make_all=True, max_workers=ncores, run_id=test_run_id_nT)
```
WenzDaniel, line 282: Oh wow indeed quite a bug. Nice catch!

WenzDaniel (approved):
Looks fine. Just one comment which you may want to consider. We have here a bit of duplicated code regarding utilix_is_configured. We could also make it a decorator at some point.

tests/test_cmt.py
JoranAngevaare: Thanks, I think that makes sense, want to check if my changes are similar to what you suggested?

PR #669, opened 2021-09-20
Add posdiff plugin

boblsc-pd:

This creates a variable with x,y mean and std for position reconstruction algorithm used.  See note: https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:shengchao:sr0:reconstruction_quality_update
Look for pos recon algs and get the mean and STD for s2/alt_s2 in an event. 
I can also provide the notebook to make this plot. 

### _Notes on testing_

### DIFF
--- a/straxen/plugins/position_reconstruction.py
+++ b/straxen/plugins/position_reconstruction.py
@@ -302,3 +302,4 @@ class S2ReconPosDiff(strax.Plugin):
                 result[t_+'_recon_pos_diff'][cur_s2_bool] = r_std
                 result[t_+'_recon_avg_x'][cur_s2_bool] = x_avg
                 result[t_+'_recon_avg_y'][cur_s2_bool] = y_avg
+                
--- a/straxen/plugins/position_reconstruction.py
+++ b/straxen/plugins/position_reconstruction.py
@@ -295,4 +295,3 @@ class S2ReconPosDiff(strax.Plugin):
                 result[peak_type + '_recon_pos_diff'][cur_s2_bool] = r_std
                 result[peak_type + '_recon_avg_x'][cur_s2_bool] = x_avg
                 result[peak_type + '_recon_avg_y'][cur_s2_bool] = y_avg
-                
--- a/straxen/plugins/position_reconstruction.py
+++ b/straxen/plugins/position_reconstruction.py
@@ -4,6 +4,7 @@ import os
 import tempfile
 import tarfile
 import numpy as np
+import numba
 import strax
 import straxen
 from warnings import warn
@@ -294,4 +295,4 @@ class S2ReconPosDiff(strax.Plugin):
                 result[t_+'_recon_avg_x'][cur_s2_bool] = x_avg
                 result[t_+'_recon_avg_y'][cur_s2_bool] = y_avg
                 
-                
+
--- a/straxen/plugins/position_reconstruction.py
+++ b/straxen/plugins/position_reconstruction.py
@@ -253,8 +253,8 @@ class S2ReconPosDiff(strax.Plugin):
     def cal_avg_and_std(self, values, axis = 1):
         
         average = np.mean(values, axis = axis)
-        variance = np.var(values, axis = axis)
-        return (average, np.sqrt(variance))
+        std = np.std(values, axis = axis)
+        return average, std
 
     def eval_recon(self, data, name_x_list, name_y_list, cur_s2_bool):
     
--- a/straxen/plugins/position_reconstruction.py
+++ b/straxen/plugins/position_reconstruction.py
@@ -256,10 +256,10 @@ class S2ReconPosDiff(strax.Plugin):
         std = np.std(values, axis = axis)
         return average, std
 
-    def eval_recon(self, data, name_x_list, name_y_list, cur_s2_bool):
-    
-        x_avg, x_std = self.cal_avg_and_std(np.array(data[name_x_list][cur_s2_bool].tolist())) #lazy fix to delete field name in array 
-        y_avg, y_std = self.cal_avg_and_std(np.array(data[name_y_list][cur_s2_bool].tolist()))
+    def eval_recon(self, data, name_x_list, name_y_list):
+       """A short description of the function might be usefull"""
+        x_avg, x_std = self.cal_avg_and_std(np.array(data[name_x_list].tolist())) #lazy fix to delete field name in array 
+        y_avg, y_std = self.cal_avg_and_std(np.array(data[name_y_list].tolist()))
         r_std = np.sqrt( x_std**2 + y_std**2 )
         res = x_avg, y_avg, r_std
         return res
--- a/straxen/plugins/position_reconstruction.py
+++ b/straxen/plugins/position_reconstruction.py
@@ -257,8 +257,9 @@ class S2ReconPosDiff(strax.Plugin):
         return average, std
 
     def eval_recon(self, data, name_x_list, name_y_list):
-       """A short description of the function might be usefull"""
-        x_avg, x_std = self.cal_avg_and_std(np.array(data[name_x_list].tolist())) #lazy fix to delete field name in array 
+       """This function reads the name list based on s2/alt_s2 and all recon algorithm registered """
+       """Each row consists the reconstructed x/y and their average and standard deviation is calculated"""
+        x_avg, x_std = self.cal_avg_and_std(np.array(data[name_x_list].tolist())) #lazy fix to delete field name in array, otherwise np.mean will complain 
         y_avg, y_std = self.cal_avg_and_std(np.array(data[name_y_list].tolist()))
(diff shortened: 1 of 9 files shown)


skazama:
@boblsc-pd  Hi Shengchao, thanks for the PR. As we discussed in a slack chat, I had a problem when running your plugin, but the problem disappeared  if I comment out  `@numba.njit`. Maybe some experts like @WenzDaniel can comment?

WenzDaniel:
@boblsc-pd to test which function is the offending one could you try to comment out the numba.njit decorator like Shingo did in a sequential manner. Fell free to also write to me in slak ksuch that I can help you debugging your code.  I am afraid you are using many things which are not actually supported by numba. In addition is the compute-method neither a staticmethod nor can it be numba jitted.

JoranAngevaare:
@skazama @WenzDaniel @boblsc-pd numba needs to be imported 😉  (76697be)

skazama:
@JoranAngevaare Yes, when debugging his code, I manually imported numba, but the error still remained...

JoranAngevaare:
I see, you might now be able 

> 
> 
> @JoranAngevaare Yes, when debugging his code, I manually imported numba, but the error still remained...

Ah I see, you might be able to get the info form the various tests that are now running:
- https://github.com/XENONnT/straxen/pull/669/checks?check_run_id=3697431591  (numba disabled)
- https://github.com/XENONnT/straxen/pull/669/checks?check_run_id=3697431649 (numba enabled)

boblsc-pd:
Hi all,  thank you for all the helpful suggestions. My initial test used the same run as I was developing the cut, so the bug escaped my radar, as I was just loading some saved files without having the bug. Sorry! 
I will look more into numba.

WenzDaniel:
Sure no problem if you need any help or advice just ping me in slack.

straxen/plugins/position_reconstruction.py
```diff
@@ -203,3 +204,95 @@ def compute(self, peaks):
...
+            for name in self.config['recon_alg_included']:
+                cur_s2_bool &= ~np.isnan(events[t_+'_x'+name])
+                cur_s2_bool &= ~np.isnan(events[t_+'_y'+name])
+            
+            # default value is nan, it will be ovewrite if the event satisfy the requirments
+            result[t_+'_recon_pos_diff'][:] = np.nan
+            result[t_+'_recon_avg_x'][:] = np.nan
+            result[t_+'_recon_avg_y'][:] = np.nan
+            
+            if np.sum(cur_s2_bool):
```
WenzDaniel, line 291: I think you can use `if np.any()` here is not it? This would be faster since you are returning as soon True as the first element is True. While in case of a sum you always must loop over everything.

straxen/plugins/position_reconstruction.py
```diff
@@ -203,3 +204,95 @@ def compute(self, peaks):
...
+        variance = np.var(values, axis=axis)
+        return (average, np.sqrt(variance))
+
+    def eval_recon(self, data, name,cur_s2_bool):
+        alg_list=self.config['recon_alg_included']
+        name_x_list=[]
+        name_y_list=[]
+        for alg in alg_list:
+            name_x_list.append(name+'_x'+alg)
+            name_y_list.append(name+'_y'+alg)
```
WenzDaniel, line 266: This is a bit of duplication with `compute_pos_diff` would not it be simpler to compute the field names once in the beginning and to store them in a variable?

straxen/plugins/position_reconstruction.py
```diff
@@ -203,3 +204,95 @@ def compute(self, peaks):
...
+        x_avg,x_std= self.cal_avg_and_std(np.array(data[name_x_list][cur_s2_bool].tolist())) #lazy fix for data types 
+        y_avg,y_std= self.cal_avg_and_std(np.array(data[name_y_list][cur_s2_bool].tolist()))
```
WenzDaniel, line 268: Those lines I do not understand `data` aka `events` should be already an array is not it. I do not understand why you transform it here back and forth.

straxen/plugins/position_reconstruction.py
```diff
@@ -203,3 +204,95 @@ def compute(self, peaks):
...
+        
+        result = np.zeros(len(events), dtype=self.dtype)
+        result['time'] = events['time']
+        result['endtime'] = strax.endtime(events)
+        # Computing position difference
+        self.compute_pos_diff(events, result)
+        return result  
+
+    def cal_avg_and_std(self, values, axis=1):
+        average = np.average(values,axis=axis)
```
WenzDaniel, line 256: Average is usually associated with a weighted mean. Since you are not using any weights here I would use for clarity `np.mean` instead.

WenzDaniel (commented):
@boblsc-pd nice PR. I have a few comments and many style suggestions. We are trying to keep some standard in strax/straxen to improve readability. I made some suggestions but there are many more things I found. Could you please check if

- you put commas in the correct way like `a, b` and not `a ,b` or `a,b`
- Have an empty space around operators.  
- Have a single empty line at the end of your code.

straxen/plugins/position_reconstruction.py
boblsc-pd: np.average or np.mean does not use np.array with filed name as an input. After this operation, the filed name was neglected which it gets around the error.

straxen/plugins/position_reconstruction.py
boblsc-pd: Good point, changed.

JoranAngevaare (commented):
Thanks @boblsc-pd , looks good, thank you for your work! 

Apart from a few minor comments/suggestions below, I also have one general question
 - Could you register your plugin around here in the context:
   https://github.com/XENONnT/straxen/blob/master/straxen/contexts.py#L78

straxen/plugins/position_reconstruction.py
```diff
@@ -203,3 +204,96 @@ def compute(self, peaks):
...
+        result['endtime'] = strax.endtime(events)
+        # Computing position difference
+        self.compute_pos_diff(events, result)
+        return result  
+
+    def cal_avg_and_std(self, values, axis = 1):
+        
+        average = np.mean(values, axis = axis)
+        variance = np.var(values, axis = axis)
+        return (average, np.sqrt(variance))
```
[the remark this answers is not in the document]
boblsc-pd: No, it should be the same.

straxen/plugins/position_reconstruction.py
```diff
@@ -203,3 +204,96 @@ def compute(self, peaks):
...
+            
+            if np.any(cur_s2_bool):
+                name_x_list = []
+                name_y_list = []
+                for alg in alg_list:
+                    name_x_list.append(peak_type + '_x' + alg)
+                    name_y_list.append(peak_type + '_y' + alg)
+
+                # Calculating average x,y, and position difference
+                x_avg, y_avg, r_std = self.eval_recon(events, name_x_list, name_y_list, cur_s2_bool)
```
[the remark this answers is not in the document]
boblsc-pd: Thank you for your comments!  All suggested changes has been made.

JoranAngevaare:
Thanks @boblsc-pd , it looks good! Thanks for your work 👍 

Let's double check if the tests run the code 😄

JoranAngevaare (approved):

straxen/plugins/position_reconstruction.py
WenzDaniel: `np.mean(some_structure_array['some_field_name'])` should work.

WenzDaniel (approved):
Looks now fine to me too. Maybe one last question have you tested how long it takes to compute this data type for a single background and a single calibration run?

straxen/plugins/position_reconstruction.py
JoranAngevaare: @WenzDaniel , if I understand the problem correctly, this is the issue (there are multiple fields):
```python
>>> import numpy as np
>>> a= np.ones(2, dtype=[('bla', float), ('foo', float)])
>>> a['foo'] = [2, 2]
>>> a
array([(1., 2.), (1., 2.)], dtype=[('bla', '<f8'), ('foo', '<f8')])
>>> np.mean(a)
  File "<stdin>", line 1, in <module>
  File "<__array_function__ internals>", line 5, in mean
  File "/home/joran/miniconda3/lib/python3.9/site-packages/numpy/core/fromnumeric.py", line 3372, in mean
    return _methods._mean(a, axis=axis, dtype=dtype,
  File "/home/joran/miniconda3/lib/python3.9/site-packages/numpy/core/_methods.py", line 160, in _mean
    ret = umr_sum(arr, axis, dtype, out, keepdims)
TypeError: cannot perform reduce with flexible type
>>> np.mean(a.tolist())
1.5
```

boblsc-pd:
@WenzDaniel The run time is on microsecond scale

PR #671, opened 2021-09-28
S1 posrec events

matteoguida:
This PR adds **S1-based position reconstruction** methods at event level. 

It was decided that implementing it at peak level was not necessary. The script is copied from the [standard position reconstruction](https://github.com/XENONnT/straxen/blob/master/straxen/plugins/position_reconstruction.py), so that different models can be menaged in the future. 

At the current stage only the **CNN data-driven** (see [here](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:analysis:reconstruction_team:s1_xyz_rec_with_data_driven)) is included. 

The **runtime** per event should be bounded by more than one order of magnitude by "event_area_per_channel" and it results now in ≈ 6 ms per event. 
 
Possible applications: improve event quality selections, to map possible charge insensitive regions
of the detector and to analyse events without a reliable S2 signal.

Loaded the trained model, for each event the (x,y,z) position in cm is returned taking as input the PMT hit patterns of the S1 signal, after that they have been normalized by dividing by the channel with maximum area. 

Example of predition:
example_prediction

### _Notes on testing_

### DIFF
--- a/straxen/plugins/s1_position_reconstruction.py
+++ b/straxen/plugins/s1_position_reconstruction.py
@@ -9,6 +9,7 @@ import straxen
 from warnings import warn
 export, __all__ = strax.exporter()
 
+# Current defaul algorithm: data-driven CNN. 
 DEFAULT_S1POSREC_ALGO_OPTION = tuple([strax.Option("default_s1reconstruction_algorithm",
                  help="default S1 reconstruction algorithm that provides (x,y,z)",
                  default="cnn_s1",
@@ -97,6 +98,7 @@ class S1EventPositionBase(strax.Plugin):
         # Getting actual position reconstruction
         _in = events['s1_area_per_channel']
         with np.errstate(divide='ignore', invalid='ignore'):
+        # Normalise patters by dividing by largest PMT output between the two arrays. 
             _in = _in / _in.max(axis=1,keepdims=True)
         _out = self.model.predict(_in)
 
--- a/straxen/plugins/s1_position_reconstruction.py
+++ b/straxen/plugins/s1_position_reconstruction.py
@@ -1,0 +1,183 @@
+"""S1-based 3D position reconstruction for Xenon-nT"""
+
+import os
+import tempfile
+import tarfile
+import numpy as np
+import strax
+import straxen
+from warnings import warn
+export, __all__ = strax.exporter()
+
+DEFAULT_S1POSREC_ALGO_OPTION = tuple([strax.Option("default_s1reconstruction_algorithm",
+                 help="default S1 reconstruction algorithm that provides (x,y,z)",
+                 default="cnn_s1",
+                 )])
+
+@export
+@strax.takes_config(
+        strax.Option("s1_cnn_model_path", 
+                     help="Path to the S1 CNN model", 
+                     default=("/project2/lgrandi/guidam/CNN_S1_XYZ_SAVED_MODELS/version_datadriven_00_080921")
+                    ),
+    strax.Option('n_top_pmts', default=straxen.n_top_pmts,
+                 help="Number of top PMTs")
+)
+
+
+class S1EventPositionBase(strax.Plugin):
+    """
+    Base class for the alternative S1 three-dimensional position reconstruction.
+    This class should only be used when subclassed for the different algorithms. 
+    """
+
+    depends_on = ('events','event_area_per_channel',)
+    algorithm = None
+    compressor = 'zstd'
+    parallel = True
+    __version__ = '0.0.0'
+
+    def infer_dtype(self):
+        if self.algorithm is None:
+            raise NotImplementedError(f'Base class should not be used without '
+                                      f'algorithm as done in {__class__.__name__}')
+        dtype = [('x_' + self.algorithm, np.float32,
+                  f'Reconstructed {self.algorithm} S1 X position (cm), uncorrected'),
+                 ('y_' + self.algorithm, np.float32,
+                  f'Reconstructed {self.algorithm} S1 Y position (cm), uncorrected'),
+                 ('z_' + self.algorithm, np.float32,
+                  f'Reconstructed {self.algorithm} S1 Z position (cm), uncorrected')
+                  ]
+
+        dtype += strax.time_fields
+        return dtype
+
+    def setup(self):
+        self.model_file = self._get_model_file_name()
+        if self.model_file is None:
+            warn(f'No file provided for {self.algorithm}. Setting all values '
+                 f'for {self.provides} to None.')
+            # No further setup required
+            return
+
+        # Load the tensorflow model
+        import tensorflow as tf
+        if os.path.exists(self.model_file):
+            print(f"Path is local. Loading {self.algorithm} TF model locally "
+                  f"from disk.")
+        else:
+            downloader = straxen.MongoDownloader()
+            try:
+                self.model_file = downloader.download_single(self.model_file)
+            except straxen.mongo_storage.CouldNotLoadError as e:
+                raise RuntimeError(f'Model files {self.model_file} is not found') from e
+        with tempfile.TemporaryDirectory() as tmpdirname:
+        # Allow to load the model both from folder or from file.
+            try:
+                tar = tarfile.open(self.model_file, mode="r:gz")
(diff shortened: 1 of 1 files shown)


jarongrigat:
Looks good @matteoguida ! I'll review and add my stuff until the end of the week.

jarongrigat:
Still need to do some testing...
We should think about a reasonable value for `'min_reconstruction_area'` 10k, 20kPE ...??

matteoguida:
> Still need to do some testing... We should think about a reasonable value for `'min_reconstruction_area'` 10k, 20kPE ...??

@jarongrigat I suggest, _for the moment_, a value of 10k [PE]. I studied lower energies in my MSc thesis with BiPo coincidence, the results can be seen, currently, [here](https://drive.google.com/drive/folders/1tOrDxvYs8wHHborjr6B6e9TWG9E6KuyK?usp=sharing). The MSc thesis will be soon posted on the XENONnT wiki. BTW, thanks for the added lines 😃.

matteoguida:
Sorry guys the management of this PR has been really bad (was also my first one). I've always left the PR as a draft because I was hoping to update the model, making it more robust and with better performance for less energetic events (S1_area < 20 k), but it's never been a priority and months went by. I think it is worthy, being that the model has been used besides me, also by @hammannr, to try to merge the current state and in the following months hopefully update it. I strongly recommend reading the 3 recommendations at the beginning of the example notebook.

hammannr (commented):
Hey Matteo, sorry for the very long delay! I'm now trying to run the code following roughly the instructions in the linked notebook. Right now I'm running into `ValueError: Two plugins have a different default value for the same option. The option "min_reconstruction_area" in "PeakPositionsCNN" takes as a default "10"  while in "S1EventPositionCNN" the default value is set to "1000". Please change one of the defaults.` when trying to register the plugin as `st.register(straxen.plugins.s1_position_reconstruction.S1EventPositionCNN)`.
But maybe before testing all this it would anyways make sense to merge the current master so that we actually check how it would run if it were merged, right?

matteoguida:
Hi @hammannr the issues that you are indicating as you can see they were introduced by changes made by @jarongrigat, please see [this commit](https://github.com/XENONnT/straxen/pull/671/commits/bb014d63fea708f2b10ca48ab3d8d570972812ba).The problem is that the changes made seem to have not been tested. I therefore restore the old script which only contained CNN, which was intensely tested by you and that it worked. I happily added the part suggested by @jarongrigat about returning nan as a prediction for events with s1_area <1000. I would suggest to do another PR to @jarongrigat that I am happy to review in a short time in which he makes the changes to insert the MLP model, as well as a example notebook to test it etc. Honestly I don't understand why the changes in this branch were made without saying anything...

jarongrigat:
Ciao Matteo. Yes, sorry, I forgot about this PR along the way - Thanks for picking up the work again. I agree with your suggestion that you finish the PR with your work only. Let me know if I can help somehow. Jaron

hammannr:
Hey @matteoguida ! This fixed it, now everything runs smoothly! I loaded some Ar data (development container with local straxen version), and then checked the S1-based prediction similar to what you did in the linked notebook. To me, everything looks fine and behaves as expected.
I can't comment on the implementation though, for this straxen gurus should comment :innocent:

matteoguida:
@JoranAngevaare or @WenzDaniel do you have some kind comment to make referring to @hammannr last message?

JoranAngevaare:
Hi Matteo, thanks, there are a couple of things that are required
- fix all the hardcoded paths to something that we can e.g. pull from the database (you can upload that here https://github.com/XENONnT/private_nt_aux_files)
- Register the plugin to the nT context
- Write a test for the plugin, probably here https://github.com/XENONnT/straxen/blob/master/tests/plugins/posrec_plugins.py , there is a whole bunch of [examples how you can write such a test](https://github.com/XENONnT/straxen/tree/master/tests/plugins). Probably you can re-use some of the material in the example notebook
- You might give some of the style suggestions some thought: https://github.com/XENONnT/straxen/pull/671/files
- Using the URL config to open the tensorflow models will simplify the code a lot, I'd urge using those options (see the posrec plugins for some examples) this also removes the need to hardcode the default algorithm
- From the discussion here (and given how long this PR has been open), it seems that we might not have three plugins soon, so I'd probably not copy over the whole merging process which we have for the S2 posrec (yet)
- Does it make sense to run this plugin on events only? We exclude S1s from triggering the event building process, I'm not sure exactly where this PR will be used for most but if it's useful for S1 only data, we might want to reconsider either of these choices

matteoguida:
Hi Joran thanks a lot for your valuable inputs, apologies for late reply mainly caused by vacation absence. I hope to address technical points soon. 
Concerning the last two of your most conceptual ones:

- We have two plugin currently CNN and MLP: the MLP from Jaron (see above) will be merged in a next separate PR when this will be merged. So I would create (as was done) a parallel structure as the one of S2 posrec, instead of merging this code into the S2 posrec one (which AFAIU is your proposal), I believe is tidy and clean in this way and in the future maybe it allows to add versions based on simulated S1 hit patterns instead of data (approach with less biases for now). Let me know

The net is applied in the [charge insensitive volume study](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:analysis:ntsciencerun0:charge_insensitive_volume_data_driven_summary) mainly, which is [again required](https://xenonnt.slack.com/archives/C0177FXBEHX/p1661772147703259) for SR1 etc. 

- Originally I implemented it for peaks, however one should remind that the processing time to have a [x,y,z] prediction is way slower than s2-based pos-rec (320 [μs/S2 hit pattern] vs 30 [μs/ S1 hit patterns]), mainly due to the fact that bottom and top PMT hit patterns are needed to have it, this led us to conclude that for now it was enough to implement it at a “higher level”. 

I understood that you are underling the role of `exclude_s1_as_triggering_peaks` [here](https://github.com/XENONnT/straxen/blob/master/straxen/plugins/event_processing.py), the solutions could be:

1. As you said implementing it at lower level with however severe impact on time loading, I guess.
2. When one want to perform an S1 only analysis anyway one wants to set exclude_s1_as_triggering_peaks = True in order to deploy the higher part of straxen chain but including also lone S1s. As I guess it is just a line when the data are loaded one could simply make the user aware with a WARNING message when S1 posrec is used and underline it many times in the tutorial.
3. As S1 posrec is used only for S1 hit patterns with total area > 1000 PE due to unavoidable performance degradation for lower energy one could set that by default S1 can be the triggering peaks if s1_area > 1000 PE. I am not sure about the following statement but I guess that this should not be a relevant change for the science analyzes.  

Let me know which one you consider more reasonable.

Even without any change one can use it to see the location of events where S1 is coupled to a random lone S2.

JoranAngevaare:
Hi Matteo,
Thanks! I see only reasonable replies here, all sounds good to me 👍. 

>The net is applied in the [charge insensitive volume study](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:analysis:ntsciencerun0:charge_insensitive_volume_data_driven_summary) mainly,

Ah this makes sense - much more sense than doing an S1-only analysis.

> Originally I implemented it for peaks, however one should remind that the processing time to have a [x,y,z] prediction is way slower 

Right, I see this point. Is this still an issue if you consider S1s only above 1000 PE? I'd think that 10x more CPU time for 100x less peaks might already be okay. On the other hand, there are probably not going to be that many S1-peaks >1000 PE that are _not_ part of an event.
Just out of curiosity how does computing this plugin compare to one of the S2-xy posrec plugins? In case you want a [tutorial on a simple method of profiling, here is a link.](https://github.com/XENONnT/workshop_zoom_2021_04_19/blob/master/1_xenonnt_data/1_Data_Structure_and_processing_tutorial.ipynb)

JoranAngevaare:
Hi @matteoguida , when you are able to get back to this PR. Could you follow the new plugin structure from https://github.com/XENONnT/straxen/pull/1094 and URL configs as in https://github.com/XENONnT/straxen/pull/1079? I would consider closing this PR and re-opening a fresh one.

matteoguida:
@JoranAngevaare sure, I apologize, as is not urgent it always ends at the bottom of my todo-list

PR #673, opened 2021-09-30
Add source to runs table

JoranAngevaare:
Per @kdund 's suggestion add the 'source' field to the rundb query

Tell strax to extract the source if any from the rundoc

```python
st = straxen.contexts.xenonnt_online(_minimum_run_number=20000)
st.context_config['store_run_fields'] = ('name', 'number', 'start', 'end', 'livetime', 'mode', 'source')
st.select_runs()
```


skazama (approved):
Thanks Joran, looks fine to me.

WenzDaniel (approved):

PR #674, opened 2021-09-30
S2 local minimum information

JYangQi00:
A plugin for looking at the S2 local minimum information. This plugin provides the depth and width of the local minimum's valley, as well as the max goodness of split of the S2 peak after merging. Hopefully this information can be used to do some cuts for multiple scatters.

This PR adds a LoopPlugin over the main S2 peaks and provides information about the S2 local minimum (if it has one). The purpose of this is to look for which S2s are bi-modal peaks, as this may be due to a double scatter that happens close in x-y but is delayed in z.
This plugin smooths the main S2 peak and provides the max goodness of split, the depth of the local minimum, and the width of the local minimum. These three parameters can be studied as figures of merit in order to make multiple-scatter cuts for those peaks which don't show two distinct populations in their hitpattern, but do show a hint at two separate peaks in their time signature.
Please see the last section of this note:

https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:jianyang:s2_split_tagging

### _Notes on testing_


skazama:
@JYangQi00 Is this PR still up-to-date?

JYangQi00:
Hi Shingo, sorry about the late reply. I updated it and tested it out (merged with the latest master branch of straxen) and it works and is up to date.

JoranAngevaare:
Hi @JYangQi00 looks like a nice plugin! 

Can you make sure the code is tested and do you maybe have some numbers on performance, how long does it take to process a typical run?

Maybe also keep in mind that the options names propagate to the entire context, should we rename them a bit such that they aren't so general, e.g. `smoothing_power` could mean many things if you see it somewhere popping up in the context config?

straxen/plugins/local_minimum_info.py
```diff
@@ -0,0 +1,196 @@
...
+    :param smoothp: The smoothed peak
+    :param max_loc: Location of every local maximum of the peak
+    :param min_loc: Location of every local minimum of the peak
+    :param pv: "Percent value", the percent of the valley height for which to calculate the gap
+    :param dt: The time of one sample in ns
+    :return: The gap in ns, the depth in PE
+    """
+    n_gap = len(min_loc)
+    p_length = len(smoothp)
+    if (len(max_loc) - n_gap) == 1 and len(min_loc):
```
WenzDaniel, line 124: You could invert this statement and directly return 0,0. In this way you can remove the additional else and indentation.

WenzDaniel (commented):
Hej @JYangQi00, the PR looks already quite nice and clean. I have not worked through logic of the code though. I think it would be good if we could add some unit tests to the function you wrote. This will help us to spot possible bugs or bugs when making changes in the future.

JYangQi00:
> Hi @JYangQi00 looks like a nice plugin!
> 
> Can you make sure the code is tested and do you maybe have some numbers on performance, how long does it take to process a typical run?
> 
> Maybe also keep in mind that the options names propagate to the entire context, should we rename them a bit such that they aren't so general, e.g. `smoothing_power` could mean many things if you see it somewhere popping up in the context config?

Hi Joran, thanks for your comments. I changed the config options to have a "_localmin" after them to avoid the confusion. I was unable to test this on a background run since I am using a local directory. However, I tested it on the Kr83 run "018161" and after all the dependent data types were made, it took 2 minutes and 13 seconds to process event_local_min_info.

JYangQi00:
> Hej @JYangQi00, the PR looks already quite nice and clean. I have not worked through logic of the code though. I think it would be good if we could add some unit tests to the function you wrote. This will help us to spot possible bugs or bugs when making changes in the future.

Hi Daniel, thanks for your suggestions. I made the appropriate adjustments as suggested. It took 2 minutes 13 seconds to process the run "018161". Unfortunately, I am using a local installation of straxen to test out this branch, so I can't load background runs, the one I processed was a Kr83m run.

JoranAngevaare:
> > Hi @JYangQi00 looks like a nice plugin!
> > Can you make sure the code is tested and do you maybe have some numbers on performance, how long does it take to process a typical run?
> > Maybe also keep in mind that the options names propagate to the entire context, should we rename them a bit such that they aren't so general, e.g. `smoothing_power` could mean many things if you see it somewhere popping up in the context config?
> 
> Hi Joran, thanks for your comments. I changed the config options to have a "_localmin" after them to avoid the confusion. I was unable to test this on a background run since I am using a local directory. However, I tested it on the Kr83 run "018161" and after all the dependent data types were made, it took 2 minutes and 13 seconds to process event_local_min_info.

Thanks @JYangQi00! Sorry, I meant if you could also write a unit-test for some of the functions you wrote (see for examples the `tests/` folder on github.

Thanks for checking out the speed as well. 2 minutes is more than I'd have expected for a high level plugin. Do you know where the bottleneck resides? With some [profiling](https://pypi.org/project/line-profiler/) you might be able to tell. What also might help is writing it as an ordinary plugin (like `EventBasics`) such that you can use numpy vectorization and/or numba more.

PR #685, opened 2021-10-04
Pbar patch for rundb query

JoranAngevaare:
Small patch to the pbar similar to https://github.com/AxFoundation/strax/pull/538


WenzDaniel (approved):
Thanks, day of the pbar :D

PR #686, opened 2021-10-04
updated S2 corrected area

Jianyu010:

Updated under event_processing the CorrectedArea to enable S2 separated corrections. 

7 new fields are added (updated): cS2top, cS2bot, cS2tot, together with corresponding alternative S2 fields.  Also the new cS2_area_fraction_top is added by cS2top/cS2tot. 

### _Notes on testing_

### DIFF
--- a/straxen/plugins/event_processing.py
+++ b/straxen/plugins/event_processing.py
@@ -598,8 +598,24 @@ def get_veto_tags(events, split_tags, result):
             (0, pax_file('XENON1T_s1_xyz_lce_true_kr83m_SR0_pax-680_fdc-3d_v0.json')),  # noqa
             (first_sr1_run, pax_file('XENON1T_s1_xyz_lce_true_kr83m_SR1_pax-680_fdc-3d_v0.json'))]),  # noqa
     strax.Option(
-        's2_xy_correction_map',
-        help="S2 (x, y) correction map. Correct S2 position dependence "
+        's2_xy_correction_map_top',
+        help="S2top (x, y) correction map. Correct S2 position dependence "
+             "manly due to bending of anode/gate-grid, PMT quantum efficiency "
+             "and extraction field distribution, as well as other geometric factors.",
+        default_by_run=[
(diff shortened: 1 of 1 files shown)


jingqiangye:
Performance check can be found [here](https://xenonnt.slack.com/archives/C020GRDUNN8/p1634016402184600?thread_ts=1634016138.184500&cid=C020GRDUNN8).

straxen/plugins/event_processing.py
```diff
@@ -653,8 +665,10 @@ def setup(self):
 
         self.s2_map = InterpolatingMap(
             get_cmt_resource(self.run_id,
-                             tuple([*self.config['s2_xy_correction_map']]),
-                             fmt='text'))
+                             tuple(['suffix',
```
WenzDaniel, line 668: Why do we need this `suffix` this looks somewhat odd to me?

WenzDaniel: No not for the moment, but in my opinion CMT must be reworked. It has become one of the most complicated parts in straxen.

ahiguera-mx: We need the suffix because now s2 maps are position dependent, and in order to use only one single option for three different maps i.e. s2_map_mlp, s2_map_gcn, s2_map_cnn we add that suffix, then at the end this works with one single straxen option `self.config['s2_xy_correction_map']`. Is either that or we modified straxen to set every configuration option for every position algorithm

WenzDaniel:
Looks fine to me, but I cannot merge without all tests passing as I am not an admin of our repository.

JoranAngevaare:
Right, we also don't want to merge if the tests are passing. Do you know when the new version would be fixed?

jingqiangye:
To my knowledge, there are 1 or 2 things that need to be finalized before a new CMT global version can be fixed. I am checking with them. On the other hand, what is the plan for the straxen release? I have the impression that it was planned this week but not sure.

Issue #670, opened 2021-09-27
Rucio front end looks for local files if one with the same name exists.


michaweiss89:
**Describe the bug**

**To Reproduce**
Insert the MWE of how to reproduce the error
```python
YOUR CODE GOES HERE
```

**Expected behavior**

**Screenshots**

**Versions**
Please add the output of:
```
straxen.print_versions()
```

michaweiss89:
I tried to download a certain chunk with rucio front end but it was looking at local files and did not find it. 
There was an incomplete raw record file of the same run and only after deleting it did rucio downloaded the chunk.

run_id, target = '020315', 'raw_records'
s2_time = 1621897731731375152
st = straxen.contexts.xenonnt(
    use_rucio=True,
    use_rucio_remote=True,
   # output_folder = os.path.join(os.environ['SCRATCH'], './rucio_staging_dir'),
)
st.storage[-1].backends[-1].download_heavy = True
for chunk in st.get_iter(run_id, target, # or touching
             time_range=(
                  # get the s1 from this event
                 int(s2_time-4*30000),
                 int(s2_time+4*60000),
            )):
    print(chunk)

output:

FileNotFoundError: [Errno 2] No such file or directory: './strax_data/020315-raw_records-rfzvpzj4mf/raw_records-rfzvpzj4mf-000019'

Expected behavior:
If there is a raw record file with the same name if it does not find the chunk it should download the missing chunk with rucio.

Working on midway2-0424.rcc.local with the following versions and installation paths:
python	v3.8.11	(default, Aug  3 2021, 15:09:35) [GCC 7.5.0]
strax	v1.0.0	/opt/XENONnT/anaconda/envs/XENONnT_development/lib/python3.8/site-packages/strax
straxen	v1.0.0	/opt/XENONnT/anaconda/envs/XENONnT_development/lib/python3.8/site-packages/straxen
cutax	v0.2.0	/home/weiss89/.local/lib/python3.8/site-packages/cutax

JoranAngevaare:
Same issue as https://github.com/XENONnT/straxen/issues/586

JoranAngevaare:
Should be fixed in #688, #694, please reopen if persistent

Issue #586, opened 2021-07-06
Rucio download with st.is_stored results in partial downloads


JoranAngevaare:
**Describe the bug**
st.is_stored and downloading using rucio is an previously untested edge case.

**To Reproduce**
Insert the MWE of how to reproduce the error
```python
st = straxen.contexts.xenonnt(cmt_version='global_v3', use_rucio=True, output_folder = './rucio_staging_dir3')
run_id = '024345'
print(st.is_stored(run, 'pulse_counts'))

pc = st.get_array(run, 'pulse_counts')
print(pc.nbytes)
```
Which yields
```
Downloading xnt_024345:pulse_counts-niuq6un3aa-metadata.json

TrueException in thread CRITICAL:ThreadedMailboxProcessor:Target Mailbox (pulse_counts) killed, exception <class 'strax.mailbox.MailboxKilled'>, message (<class 'FileNotFoundError'>, FileNotFoundError(2, 'No such file or directory'), <traceback object at 0x7f47f3d21740>)
load:pulse_counts:
Traceback (most recent call last):
  File "/opt/XENONnT/anaconda/envs/XENONnT_development/lib/python3.8/threading.py", line 932, in _bootstrap_inner
    self.run()
  File "/opt/XENONnT/anaconda/envs/XENONnT_development/lib/python3.8/threading.py", line 870, in run
    self._target(*self._args, **self._kwargs)
  File "/opt/XENONnT/anaconda/envs/XENONnT_development/lib/python3.8/site-packages/strax/mailbox.py", line 267, in _send_from
    self.kill_from_exception(e)
  File "/opt/XENONnT/anaconda/envs/XENONnT_development/lib/python3.8/site-packages/strax/mailbox.py", line 186, in kill_from_exception
    raise e
  File "/opt/XENONnT/anaconda/envs/XENONnT_development/lib/python3.8/site-packages/strax/mailbox.py", line 254, in _send_from
    x = next(iterable)
  File "/opt/XENONnT/anaconda/envs/XENONnT_development/lib/python3.8/site-packages/strax/storage/common.py", line 456, in loader
    yield self._read_and_format_chunk(**read_chunk_kwargs)
  File "/opt/XENONnT/anaconda/envs/XENONnT_development/lib/python3.8/site-packages/strax/storage/common.py", line 474, in _read_and_format_chunk
    data = self._read_chunk(backend_key,
[...]

PR #688, opened 2021-10-06
Smart RSE selection in RucioRemoteBackend

ershockley:
Previously we always assumed that any data we would want to download was on a specific RSE (UC_OSG_USERDISK), while that obviously will not always be the case. This update aims to select the best RSE based on where the host where straxen is running. 

This update uses the hostname (via `socket.getfqdn()`) along with specific environment variables associated with OSG jobs to determine where the code is running, and then determines if the data in question is on a preferred RSE for that location. 

I tested [this code snippet](https://gist.github.com/ershockley/19d38769bb4fd9b3aa425f2cfcf71dfd) on both Midway and Expanse (in development container). It looks for a run on SDSC_USERDISK (which I happen to know are all on UC_OSG_USERDISK at this time also) and then load 2 seconds of raw_records. Depending on which site I run on, the data is downloaded from different endpoints: 

On Midway:
```Singularity> hostname -f; python test_remote.py
dali-login1.rcc.local
Downloading xnt_018159:raw_records-rfzvpzj4mf-metadata.json from UC_OSG_USERDISK
Loading raw_records: |                                                                            | 0.00 % [00:00<?]Downloading xnt_018159:raw_records-rfzvpzj4mf-000000 from UC_OSG_USERDISK
Loading raw_records: |▏                                                | 0.35 % [00:06<29:05], #1 (6.05 s). 7.5 MB/s
```

On Expanse:

```
Singularity> hostname -f; python test_remote.py
login02.expanse.sdsc.edu
Downloading xnt_018159:raw_records-rfzvpzj4mf-metadata.json from SDSC_USERDISK
Loading raw_records: |                                                                            | 0.00 % [00:00<?]Downloading xnt_018159:raw_records-rfzvpzj4mf-000000 from SDSC_USERDISK
Loading raw_records: |▏                                                | 0.35 % [00:04<22:00], #1 (4.58 s). 9.9 MB/s
```

You can see that when running on Midway, it grabs the data from UC_OSG_USERDISK, whereas on Expanse it grabs it from SDSC_USERDISK.

### DIFF
--- a/straxen/rucio.py
+++ b/straxen/rucio.py
@@ -248,6 +248,17 @@ class RucioRemoteBackend(strax.FileSytemBackend):
     # datatypes we don't want to download since they're too heavy
     heavy_types = ['raw_records', 'raw_records_nv', 'raw_records_he']
 
+    preferred_rses = {'midway': ['UC_DALI_USERDISK', 'UC_OSG_USERDISK', 'SDSC_USERDISK'],
+                      'expanse': ['SDSC_USERDISK', 'UC_OSG_USERDISK', 'UC_DALI_USERDISK'],
+                      'in2p3': ['CCIN2P3_USERDISK', 'NIKHEF2_USERDISK', 'CNAF_USERDISK'],
+                      'nikhef': ['NIKHEF2_USERDISK', 'SURFSARA_USERDISK', 'CNAF_USERDISK'],
+                      'surf': ['SURFSARA_USERDISK', 'NIKHEF2_USERDISK', 'CNAF_USERDISK'],
+                      'osg_northamerica': ['UC_OSG_USERDISK', 'SDSC_USERDISK', 'UC_DALI_USERDISK'],
+                      'osg_europe': ['NIKHEF2_USERDISK', 'CNAF_USERDISK', 'SURFSARA_USERDISK'],
+                      }
+
+    dset_cache = {}
+
     def __init__(self, staging_dir, download_heavy=False, **kwargs):
         """
         :param staging_dir: Path (a string) where to save data. Must be a writable location.
@@ -271,9 +282,11 @@ class RucioRemoteBackend(strax.FileSytemBackend):
         self.download_heavy = download_heavy
         # Do it only when we actually load rucio
         from rucio.client.downloadclient import DownloadClient
+        from rucio.client.client import Client
         self.download_client = DownloadClient()
+        self.rucio_client = Client()
 
-    def get_metadata(self, dset_did, rse='UC_OSG_USERDISK', **kwargs):
+    def get_metadata(self, dset_did, **kwargs):
         base_dir = os.path.join(self.staging_dir, did_to_dirname(dset_did))
 
         # define where the metadata will go (or where it already might be)
@@ -283,13 +296,20 @@ class RucioRemoteBackend(strax.FileSytemBackend):
 
         # download if it doesn't exist
         if not os.path.exists(metadata_path):
+            if dset_did in self.dset_cache:
+                rses = self.dset_cache[dset_did]
+            else:
+                rses = self.get_rses(dset_did)
+                self.dset_cache[dset_did] = rses
+
+            rse = self.find_best_rse(rses)
             metadata_did = f'{dset_did}-metadata.json'
             did_dict = dict(did=metadata_did,
                             base_dir=base_dir,
                             no_subdir=True,
                             rse=rse
                             )
-            print(f"Downloading {metadata_did}")
+            print(f"Downloading {metadata_did} from {rse}")
             self._download([did_dict])
 
         # check again
@@ -299,7 +319,7 @@ class RucioRemoteBackend(strax.FileSytemBackend):
         with open(metadata_path, mode='r') as f:
             return json.loads(f.read())
 
-    def _read_chunk(self, dset_did, chunk_info, dtype, compressor, rse="UC_OSG_USERDISK"):
+    def _read_chunk(self, dset_did, chunk_info, dtype, compressor):
         base_dir = os.path.join(self.staging_dir, did_to_dirname(dset_did))
         chunk_file = chunk_info['filename']
         chunk_path = os.path.join(base_dir, chunk_file)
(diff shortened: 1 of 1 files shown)


straxen/rucio.py
```diff
@@ -248,6 +248,17 @@ class RucioRemoteBackend(strax.FileSytemBackend):
...
     heavy_types = ['raw_records', 'raw_records_nv', 'raw_records_he']
 
+    preferred_rses = {'midway': ['UC_DALI_USERDISK', 'UC_OSG_USERDISK', 'SDSC_USERDISK'],
+                      'expanse': ['SDSC_USERDISK', 'UC_OSG_USERDISK', 'UC_DALI_USERDISK'],
+                      'in2p3': ['CCIN2P3_USERDISK', 'NIKHEF2_USERDISK', 'CNAF_USERDISK'],
+                      'nikhef': ['NIKHEF2_USERDISK', 'SURFSARA_USERDISK', 'CNAF_USERDISK'],
+                      'surf': ['SURFSARA_USERDISK', 'NIKHEF2_USERDISK', 'CNAF_USERDISK'],
+                      'osg_northamerica': ['UC_OSG_USERDISK', 'SDSC_USERDISK', 'UC_DALI_USERDISK'],
+                      'osg_europe': ['NIKHEF2_USERDISK', 'CNAF_USERDISK', 'SURFSARA_USERDISK'],
+                      }
```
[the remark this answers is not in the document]
ershockley: Putting it in a config is definitely more aesthetic, so I'm not totally against it. From a practical standpoint I don't think it helps much though. Downloading some config from the database seems like an unnecessary step to me, especially at this time.

ershockley:
Thanks for the quick feedback @JoranAngevaare! I will clean up a bit today. I take your points about the hard coding etc, but I'm not sure the best way to handle it at the moment. Maybe we could just push through the admix updates soon and use that, which would definitely be the cleanest but I worry about potential issues in adding a newly updated package as a dependency here in straxen. I will think about it.

straxen/rucio.py
```diff
@@ -5,13 +5,16 @@
...
 import glob
 import hashlib
-import time
 from utilix import xent_collection
 import strax
+import warnings
+try:
+    import admix
+except (ImportError, RuntimeError):
+    print("Problem importing admix. You will not be able to use rucio tools.")
```
JoranAngevaare, line 14: Can we make this similar to https://github.com/XENONnT/straxen/pull/592? Otherwise we're getting an annoying message every time we do something outside of the container, e.g. for developers or on the DAQ.

straxen/rucio.py
```diff
@@ -5,13 +5,16 @@
 import os
 import glob
 import hashlib
-import time
 from utilix import xent_collection
 import strax
+import warnings
```
JoranAngevaare, line 10: looks like you aren't using `warnings`?

straxen/rucio.py
```diff
@@ -268,44 +274,33 @@ def __init__(self, staging_dir, download_heavy=False, **kwargs):
...
+            rse = self.dset_cache[dset_did]
+        else:
+            rses = admix.rucio.get_rses(dset_did)
+            rse = admix.downloader.determine_rse(rses)
+            self.dset_cache[dset_did] = rse
+
+        metadata_did = f'{dset_did}-metadata.json'
+        downloaded = admix.download(metadata_did, rse=rse, location=self.staging_dir)
+        assert len(downloaded) == 1, f"{metadata_did} should be a single file. We found {len(downloaded)}."
+        metadata_path = downloaded[0]
```
JoranAngevaare, line 292: while at it, do you see a way of fixing https://github.com/XENONnT/straxen/issues/586 and https://github.com/XENONnT/straxen/issues/670 which are caused by this piece of the code?

straxen/rucio.py
ershockley: Ok this should be better now, thanks for the suggestion.

straxen/rucio.py
ershockley: Done.

straxen/rucio.py
ershockley: Ok I think this should be semi-addressed now. The problem seemed to be with the other storage plugins. If I only loaded the rucio frontend it worked fine. I still don't fully understand the reason yet but I think it probably had to do with the fact that the rucio backend was downloading data be in same place that the other storage frontends were looking? 

Anyway after a few minor improvements and putting the rucio frontend to the front of the line,  this snippet now works for me: 
```
st = straxen.contexts.xenonnt(cmt_version='global_v3',
                              download_heavy=True
                             )
print(st.storage)
#st.storage = [st.storage[-1]]

run_id = '024345'
print(st.is_stored(run_id, 'raw_records'))

pc = st.get_array(run_id, 'raw_records', seconds_range=(0, 2))
print(pc.nbytes)
```

JoranAngevaare: Ah you are right, thanks a lot! 
The issue comes from the fact that we are using the same folder in two back ends with different frontends. Whereas the DataDirectory frontend considered something stored if both the folder is named correctly (since a complete write of the folder removes the _temp of the folder name), the logic for the RucioFrontend is different, it checks all the chunks exist.

I think the most robust solution is to just not re-use the exact same path in the two backends, would this be fine with you?

straxen/rucio.py
```diff
@@ -29,12 +37,10 @@ class RucioFrontend(strax.StorageFrontend):
...
-    local_rucio_path = None
-
-    # Some attributes to set if we have the remote backend
-    _did_client = None
-    _id_not_found_error = None
-    _rse_client = None
+    path = None
+    local_prefixes = {'UC_DALI_USERDISK': '/dali/lgrandi/rucio/',
+                      'SDSC_USERDISK': '/expanse/lustre/projects/chi135/shockley/rucio',
+                      }
```
WenzDaniel, line 43: If this is for local use only could we put it in some local file then?

straxen/rucio.py
```diff
@@ -91,32 +101,10 @@ def __repr__(self):
...
-                continue
-            did = key_to_rucio_did(key)
-            if self.did_is_local(did):
-                ret.append(('RucioLocalBackend', did))
-            else:
-                ret.append(False)
-        return ret
+        # for performance, dont do find_several with this plugin
+        # we basically do the same query we would do in the RunDB plugin
+        return np.zeros_like(keys, dtype=bool).tolist()
```
WenzDaniel, line 107: You will get here a different behavior whether keys is an iterable or not. 

```
np.zeros_like(1, dtype=bool).tolist()  # return False
np.zeros_like((1, ), dtype=bool).tolist()  # returns [False]
```

WenzDaniel (approved):
To me it looks fine, just two minor comments. But I cannot judge the admix part. To me things looked at least logical.

straxen/rucio.py
JoranAngevaare: Find several is always called with an iterable:
https://github.com/AxFoundation/strax/blob/64a371384a14d81038d1b573f9ce99348caa569b/strax/run_selection.py#L35

straxen/rucio.py
JoranAngevaare: Generally you are correct.

ershockley: Ok @WenzDaniel  I see your point. Those paths are quite secure I think but in principle I agree. I will look into other options, perhaps using utilix as @JoranAngevaare mentioned.

PR #689, opened 2021-10-07
Skips superruns in rucio frontend

WenzDaniel:
Adds check which skips superruns in rucio front-end. Requires: https://github.com/AxFoundation/strax/pull/532

### DIFF
--- a/straxen/rucio.py
+++ b/straxen/rucio.py
@@ -108,6 +108,10 @@ class RucioFrontend(strax.StorageFrontend):
 
         ret = []
         for key in keys:
+            if not self._support_superruns(key.run_id):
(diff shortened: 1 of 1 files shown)


straxen/rucio.py
```diff
@@ -108,6 +108,10 @@ def find_several(self, keys, **kwargs):
 
         ret = []
         for key in keys:
+            if not self._support_superruns(key.run_id):
+                ret.append(False)
+                continue
+
```
JoranAngevaare, line 114:
```suggestion

```

JoranAngevaare (approved):
Thanks!

PR #690, opened 2021-10-11
Version bump of hitlets

WenzDaniel:

Version bump because of https://github.com/AxFoundation/strax/pull/549.

### DIFF
--- a/straxen/common.py
+++ b/straxen/common.py
@@ -300,7 +300,7 @@ def get_secret(x):
 @export
 def download_test_data():
     """Downloads strax test data to strax_test_data in the current directory"""
-    blob = get_resource('https://raw.githubusercontent.com/XENONnT/strax_auxiliary_files/609b492e1389369734c7d2cbabb38059f14fc05e/strax_files/strax_test_data_straxv0.9.tar',  #  noqa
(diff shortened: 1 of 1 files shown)


WenzDaniel:
@ershockley I did a bug fix in hitlet processing. This will require reprocessing of `hitlets_nv` and `hitlets_mv`. I have a question regarding managing reprocessing. Currently, the muon-veto plugins are set-up such that they are inherited from the neutron-veto plugins. Hence, their lineage also has the version of the corresponding neutron-veto plugin encoded. This means whenever I change the version of the neutron-veto plugin this will also change the lineage/hash of the muon-veto plugin, indicating that reprocessing is needed. 
However, from a human perspective would you prefer that I bump the version of the muon-veto plugins, too? Would this help you to keep track for reprocessing?

JoranAngevaare (changes requested):
Perhaps we should put strax v1.0.1 in the requirements file? It won't work now but this PR can be merged after releasing the new strax version

ershockley:
@WenzDaniel I don't think it is needed to manually up the versions. We store the hashes for each context/straxen version that goes into base environment and don't really track what individual plugin versions are. So I think it is fine like this.

ershockley (approved):

PR #691, opened 2021-10-12
Add n saturated channels

WenzDaniel:
In this PR we do 3 things:

- Add `n_saturated_channel` information to peak_basics and propagate the information upwards to event_basics.
- Propagate the information about `tight_coincidence` up to event_basics
- Add a new `tight_coincidence_channel` field which contains the tight coincidence based on number of contributing PMTs instead of hits.

For the last change I modified `get_tight_coincidence` and I also updated the peak_merging accordingly in https://github.com/AxFoundation/strax/pull/551. Further, I added a test for `get_tight_coincidence` which tests both the normal `tight_coincidence` and the channel based one. 

Some of the test in this PR will currently, fail as we have to merge https://github.com/AxFoundation/strax/pull/551 first and updated our peak-level test data.

Some disadvantage of this PR: We need to reprocess starting with peaklets if we want to include a channel based tight coincidence.

### DIFF
--- a/straxen/plugins/event_processing.py
+++ b/straxen/plugins/event_processing.py
@@ -135,7 +135,7 @@ class EventBasics(strax.Plugin):
     The main S2 and alternative S2 are given by the largest two S2-Peaks
     within the event. By default this is also true for S1.
     """
-    __version__ = '1.1.1'
+    __version__ = '1.2.0'
 
     depends_on = ('events',
                   'peak_basics',
@@ -195,7 +195,10 @@ class EventBasics(strax.Plugin):
             ('range_50p_area',    np.float32, 'width, 50% area [ns]'),
             ('range_90p_area',    np.float32, 'width, 90% area [ns]'),
             ('rise_time',         np.float32, 'time between 10% and 50% area quantiles [ns]'),
-            ('area_fraction_top', np.float32, 'fraction of area seen by the top PMT array')
+            ('area_fraction_top', np.float32, 'fraction of area seen by the top PMT array'),
+            ('tight_coincidence', np.int16, 'Hits within tight range of mean'),
+            ('n_saturated_channels',      np.int16, 'Total number of saturated channels'),
+            ('tight_coincidence_channel', np.int16, 'PMT channel within tight range of mean'),
         )
 
     @staticmethod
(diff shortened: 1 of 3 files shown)


straxen/plugins/event_processing.py
```diff
@@ -195,7 +195,10 @@ def _set_dtype_requirements(self):
             ('range_50p_area',    np.float32, 'width, 50% area [ns]'),
             ('range_90p_area',    np.float32, 'width, 90% area [ns]'),
             ('rise_time',         np.float32, 'time between 10% and 50% area quantiles [ns]'),
-            ('area_fraction_top', np.float32, 'fraction of area seen by the top PMT array')
+            ('area_fraction_top', np.float32, 'fraction of area seen by the top PMT array'),
+            ('tight_coincidence', np.int16, 'Hits within tight range of mean'),
+            ('n_saturated_channels',      np.int16, 'Total number of saturated channels'),
+            ('tight_coincidence_channel', np.int16, 'PMT channel within tight range of mean'),
```
JoranAngevaare, line 201: should be # of channels right (the description)? I think these things are quire similar so we should make it easy for people to understand the difference

JoranAngevaare, line 201: every field we add here is multiplied by 4 in the total even-basics dtype (alt, main, S1, S2). Probably there is some story to as to why we need all these fields but we should carefully think about adding fields here because no-one ever removes fields from event_basics

JoranAngevaare (commented):
Looks like a simple change, seems technically sound.

Like you say, it requires a reprocessing of peaklets and is breaking change as such.

The other thing I'd be cautious about is making event_basics ever expanding, this PR adds another 12 fields, perhaps it's good to explain in the PR description the envisaged use cases.

straxen/plugins/event_processing.py
WenzDaniel: Yes, true nice catch I will updated it.

WenzDaniel:
Motivating the fields:

- tight_coincidence and tight_coincidence_channel are required in almost any science analysis and hence I think it is good to give analysts the possibility to access those quantities easily without loading weeks of peak_basic data first. Which might either become challenging due to its data-size or tedious if loaded run by run and stored separately. 
- For n_saturated channel it is kind of the same argument, but it was primarily requested by different analysts working on high energy and alpha decay studies. For this quantity it was even worse before as it was not even stored in peak_basics.

I think in general it is fine to add new quantities in event_basics if they are used by a multitude of different analyses. Storage-wise it does not add much. But it might save analysts dozens of hours time (in total across all analyses). I also do agree that it does not make sense to store every  single quantity in the end.
 
Maybe also our ACs can comment about this.

WenzDaniel:
Just for the record I patched in the he support in https://github.com/XENONnT/straxen/commit/72f92d362cd2c7a52e37de36b8149bd3df2a9405.

PR #692, opened 2021-10-13
add small tool to extract run comments from database

JoranAngevaare:
##What does the code in this PR do / what does it improve / Can you briefly describe how it works?
Load the comments field from the database and extract the latest comment.

afbeelding

### DIFF
--- a/straxen/misc.py
+++ b/straxen/misc.py
@@ -244,6 +244,26 @@ class TimeWidgets:
         return time, time_ns
 
 
+@strax.Context.add_method
+def extract_latest_comment(self):
+    """
+    Extract the latest comment in the runs-database. This just adds info to st.runs
+
+    Example:
+        st.extract_latest_comment()
+        st.select_runs(available=('raw_records'))
+    """
+    if self.runs is None or 'comments' not in self.runs.keys():
+        self.context_config['store_run_fields'] = tuple(
+            list(self.context_config['store_run_fields']) + ['comments']
+        )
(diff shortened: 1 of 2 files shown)


straxen/misc.py
```diff
@@ -244,6 +244,26 @@ def _convert_to_datetime(time_widget, time_zone):
...
+    """
+    if self.runs is None or 'comments' not in self.runs.keys():
+        self.scan_runs(store_fields=('comments',))
+        latest_comments = _parse_to_last_comment(self.runs['comments'])
+        self.runs['comments'] = latest_comments
+    return self.runs
+
+
+def _parse_to_last_comment(comments):
+    return [(c[-1]['comment'] if hasattr(c, '__len__') else '') for c in comments]
```
WenzDaniel, line 264: This guy is a bit complicated and take a sec to understand but I guess it is fine. I do not see any better solution.

WenzDaniel (approved):
Looks good. It is a nice addition, although I doubt many people will use it. One question though, why is comments not added to the dataframe by default? Some performance reason?

JoranAngevaare:
> 
> 
> Looks good. It is a nice addition, although I doubt many people will use it. One question though, why is comments not added to the dataframe by default? Some performance reason?

It's because it isn't a flat array, rather you have to query and unpack the nested structure of the comments field.

> although I doubt many people will use it.

Fair, it's not so bad to have though indeed, it's much easier to point people to than learning to query a mongo database with projections and all.

PR #693, opened 2021-10-13
Warn about non-loadable loggers

JoranAngevaare:
Small change but logging was stopped briefly because a dependency wasn't installed in daqnt


JoranAngevaare:
It was mostly added for the sake of failing nfs mounts or issues in daqnt as it's used for other purposes as well. Back in the days, we didn't use to log bootstrax other than the logs to the database

PR #708, opened 2021-10-27
Update for peak classification based on manual boundaries

yuema137:
This PR only does one thing: applicate the manual boundaries explored in [this note](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:yue:classification_manual_boundary) in the peak classification. The boundaries are displayed in Fig.5 in the note.
This change is just for small peaks (below 100PE) and the main affect is increasing the power of separating single electrons from the small S1 population. So in principle this update could lower the rate of AC background considerably and hence help us get cleaner data. 
Further investigations and tests are needed to optimize the empirical parameters so welcome to test it and thanks a lot!


skazama (approved):
As we are discussing in slack, the boundary Yue optimized should be fine according to Daniel/Tianyu's plots (See [here](https://xenonnt.slack.com/archives/C016USUN64A/p1636364717029100?thread_ts=1635297704.012700&cid=C016USUN64A)). I also tested this plugin and it worked without any errors, so can be merged.

JoranAngevaare (commented):
Hi Yue,
thanks for the PR, based on simulation, the results are quite similar to what we currently have but I guess we will discuss on Thursday a bit.

One general question is if we can simplify the code a bit? E.g. by naming things like `norm, const, tau = self.config['s1_rise_time_parameters']` and adding a link maybe to this PR/note to explain the `A, B, T, k and b` you are using.

Perhaps changing the style as below a bit already helps a bit to keep it readable, let me know if I can help 😉 : 
```python
 has_s1_like_risetime = rise_time <= (norm*np.exp(-peaks['area']/tau)+ const)
 has_s1_like_risetime |= (rise_time <= self.config['s1_max_rise_time_post100']) & (peaks['area'] > 100)
 ....
 is_s1 = has_s1_like_rise_time & has_s1_like_aft
```

skazama:
@yuema137 Your current branch seems to conflict with the current straxen `(straxen/plugins/peaklet_processing.py
)`. Can you please update your branch?

WenzDaniel (commented):
In addition of fixing the conflict, would you mind to also implement the suggestions by Joran. 

I agree with him, that classification is currently hard to read. Here are some suggestions to make it simpler:

- Move your boundaries e.g. `(A*np.exp(-peaks['area']/T)+B)` into separate functions.  And call them like e.g. `upper_rise_time_boundary(peaks['area'], *self.config['s1_rise_time_parameters'])` Please do the same for your aft boundary.
- Since this plugin is not very computational intense. I think we do not have to do all the S1 classification at once. Usually you can split logic statements when using an logical-or.  E.g. you can divide your classification into above and below 100 PE.

ershockley:
@yuema137 would you be able to address the above comments today?

yuema137:
Thanks for the suggestions! I will update my branch and address the comments today.

yuema137:
I tried to merge the changes in the current straxen to my repository but there are some conflicts which must be done manually. So can I just open a new pull request and move the main points in the comments to there? I have created a new branch including the updates in the original repository.

WenzDaniel:
I think this is up to you. You can make a new PR and close this one. But in this case please tag this PR with the new one and vis-versa.

yuema137:
I have made a new PR #775 and closed this one. Thanks for the comments!

PR #719, opened 2021-10-29
Remove RuntimeError in RucioFrontend

ershockley:
Many tests are failing over in [base_environment](https://github.com/XENONnT/base_environment/pull/757) since we actually test there in the analysis environment (particularly containing rucio). I think this try except block should fix.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.1.1
+current_version = 1.1.2
 files = setup.py straxen/__init__.py
 commit = True
 tag = True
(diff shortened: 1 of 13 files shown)


WenzDaniel:
Could we use something similar to https://github.com/XENONnT/straxen/blob/63a1561e2a56b1fe538a2db1a2a205e8c8999eb1/tests/test_rucio.py#L31 instead? This makes it more clear if a certain test is actually executed or not.

JoranAngevaare (changes requested):
I think it would be much better to not raise runtime errors in the Rucio Frontend, just that it cannot provide you data shouldn't prevent you from initializing the context (or event the frontend itself). For example, you can also initialize the storage front-ends without having the `/dali/bla` paths.

ershockley:
Ok we can also do this. In my view it doesn't make much sense to register storage frontends that can't be used, but it would make this exception handling much better.

JoranAngevaare:
Thanks Evan

PR #720, opened 2021-10-30
Online monitor test

JoranAngevaare:
It's something to test, but we could try to online monitor these


JoranAngevaare:
Would be nice but with the 20s chunks computing requested, this is not possible

PR #740, opened 2021-11-01
Add a few special cases

jjakob03:
The renaming for variables starting with cs1/cs2 is made via explicitly written special cases. Those were not updated for a few added fields. 
See https://straxen.readthedocs.io/en/latest/reference/datastructure_nT.html#event-info-double

### DIFF
--- a/straxen/plugins/double_scatter.py
+++ b/straxen/plugins/double_scatter.py
@@ -57,7 +57,7 @@ class EventInfoDouble(strax.MergeOnlyPlugin):
                          'cs1': 'cs1_a',
                          'cs2': 'cs2_a',
                          'alt_s1_delay': 'ds_s1_dt',
-                         'alt_s2_delay': 'ds_s2_dt'
+                         'alt_s2_delay': 'ds_s2_dt',
                          'cs2_wo_elifecorr': 'cs2_a_wo_elifecorr',
                          'alt_cs2_wo_elifecorr': 'cs2_b_wo_elifecorr',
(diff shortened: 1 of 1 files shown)


JoranAngevaare (changes requested):
Thanks @jjakob03 ! Indeed, let's do add these.

Could you increment the version of the plugin such that we know the code changed?

jjakob03:
Hey @JoranAngevaare, I always guessed so, but actually I do not know for sure. Probably @DonNabla knows more. I just added a space in these two lines for conformity.

JoranAngevaare (approved):

JoranAngevaare:
> 
> 
> Hey @JoranAngevaare, I always guessed so, but actually I do not know for sure. Probably @DonNabla knows more. I just added a space in these two lines for conformity.

Right, sometimes github tricks me in believing these lines are new whereas they are not really 😉 . Thanks, yeah let's certainly not change it when it has always been like this (still interested though)..

Thanks again, as far as I'm concerned, we can merge

DonNabla:
Hello, the name `ds_s1_dt` is an old notation, which could be indeed renamed: S1_delay_time. But a lot of people are used to ds_s1_dt, and are doing analysis based on this name, not sure if it is really necessary to change the name. (for me it would be like rename cS1 into cor_S1) :)

JoranAngevaare:
@DonNabla , definitely let's not break things just because they confuse me :wink: thanks for explaining!

PR #745, opened 2021-11-03
use channel tight coincidence level

JoranAngevaare:
Change the tight coincidence parameter used in the peaklet classification

## Alternatives
We could also revert https://github.com/XENONnT/straxen/pull/691 since it might probably be more confusing to keep two parameters around


WenzDaniel (approved):
Thanks

ershockley:
@JoranAngevaare @WenzDaniel Are we waiting on anything else for this one?

WenzDaniel:
Yes, as discussed within the Team-Leader channel I would like to rename things. `tight_coincidence -> tight_coincidence_hits` and `tight_coincidence_channel -> tight_coincidence` (at least I had the impression that we converged on this one see https://app.slack.com/client/T01602WL334/G01DGC77CAW/thread/G01DGC77CAW-1637223273.020600). But this requires reprocessing `>= peaklets`. So I would like to wait for `global_v6`.

WenzDaniel:
I am also wondering, if we should not wait in general with all the PRs which change the lineage till the end. Might be that some analysts would like to use a certain bug fix or patch which is already merged with master.

ershockley:
Ah right, agreed. THanks!

PR #750, opened 2021-11-07
Explicitly set infer_dtype=False for all Options

jmosbacher:
Due to a bug fixed in strax PR [#569](https://github.com/AxFoundation/strax/pull/569) many config options in straxen will now produce warnings. This PR suppresses these warning by explicitly setting the infer_dtype to False wherever the type of the config Option would have been inferred from the default value in the future.
This breaks backward compatibility with older strax versions since the infer_dtype keyword did not exist before, so not sure this is the best way to suppress these warnings. Maybe a better option would be to have a global setting in strax that suppresses them. Its open for discussion, but the PR is here in case we decide to use it.


WenzDaniel:
Looks fine to me. We can updated the individual option whenever we change something else. I think for most of the int/float options you could also set it already to true. Btw I also noticed that you forgot to updated the doc-string here: https://github.com/AxFoundation/strax/blob/d23a605ab8171d89209aeb10f9d2f86b456a212e/strax/config.py#L66

jmosbacher:
@WenzDaniel Indeed you are correct! I also realized when adding the docstring that infer_dtype is a bit misleading and renamed it to infer_type in this [draft](https://github.com/AxFoundation/strax/pull/587). Until that is resolved, ill convert this to a draft.

JoranAngevaare:
Thanks @jmosbacher , let's merge sooner rather than later to prevent conflicts, I don't see any issue with this 👍

JoranAngevaare (approved):

WenzDaniel:
Any idea why it is just failing for this DAQReader option. I cannot spot the mistake.

jmosbacher:
@WenzDaniel I think the installation test is done with the latest version of strax and not master right? so wouldn't expect it to succeed

JoranAngevaare:
@jmosbacher maybe we should also do this on WFSim, the logs also contain quite a few warnings there.

jmosbacher:
Good point, ok ill take a look at the WFSim code

JoranAngevaare:
@jmosbacher we are also going to have some fun in cutax

JoranAngevaare:
@jmosbacher I'm merging this for now, although I still fill feel inclined to just not type-check options but I want to get a release out.

PR #751, opened 2021-11-07
pyroscope online profiling

jmosbacher:
Add optional pyroscope profiling for bootstrax. It may prove useful to add online profiling since the bottlenecks can change depending on TPC rates/chunk size and DAQ/bootstrax settings. If we start using pyroscope we can set up a grafana dashboard showing the profiling in real time with minimal overhead (5-10%).


bin/bootstrax
```diff
@@ -95,6 +95,12 @@ actions.add_argument(
 actions.add_argument(
     '--undying', action='store_true',
     help="Except any error and ignore it")
+actions.add_argument(
+    '--pyroscope', action='store_true',
+    help="Use pyroscope profiler")
+parser.add_argument(
+    '--pyro_address', type=str, default='http://pyroscope:4040',
```
darrylmasson, line 102: What's involved with hosting the server backend? I assume it's something we can set up ourselves on eb2.

darrylmasson (commented):
I don't know how much use the profiling will get but it's probably a useful thing to have.

bin/bootstrax
jmosbacher: Its pretty lightweight so we can run it anywhere that both bootstrax and grafana (grafana in on the above ground lngs network) can access as long as that machine can spare a bit of CPU and network bandwidth.

JoranAngevaare: CPU and network are almost never the issue, if it were RAM it might have been a different story

JoranAngevaare (approved):
Thanks Yossi, seems like a cool feature.

Looks fine, let's run a test sometime soon to see how it works

JoranAngevaare:
@jmosbacher am I overlooking something or does this [require python2](https://pyrocore.readthedocs.io/en/latest/installation.html#installing-dependency-packages)? I'm not a big fan of installing that on the eventbuilders.

JoranAngevaare:
I've opened a thread on this on slack:
https://xenonnt.slack.com/archives/C016UJZ090B/p1637225256202700

PR #753, opened 2021-11-11
Implement peak-level shadow and event-level shadow refactor

dachengx:

Previously the event-level shadow is defined only considering the time window before the center time of an event. In the PR, we defined the peak-level shadow and redefined the event-level shadow depending on the peak-level shadow. So the definition here is more consistent. 

The peaks always have a `time` field. In some time window before `time`, there will be several so-called 'big previous S2'. The peak shadow is defined as the maximum `s2_area`/`delta t`, where `delta t` is the time difference of current `time` of peak and `center_time` of 'big previous S2'. 

Then the event shadow is defined as the `shadow` of the 1st S2 peak in an event because an event will have at least one S2 peak most of the time. 

We may find more figures in this [note](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:kexin:sr0_shadow_cut). 


ershockley:
@dachengx can you please keep analysis figures in the wiki notes? Remember straxen is a public repo.

dachengx:
> @dachengx can you please keep analysis figures in the wiki notes? Remember straxen is a public repo.

SORRY. I forget it again. Already deleted.

dachengx:
Currently, I can not figure out what is the cause of the remaining failing checks.

straxen/plugins/peak_shadow.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+@numba.njit()
+def _compute_shadow(peaks, split_peaks, res):
+    for p_i, p_a in enumerate(peaks):
+        new_shadow = 0
+        for s2_a in split_peaks[p_i]:
+            if p_a['time'] - s2_a['center_time'] <= 0:
+                continue
+            new_shadow = s2_a['area'] / (p_a['time'] - s2_a['center_time'])
```
skazama, line 72: `p_a['time']` should be `p_a['center_time']`, right?

straxen/plugins/peak_shadow.py
```diff
@@ -0,0 +1,78 @@
...
+            if new_shadow > res['shadow'][p_i]:
+                res['shadow'][p_i] = new_shadow
+                res['pre_s2_area'][p_i] = s2_a['area']
+                res['shadow_dt'][p_i] = p_a['time'] - s2_a['center_time']
+                res['pre_s2_x'][p_i] = p_a['x']
```
skazama, line 77: `p_a` should be `s2_a`?

dachengx: Thanks. This is a mistake. I will look into its impact and correct the codes.

straxen/plugins/peak_shadow.py
dachengx: Originally I have compared them and the difference in `shadow` and `shadow_distance` is tiny. But it is true that `time` is not as good as `center_time`. I will change it.

straxen/plugins/peak_shadow.py
```diff
@@ -0,0 +1,77 @@
...
+export, __all__ = strax.exporter()
+
+
+@export
+@strax.takes_config(
+    strax.Option('pre_s2_area_threshold', default=1000,
+                 help='Only take S2s large than this into account'
+                 ' when calculating PeakShadow [PE]'),
+    strax.Option('time_window_backward', default=int(3e9),
+                 help='Search for S2s causing shadow in this time window [ns]'),
```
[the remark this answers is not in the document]
WenzDaniel: Okay, after reading more thoroughly the code I noticed I am wrong you do it indeed the other way around. But I am a bit confused why as it seems to be to me computational much much heavier. 

Just to be sure that I understand it now. Here is what you do:

1. Select all `n` S2 larger 1000 pe
2. Search in a window of 3 ns for each of these peaks all `m_i` other peaks
3.  Iterate for each of these `n` peaks over the corresponding `m_i` peaks to compute the shadow parameter. 

Would not it be sufficient to select first all large S2s and compute only for those the shadow? In you note you speak about a fairly high threshold for those large S2s.

straxen/plugins/peak_shadow.py
```diff
@@ -0,0 +1,72 @@
...
+        res['time'] = peaks['time']
+        res['endtime'] = strax.endtime(peaks)
+
+        return res
+
+def compute_shadow(peaks, split_peaks, res):
+    if len(res):
+        return _compute_shadow(peaks, split_peaks, res)
+
+@numba.njit()
```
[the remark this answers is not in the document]
WenzDaniel: Sorry, otherwise I cannot comment on this. Could you please cache the corresponding numba function `cache=True`?

WenzDaniel (commented):
Hej @dachengx and @chnlkx thanks for the update. Before I continue with the event level plugin I have already some comments and concerns regarding the peak level plugin. Would it be possible to change the logic as suggested below.

straxen/plugins/peak_shadow.py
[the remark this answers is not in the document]
dachengx: I think the logic order is like this:

1. For any given time `t`, search all s2 peaks which are larger than a threshold(Currently 1000 PE) in [`t`-3s, `t`]. 
2. Calculate all `s2_area`/(`center_time` - `t`), where `center_time` are the center time of those s2 peaks. 
3. Select the largest `s2_area`/(`center_time` - `t`) as the `shadow` of the time `t`

Noted that these steps are for any given time `t`. For peak_shadow plugin, the `t` are just the center time of each given peak.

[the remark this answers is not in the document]
dachengx: Yes, we indeed select those large S2s [here](https://github.com/XENONnT/straxen/blob/69e3ad71b641a199764fa068afdeb6c131a5ec05/straxen/plugins/peak_shadow.py#L45).

straxen/plugins/peak_shadow.py
```diff
@@ -0,0 +1,77 @@
...
+@export
+@strax.takes_config(
+    strax.Option('pre_s2_area_threshold', default=1000,
+                 help='Only take S2s large than this into account'
+                 ' when calculating PeakShadow [PE]'),
+    strax.Option('time_window_backward', default=int(3e9),
+                 help='Search for S2s causing shadow in this time window [ns]'),
+    strax.Option('skip_drift_time', default=0,
+                 help='Deduct max drift time to avoid peak interference in a single event [ns]'))
+class PeakShadow(strax.Plugin):
```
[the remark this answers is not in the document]
dachengx: Thanks, I will add them. Also the numba cache.

straxen/plugins/peak_shadow.py
```diff
@@ -0,0 +1,77 @@
...
+    depends_on = ('peak_basics', 'peak_positions')
+    provides = 'peak_shadow'
+    save_when = strax.SaveWhen.EXPLICIT
+
+    def infer_dtype(self):
+        dtype = [('shadow', np.float32, 'previous s2 shadow [PE/ns]'),
+                 ('pre_s2_area', np.float32, 'previous s2 area [PE]'),
+                 ('shadow_dt', np.int64, 'time diffrence to the previous s2 [ns]'),
+                 ('pre_s2_x', np.float32, 'x of previous s2 peak causing shadow [cm]'),
+                 ('pre_s2_y', np.float32, 'y of previous s2 peak causing shadow [cm]')]
```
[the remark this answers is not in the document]
dachengx: You are right. They are just 'previous big S2'.

straxen/plugins/peak_shadow.py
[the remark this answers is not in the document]
dachengx: I think if we first select the big S2s, we still need to decide which peaks are in the shadow of each big S2s. In that case, we still need a time window and find the peaks in it. 

Additionally, for each `t` we will not know how many big S2s are shading it. So maybe we need to save a list of shadows(and related information) for each `t`. Maybe space-consuming. 

Indeed the latest code calculates the same big S2's shadow for different `t`, but I think this is the straightforward way.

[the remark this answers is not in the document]
WenzDaniel: Okay if this is your new threshold than indeed it does not matter which way around you do it. I thought about the threshold you specify in the abstract of this [note](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:kexin:sr0_shadow_cut).

straxen/plugins/peak_shadow.py
```diff
@@ -0,0 +1,79 @@
...
+
+@export
+@strax.takes_config(
+    strax.Option('pre_s2_area_threshold', default=1000,
+                 help='Only take S2s large than this into account'
+                 ' when calculating PeakShadow [PE]'),
+    strax.Option('time_window_backward', default=int(3e9),
+                 help='Search for S2s causing shadow in this time window [ns]'),
+    strax.Option('skip_drift_time', default=0,
+                 help='Deduct max drift time to avoid peak interference in a single event [ns]'))
```
WenzDaniel, line 15: Maybe we should change this in a similar way as in: 

https://github.com/XENONnT/straxen/blob/090e3faa152c9ab5eaf795ea7b7c16fa1b5eccef/straxen/plugins/event_processing.py#L30-L38

and 

https://github.com/XENONnT/straxen/blob/090e3faa152c9ab5eaf795ea7b7c16fa1b5eccef/straxen/plugins/event_processing.py#L70-L74

Then you can add a another option exclude_drift_time True/False

straxen/plugins/peak_shadow.py
```diff
@@ -0,0 +1,79 @@
...
+    It also gives the area and position infomation of these previous S2s.
+    """
+
+    __version__ = '0.0.2'
+    depends_on = ('peak_basics', 'peak_positions')
+    provides = 'peak_shadow'
+    save_when = strax.SaveWhen.EXPLICIT
+
+    def get_window_size(self):
+        return self.config['time_window_backward']
```
WenzDaniel, line 29: You can also make the window a bit larger than actually needed. In this way you can avoid some boundary conditions more easily.

straxen/plugins/peak_shadow.py
```diff
@@ -0,0 +1,72 @@
...
+        res['pre_s2_x'] = np.nan
+        res['pre_s2_y'] = np.nan
+        compute_shadow(peaks, split_peaks, res)
+
+        res['time'] = peaks['time']
+        res['endtime'] = strax.endtime(peaks)
+
+        return res
+
+def compute_shadow(peaks, split_peaks, res):
```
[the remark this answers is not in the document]
WenzDaniel: I agree with review doc that having a doc string here would be nice.

straxen/plugins/peak_shadow.py
```diff
@@ -0,0 +1,79 @@
...
+    if len(res):
+        return _compute_shadow(peaks, split_peaks, res)
+
+
+@numba.njit(cache=True)
+def _compute_shadow(peaks, split_peaks, res):
+    for p_i, p_a in enumerate(peaks):
+        new_shadow = 0
+        for s2_a in split_peaks[p_i]:
+            if p_a['center_time'] - s2_a['center_time'] <= 0:
```
WenzDaniel, line 71: == 0 I can see, but < 0 should actually never happen is not it?

straxen/plugins/event_shadow.py
```diff
@@ -4,67 +4,48 @@
...
+        for event_i, sp in enumerate(split_peaks):
+            if (sp['type'] == 2).sum() > 0:
+                # Define event shadow as the first S2 peak shadow
+                first_s2_index = np.argwhere(sp['type'] == 2)[0]
+                res['shadow_index'][event_i] = first_s2_index
+                res['shadow'][event_i] = sp['shadow'][first_s2_index]
+                res['pre_s2_area'][event_i] = sp['pre_s2_area'][first_s2_index]
+                res['shadow_dt'][event_i] = sp['shadow_dt'][first_s2_index]
+                res['pre_s2_x'][event_i] = sp['pre_s2_x'][first_s2_index]
+                res['pre_s2_y'][event_i] = sp['pre_s2_y'][first_s2_index]
```
WenzDaniel, line 47: This part I do not understand. What is the motivation for having the first peaks shadow to be the one of the event? Because it is by definition of the shadow the largest one? But in this way you can also have the event itself casting a shadow on itself is not it? Should not it be therefore the first shadow for which `shadow_dt` corresponds to a time outside of the event?
In addition I do not understand why you are only looking for S2s. In principle also a S1 can be in the shadow of an S2 is not it?

WenzDaniel (commented):
Sorry, but I have a few more questions. Sorry, I am a bit slow recently :D

straxen/plugins/peak_shadow.py
dachengx: Thanks. I will change it.

straxen/plugins/event_shadow.py
dachengx: 1. The reason we deliberately choose the first S2's shadow as the event's shadow is exactly as you say. I will check whether the performance of shadow variables changes after I change the peaks from the first to the peak with a `shadow_dt` out of the event. 
2. For the second question I think @chnlkx will have some comments?

straxen/plugins/peak_shadow.py
dachengx: I did not check whether there are <0. If the computing consumption goes the same for ==0 and <0, I think this is fine?

WenzDaniel: Yes, I just wanted to point it out. In case you got negative values I think something might be fishy.

straxen/plugins/event_shadow.py
chnlkx: Yes, we believe large S1s might also cause isolated events, and possibly be different from the S2 shadow. We do plan to investigate the S1 shadow, but in the meantime, we would carry on the AC BG analysis with only S2 shadow and shadow distance first.

WenzDaniel: Sorry, I did not mean that an S1 casts a shadow, but that an S1 is inside an S2 shadow.

dachengx: Yes, you are right. But most events have S2 but some events have no S1. Since peak-level shadow cut will suppress isoS1s, it is not necessary to define the shadow of an event as the shadow of any S1 peak in the event.

dachengx: Additionally, considering max drift time is small, the time window is almost the same for each peak in an event. So the first peak will always be the one with the largest shadow(have checked, for more than 99% of events, the first shadow is the largest). So I just defined shadow as the first S2 peak's shadow.

WenzDaniel: Okay yes this make sense. Thanks for the clarification. Nice work.

WenzDaniel (approved):
Thanks a lot for being so patience with my review. It is a nice work thanks a lot.

WenzDaniel:
Ah okay there is now some conflict in /event_shadow.py which have to be resolved first.

dachengx:
> Ah okay there is now some conflict in /event_shadow.py which have to be resolved first.

I just resolved it.

PR #754, opened 2021-11-11
Disable scada tests

JoranAngevaare:
Scada is failing, I don't feel comfortable disabling verification in the scada interface so until fixed, I propose to not test this piece of code

### DIFF
--- a/tests/test_scada.py
+++ b/tests/test_scada.py
@@ -2,83 +2,83 @@ import numpy as np
 import straxen
 import unittest
 
-@unittest.skipIf(not straxen.utilix_is_configured('scada','scdata_url',),
-                 "Cannot test scada since we have no access to xenon secrets.')")
-def test_query_sc_values():
-    """
-    Unity test for the SCADAInterface. Query a fixed range and check if 
-    return is correct.
-    """
-    print('Testing SCADAInterface')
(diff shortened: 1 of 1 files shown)


jmosbacher (approved):
Obviously im not opposed to this, but my first instinct would be to add another skipIf condition that tests whether scada is reachable instead of commenting out the code? that way your dont have to make a new PR when its fixed... and even better would be to add a flag that disables verification if the tester requests it

WenzDaniel:
A better solution would be https://github.com/XENONnT/straxen/pull/755. But somehow the response from the DAQ server is inconsistent. Sometimes it responds that the verification of the SSL certificate failed, but sometimes it also returns a valid token without raising any issues.

WenzDaniel:
At least locally I had the issue one or two times now.

WenzDaniel:
Should be the better solution https://github.com/XENONnT/straxen/pull/755. However, due to the unreliable answers from the server side I flagged the tests to be always skipped for the moment.

jmosbacher:
@WenzDaniel agreed. But i think if you make e.g 5 requests in series and some fail its safe to conclude that the API is not reliable enough at the moment to test the straxen logic which assumes the existence of a reliable api (due to overload/certificates/some other issue)

PR #755, opened 2021-11-12
Added resource exception

WenzDaniel:

Adds setup which checks if SSL certificate is expired or not.

### DIFF
--- a/tests/test_scada.py
+++ b/tests/test_scada.py
@@ -1,87 +1,107 @@
 import numpy as np
 import straxen
 import unittest
-
-@unittest.skipIf(not straxen.utilix_is_configured('scada','scdata_url',),
-                 "Cannot test scada since we have no access to xenon secrets.')")
-def test_query_sc_values():
-    """
-    Unity test for the SCADAInterface. Query a fixed range and check if 
-    return is correct.
-    """
-    print('Testing SCADAInterface')
-    sc = straxen.SCADAInterface(use_progress_bar=False)
-
-    print('Query single value:')
-    # Simple query test:
-    # Query 5 s of data:
-    start = 1609682275000000000
-    # Add micro-second to check if query does not fail if inquery precsion > SC precision
-    start += 10**6
-    end = start + 5*10**9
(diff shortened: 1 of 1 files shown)


WenzDaniel:
Okay so it is the same here. Sometimes it passes and sometimes it fails.

WenzDaniel:
Okay changed it to always skip for now.

jmosbacher:
I guess [this](https://docs.pytest.org/en/latest/example/simple.html#incremental-testing-test-steps) would be the proper way to handle these types of tests where the test logic depends on some resource which may be available and me not at runtime right? but that's maybe a bit overkill for this small quick fix you are proposing.

tests/test_scada.py
```diff
@@ -1,87 +1,105 @@
...
+        Scada interface as self.sc.
+        """
+        if not straxen.utilix_is_configured('scada','scdata_url',):
+            self.skipTest("Cannot test scada since we have no access to xenon secrets.)")
+        
+        try:
+            self.sc = straxen.SCADAInterface(use_progress_bar=False)
+            self.sc.get_new_token()
+        except requests.exceptions.SSLError:
+            self.skipTest("Cannot reach database since HTTPs certifcate expired.")
```
WenzDaniel, line 105: @jmosbacher does the same as what you are suggesting. The only problem is that not each query triggers the SSLError for some reason. In a few cases you get back a valid token to query parameters. I do not know why, but this I cannot fix on the client side.

WenzDaniel: (But in a later query step the server tells you that the certificate has expired.)

JoranAngevaare: This is exactly the issue right:
https://github.com/XENONnT/straxen/pull/754#issuecomment-966905978

WenzDaniel: Yes, this is why I put a skip flag at the very beginning of the test class so at the moment they are always skipped until the web API is fixed.

WenzDaniel:
@jmosbacher @JoranAngevaare do you guys approve?

jmosbacher:
Im fine with this

PR #757, opened 2021-11-12
Update definition array_valued

jjakob03:
This is an update of the part of the code that recognizes a map as `array_valued = True`. It allows to use InterpolatingMap for array-valued maps not only if they are stored as a histogram, but also if they are stored as a list, as it is suggested by the docstring of InterpolatingMap.

Before checking if `array_valued = True`, there is inserted a check if the map is stored as a list. If this is the case, the previous definition of array_valued does not give the correct result. With the added definition, array_valued is set to True if the map values in the list are also a list. If the map is not stored as a list, the definition remains unchanged.

First, create a dummy map with array valued entries. Then store it in a json file, read it and interpolate at two example positions.:
```
import json
from edit_interpolate import InterpolatingMap, InterpolateAndExtrapolate 
# (if edit_interpolate.py contains the updated function)

_map = {'coordinate_system': [[-18.3, -31.7, -111.5],
  [36.6, -0.0, -111.5],
  [-18.3, 31.7, -111.5],
  [-18.3, -31.7, -37.5],
  [36.6, -0.0, -37.5],
  [-18.3, 31.7, -37.5]],
 'description': 'Array_valued dummy map with lists',
 'name': 'Dummy map',
 'map': [[1.7, 11.1],
  [1.7, 11.1],
  [1.7, 11.0],
  [3.3, 5.7],
  [3.3, 5.8],
  [3.3, 5.7]]}
with open('dummymap.json', 'w') as outfile:
            json.dump(_map, outfile)
        
IntMap=InterpolatingMap(open('dummymap.json','r').read())
print(IntMap([[0,0,0],[0,0,-140]]))
```


l-althueser:
Thanks, looks like a good addition!

zhut19:
Thanks, @jjakob03. This pr should cover the case when the map data is in the list format. The alternative is to reshape the map data e.g. reshape s1 per channel LCE to (x, y, z, n channel).

zhut19 (approved):

WenzDaniel:
Okay, I am happy to merge as you approved, but do you want to do something about the `complex method` warning? (not a high priority though)

PR #758, opened 2021-11-14
Add URL based configs

jmosbacher:
This uses strax PR [#550](https://github.com/AxFoundation/strax/pull/550) to add URL-like configuration options in our plugins.
The main goals of this new approach to configuration

- More flexibility in switching between CMT, get_resource, and static configuration values.
- Remove logic of how to fetch and construct configuration objects from the plugin to improve purity (computational logic only) and maintainability of the plugins.
- Make unit testing easier by separating the logic that uses the configuration from the logic that fetches its current value.
- Increase the expressivity of the CMT values (descriptive string instead of opaque tuple)
- Remove need for hardcoding of special treatment for each correction in CMT when reading values.

# URLs as config options
(From wikipedia) A Uniform Resource Locator (URL), colloquially termed a web address, is a reference to a web resource that specifies its location on a computer network and a mechanism for retrieving it.

URI = scheme ":" ["//" authority] path ["?" query] ["#" fragment]

A concrete plugin example
-----------------------------------

**Current implementation using CMT**

```python
@export
@strax.takes_config(
    strax.Option('min_reconstruction_area',
                 help='Skip reconstruction if area (PE) is less than this',
                 default=10),
    strax.Option('n_top_pmts', default=straxen.n_top_pmts,
                 help="Number of top PMTs")
)
class PeakPositionsBaseNT(strax.Plugin):

    def setup(self):
        self.model_file = self._get_model_file_name()
        if self.model_file is None:
            warn(f'No file provided for {self.algorithm}. Setting all values '
                 f'for {self.provides} to None.')
            # No further setup required
            return

        # Load the tensorflow model
        import tensorflow as tf
        if os.path.exists(self.model_file):
            print(f"Path is local. Loading {self.algorithm} TF model locally "
                  f"from disk.")
        else:
            downloader = straxen.MongoDownloader()
            try:
                self.model_file = downloader.download_single(self.model_file)
            except straxen.mongo_storage.CouldNotLoadError as e:
                raise RuntimeError(f'Model files {self.model_file} is not found') from e
        with tempfile.TemporaryDirectory() as tmpdirname:
            tar = tarfile.open(self.model_file, mode="r:gz")
            tar.extractall(path=tmpdirname)
            self.model = tf.keras.models.load_model(tmpdirname)
            
    def _get_model_file_name(self):
        config_file = f'{self.algorithm}_model'
        model_from_config = self.config.get(config_file, 'No file')
        if model_from_config == 'No file':
            raise ValueError(f'{__class__.__name__} should have {config_file} '
                             f'provided as an option.')
        if isinstance(model_from_config, str) and os.path.exists(model_from_config):
            # Allow direct path specification
            return model_from_config
        if model_from_config is None:
            # Allow None to be specified (disables processing for given posrec)
            return model_from_config

        # Use CMT
        model_file = straxen.get_correction_from_cmt(self.run_id, model_from_config)
        return model_file

```
Notice how all the details on how to fetch the model file and convert it to a python object that is actually needed, is all hardcoded into the plugin. This is not desirable, the plugin should contain processing logic only.

**How this would be refactored using `strax.URLConfig`:**

```python
# These protocols would be defined in straxen for use with 
# all plugins or at the beginning of the script/notebook when doing custom stuff.

@strax.URLConfig.register('cmt')
def get_correction(name, run_id=0, version='ONLINE', detector='nt', **kwargs):
    return straxen.get_correction_from_cmt(run_id, ( name, version, detector=='nt'))

@strax.URLConfig.register('resource')
def get_resource(name, fmt='text', **kwargs):
    return straxen.get_resource(name, fmt=fmt)

@strax.URLConfig.register('format')
def format_arg(arg, **kwargs):
    return arg.format(**kwargs)
    
@strax.URLConfig.register('keras')
def load_keras_model(content, **kwargs):
    fd = io.BytesIO(content)
    with tempfile.TemporaryDirectory() as tmpdirname:
        tar = tarfile.open(fileobj=fd, mode="r:gz")
        tar.extractall(path=tmpdirname)
        model = tf.keras.models.load_model(tmpdirname)
    fd.close()
    return model
```

**Plugin code** 

```python

"""
Then the plugin would just declare that it needs a keras model to function, with no details on where 
the model comes from hardcoded in the plugin itself.
"""
class PeakPositionsBaseNT(strax.Plugin):
    model = strax.URLConfig(type=tf.keras.Model, cache=True)
```

 **The details of where the model object is taken from can be determined by setting the model key of the context config**
The URL is the object being hashed, so it is important to only use pure URLs i.e the same URL should always refer to the same resource.
```python
config = {
    'model': 'keras://resource://cmt://format://{algorithm}_model?version=v1&run_id=plugin.run_id&fmt=binary&algorithm=plugin.algorithm',
}

st = strax.Context(config=config)

```

 When plugin.model is accessed by the compute logic, the algorithm attribute 
of the plugin will be passed to the `format_arg` function, the resulting correction 
name will be passed to the `get_correction` function with the run_id and 
version as keyword arguments, which will return a filename that will be passed 
to the `get_resource` function along with the fmt=binary keyword argument,
 the resulting bytestring will be passed to the `load_keras_model` function that will return the 
model for use. since loading a model is usually an expensive operation we enable caching.

The URL is evaluated recursively:
  1) **keras://** - Loads a keras model from a bytestring 
  2) **resource://** - Loads a xenon resource by name (can also load web URLs), in this case returns a bytestring encoding a keras model.
  3) **cmt://** - Loads value from CMT, in this case it loads the name of the resource encoding the keras model.
  4) **format://** - Calls python format function on input string, in this case it will replace {algorithm} with the name of the algorithm being used.
  5) **{algorithm}_model** - No protocol, is returned as is (just a string).
  6) **?version=v1&run_id=plugin.run_id&fmt=binary&algorithm=plugin.algorithm** - Query is parsed and substituted (plugin.* are replaced with plugin attributes as evaluated at runtime) the values are then passed as keyword arguments to any protocols that include them in their signature. Everythin after the rightmost `?` character is considered the keyword arguments for the protocols.


jmosbacher:
Moved from strax to here where it is more appropriate.

straxen/url_config.py
```diff
@@ -0,0 +1,201 @@
...
+    return val
+
+
+@export
+class URLConfig(strax.Config):
+    """Dispatch on URL protocol.
+    unrecognized protocol returns identity
+    inspired by dasks Dispatch and fsspec fs protocols.
+    """
+
```
JoranAngevaare, line 31:
```suggestion

```

straxen/url_config.py
```diff
@@ -0,0 +1,201 @@
...
+    unrecognized protocol returns identity
+    inspired by dasks Dispatch and fsspec fs protocols.
+    """
+    
+    _LOOKUP = {}
+    SCHEME_SEP = '://'
+    QUERY_SEP = '?'
+    PLUGIN_ATTR_PREFIX = 'plugin.'
+
+    def __init__(self, cache=False, **kwargs):
```
JoranAngevaare, line 37: perhaps add to a docstring that this is in bytes (and/or do a byes to MB conversion in the `if cache` block and rename the variable to `max_cache_mb`)

straxen/url_config.py
```diff
@@ -0,0 +1,201 @@
...
+                # kwarg is referring to a plugin attribute, lets fetch it
+                kwargs[k] = getattr(plugin, v[len(self.PLUGIN_ATTR_PREFIX):], v)
+            else:
+                # kwarg is a literal, add its value to the kwargs dict
+                kwargs[k] = v
+        
+        return self.dispatch(url, **kwargs)
+ 
+@strax.URLConfig.register('cmt')
+def get_correction(name, run_id=0, version='ONLINE', detector='nt', **kwargs):
```
JoranAngevaare, line 162: Maybe set run_id to None and warn by default if the argument is not specified?

bin/bootstrax
```diff
@@ -95,6 +95,12 @@ actions.add_argument(
 actions.add_argument(
     '--undying', action='store_true',
     help="Except any error and ignore it")
+actions.add_argument(
```
JoranAngevaare, line 98: I think this commit was added by accident?

straxen/url_config.py
```diff
@@ -0,0 +1,201 @@
...
+
+from strax.config import OMITTED
+
+export, __all__ = strax.exporter()
+
+
+def parse_val(val):
+    try:
+        val = literal_eval(val)
+    except:
```
JoranAngevaare, line 20:
```suggestion
    except Exception:
```
To prevent excepting baseExceptions:
```
[pep8] reported by reviewdog 🐶
B001 Do not use bare `except:`, it also catches unexpected events like memory errors, interrupts, system exit, and so on.  Prefer `except Exception:`.  If you're sure what you're doing, be explicit and write `except BaseException:`.
```

JoranAngevaare (approved):
Thanks Yossi! I've already reviewed in straxen and as such have close to no comments.

straxen/url_config.py
```diff
@@ -0,0 +1,211 @@
...
+        
+        # seperate the protocol name from the path
+        protocol, _, path =  url.partition(self.SCHEME_SEP)
+
+        # find the corresponding protocol method
+        meth = self._LOOKUP.get(protocol, None)
+        if meth is None:
+            # unrecongnized protocol
+            # evaluate as string-literal
+            return url
```
WenzDaniel, line 90: Given lines 144-152 in fetch, I am wondering if this should not raise an Error. I think here you took already care of all the cases which do not follow your url logic, did not you?.

WenzDaniel (approved):
Thanks Yossi looks very good. Your comments in the code help a lot. Just one question/comment below.

jmosbacher:
> Given lines 144-152 in fetch, I am wondering if this should not raise an Error. I think here you took already care of all the cases which do not follow your url logic, did not you?
@WenzDaniel the protocols are stripped off one by one recursively so this is meant to handle when we reach the end of the recursion i.e the argument of the protocol

WenzDaniel:
Okay I see, but should not we distinguish the two cases "protocol does not exist" and "return just the value". In the first case we should maybe raise an Error otherwise this may lead to unwanted side effects and harder to trace errors. 

I guess you can distinguish the two cases with: 

```
protocol, _, path = url.partition(self.SCHEME_SEP)
```

In the latter the protocol does not exist  and in the latter former `protocol` is empty or none I guess. If you have such a check already somewhere else, never mind.

jmosbacher:
> Okay I see, but should not we distinguish the two cases "protocol does not exist" and "return just the value". In the first case we should maybe raise an Error otherwise this may lead to unwanted side effects and harder to trace errors.

I cant think of any way to distinguish between a user typo and the argument just being a URL so it happens to have the :// characters in it. Im afraid that this is the price we pay for using a standard way of expressing the protocol. On one hand its very self-explanatory because everyone is used to the URL format, but on the other hand we cannot assume that the argument is not a URL meant for the protocol itself to interpret e.g when using the fsspec protocol, its argument will be a URL formatted identically to ours just the protocol names will be unregistered in our system so the entire URL needs to be passed as is when calling the fsspec protocol.

jmosbacher:
@WenzDaniel if we see that it becomes an actual issue, we can always change the URLConfig.SCHEME_SEP to something non standard but it would look a bit weird. alternatively we can require the the argument of the protocol be inside curly brackets like resource://{name_of_resoure} but again we pay the price of users needing to learn a new format instead of using one they are already familiar with which is a big advantage in my opinion.

JoranAngevaare:
sounds good!

PR #761, opened 2021-11-17
Events synchronize

WenzDaniel:

In this PR we adding a new plugin which computes synchronized timestamps between TPC and vetos. Currently, this value is computed based on the delay set at the digitizer level. In an later update we will use CMT to query for the correct time delay values. For the muon-veto we only have currently a dummy plugin which is required to update the computation of the veto-regions and tagged TPC events consistently.   
Last but not least I removed some remnants of an earlier child_plugin implementation (hitlets_** and events_**).


JoranAngevaare:
@WenzDaniel did you find a reviewer?

straxen/plugins/veto_events.py
```diff
@@ -473,3 +461,99 @@ def get_window_size(self):
...
+    """
+    try:
+        import utilix
+    except ModuleNotFoundError:
+        return np.nan
+
+    delay = np.nan
+    if straxen.utilix_is_configured():
+        run_db = utilix.DB()
+        run_meta = run_db.get_doc(run_id)
```
JoranAngevaare, line 516: is there a strong reason why you are the API instead of mongo itself. It sounds unnecessarily slow?
https://github.com/XENONnT/utilix/blob/master/utilix/rundb.py#L177

JoranAngevaare, line 516: we should add a projection. E.g.
```python
>>> import utilix
>>> c = utilix.xent_collection()
>>> c.find_one({'number':1234}, projection = {'number':True})
{'_id': ObjectId('5da90710a72ff248fc8e8f7b'), 'number': 1234}
>>>
```
so here you would need to have the `daq_config` as a porjection

straxen/plugins/veto_events.py
```diff
@@ -473,3 +461,99 @@ def get_window_size(self):
...
+    def compute(self, events_nv):
+        events_sync_nv = np.zeros(len(events_nv), self.dtype)
+        events_sync_nv['time'] = events_nv['time']
+        events_sync_nv['endtime'] = events_nv['endtime']
+        events_sync_nv['time_sync'] = events_nv['time'] + self.total_delay
+        events_sync_nv['endtime_sync'] = events_nv['endtime'] + self.total_delay
+        return events_sync_nv
+
+
+def get_delay(run_id):
```
JoranAngevaare, line 503: - I think we shouldn't have to use nested import here right?
 - I don't think we should just return NaNs without warning is really smart?
 - Perhaps make a check that we are doing a linked mode run (i.e. detector field comprises TPC+X) otherwise I'm not sure if it's super useful to correct timestamps?

straxen/plugins/veto_veto_regions.py
```diff
@@ -33,13 +33,12 @@ class nVETOVetoRegions(strax.OverlapWindowPlugin):
     """
     __version__ = '0.0.1'
```
JoranAngevaare, line 34:
```suggestion
    __version__ = '0.0.2'
```

straxen/plugins/veto_veto_regions.py
```diff
@@ -33,13 +33,12 @@ class nVETOVetoRegions(strax.OverlapWindowPlugin):
...
 
-    depends_on = 'events_nv'
+    depends_on = ('events_nv', 'events_sync_nv')
```
JoranAngevaare, line 36: Do you foresee any issues with the non-linked runs?

JoranAngevaare (changes requested):
Thanks looks rather straightforward, thanks for going through the process!

Maybe we should keep in mind to also have NV independent selections you mention above just to make sure that we can also make selections for runs where we did not run in linked mode?

Some technicalities but I think we should change it:

straxen/plugins/veto_events.py
JoranAngevaare: This decreases the load on the hosts and prevents moving around data that you aren't interested in anyway

WenzDaniel:
Let me answer down here since your comment are all somewhat related to the intermediate plans for this plugin.  In the long run I would like to use the measured delay values by Andrea which we are planing to put into CMT. Hence the entire run-doc querying will be removed at some point. This should also help with the unlinked data which we can not correct otherwise. 

At the moment this plugin is only required for the selections done in cutax such that Team E can do their studies and selections. And here we just have linked mode data.

PR #762, opened 2021-11-17
Event Positions bug corrected z

cfuselli:
the same z_obs variable was returned twice, now changed to z_cor

### _Notes on testing_

### DIFF
--- a/straxen/plugins/event_processing.py
+++ b/straxen/plugins/event_processing.py
@@ -532,7 +532,7 @@ class EventPositions(strax.Plugin):
                        'r_field_distortion_correction': delta_r,
                        'theta': np.arctan2(orig_pos[:, 1], orig_pos[:, 0]),
                        'z_naive': z_obs,
-                       'z': z_obs,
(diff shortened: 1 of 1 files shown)


WenzDaniel (approved):
Oh wow thanks for spotting. Can you sent in slack tech-support a plot z/z_naive versus r**2 to show the difference of the two parameters. We should make people aware.

l-althueser:
No, the z_obs was the one intended! :)

l-althueser:
We should leave a comment there but we actually changed to z_obs since the z correction with FDC was found to be not working correctly and should not be used until fixed (which might never happen). I will find the PR and link here

l-althueser:
See https://github.com/XENONnT/straxen/issues/527

WenzDaniel:
Do you mind to open another PR putting a comment? I agree that it is quite dangerous right now to be mistaken as a bug.

cfuselli:
Oh, I see! Nevermind!

PR #768, opened 2021-11-18
cleanup bootstrax logic for target determiniation

JoranAngevaare:
The infer mode had some issues, let's use fnmatch and clearer code to determine targets.


JoranAngevaare:
nice

PR #770, opened 2021-11-20
Add main peaks' shadow for event shadow

dachengx:

Specify the main peaks' shadow because later some cuts may implement on S1 or S2 peaks. 

Find the main S1 and S2 peaks and save them in new fields: `s1_shadow` and `s2_shadow`. This PR does not change the definition of already existing fields. 

[Note](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:kexin:sr0_shadow_cut)

### DIFF
--- a/straxen/plugins/event_shadow.py
+++ b/straxen/plugins/event_shadow.py
@@ -20,7 +20,7 @@ class EventShadow(strax.Plugin):
 
     def infer_dtype(self):
         dtype = [('s1_shadow', np.float32, 'main s1 shadow [PE/ns]'),
-                 ('s2_shadow', np.float32, 'mian s2 shadow [PE/ns]'),
+                 ('s2_shadow', np.float32, 'main s2 shadow [PE/ns]'),
                  ('shadow', np.float32, 'shadow of event [PE/ns]'),
                  ('pre_s2_area', np.float32, 'previous s2 area [PE]'),
                  ('shadow_dt', np.int64, 'time diffrence to the previous s2 [ns]'),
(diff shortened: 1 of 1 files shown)


kdund:
Is this required for the version if the cut in cutax or can that be added before?

dachengx:
> Is this required for the version if the cut in cutax or can that be added before?

Yes, it is.

straxen/plugins/event_shadow.py
```diff
@@ -9,16 +9,19 @@ class EventShadow(strax.Plugin):
...
-    __version__ = '0.0.7'
+    __version__ = '0.0.8'
     depends_on = ('event_basics', 'peak_basics', 'peak_shadow')
     provides = 'event_shadow'
     save_when = strax.SaveWhen.EXPLICIT
 
     def infer_dtype(self):
-        dtype = [('shadow', np.float32, 'previous s2 shadow [PE/ns]'),
+        dtype = [('s1_shadow', np.float32, 'main s1 shadow [PE/ns]'),
+                 ('s2_shadow', np.float32, 'mian s2 shadow [PE/ns]'),
```
kdund, line 23: mian->main

kdund (approved):
after typo-correction, I approve :)

straxen/plugins/event_shadow.py
dachengx: Thanks. I just corrected it.

WenzDaniel (approved):
Looks fine, we should see after a while if we really need all three "shadow" values. I still have to get used to the meaning of shadow here. For me a shadow is always shaded on something, but not a property of the thing itself. In your definition it is somewhat both (large peaks casts shadow and other peaks within the shadow have the property shadow).

PR #775, opened 2021-11-24
Two manual boundaries (updated 11/24/2021)

yuema137:
**This PR implements the manual boundaries proposed in the note:**
https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:yue:classification_manual_boundary_update_1
Thanks for the suggestions from @JoranAngevaare and @WenzDaniel. I have moved the boundaries into separate functions as they suggested.
Besides, I closed the old PR (#708) because it was based on the old straxen version (v1.1.2) and made this new PR which is consistent with the latest one (v1.1.3)


WenzDaniel:
@yuema137 I did a minor refactor and some style changes mentioned by the review dog. Can you quickly check if I did not do any mistake during the refactoring?

yuema137:
> @yuema137 I did a minor refactor and some style changes mentioned by the review dog. Can you quickly check if I did not do any mistake during the refactoring?

Thanks Daniel! It is much more clear now and I didn't find any mistakes.

straxen/plugins/peaklet_processing.py
```diff
@@ -543,13 +543,21 @@ def compute(self, records_he, start, end):
...
 @strax.takes_config(
-    strax.Option('s1_max_rise_time', default=110, infer_type=False,
-                 help="Maximum S1 rise time for < 100 PE [ns]"),
-    strax.Option('s1_max_rise_time_post100', default=200, infer_type=False,
+    strax.Option('s1_risetime_area_parameters', default=(50, 80, 12), type=(list, tuple),
+                 help="norm, const, tau in the empirical boundary in the risetime-area plot"),
+    strax.Option('s1_risetime_aft_parameters', default=(-1, 2.6), type=(list, tuple),
+                 help=("Slope and offset in exponential of emperical boundary in the rise time-AFT "
+                      "plot. Specified as (slope, offset)")),
+    strax.Option('s1_flatten_threshold_aft', default=(0.7, 100), type=(tuple, list),
```
skazama, line 551: @yuema137 sorry Yue, but it's not clear to me which values (100ns or 80ns) should use be used here. I'm asking because in[ your note](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:yue:classification_manual_boundary) it is defined as 80ns, but here it is defined as 100ns. Which one is correct?

straxen/plugins/peaklet_processing.py
```diff
@@ -543,13 +543,21 @@ def compute(self, records_he, start, end):
...
+                 help=("Threshold for AFT, above which we use a flatted boundary for rise time" 
+                       "Specified values: (AFT boundary, constant rise time).")),
+    strax.Option('n_top_pmts', default=straxen.n_top_pmts, type=int,
+                 help="Number of top PMTs"),
+    strax.Option('s1_max_rise_time_post100', default=200, type=(int, float),
```
skazama, line 556: This current 200 ns cut seems bit loose to me. Do you (or @zhut19 / @WenzDaniel ) have any suggestions?

straxen/plugins/peaklet_processing.py
```diff
@@ -543,13 +543,21 @@ def compute(self, records_he, start, end):
 
 @export
 @strax.takes_config(
-    strax.Option('s1_max_rise_time', default=110, infer_type=False,
-                 help="Maximum S1 rise time for < 100 PE [ns]"),
-    strax.Option('s1_max_rise_time_post100', default=200, infer_type=False,
+    strax.Option('s1_risetime_area_parameters', default=(50, 80, 12), type=(list, tuple),
```
skazama, line 546: Looking at ` upper_rise_time_area_boundary(area, norm, tau, const)`, `const` should 80 and `tau` should be 12, no?

straxen/plugins/peaklet_processing.py
skazama: OK, now I found [your new note](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:yue:classification_manual_boundary_update_1), indicating that 100 ns should be the correct one, but here `s1_flatten_threshold_aft` value of 0.6 is suggested to use. Which one (0.6 or 0.7) is correct?

straxen/plugins/peaklet_processing.py
yuema137: I personally think we can just set it as the constant (B=80ns) in the <100 PE classification boundary, as we know the fluctuation is less for large peaks. But it depends on whether we want to keep the gas S1s.

straxen/plugins/peaklet_processing.py
yuema137: 0.6 is correct and I have fixed that. 
As we discussed earlier, it is more safe to use a loose boundary for now. So I changed it from (0.7, 80) to (0.6, 100). Sorry for the mistake.

straxen/plugins/peaklet_processing.py
skazama: Thanks, I agree that we probably need other inputs to optimize this. So it's fine as it is now at the momement

skazama (approved):

JoranAngevaare:
Based on https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:j_angevaare:pema_efficiency#c_c_update_in_pr775 I think this is fine to go ahead. Thanks Yue!

JoranAngevaare (approved):

JoranAngevaare:
> Actually, @JoranAngevaare can you wait a tiny bit to merge. Since this is very likely the last version, I am doing some quick final checks.

Of course, there are also pending other PRs with changes to C&C, I just wanted to approve 👍

WenzDaniel:
Thanks for spotting and fixing. Yes I think we should only merge those lineage changing PRs shortly before the new release.

straxen/plugins/peaklet_processing.py
```diff
@@ -574,7 +574,7 @@ def upper_rise_time_area_boundary(area, norm, const, tau):
         Function which determines the upper boundary for the rise-time
         for a given area.
         """
-        return norm*np.exp(-area/tau) + const
+        return norm*np.exp(area/tau) + const
```
skazama, line 577: sorry @zhut19, don't we need minus here, no?

zhut19:
Sorry, I made the mistake, it should be the other minus sign

straxen/plugins/peaklet_processing.py
```diff
@@ -574,15 +574,15 @@ def upper_rise_time_area_boundary(area, norm, const, tau):
...
+        return norm*np.exp(-area/tau) + const
 
     @staticmethod
     def upper_rise_time_aft_boundary(aft, slope, offset, aft_boundary, flat_threshold):
         """
         Function which computes the upper rise time boundary as a function
         of area fraction top.
         """
-        res = 10**(-slope * aft + offset)
+        res = 10**(slope * aft + offset)
```
skazama, line 585: sorry @zhut19, still the slope is initialized as "-1", so we don't need to change, no?

WenzDaniel: I think the current version is the correct one is not it? Or what do you mean @skazama ?

WenzDaniel: Okay now I got it sorry.

PR #776, opened 2021-11-25
Set shadow plugins as SaveWhen.ALWAYS

dachengx:

This PR suggests setting PeakShadow and EventShadow plugin as `strax.SaveWhen.ALWAYS`. Additionally, set `infer_type` explicitly. 

### DIFF
--- a/straxen/plugins/event_shadow.py
+++ b/straxen/plugins/event_shadow.py
@@ -16,7 +16,7 @@ class EventShadow(strax.Plugin):
     __version__ = '0.0.8'
     depends_on = ('event_basics', 'peak_basics', 'peak_shadow')
     provides = 'event_shadow'
-    save_when = strax.SaveWhen.EXPLICIT
+    save_when = strax.SaveWhen.ALWAYS
 
     def infer_dtype(self):
         dtype = [('s1_shadow', np.float32, 'main s1 shadow [PE/ns]'),
--- a/straxen/plugins/peak_shadow.py
+++ b/straxen/plugins/peak_shadow.py
@@ -9,20 +9,20 @@ export, __all__ = strax.exporter()
 
 @export
 @strax.takes_config(
-    strax.Option(name='pre_s2_area_threshold', default=1000,
+    strax.Option(name='pre_s2_area_threshold', default=1000, infer_type=False,
                  help='Only take S2s large than this into account'
                  ' when calculating PeakShadow [PE]'),
-    strax.Option(name='deltatime_exponent', default=-1.0,
+    strax.Option(name='deltatime_exponent', default=-1.0, infer_type=False,
                  help='The exponent of delta t when calculating shadow'),
-    strax.Option('time_window_backward', default=int(3e9),
+    strax.Option('time_window_backward', default=int(3e9), infer_type=False,
                  help='Search for S2s causing shadow in this time window [ns]'),
     strax.Option(name='electron_drift_velocity',
-                 default=('electron_drift_velocity', 'ONLINE', True),
+                 default=('electron_drift_velocity', 'ONLINE', True), infer_type=False,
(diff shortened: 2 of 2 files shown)


WenzDaniel (commented):
Why do you always have to save it? I understood that computation would be rather quick. 

In addition can you please set the proper types of your options? Setting infer_type to false was just a quick and dirty solution for all other plugins as Yossi fixed a bug in strax's type checking. You can set the type by adding `type=str` or `type=(int, float)` etc. to your options.

WenzDaniel:
* save_when.EXPLICIT means that you still can save the data if you call make like `st.make(run_id, data_type, save=(data_type,))`

dachengx:
> Why do you always have to save it? I understood that computation would be rather quick.
> 
> In addition can you please set the proper types of your options? Setting infer_type to false was just a quick and dirty solution for all other plugins as Yossi fixed a bug in strax's type checking. You can set the type by adding `type=str` or `type=(int, float)` etc. to your options.

- The reason for saving shadow variables during reprocessing is to save time for analysts when they want to implement shadow cut and shadow distance cut. This strategy is similar to `event_pattern_fit` and its related cuts. 
- Yes, I will set the proper type to these options.

JoranAngevaare:
> > Why do you always have to save it? I understood that computation would be rather quick.
> > In addition can you please set the proper types of your options? Setting infer_type to false was just a quick and dirty solution for all other plugins as Yossi fixed a bug in strax's type checking. You can set the type by adding `type=str` or `type=(int, float)` etc. to your options.
> 
>     * The reason for saving shadow variables during reprocessing is to save time for analysts when they want to implement shadow cut and shadow distance cut. This strategy is similar to `event_pattern_fit` and its related cuts.
> 
>     * Yes, I will set the proper type to these options.

Hi @dachengx , thanks, do you have the numbers for a typical run? If I recall correctly, this was O (minutes) for (event_pattern_fit) and O seconds for EventShadow but it might be good to check and compare.

Small note this discussion is fully about code so can be perfectly held on github 😉

dachengx:
> > > Why do you always have to save it? I understood that computation would be rather quick.
> > > In addition can you please set the proper types of your options? Setting infer_type to false was just a quick and dirty solution for all other plugins as Yossi fixed a bug in strax's type checking. You can set the type by adding `type=str` or `type=(int, float)` etc. to your options.
> > 
> > 
> > ```
> > * The reason for saving shadow variables during reprocessing is to save time for analysts when they want to implement shadow cut and shadow distance cut. This strategy is similar to `event_pattern_fit` and its related cuts.
> > 
> > * Yes, I will set the proper type to these options.
> > ```
> 
> Hi @dachengx , thanks, do you have the numbers for a typical run? If I recall correctly, this was O (minutes) for event_pattern_fit and O (seconds) for EventShadow but it might be good to check and compare.
> 
> Small note this discussion is fully about code so can be perfectly held on github 😉

Sorry for the late reply. I checked 10 bkg runs and 10 radon runs which are all ~30min. For bkg run, the average time consumption is ~2min, while for radon runs, ~7min.

skazama:
@dachengx Thanks Dacheng, as you said above it took significant amount of time to use shadow cut for Rn220 ER-leakage studies (me and @jingqiangye ), so it's nice if we can keep these plugins like event_pattern_fit. @JoranAngevaare Do you think the time Dacheng sent above is reasonable?

JoranAngevaare (changes requested):
@skazama @dachengx sorry, I did not pay full attention to https://github.com/XENONnT/straxen/pull/753 but these processing times are __really__ long. 

Let me open a PR to fix this because this plugin shouldn't be this is so slow

WenzDaniel:
Closed due to https://github.com/XENONnT/straxen/pull/806

PR #780, opened 2021-11-29
S2 pattern fit

shenyangshi:
It adds a data-driven version of s2 pattern likelihood in event_pattern_fit, for s2 pattern likelihood cut development in cutax.

A tensorflow model fetched from strax_auxillary file is loaded, with position and pattern of S2, the chi2 can be calculated.

The user manual is 
https://github.com/XENONnT/analysiscode/blob/master/event_patternfit/neural_net_patternfit/Plugin_Development.ipynb, while the supporting note is https://xe1t-wiki.lngs.infn.it/doku.php?id=shenyang:s2_pattern_likelihood.


straxen/plugins/event_patternfit.py
```diff
@@ -305,7 +316,42 @@ def compute_s2_llhvalue(self, events, result):
...
+            tar = tarfile.open(self.model_file, mode="r:gz")
+            tar.extractall(path=tmpdirname)
+
+            import tensorflow as tf
+            def _logl_loss(patterns_true, likelihood):
+                return likelihood / 10.
+            self.model = tf.keras.models.load_model(tmpdirname,
+                                                    custom_objects={"_logl_loss": _logl_loss})
+            self.model_chi2 = tf.keras.Model(self.model.inputs,
+                                             self.model.get_layer('Likelihood').output)
```
WenzDaniel, line 334: I think rather than getting the file during run time we should do it only once during the setup of the plugin like it is done for the peak position reconstruction: https://github.com/XENONnT/straxen/blob/b73bd26feef431be55cf963e19a3d204739e1054/straxen/plugins/position_reconstruction.py#L52
Or am I missing something why we have to redo it for each chunk of data?

shenyangshi: Thank you for pointing this out, Daniel! I'll make this right in the next commit.

straxen/plugins/event_patternfit.py
```diff
@@ -305,7 +331,27 @@ def compute_s2_llhvalue(self, events, result):
...
+
+    def compute_s2_neural_llhvalue(self, events, result):
+        for t_ in ['s2', 'alt_s2']:
+            # Selecting S2s for pattern fit calculation
+            # - must exist (index != -1)
+            # - must have total area larger minimal one
+            # - must have positive AFT
+            x, y = events[t_ + '_x'], events[t_ + '_y']
+            cur_s2_bool = (events[t_ + '_area'] > self.config['s2_min_area_pattern_fit'])
+            cur_s2_bool &= (events[t_ + '_index'] != -1)
```
jingqiangye, line 343: It's not from this PR, but isn't that redundant with the minimum S2 area requirement?

jingqiangye (commented):
@shenyangshi Thanks a lot for the PR. I used the notebook you provided to load one dataset, and didn't find anything suspicious. The processing speed is about 10 ms/event. If the speed is ok, then I think we can approve the PR so that Shenyang can work on the following study. @WenzDaniel

straxen/plugins/event_patternfit.py
shenyangshi: I didn't quite get the line of 343 actually. In which case can an s2_index be -1?

jingqiangye: If s2 doesn't exist, then `s2_index` is -1. However, since we already impose a minimum area for s2, the existence requirement seems to be redundant to me.

shenyangshi: Sure, thanks for pointing this out! I deleted this line and tested it.

WenzDaniel (approved):
Looks good just two minor comments regarding readability. 

Concerning the speed. event_pattern_fit is already awfully slow. 10 ms/event is super terrible, so I think there is no other way then processing it on the ebs. @JoranAngevaare what do you think?

straxen/plugins/event_patternfit.py
```diff
@@ -305,7 +330,26 @@ def compute_s2_llhvalue(self, events, result):
...
-                    
+
+    def compute_s2_neural_llhvalue(self, events, result):
+        for t_ in ['s2', 'alt_s2']:
+            # Selecting S2s for pattern fit calculation
+            # - must exist (index != -1)
+            # - must have total area larger minimal one
+            # - must have positive AFT
+            x, y = events[t_ + '_x'], events[t_ + '_y']
+            cur_s2_bool = (events[t_ + '_area'] > self.config['s2_min_area_pattern_fit'])
```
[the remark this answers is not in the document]
shenyangshi: Thanks for the comments! I've updated these in the commit.

Issue #769, opened 2021-11-19
URL configs need testing


JoranAngevaare:
After PR #758

jmosbacher:
Tests added in #781

PR #781, opened 2021-11-30
URLConfig not in strax

ershockley:
There is a small bug here, I presume from this PR originating in strax or vice versa? 

In current release, I get:

```
>>> from straxen.url_config import URLConfig
[lots of UserWarnings...]
  File "<stdin>", line 1, in <module>
  File "/cvmfs/xenon.opensciencegrid.org/releases/nT/development/anaconda/envs/XENONnT_development/lib/python3.8/site-packages/straxen/url_config.py", line 168, in <module>
    @strax.URLConfig.register('cmt')
AttributeError: module 'strax' has no attribute 'URLConfig'
```

This PR should fix this bug.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,6 @@
 numpy~=1.19.5
 packaging~=20.8
+pymongo<=3.12.0
 strax>=1.1.2
 utilix>=0.5.3
 # tensorflow>=2.3.0  # Optional, to (re)do posrec
--- a/tests/test_url_config.py
+++ b/tests/test_url_config.py
@@ -1,7 +1,9 @@
 
+import json
 import strax
 import straxen
 import numpy as np
+import fsspec
 from warnings import warn
 from straxen.test_utils import nt_test_context, nt_test_run_id
 import unittest
@@ -15,7 +17,6 @@ class TestPlugin(strax.Plugin):
     
     def compute(self):
         pass
-
 
 class TestURLConfig(unittest.TestCase):
     def setUp(self):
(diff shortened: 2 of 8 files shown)


ershockley:
I should add that this is in the current development container.

jmosbacher (approved):
Thanks Evan for catching this! we moved this code from strax to straxen rather hastily prior to merging and somehow this got unnoticed. I also added an import of the url_config in the package init file.

JoranAngevaare:
Maybe this would also be a good time to add some tests 👼 ?

jmosbacher:
> Maybe this would also be a good time to add some tests angel ?
Yes indeed!

jmosbacher:
@JoranAngevaare I added some basic tests. for now no fancy hypothesis testing, just some hardcoded examples.

jmosbacher:
No addresses #769

jmosbacher:
now addresses #782

jmosbacher:
OK tests are finally passing. The testing envs were running pymongo 4.0 which introduced some breaking changes. Hopefully I found all of the effects on our code.

requirements.txt
```diff
@@ -1,5 +1,6 @@
 numpy~=1.19.5
 packaging~=20.8
+pymongo<=3.12.0
```
JoranAngevaare, line 3: Is this still needed after the changes you made?

tests/test_url_config.py
```diff
@@ -0,0 +1,55 @@
...
+    depends_on = ()
+    dtype = strax.time_fields
+    provides = ('test_data',)
+    test_config = straxen.URLConfig(default=42,)
+    
+    def compute(self):
+        pass
+
+
+class TestURLConfig(unittest.TestCase):
```
JoranAngevaare, line 20: Maybe also add a test for the `protocol_descr` and `fsspec`, mostly because I think these tests are actually super instructive examples of how the code should be used.

JoranAngevaare (approved):
Thanks Yossi, looks good and it's nice to have some tests and examples in place!

requirements.txt
jmosbacher: only if you want to be extra careful. If I successfully found all the breaking changes then its no longer needed.

tests/test_url_config.py
jmosbacher: added a test for fsspec protocol. not sure how to test protocol_descr method since its just a helper method to quickly see what protocols are registered and their call signature

requirements.txt
JoranAngevaare: based on strax' tests I think it's indeed good to add this requirement

JoranAngevaare:
@jmosbacher , can we merge?

jmosbacher:
@JoranAngevaare yeah im not planning on adding anything. @ershockley did you want to add anything else before we merge this?

ershockley:
Sorry I did not see this until now. It looks good -- thanks a lot!

PR #785, opened 2021-11-30
Update EnergyEstimates

ershockley:
This updates the EnergyEstimates plugin to use the new [descriptor config](https://github.com/AxFoundation/strax/pull/550) format, and also adds a few more variables which might be useful parameter spaces to plot things in: 

 - `n_electrons` which is cs2/g2, and includes relative changes in SE gain and extraction efficiency. 
 - `n_photons` which is cs1/g1
 - `tcs2` which is a 'time-corrected' S2 which allows for relative changes in SE gain and extraction efficiency to be corrected for at S2 level. 

It also adds a variable to the CMT hardcoded list. 

The goal is to make it a bit easier to compare data from times when detector conditions might be slightly different, like described in @Jianyu010's [note](https://xe1t-wiki.lngs.infn.it/doku.php?id=jlong:sr0_ramp_up_kr_se_study)

It's rather straightforward, just adds a few more config options (which will go through CMT, work in progress) and then makes correction.

I tested it locally when modifying this plugin on top of straxen v1.1.3.  I still need to test it on straxen master, but I don't anticipate issues.

### DIFF
--- a/straxen/plugins/event_processing.py
+++ b/straxen/plugins/event_processing.py
@@ -626,8 +626,10 @@ class CorrectedAreas(strax.Plugin):
     """
     Plugin which applies light collection efficiency maps and electron
     life time to the data.
+
     Computes the cS1/cS2 for the main/alternative S1/S2 as well as the
     corrected life time.
+
     Note:
         Please be aware that for both, the main and alternative S1, the
         area is corrected according to the xy-position of the main S2.
@@ -652,7 +654,7 @@ class CorrectedAreas(strax.Plugin):
                        f'Fraction of area seen by the top PMT array for corrected {peak_name} S2'),
                       (f'{peak_type}cs2_bottom', np.float32,
                        f'Corrected area of {peak_name} S2 in the bottom PMT array [PE]'),
-                      (f'{peak_type}cs2', np.float32, f'Corrected area of {peak_name} S2 [PE]'), ]
+                      (f'{peak_type}cs2', np.float32, f'Corrected area of {peak_name} S2 [PE]'),]
 
         return dtype
 
--- a/straxen/plugins/event_processing.py
+++ b/straxen/plugins/event_processing.py
@@ -7,6 +7,7 @@ from .position_reconstruction import DEFAULT_POSREC_ALGO_OPTION
 from straxen.common import pax_file, get_resource, first_sr1_run, pre_apply_function
 from straxen.get_corrections import get_correction_from_cmt, get_cmt_resource, is_cmt_option
 from straxen.itp_map import InterpolatingMap
+from straxen.url_config import URLConfig
 export, __all__ = strax.exporter()
 
 
@@ -625,14 +626,12 @@ class CorrectedAreas(strax.Plugin):
     """
     Plugin which applies light collection efficiency maps and electron
     life time to the data.
-
     Computes the cS1/cS2 for the main/alternative S1/S2 as well as the
     corrected life time.
-
     Note:
         Please be aware that for both, the main and alternative S1, the
         area is corrected according to the xy-position of the main S2.
-        
+
         There are now 3 components of cS2s: cs2_top, cS2_bottom and cs2.
         cs2_top and cs2_bottom are corrected by the corresponding maps,
         and cs2 is the sum of the two.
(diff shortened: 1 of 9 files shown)


straxen/plugins/event_processing.py
```diff
@@ -727,46 +728,56 @@ def compute(self, events):
...
-        ('e_light', np.float32, 'Energy in light signal [keVee]'),
-        ('e_charge', np.float32, 'Energy in charge signal [keVee]'),
-        ('e_ces', np.float32, 'Energy estimate [keVee]')
-    ] + strax.time_fields
+    dtype = [('e_light', np.float32, 'Energy in light signal [keVee]'),
+             ('e_charge', np.float32, 'Energy in charge signal [keVee]'),
+             ('e_ces', np.float32, 'Energy estimate [keVee]'),
+             ('n_electrons', np.float32, 'Estimated number of electrons produced'),
+             ('n_photons', np.float32, 'Estimated number of photons produced'),
+             ('tcs2', np.float32, 'Time-corrected S2 (includes SE gain + extraction eff)'),
```
JoranAngevaare, line 742: Perhaps make this a bit more explicit with something like `time_cor_s2` because I thought it was something with `cs2`

straxen/plugins/event_processing.py
```diff
@@ -727,46 +728,56 @@ def compute(self, events):
...
+             ] + strax.time_fields
     save_when = strax.SaveWhen.TARGET
 
+    # config options
+    g1 = straxen.URLConfig(default=0.1426,
+                           help="S1 gain in PE / photons produced")
+    g2 = straxen.URLConfig(default=11.55 / (1 - 0.63),
+                           help="S2 gain in PE / electrons produced")
```
JoranAngevaare, line 750: Looks good, perhaps we can change to nT values? (could even use bodega for that if needed)

Additionally, it would be nice if we can setup the 1T values with the run dependent defaults which should be easy with the URL configs

straxen/plugins/event_processing.py
```diff
@@ -727,46 +728,56 @@ def compute(self, events):
...
+                                   help='Relative extraction efficiency for this run (allows for time dependence)')
+
     def compute(self, events):
-        el = self.cs1_to_e(events['cs1'])
-        ec = self.cs2_to_e(events['cs2'])
+        nph = self.get_nph(events['cs1'])
+        ne = self.get_ne(events['cs2'])
+        el = nph / self.lxe_w
+        ec = ne / self.lxe_w
+        tcs2 = ne * self.g2
```
JoranAngevaare, line 766: especially here, a bit longer naming of the variables doesn't hurt

straxen/plugins/event_processing.py
```diff
@@ -727,46 +728,56 @@ def compute(self, events):
...
-    dtype = [
-        ('e_light', np.float32, 'Energy in light signal [keVee]'),
-        ('e_charge', np.float32, 'Energy in charge signal [keVee]'),
-        ('e_ces', np.float32, 'Energy estimate [keVee]')
-    ] + strax.time_fields
+    dtype = [('e_light', np.float32, 'Energy in light signal [keVee]'),
+             ('e_charge', np.float32, 'Energy in charge signal [keVee]'),
+             ('e_ces', np.float32, 'Energy estimate [keVee]'),
+             ('n_electrons', np.float32, 'Estimated number of electrons produced'),
+             ('n_photons', np.float32, 'Estimated number of photons produced'),
```
JoranAngevaare, line 741: perhaps add that (for the n_photons) you mean prompt scintillation photons produced in the S1 (just to avoid confusion with the S2 also producing photons)

JoranAngevaare (commented):
Thanks Evan, yeah I'm all for changing to the URL configs! 

Perhaps it's better to recreate this PR since it's added on top of #781.

I only have minor suggestions and don't see a big problem with adding these fields.

One personal preference might be that some are trivially computed from each other and I'm always inclined not save `a, b and c` if `a=b/c`. Just because it breaks everything if you later realize that we have a 500 field event_info and we want to get rid of a few fields. I'm not advocating this PR is wrong, it's just my personal preference 😉

ershockley:
Re-made these changes in #817

PR #806, opened 2021-12-01
Speedup shadow plugins by ~100x

JoranAngevaare:
Fix the processing time for the shadow plugins. This replaces [the need ](https://github.com/XENONnT/straxen/pull/776/files) for saving them as computation is trivial

In https://github.com/XENONnT/straxen/pull/753, the function `strax.split_by_touching_windows`  was used with an **extreme window of 3e9 ns**. This was passed on to a numba function. The result is that you are passing for each peak in the dataset a list of all peak within 3e9 ns which is really tough for your memory management, especially since we are also passing all the peaks as an argument as well.

Rather than passing all those peaks and lists of peaks, keep track of what peaks you need to have using the raw output of touching windows.

The effect of this is a speed increase of `~100 x` for run `033341`


WenzDaniel:
Are the actual changes in https://github.com/XENONnT/straxen/pull/806/commits/b705e3c275c0a223c502357d16668fa78d4653a1 ? Since you copied things it is hard to tell.

straxen/plugins/peak_processing.py
```diff
@@ -332,6 +332,103 @@ def find_n_competing(peaks, windows, fraction):
...
+        res['pre_s2_y'] = np.nan
+        if len(peaks):
+            self.compute_shadow(peaks, peaks[mask_pre_s2], split_peaks, self.exponent, res)
+
+        res['time'] = peaks['time']
+        res['endtime'] = strax.endtime(peaks)
+        return res
+
+    @staticmethod
+    @numba.njit
```
WenzDaniel, line 409: Any reason why we are not caching this guy?

JoranAngevaare: It did not speed it up and I think we should only cache if there is a good reason to do so. It can create issues when you are running many jobs at once.

These are the times for computing the plugin:

With caching:
```2.79 s ± 28.2 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)```

without caching:
```2.73 s ± 32.2 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)```

JoranAngevaare:
> Are the actual changes in [b705e3c](https://github.com/XENONnT/straxen/commit/b705e3c275c0a223c502357d16668fa78d4653a1) ? Since you copied things it is hard to tell.

yeah the changes are mostly in there. It's also explained in the PR description

dachengx:
I have tested the plugins and they give the same results as the ones in [master](https://github.com/XENONnT/straxen/blob/1036f050e78b5eb8973f56000821516b8c2caadf/straxen/plugins/peak_shadow.py) give.

dachengx (approved):

straxen/plugins/peak_processing.py
WenzDaniel: Agree

WenzDaniel (approved):
Looks good thanks.

kdund:
I think with two approvals, we can merge this :)

PR #809, opened 2021-12-02
can we disable codefactor please

JoranAngevaare:
Codefactor drives me nuts, it's always complaining if someone is using their own fork and makes pushes from there.


jmosbacher:
I think this is the first PR I've seen where CodeFactor approves :)

PR #813, opened 2021-12-10
Itp test

JoranAngevaare:
Add a test that fails:
https://github.com/XENONnT/base_environment/runs/4483785976?check_suite_focus=true


WenzDaniel:
Okay I reverted https://github.com/XENONnT/straxen/pull/757 feel free to merge it again with master to see if the test passes now. Can merge afterwards.

JoranAngevaare:
@WenzDaniel thanks, actually I would like to see this test fail. I also feel like we should have a few extra test for this function. This test is the bare minimum of course 😉

WenzDaniel:
The current ones running should fail hopefully :D

JoranAngevaare:
okay, they just did, I'm updating with master now and merge if they succeed now

zhut19:
Hi, @JoranAngevaare I think it is caused by one of my mistakes, a typo in the s2 correction causing a metadata field to be considered as a map. I will change the file later but I think it has been excluded from default tests since https://github.com/XENONnT/private_nt_aux_files/pull/129, so the test should pass now.

JoranAngevaare:
@zhut19 , ah okay, good to know thanks for fixing it, I guess fixing the map is one method.

I did see this issue later than https://github.com/XENONnT/private_nt_aux_files/pull/129 (I think) but okay, if you are fixing the map, there is no issue :)

PR #814, opened 2021-12-10
Restore #757

JoranAngevaare:
#757 was merged but it caused errors downstream. Maybe we need some more testing on this class anyway.


JoranAngevaare:
@zhut19 can it be that the map wasn't updated in privevat nt files repo?

l-althueser:
Thanks for the checks!

PR #816, opened 2021-12-13
Upgrade CMT

jmosbacher:
## Goals

This PR is aimed at addressing a number of limitations in the current CMT implementations:

- Rigid indexing - all corrections are indexed by time only, requiring individual corrections for each PMT in the case of the pmt gains. This also limits the corrections database to only store time-dependent values, requiring other solutions for storing corrections without a time dependence.
-  Shared indexing - all versions of a single correction share the same time index, requiring special values (null) as a hack to have independent intervals of validity for each version. This complicates updating interpolated values like the pmt gains.
- Inefficient use of MongoDB - the database is used as an object store for storing pandas dataframes, requiring the upload of the entire dataframe (for all times/versions) when updating a correction.
- Hardcoded corrections - adding a new correction requires both adding a new global version and releasing a new straxen version with the plugin code adapted to using the new value from CMT.
-  Single value storage - Each correction can only hold a single value, this prevent adding extra data fields (eg errors) and metadata fields (eg analyst name, creating date, description etc ) requiring a separate git repo just to store the metadata.
- Global versions as strings - global versions are stored as json strings in a dedicated collection, no time dependence and adding new corrections requires a new global version. global versions have special status in CMT requiring extra code just to manage this collection. 

All these and more will hopefully be solved by this PR.

## Defining correction schemas

Corrections schemas can be defined by subclassing the `BaseCorrectionSchema` or one of its subclasses ( similar to defining plugins/analyses in straxen).

```python
class SimpleCorrection(straxen.BaseCorrectionSchema):
    _NAME = 'simple_correction'

    # The simplest index, matches on exact value. 
    # This is how we define a versioned document without 
    # time dependence
    version: str = rframe.Index()

    # we can define as many fields as we like, each with its own type
    # defined using standard python annotations
    value: str

   def pre_insert(self, db):
        # the base class implementation checks if any documents already exist at the index 
       # and raises an error if it does
        super().pre_insert(db)
        # here you can add any special logic to perform prior to inserts

   def pre_update(self, db, new):
        # the base class implementation raises an error if new values dont match
        # the current ones for this index. we allow replacing the current values
        # with identical ones because the current values may be inferred (i.e interpolated)
        # in which case we allow new documents with the interpolated values since that wont
        # change any interpolated values.
        super().pre_update(db, new)
        # add any extra logic/checks to perform here 

class SomeTimeIntervalCorrection(straxen.TimeIntervalCorrection):
    name = 'time_interval_correction'
    value: int
    error: float
    analyst: str
    description: str
        
class SomeSampledCorrection(straxen.TimeSampledCorrection):
    name = 'sampled_correction'
    value: int

```
There is nothing special about `straxen.TimeSampledCorrection` and `straxen.TimeIntervalCorrection`, they are just implementations of the two most frequent use cases I would expect for time-dependent corrections. They just include a pre-defined index with version and time fields and rules for when you can apply updates.

## Querying values (experts)

Correction documents can then be queried/set from a data source by their index, right now only MongoDB and pandas data sources are supported but support for other sources can be added as needed (eg json, sqlite etc).

```python
import pymongo
import pandas as pd

db = pymongo.MongoClient()['cmt2']
datasource = db["sampled_corrections"]
# or 
datasource = pd.read_csv("sampled_corrections.csv")

SomeSampledCorrection.find(datasource, time=..., version=...)

# Or query the default mongo database
SomeSampledCorrection.find(time=..., version=...)
```

## Querying values (analysts)

For convenience the `straxen.CorrectionDataframes` can be used to easily access all the corrections by name without having to worry about correction classes etc. The `straxen.CorrectionsClient` dynamically loads all subclasses of the corrections class so as soon as a correction is defined in your current python process it is immediately accessible through the client (including tab completion).

### Get a client for a specific database
Of course in most case people will just use the default database dataframes defined e.g. as `straxen.cmt`, but in cases where the MongoDB is not accessible or custom values need to be used for eg testing you may want to load data from an alternative database/dataframe

```python

rdfs = straxen.CorrectionDataframes.from_mongodb(url='localhost', dbname='cmt2')
```

### Access values using the RemoteDataframe interface
Values can be accessed using pandas-like indexing `df.loc[idx]`, `df.at[idx, column]`, df[column].loc[idx] or with the xarray-like `.sel(index_name=idx)` method
```python

# specific remote dataframes can be accessed via dict-like access or attribute access
rdf = rdfs.sampled_correction
# or
rdf = rdfs['sampled_correction']

# These methods return a pandas dataframe
rdf.loc[version,time]
# or
df = rdf.sel(version=version, time=time)
# or
df = rdf.loc[version,time]
# or
df = rdf['value'].loc[version,time]

# pandas-style scalar lookup returns a scalar
value = rdf.at[(version,time), 'value']
```

#### Slicing
You can also omit indices to get results back matching all values of the omitted index

```python
df = rdf.sel(version=version)
# or
df = rdf.loc[version, :]
# or
df = rdf.loc[version]
```
or pass a list a values you want to match on:
```python
df = rdf.sel(version=[0,1], time=time)
```
Slicing is also supported
```python
df = rdf.sel(version=slice(2,10), time=time)

```

The interval index also supports passing a tuple/slice/Interval to query all intervals overlapping the given interval
```python
rdf = rdfs.time_interval_correction
 df = rdf.sel(version=version, time=(time1,time2))
 df = rdf.loc[version, time1:time2]
 df = rdf.sel(version=version, time=pd.Interval(left=time1, right=time2) )

```

## Global versions

Global versions can easily be implemented in this framework as just another correction like all others (multiple collections of global versions can also be defined eg for MC-specific or xenon1t corrections, each with its own unique index/fields):

```python
class GlobalVersions(straxen.BaseCorrectionSchema):
    _NAME = 'global_versions'
    name: str = straxen.Index()
    version: str = straxen.Index()

    correction: str
    labels: dict
  

```
that's it! we now have a correction that references other correction versions. Each document is indexed by the global version number and the correction name. No special logic needs maintained for the global versions, they use the same framework as the other corrections. Adding a new correction to an existing global version is trivial, you just add a new document, and since it references a new correction name it wont clash with any of the existing corrections in that version. The set of all index values defines the uniqueness of a document. The db you are querying either has the new correction or it doesn't, just like any other correction may be defined for the run you are processing and it may not. The behavior is consistent between all missing data, whether its a correction in a global version or a time range in a specific version of a correction, you simply dont get back any results from your query.

But why stop there? Since there is nothing special about a global version, why not make it time dependent?
```python
class CorrectionReference(TimeIntervalCorrection):
    _NAME = 'global_versions'
    
    version: str = rframe.Index()
    name: str = rframe.Index()
    time: rframe.Interval[datetime.datetime] = rframe.IntervalIndex()

    correction: str
    labels: dict

    def get_corrections(self, datasource=None, **overrides):
        labels = dict(self.labels, **overrides)
        if self.correction not in BaseCorrectionSchema._SCHEMAS:
            raise KeyError(f'Reference to undefined schema name {self.correction}')
        schema = BaseCorrectionSchema._SCHEMAS[self.correction]
        return schema.find(datasource, **labels)

```
Now we can have different local versions for different time intervals for a given global version!

## get_correction like functions

A reference implementation of a cmt2 using this framework is added using a local db for testing. Will switch to xenon DB when PR is ready.

```python
corrections_db = straxen.CorrectionsClient.local_mongo()

@export
def cmt2(name, version='ONLINE', label=None ,**kwargs):
    dtime = extract_time(kwargs)
    if label is None:
        return corrections_db[name].sel(time=dtime, version=version, **kwargs)
    return return corrections_db[name].at[(version,time), label]

straxen.cmt2('simple_correction', version=2)
>>> {'value': 'some string', 'version': 2}
```
This is just an example showing that the framework can reproduce the current uses of CMT, the goal is to expand the use cases, eg to add all corrections including ones with no time-dependence. One thing that's also easy to add is for the user to choose which database to take the values from (eg a local file for testing or an alternate MongoDB database with MC values). The data source is meant to be flexible and pluggable.

## Interface with URLConfigs
While its trivial to create a single protocol that accepts a correction name and kwargs and returns the selection e.g.:
```python
config = {
    'elife': 'cmt2://elife?version=0&run_id=plugin.run_id'
}
```
The framework's flexibility opens up other possibilities.
 Since RemoteDataframes are callable, this framework also opens up the possibility to have all correction dataframes registered as protocols to the URLConfig class. If we do that, a URLConfig correction would be configured with e.g.:
```python
config = {
    'elife': 'elife://value?version=0&run_id=plugin.run_id'
}
```
Then you would not have a situation where the cmt protocol was called to fetch a non-existent correction in the case of a typo or straxen version mismatch. If the protocol doesn't exist the plugin would just get url string back (maybe we can add a warning in such cases), making it easier to debug such an error.


jmosbacher:
@JoranAngevaare I dont think its ready for review yet, still a draft. I still haven't added the documentation (most important for review) and the tests (also important for review) 
I tagged people mostly so you can read the example code I put here in the discussion and comment if you think the behavior should be different or notice something that I missed (still haven't added everything I'm planning but still good to mention If you see some feature you want that's missing)

JoranAngevaare:
Thanks Yossi, makes sense, already three suggestions that would help with the review (as you mention too):
 - A small test that does the entire uploading/downloading on a dummy db as we do [for this test ](https://github.com/XENONnT/straxen/blob/master/tests/test_database_frontends.py#L16). you can use the same address
 - Before we merge this PR to the master branch of straxen, we should run the tests also on the XENONnT fork, otherwise we're not going to be able to fully exploit the e.g. the coveralls tool to make sure that the new code is tested
 - I was a moment a bit confused by `straxen/records` not sure if it's hard but maybe a different name would be okay to prevent confusion with `records` 😂 ?

jmosbacher:
The remote_dataframe stuff was moved to an independent package to make reviewing more focused on straxen specific things

JoranAngevaare:
@jmosbacher should we close this in favor of xedocs?

jmosbacher:
@JoranAngevaare Yeah, forgot to close this PR :)

Issue #784, opened 2021-11-30
Plugin options could be streamlined


ershockley:
Right now the default options of our plugins vary from e.g. online CMT corrections (so nT) and old 1T corrections from when straxen first was developed. I think it might be nice at some point to have all defaults be somewhat consistent, ideally with a focus towards nT over 1T. 

I also would be inclined to move to the [descriptor config](https://github.com/AxFoundation/strax/pull/550) format for specifying options, but that's just a personal opinion. 

This is somewhat related to #328. 

Super low priority of course.

JoranAngevaare:
I agree 👍 . Just to explain what the current state is:
 - Set default values for nT, any divergences with 1T should go in the [1T common config](https://github.com/XENONnT/straxen/blob/64bd823a11800ab7d554b5aedc4114dbfd61cfbf/straxen/contexts.py#L422)
 - However, due to the fact that 1T uses per run defaults and we had this issue https://github.com/AxFoundation/strax/issues/400 with that, some options are overwritten in the [**nT** common config](https://github.com/XENONnT/straxen/blob/64bd823a11800ab7d554b5aedc4114dbfd61cfbf/straxen/contexts.py#L33). Ideally, that config should not be there (which is the same as saying that defaults have to reflect nT).

Before the recently added [descriptor config](https://github.com/AxFoundation/strax/pull/550), there wasn't really much we could do about the per_run_defaults options used for 1T.

PR #817, opened 2021-12-14
Update CorrectedAreas (instead of EnergyEstimates)

ershockley:
This replaces #785, putting the changes in question into CorrectedAreas instead of EnergyEstimates, and removing a few columns that are trivial to calculate in response to the useful comments from @JoranAngevaare. 

So in the end, we only add a single new column, `tcs2` which is a time-corrected cs2 variable that can account for changes in single e- gain and extraction efficiency. (@JoranAngevaare I would prefer to keep the variable name as short as possible, but let me know if you feel strongly about having a longer name)

It still uses the descriptor configs because I think they are cool, and I left the changes in EnergyEstimates since the lineage is going to change anyway. 

Thanks to @Jianyu010, @yabbyhome, and @ahiguera-mx, the required variables are in CMT now. So we just add a few lines in places to add this correction.

I think it is useful to keep this `tcs2` variable separate from the main `cs2`, at least for now until it is validated with larger reprocessing campaign.

I've tested with run 18162, which has raw data on dali. I set up my context like this:

```
data_dir = '/dali/lgrandi/shockley/nt/strax_data'
st = cutax.contexts.xenonnt_v5()
st.set_config(dict(gain_model=("to_pe_model", "v6", True)))
st.storage += [strax.DataDirectory(data_dir, readonly=True)]
```


straxen/plugins/event_processing.py
```diff
@@ -643,15 +618,46 @@ class CorrectedAreas(strax.Plugin):
...
+    s1_map = straxen.URLConfig(default=f'interpolatingmap://resource://cmt://s1_xyz_map_{DEFAULT_POSREC_ALGO}?'
+                                       'version=ONLINE'
+                                       '&run_id=plugin.run_id'
+                                       '&fmt=json')
+
+    # online s2_xy map
+    s2_map = straxen.URLConfig(default=f'interpolatingmap://resource://cmt://s2_xy_map_{DEFAULT_POSREC_ALGO}?'
+                                       'version=ONLINE'
+                                       '&run_id=plugin.run_id'
+                                       '&fmt=json')
```
jmosbacher, line 644: Not sure how long it takes to load an interpolating map, but if its not super quick I would recommend adding `cache=True` or `cache=MAXSIZE` here (where MAXSIZE is an integer defining the maximum maps you want it to cache)

straxen/url_config.py
```diff
@@ -229,3 +229,8 @@ def format_arg(arg: str, **kwargs):
     '''apply pythons builtin format function to a string
     '''
     return arg.format(**kwargs)
+
+
+@URLConfig.register('interpolatingmap')
+def load_map(some_map):
```
jmosbacher, line 235: InterpolatingMap accepts kwargs, why not have the protocol accept kwargs and pass them along to InterpolatingMap?

jmosbacher (commented):
Im really glad to see URLConfig being used here with the interpolating map! Any idea why the tests are failing? Im guessing the renaming of `elife_conf` to `elife`?

JoranAngevaare (changes requested):
@ershockley yeah Yossi is right, it's the breaking change of the `elife` config which you are introducing here which is breaking things and probably will also do downstream.

One note on testing, you are disabling the testing for no-db access which I think we should prevent if we can help it, it's why we have the testing utils here:
https://github.com/XENONnT/straxen/blob/master/straxen/test_utils.py#L20-L62
So instead of preventing people to make [these kind of PRs from their own ](https://github.com/XENONnT/straxen/pull/816) forks, we use some dummy values. It's not really the prettiest but the alternative of not testing or sharing our secrets across forks seems worse to me.

Finally, did you also check if this doesn't break e.g. the global v5 context logic? I would guess it does similar to how it's breaking the sim contexts, it's a bit of a pity we don't test these kind of things anywhere

l-althueser:
Do we maybe just want to save the correction factor cS2/tcS2 as one value and a cS2 including the time dependent correction already? This way we bring the correction to all analyses directly and can still remove it if needed by applying the factor.
Otherwise I would suggest to use "cS2" also including time dependence and call the other "cS2_xy" or similar.

I obviously would like the idea of making the new correction directly the default algorithm. Will avoid confusions in the future if we keep it .. but is confusing in another way right now. Which should be ok since we also have completely new global version and thus a hard cut in the work of the analysts.

JosephJHowlett:
I'm with @l-althueser -- I think an e.g. 5keV ER _anywhere_ in the TPC _any time_ during the SR should have the _same cS2_. I'm OK with adding a "tcS2" where the "t" means "time-dependent" (inverse of current PR), or including the ratio as lutz suggests, but isn't the ratio easily calculated using the CMT's EE(t) & SE(t) functions? If so, why do we need a new field? If we don't trust the corrections yet, can't they stay in the CMT until validated?

Jianyu010:
@JosephJHowlett In that case I agree more with Lutz that we should make the all-correction-implemented value the default cS2 to minimize confusion, since it is the closest to the idea of a constant yield from a same energy deposition. For me it was always not as convenient to draw values from CMT as an extra field in event_info, and since these two correction factors are quite new people might frequently visit them, it might be better to keep the values explicit (and also to notify people they exist).

straxen/plugins/event_processing.py
ershockley: Ah I thought caching was turned on by default. So just add this to the url string like so? `.'..&cache=true'` ?

ershockley:
@JoranAngevaare 

>@ershockley yeah Yossi is right, it's the breaking change of the elife config which you are introducing here which is breaking things and probably will also do downstream.

I can see why you think that but I tried reverting this back to `elife_conf` and the tests still broke, so I don't think it's just that. But I will try again with Yossi's suggestions on CMT side. 

>One note on testing, you are disabling the testing for no-db access which I think we should prevent if we can help it, it's why we have the testing utils here:
https://github.com/XENONnT/straxen/blob/master/straxen/test_utils.py#L20-L62
So instead of preventing people to make these kind of PRs from their own forks, we use some dummy values. It's not really the prettiest but the alternative of not testing or sharing our secrets across forks seems worse to me.

I found this file just now, will try using that. Tbh I think this sort of testing with dummy values (and no database) is not particularly useful, we should only care about testing the contexts/functionality we actually use. I do see the issue with the forks though, that's tricky. 

>Finally, did you also check if this doesn't break e.g. the global v5 context logic? I would guess it does similar to how it's breaking the sim contexts, it's a bit of a pity we don't test these kind of things anywhere

It breaks the global_v5 context anyway because we are adding 2 new CMT corrections that aren't in global v5, just like situation we had before. So testing with e.g. xenonnt_v5 (which is in cutax anyway, not straxen) is not useful here, similar comment to above.

straxen/plugins/event_processing.py
jmosbacher: I guess caching by default wouldnt be a terrible idea but I'm always hesitant about introducing caching, debugging cache related issues can be a nightmare... You need to pass it as a kwarg to URLConfig, so `URLConfig(..., cache=True)`.

ershockley:
@l-althueser 

>Do we maybe just want to save the correction factor cS2/tcS2 as one value and a cS2 including the time dependent correction already? This way we bring the correction to all analyses directly and can still remove it if needed by applying the factor.

Looks like I am outnumbered here, so of course we can do it this way also. As long as it is very easy for analysts to compare plots with/out this correction I am happy, as it is the first time we will have made such a correction (that I know of at least). 


@JosephJHowlett 
> If so, why do we need a new field? If we don't trust the corrections yet, can't they stay in the CMT until validated?

At least at the beginning I agree with @Jianyu010 and think it is nice to have them both side by side in the dataframe. Yes in principle you could use CMT to get the correction factors but it's just more overhead (and same exact argument holds for all the extra cS2_* variables we have).

And to validate the corrections we need to actually use them, which we have not yet except for minimal testing.

l-althueser:
@ershockley 

I think I am most convinced of the solution of having "cS2" (xy + time correction) and another "cS2_*" variable in the end. We can then use both and do the comparison very easily. If we want to follow the same approach as we do for positions then we should actually have:
 * cS2 - the one everyone should use - S2 * xy * time * elife
 * cS2_xy - xy correction factor
 * cS2_t - time correction factor
 * cS2_elife - elife correction factor

all of those should then be also produced for "top" and "bottom". Not sure how we treat the "cs2_wo_elifecorr" variables we have right now but I would suggest to make all variables in the same fashion now. No matter how big the change is since we want to anyway do a complete overhaul. 
And if we see that something is not yet working the way we are thinking here it won't be that big of an issue since we can still do small changes in the minor releases until we actually have new S2 correction map. So we have at least one full iteration still ahead, right?

Sorry that I made it a bit more difficult now. We can also discuss in some other place if needed. Will hopefully be around for some time still.

straxen/url_config.py
ershockley: I did try it, but it doesn't work since the kwargs eventually get passed to [InterpolateAndExtrapolate](https://github.com/XENONnT/straxen/blob/master/straxen/itp_map.py#L15) which doesn't accept kwargs, so got an argument error. I allowed for passing a different `method` though. I guess if someone wants to use a different map we can reconsider this or make a new protocol?

ershockley:
Alright well it took me a full work day but looks like maybe I finally got the tests working :) 

Thanks a lot to @jmosbacher for #820! I modified things bit but that was huge help in getting it working. 

To summarize current changes in this branch: 

 - I dropped the `tcs2` variable and instead absorb all corrections into standard `cs2`. There is a `cs2_wo_timecorr` which corresponds to the "old" cs2. 
 - URLConfig now used for CorrectedAreas
 - Since the URLConfigs allow us to skip the `setup` step we had before, I also took the chance to rename the option names so that they match the collection names in CMT. **This is a breaking change** in principle, though in my view it is also much less confusing. The changes are (a) `elife_conf` --> `elife`, (b) `s1_xyz_correction_map` --> `s1_xyz_map`, (c) `s2_xy_correction_map` --> `s2_xy_map`.
 - Added a protocol to URLConfig for making InterpolatedMaps, which is used for the S1 and S2 maps. 
 - Changed `apply_cmt_version` to only give a warning if using the global_ONLINE version. It will still raise an error for the offline versions. 
 

You can test processing data with the following: 

```
data_dir = '/dali/lgrandi/shockley/nt/strax_data'
st = cutax.contexts.xenonnt_online()
st.storage += [strax.DataDirectory(data_dir, readonly=True)]
```

straxen/corrections_services.py
```diff
@@ -327,6 +326,25 @@ def get_cmt_local_versions(global_version):
...
+    args_str = "&".join(args)
+    return f'{url[:args_idx(url)]}?{args_str}'
+
+
+def correction_name_from_urlstring(url):
+    # only care about what happens after cmt
+    before_url, cmt, after_url = url.partition('cmt://')
+    tmp_config = straxen.URLConfig(default=after_url)
+    new_url, kwargs = tmp_config.split_url_kwargs(after_url)
+    return tmp_config.dispatch(new_url, **kwargs)
```
jmosbacher, line 345: nice hack! any reason you decided to put this here and not in the `get_cmt_options()` function? in that function you already have access to the URLConfig instance so you can skip creating a new one. Then you already return the actual names for anyone else that may be using `get_cmt_options()` in their code, no only this script for global_versions...
The only limitation in this hack is that now you cannot have the cmt option you use depend on the run_id or any other run-time configured things. I guess that functionality is not really needed in 99.99% of cases though, plus this is only temporary as CMT is upgraded.

straxen/url_config.py
```diff
@@ -229,3 +230,8 @@ def format_arg(arg: str, **kwargs):
     '''apply pythons builtin format function to a string
     '''
     return arg.format(**kwargs)
+
+
+@URLConfig.register('interpolatingmap')
+def load_map(some_map, method='WeightedNearestNeighbors', **kwargs):
```
jmosbacher, line 236: I guess the `method` is the most important argument. But if analysts actually do use the kwargs options in the InterpolatingMap then its an easy fix to add a cleaning step of the kwargs before they are passed to the various interpolating methods inside the __init__ method of InterpolatingMap. I have an implementation of this in the URLConfig class `URLConfig.filter_kwargs(func, kwargs)`, its a static method so can be called from the class and it will return the subset of kwargs accepted by func.

jmosbacher (commented):
Very nice hack to allow changing the field names, i would probably move it to `get_cmt_options` so you dont break other peoples code but i like it a lot.

straxen/corrections_services.py
```diff
@@ -350,20 +371,41 @@ def apply_cmt_version(context: strax.Context, cmt_global_version: str):
...
-        raise CMTVersionError(f"CMT version {cmt_global_version} is not compatible with this straxen version! "
-                              f"CMT {cmt_global_version} is missing these corrections: {failed_keys}")
+        msg = f"CMT version {cmt_global_version} is not compatible with this straxen version! " \
+              f"CMT {cmt_global_version} is missing these corrections: {failed_keys}"
+
+        # only raise a warning if we are working with the online context
+        if cmt_global_version == "global_ONLINE":
+            warnings.warn(msg)
+        else:
+            raise CMTVersionError(msg)
```
JoranAngevaare, line 409: Shouldn't we just quit if we are using the online version? I don't see why this function would be called at all of online?

straxen/url_config.py
```diff
@@ -229,3 +230,8 @@ def format_arg(arg: str, **kwargs):
     '''apply pythons builtin format function to a string
     '''
     return arg.format(**kwargs)
+
+
+@URLConfig.register('interpolatingmap')
```
JoranAngevaare, line 235: maybe semantics, but I always thought it it was interpolation map, but maybe, ìtp_map` is even better to keep stings a bit more readable? I don't think it matters much but I just want to ask it now since changing this function later is kind of defeating the purpose of URL configs since these roadmaps to the values shouldn't change.

So better ask such a question now when we introduce the method 😄

straxen/plugins/event_processing.py
```diff
@@ -766,6 +755,18 @@ class EnergyEstimates(strax.Plugin):
...
+    # g1 and g2 values default to 1T values
+    g1 = straxen.URLConfig(
+        default=0.1426,
+        help="S1 gain in PE / photons produced")
+    g2 = straxen.URLConfig(
+        default=11.55 / (1 - 0.63),
+        help="S2 gain in PE / electrons produced")
+    lxe_w = straxen.URLConfig(
+        default=13.7e-3,
+        help="LXe work function in quanta/keV")
```
JoranAngevaare, line 768: @ershockley don't we want to make a start with #784 and use the nT defaults here?

Also, the 1T defaults are not correct for all 1T runs, we should make them run dependent. @jmosbacher what would be the cleanest way of doing this with the URL configs?

JoranAngevaare (commented):
Thanks Evan, I really like the URL configs and the changes made in this PR. Thanks for going to the painstaking process of introducing them first 👍 

I made a few superficial changes in https://github.com/XENONnT/straxen/pull/817/commits/521643ad4eb912d219e1c9bd55c6f89475a2db52 with the only actual change the re-enabling of the tests

straxen/url_config.py
jmosbacher: I also like itp_map better as the protocol name, that way its also clear that its using the methods defined in itp_map.py

straxen/plugins/event_processing.py
```diff
@@ -714,11 +716,11 @@ def compute(self, events):
...
             # this is for s2-only events which don't have drift time info
-            cs2_top_wo_elifecorr = (events[f'{peak_type}s2_area'] * events[f'{peak_type}s2_area_fraction_top'] / 
-                                    self.s2_map(s2_positions, map_name=s2_top_map_name))
+            cs2_top_wo_elifecorr = (events[f'{peak_type}s2_area'] * events[f'{peak_type}s2_area_fraction_top'] /
+                                    self.s2_xy_map(s2_positions, map_name=s2_top_map_name))
             cs2_bottom_wo_elifecorr = (events[f'{peak_type}s2_area'] *
                                        (1 - events[f'{peak_type}s2_area_fraction_top']) /
-                                       self.s2_map(s2_positions, map_name=s2_bottom_map_name))
+                                       self.s2_xy_map(s2_positions, map_name=s2_bottom_map_name))
             result[f"{peak_type}cs2_wo_elifecorr"] = (cs2_top_wo_elifecorr + cs2_bottom_wo_elifecorr)
```
WenzDaniel, line 724: I think this guy is now a bit misleading as cs2 != cs2_wo_elifecorr * elife_correction. At least I find this quite confusing.

straxen/plugins/event_processing.py
jmosbacher: @JoranAngevaare maybe add a protocol called e.g. defaults1t that accepts the run_id and returns the defaults? then you can set the configs for 1t to all be `defaults1t://name?run_id=plugin.run_id`

straxen/plugins/event_processing.py
```diff
@@ -728,31 +730,18 @@ def compute(self, events):
...
-            result[f"{peak_type}cs2"] = cs2_top + result[f"{peak_type}cs2_bottom"]
-
+            cs2_top_wo_timecorr = cs2_top_wo_elifecorr * elife_correction
+            cs2_bottom_wo_timecorr = cs2_bottom_wo_elifecorr * elife_correction
+            result[f"{peak_type}cs2_wo_timecorr"] = cs2_top_wo_timecorr + cs2_bottom_wo_timecorr
+
+            # Correct for SEgain and extraction efficiency
+            seg_ee_corr = (self.se_gain / self.avg_se_gain) * self.rel_extraction_eff
+            result[f"{peak_type}cs2_bottom"] = cs2_bottom_wo_timecorr / seg_ee_corr
+            result[f"{peak_type}cs2"] = cs2_top_wo_timecorr / seg_ee_corr + result[f"{peak_type}cs2_bottom"]
```
WenzDaniel, line 740:
```suggestion
            result[f"{peak_type}cs2"] = result[f"{peak_type}cs2_wo_timecorr"] / seg_ee_corr 
```
Is a less complicated in my opinion. Follows the same structure as the line above.

straxen/plugins/event_processing.py
JoranAngevaare: Sounds good, I think such hard coding within straxen would be fine for 1T. Things aren't going to change there anyway.

In theory it could be bad if we update those `defaults1t` with different values, as the lineage wouldn't change. Then again, how big is our focus on 1T really in straxen.

straxen/corrections_services.py
WenzDaniel: Minor comment can you put this below `apply_cmt_version`. In the beginning it was not entirely clear to me why I am reading it. I think in general code should always flow from general to detail.

WenzDaniel (commented):
Few minor comments. The important stuff is already covered by Joran and Yossi.

straxen/plugins/event_processing.py
jmosbacher: Yeah if you wanted to change the defaults in the future you would have to use a new protocol name since once you set a protocol name you commiting to it always giving back the same value for the same inputs. If you think there is a reasonable chance of this happening then just add version parameter to the url  set to 0/1 and for now ignore it in the protocol but if you want to change the defaults in the future you use a new version for the new values. The upgrades to CMT in #816 should make editing the protocol to use version in the future (if needed) very easy

straxen/url_config.py
ershockley: Sure, that's fine with with me.

straxen/plugins/event_processing.py
ershockley: To be honest I don't think we should bother with the 1T stuff, it's a bit of a waste of time in my view. I left the 1T values as defaults because (I'm ashamed to say) I don't know how to use bodega to get the nT values 😕 

@JosephJHowlett or someone can you show me how to get the nT values here in a secret way?

JoranAngevaare: I can see this point of view too, but it's almost trivial to make our changes carefully and make sure that if someday someone wants to use 1T data, this person doesn't have to re-invent the wheel.

for your second question
```python
nT_numbers = straxen.get_resource("XENONnT_numbers.json", fmt="json")
g1 = nT_numbers["g1"]["v2"]["value"]
g2 = nT_numbers["g2"]["v2"]["value"]
```

ershockley: Ok thanks, I will implement the nT defaults (though we should probably talk about putting these numbers into CMT also, the functionality w/ versions etc is almost identical it seems to me). 

If you guys want to implement the 1T stuff please go ahead.

JoranAngevaare: >Ok thanks, I will implement the nT defaults (though we should probably talk about putting these numbers into CMT also, the functionality w/ versions etc is almost identical it seems to me).

From what I understood, this was indeed the idea https://github.com/XENONnT/private_nt_aux_files/pull/61#issuecomment-832988529

JoranAngevaare: > If you guys want to implement the 1T stuff please go ahead.

If our ACs don't care about these things, I'm not going to swim upstream because that would be fighting a lost battle 😂 😉

ershockley: Yeah I remember that discussion, I think maybe we should have it again. g1 and g2 in particular I feel like should be in CMT since they might have some time dependence.

l-althueser: Are those values actually used? If that is the case we should probably incorporate to the corrections workflow as kind of final result ..

straxen/corrections_services.py
ershockley: @jmosbacher, thanks that's a good suggestion. I will move

straxen/url_config.py
ershockley: I did see this filter_kwargs method but I naively assumed it was always being used. Sorry I still don't fully understand though -- what do I need to change to filter the kwargs? I agree this would be good to have

straxen/corrections_services.py
ershockley: Yeah I think it wouldn't be called typically, and that is also why I changed to warning instead of raising an error. So you're saying don't even bother with the warning?

JoranAngevaare: yeah, it's hard to figure out what be the correct way of dealing with abuse (what I would call throwing an ONLINE version into this function) of this code would be.

I would indeed just raise the error (maybe _especially_) if someone is using it on an ONLINE version.

straxen/plugins/event_processing.py
WenzDaniel: Maybe call it `cs2_only_pos_corr`?

WenzDaniel: Otherwise no idea either

straxen/plugins/event_processing.py
ershockley: @l-althueser yes I agree

straxen/corrections_services.py
ershockley: I don't think we want to raise an error for online, as if you hit this problem it just means a new correction has not been added to the global_online collection yet (so defaulting to default values, which should use online _local_ correction anyway, is not an issue). It's the fixed offline versions that need an error as those should never change so will never had the new corrections we add. 

Personally I think a warning seems like good middle ground. It notifies you that something is maybe amiss but doesn't it's not a huge deal. Plus it was nice to test this function with global_ONLINE yesterday since the new offline is not out yet :)

JoranAngevaare: ah okay, you just want to have it as a check of the `global` collection of CMT. Okay, this I would not have guessed 😉 because it looks quite funny if one does:
 - start always from `xenonnt_online` (with online local versions) 
 - for `xenonnt_vX` we do `xenonnt_vX = xenonnt_online().apply_cmt_version("global_vX")` 
 - but then you want to do: `xenonnt_vX().apply_cmt_version("global_ONLNE")`.

straxen/corrections_services.py
ershockley: @jmosbacher sorry now that I'm looking at it I think I actually am not sure what you suggest. 

`get_cmt_options` is getting the strax option/config name and the values for all options that use CMT. So this itself does not care what the actual CMT collection name is (though ideally these things would be the same, they are not).

So what do we gain by adding this to `get_cmt_options` instead of here in `apply_cmt_version` ?

ershockley: I guess maybe we can change what this `get_cmt_options` returns? It could be a nested dict that returns something like 

```
{'strax_option`: {'collection_name': 'some_name',
                             'version': 'some_local_version'
                            }
}
```

straxen/corrections_services.py
ershockley: It does look quite funny, I agree. But it's not wrong in principle right?

straxen/corrections_services.py
jmosbacher: Sorry for not being more explicit. 
What i had in mind was to have `get_cmt_options` return a tuple consistent with the cmt structure so first item is the name and second is maybe the URL string. but i guess its not very crucial, i doubt anyone else has written any code using `get_cmt_options`

jmosbacher: but not sure its worth the effort

straxen/url_config.py
jmosbacher: the filter_kwargs is applied for the protocol functions, but if the protocol accepts kwargs then all of them will be passed in. What I meant is that you can use the function to filter the kwargs that are passed on by the `__init__` method of the InterpolatingMap class. Currently it just passes along all kwargs to the various interpolating methods. I can edit the class and push it later if you think this would be useful, not sure if these extra interpolating options are ever used

straxen/corrections_services.py
ershockley: The way it works now is that it returns a dict like this: 

{'strax_option1': (some_cmt_tuple, if_not, using_urlconfig),
 'strax_option2': 'protocol://could_also_be_a_urlconfig'
}

So what the function in question is doing is grabbing the collection name (which is in CMT tuples already) from the URL string. 

So I still don't understand what the suggestion is, sorry 😕

jmosbacher: I was thinking something like this:

```python
def get_cmt_options(context):
    """
    Function which loops over all plugin configs and returns dictionary
    with option name as key and current settings as values.

    :param context: Context with registered plugins.
    """
    cmt_options = {}
    for data_type, plugin in context._plugin_class_registry.items():
        for option_key, option in plugin.takes_config.items():
            value = option.default
            if option_key in context.config:
                value = context.config[option_key]

            if isinstance(option, URLConfig) and 'cmt://' in value:
                _, _, after_url = value.partition('cmt://')
                new_url, kwargs = option.split_url_kwargs(after_url)
                cmt_name =  option.dispatch(new_url, **kwargs)
                value = (cmt_name, value, True)

            if straxen.get_corrections.is_cmt_option(value):
                cmt_options[option_key] = value
            
           
    return cmt_options

```

then we dont have to touch the `_is_cmt_option` function, so less chance of breaking any existing code somewhere.
 and int the just have to check if the version part of the `apply_cmt_version` function we just check if `cmt://` is in `tup[1]` to know if its a urlconfig and treat it accordingly.

straxen/url_config.py
jmosbacher: I just made a PR #823 for this just in case, you never know when you may need to change some parameter of the interpolation method...

straxen/corrections_services.py
ershockley: Thanks for the clear example @jmosbacher. Personally I would prefer to keep the `_is_cmt_option` how we have it now because that can be easily applied to both CMT tuple configs and url configs. What you propose would return things like this, correct? 

```
{'strax_option1': ('collection_name', 'collection_version', True),  # a tuple config 
  'strax_option2' : ('collection_name', 'cmt://collection_name?etc', True)  # url config
}
```

Where the strax_option2 is a bit nonsensical right? Sorry for being so slow/pedantic here, just still not understanding what we gain.

straxen/plugins/event_processing.py
ershockley: Nothing is actually changing here except for `cs2` and the addition of this `cs2_wo_timecorr` variable, so that would be my preference personally. But interested to hear what others think. @jingqiangye @l-althueser thoughts? Should we change the variable names to be more clear, but also break all the analyses downstream that use these variables like `cs2_wo_elifecorr`?

straxen/corrections_services.py
jmosbacher: Fair enough, to me this just seemed like the natural place to do the lookup for the cmt name. I always hate changing the behavior of an existing function because you never know who uses that function and is relying on the existing behavior in the code.

But I think your implementation is fine, and most importantly it works so I'm fine with it being there until all the configs are converted.

straxen/url_config.py
```diff
@@ -232,6 +232,13 @@ def format_arg(arg: str, **kwargs):
...
+@URLConfig.register('itp_map')
 def load_map(some_map, method='WeightedNearestNeighbors', **kwargs):
+    '''Make an InterpolatingMap'''
     return straxen.InterpolatingMap(some_map, method=method)
+
+@URLConfig.register('bodega')
+def load_value(name: str, version='v0'):
+    '''Load a number from BODEGA file'''
+    nT_numbers = straxen.get_resource("XENONnT_numbers.json", fmt="json")
+    return nT_numbers[name][version]["value"]
```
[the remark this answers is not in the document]
jmosbacher: making a PR now. In any case a more general comment is that we should limit the reuse of keyword arguments as much as possible, so here i would go with `bodega_version` or something similar since version is already use quite ofter with cmt.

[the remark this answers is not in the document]
JoranAngevaare: @ershockley sure, if we are starting it for a lot of things, it's better to add a elegant alias as you did here

I'm just trying to think ahead and prevent adding too many (unused) protocols as removing them later in an attempt to simplify is never going to happen especially if it's easy to use nice features that Yossi keeps adding: https://github.com/XENONnT/straxen/pull/826

We can always add it later if needed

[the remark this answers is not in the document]
jmosbacher: I renamed the bodega_version for now since that's a minimum needed change. Did you also want to use #826 instead of adding this protocol? I am on the fence. On one hand its completely possible to use nested access but on the other hand it reduces the readability of the config URL.
In general my approach would be that we should aim to have enough protocols that make the default configs simple and readable, and use the chaining feature for flexibility when you want to change something without releasing a new straxen version.
But on the third hand, having a million redundant protocols will make a mess for maintenance...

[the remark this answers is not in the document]
JoranAngevaare: also fine, let's wrap now then. I'll just take my loss again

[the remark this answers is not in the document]
ershockley: I vote for keeping the bodega protocol. It's very simple and clear what it is doing.

straxen/plugins/event_processing.py
jingqiangye: Hey @ershockley  my opinions will be:
* keep `cs2_wo_elifecorr` as 1) it's used in the [cS2AFT cut](https://github.com/XENONnT/cutax/blob/0ed85580c6a26e5ce4c1841d2d360bd3d8c83e7e/cutax/cuts/s2_aft.py#L50) and 2) it'll be useful for S2-only events which don't have elife correction
* save the time correction factor for SE_EE, instead of saving `cs2_wo_timecorr`, as it's not directly used in any analysis. But one can always get it by cS2 * time_correction_factor. This is the same thought as @l-althueser 

To make `cs2_wo_elifecorr` meaningful, we need to apply the time correction factor there. If people agree, I can make a commit towards this direction.

ershockley: @jingqiangye `cs2_wo_timecorr` is equivalent to the current cs2, so I think it is definitely useful to have it to compare things as we work on this new correction. Is there a reason that saving the correction factor is better? To me it seems like an unnecessary middle step but maybe I'm missing something. 

And yes good point on the order the corrections are applied. It makes most sense to apply elife last.

jingqiangye: @ershockley I think that `cs2_wo_timecorr` is just a temporary field when we are not sure of the time correction performance. But the time correction factor will be there anyway even if the correction performance is validated. I assume that when we know the correction is good, we don't want `cs2_wo_timecorr` to still be there to complicate things. For the performance validation, we can still easily compute `cs2_wo_timecorr` using the time correction factor.

However, I don't have a strong opinion about this and won't insist if people feel that the other way is better. What I would insist on is to keep `cs2_wo_elifecorr`, but this seems not a problem :)

ershockley: > I think that cs2_wo_timecorr is just a temporary field when we are not sure of the time correction performance. But the time correction factor will be there anyway even if the correction performance is validated

Why do we need the correction factor there long term?

ershockley:
@jingqiangye is the order the corrections are applied better now?

jingqiangye:
@ershockley Thanks! Yes `cs2_wo_elifecorr` makes more sense now. 

Just want to point out that `cs2_wo_timecorr` is not directly comparable to `cs2` as the former also lacks elife correction. But one can get elife correction from `cs2_wo_elifecorr` and `cs2`, and then apply to `cs2_wo_timecorr`. Is this the plan?

jingqiangye:
Maybe a better way is:
1. do S2xy correction, get something called `cs2_xy_only` or so (dont have to save this field though)
2. a) apply time correction to `cs2_xy_only` to get `cs2_wo_elifecorr`
b) apply elife correction to `cs2_xy_only` to get `cs2_wo_timecorr`
3. apply elife correction to `cs2_wo_elifecorr` to get `cs2`

so that both `cs2_wo_elifecorr` and `cs2_wo_timecorr` make sense. what do you think?

ershockley:
Thanks, good suggestion @jingqiangye!  I will implement later today. 

Are there any other open questions?

jingqiangye:
Great thanks @ershockley ! For me no :)

PR #819, opened 2021-12-14
Another update to CorrectedAreas: Use CMT tuples instead of URLConfigs

ershockley:
Same thing as #817, but doesn't use URLConfigs as they are interfering with CMT-related functions like `apply_global_version`, `get_cmt_options`, etc. I think maybe the switch to URLConfigs should wait until CMT2.0, unfortunately :/ 

I will use #817 to instead start switching everythning to URLConfigs, and we can maybe merge it later on along with #816. 

It's yet another attempt to get this `tcs2` variable in, but this time using the tuple CMT configs instead of the new URLConfig.

### DIFF
--- a/straxen/contexts.py
+++ b/straxen/contexts.py
@@ -463,6 +463,9 @@ x1t_common_config = dict(
     electron_drift_velocity=("electron_drift_velocity_constant", 1.3325e-4),
     max_drift_length=96.9,
     electron_drift_time_gate=("electron_drift_time_gate_constant", 1700),
+    avg_se_gain=('avg_se_gain_constant', 28.2),
+    se_gain=('se_gain_constant', 28.2),
+    rel_extraction_eff=('rel_extraction_eff_constant', 1.0)
 )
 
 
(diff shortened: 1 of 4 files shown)


jmosbacher:
@ershockley sorry you had to rewrite this, are you are sure its not an easy fix? see PR #820 for possible fixes for the `apply_global_version` and `get_cmt_options`, did you notice any other incompatibilities besides those two functions?

BTW you dont have to revert to using strax.Option to use the old CMT configs, the URLConfig is backward compatible so you should be able to set it to a CMT tuple for now and then switching in the future wont require changing the plugin.

ershockley:
Thanks @jmosbacher, let me see if that PR will help #817. 

And I spent much less time writing this than struggling with the test suite so it is fine ;)

ershockley:
The tests are passing now in #817 thanks to Yossi's help, so I think we probably don't need this one anymore.

JoranAngevaare:
okay, I'll close this PR

PR #820, opened 2021-12-14
Support URLConfig in apply_cmt_version

jmosbacher:
To maintain full backwards compatibility of URLConfig we need to handle URLs when settings config values from using the `context.apply_cmt_version()` method.

This PR adds a check to the `apply_cmt_version` method whether the default value is using a URL to configure a cmt value, if it is we just set the version parameter in the URL to the new version instead of replacing the entire URL with a CMT tuple. If its not using the cmt:// protocol we ignore the fact its a URLConfig and set it to a tuple.

the `_is_cmt_option` is also edited to return True if the option is a string and includes "cmt://".


jmosbacher:
Implementation moved to #817

Issue #811, opened 2021-12-09
Tensorflow attributes are non-pickable (bad for profiling and mutliprocessing)


JoranAngevaare:
Maybe, like https://github.com/AxFoundation/strax/pull/588, we should avoid setting [TF models as class ](https://github.com/XENONnT/straxen/blob/5c9a33467f362dabdffc75a1164a0bf312d92708/straxen/plugins/position_reconstruction.py) attributes.

I'm not sure how large the networks are and if we would need to clean them up after the plugin finished to avoid memory leaks.

jmosbacher:
Of course I would advocate for getting rid of the `setup()` code and just use a URLConfig with caching like the example I give in #758.
But then again, I am very biased ;)

JoranAngevaare:
Yeah I fully agree it would be nice to start using the `URLConfig`.

The only issue with these neural nets is that you need to [unpack them](https://github.com/XENONnT/straxen/blob/5c9a33467f362dabdffc75a1164a0bf312d92708/straxen/plugins/position_reconstruction.py#L72-L74). So if you just do the naive thing, you end up un-tarring the network for each chunk.

jmosbacher:
Thats exactly what the `URLConfig(cache=True)` is for, it should only be loaded once for each python process.

PR #821, opened 2021-12-15
use URL configs for NeuralNets

JoranAngevaare:
## URL configs for neural nets
In this PR we change the usage of the neural nets to make use of the nice new URL config. This adresses #784 and #811 but probably needs #817 to work well.

If we are using URL configs and make low level changes, let's also do the Neural nets #811.

## Changed plugins
Now, we set the options as an URLConfig. We needed three because we need three different configs for the three plugins.

## Added protocols:
 - `download` is just using the downloader to return a path to a file
 - `tf` which opens a TF file in a temp directory and returns the model object

## Performance
Unchanged

## Testing
afbeelding


straxen/plugins/position_reconstruction.py
```diff
@@ -33,9 +63,7 @@ class PeakPositionsBaseNT(strax.Plugin):
     depends_on = ('peaks',)
     algorithm = None
     compressor = 'zstd'
-    # Using parallel = 'process' is not allowed as we cannot Pickle
-    # self.model during multiprocessing (to fix?)
-    parallel = True
+    parallel = "process"
```
JoranAngevaare, line 66: not working in multi-process mode at this moment

straxen/plugins/position_reconstruction.py
```diff
@@ -91,98 +109,84 @@ def compute(self, peaks):
...
     algorithm = "gcn"
     __version__ = '0.0.1'
 
+    tf_model_gcn = straxen.URLConfig(
+        default=f'tf://download://cmt://{algorithm}_model?version=ONLINE&run_id=plugin.run_id',
+        help='GCN model. Should be opened using the "tf" descriptor. '
+             'Set to "None" to skip computation',
+        cache=True,
+    )
+
```
jmosbacher, line 158:
```suggestion
model = straxen.URLConfig(
        default=f'tf://download://cmt://format_arg://{algorithm}_model?version=ONLINE&run_id=plugin.run_id&algorithm=plugin.algorithm',
        help='TF model. Should be opened using a compatible protocol (e.g. tf://). '
             'Set to "None" to skip computation',
        cache=True,
    )

```

jmosbacher: Any reason youre not using the `format_arg://` to select which model name to load from CMT? I had added that protocol with this use case in mind...

jmosbacher (commented):
Nice! im a bit (a lot) biased but I think this really improves readability. Not sure why you still need the three parameters though, isn't only one of them used for a specific plugin instance?

hopefully multiprocessing will be solved with #822

JoranAngevaare:
> Not sure why you still need the three parameters though, isn't only one of them used for a specific plugin instance?

What you are suggestion doesn't work unless I don't get it but we cannot have "`model`" as the same config key for three algorithms, we want to have `tf_model_mlp` or something as the config key.

Maybe I'm missing something?

straxen/plugins/position_reconstruction.py
JoranAngevaare: Could do, but to be honest I find it less readable than
```python
'tf://download://cmt://{algorithm}_model?version=ONLINE&run_id=plugin.run_id'
```


But again, only if I understand [your general comment correctly](https://github.com/XENONnT/straxen/pull/821#issuecomment-994839106)

jmosbacher:
> Maybe I'm missing something?

hmmm now im thinking maybe I am missing something... To me it looks like the `compute()` method only references the `model` attribute right? so why not let the URLConfig do all the work of loading the appropriate model instead of hardcoding it as a plugin property? that way if you want to use a new type of model, you just change the config value instead of changing the plugin code and add more parameters.

JoranAngevaare:
> To be clear, what I'm suggesting is to only have a single URLConfig defined in the `PeakPositionsBaseNT` base class and setting its value in the context config to be eg `'tf://download://cmt://format_arg://{algorithm}_model?version=ONLINE&run_id=plugin.run_id&algorithm=plugin.algorithm'` which will evaluate to a different model depending on the subclass that is calling it (more specifically the value of plugin.algorithm set in that plugin)

Right this is exactly what we don't want, because you can e.g. not turn off one of the three without making a new CMT version?

jmosbacher:
> because you can e.g. not turn off one of the three without making a new CMT version?

Ah ok now i understand why you did it this way. But dont you anyways have the `recon_alg_included` config to define which algorithms you want to calculate?

JoranAngevaare:
> > because you can e.g. not turn off one of the three without making a new CMT version?
> 
> Ah ok now i understand why you did it this way. But dont you anyways have the `recon_alg_included` config to define which algorithms you want to calculate?

No that is only for the posdiff plugin. 

From some discussion we had long ago, (let's discuss offline if you want to know more), we concluded that the three plugins should be independent and easy to change neural nets. Otherwise I would just have made a single plugin that does all three neural nets, but there was some strong opposition against that.

straxen/plugins/position_reconstruction.py
jmosbacher: this comment is independent of the more general comment. Even if you have three different configs, you still need to format the string before its passed to cmt, otherwise cmt will just get the literal string `{algorithm}_model` as the name of the correction to load...

jmosbacher:
Ok fair enough, just wanted to make sure. But you still need will need to add the `format_arg` protocol to get the correct correction name passed to cmt. 
But if you anyways have a unique config for each net then you can just explicitly add the algorithm name in the default config value, since its just a default so its not really "hard coding" its value.

straxen/plugins/position_reconstruction.py
JoranAngevaare: @jmosbacher sorry, it is an f-string, so you get it from the class attribute (just sneakily changed my example 👿 )

jmosbacher: ah, very sneaky! ok great then this is fine :)

jmosbacher (approved):

jmosbacher:
@JoranAngevaare have you had a chance to check whether #822 solves the multi-processing issue for this plugin?

JoranAngevaare:
@jmosbacher not quite, but I figured out the issue seems to be the CMT LRU config as well. If I also turn this of it works 🎉 could we do the same caching there?

I'm not entirely sure if the caching actually works though:

afbeelding

jmosbacher:
@JoranAngevaare yeah no problem, ill take a look at the CMT code and see if its an easy fix.

Regarding the screenshot you posted, Im assuming `max_workers=2` refers to a process pool and not a thread pool? 
Each process runs its own python runtime so there are no shared objects. Each process will have its own _CACHES dict shared only among the threads belonging to that process...

straxen/plugins/position_reconstruction.py
```diff
@@ -91,98 +109,84 @@ def compute(self, peaks):
...
-            _in = _in / np.max(_in, axis=1).reshape(-1, 1)
-        _in = _in.reshape(-1, self.config['n_top_pmts'])
-        _out = self.model.predict(_in)
+            area_per_channel_top = (
+                    area_per_channel_top /
+                    np.max(area_per_channel_top, axis=1).reshape(-1, 1)
+            )
+        area_per_channel_top = area_per_channel_top.reshape(-1,
+                                                            self.config['n_top_pmts']
+                                                            )
```
WenzDaniel, line 122: FYI
```
area_per_channel_top = (area_per_channel_top.T / np.max(area_per_channel_top, axis=1)).T
```
works too and is a bit shorter. A bit like matrix multiplication, but not really :D

straxen/plugins/position_reconstruction.py
```diff
@@ -9,6 +9,36 @@
...
+
+@export
+@URLConfig.register('download')
+def download(file_name: str) -> str:
+    """
+    Download single and return path to object. Different from
+    "resource://" is that we don't to open the file.
+    """
+    downloader = straxen.MongoDownloader()
+    return downloader.download_single(file_name)
```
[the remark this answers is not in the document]
jmosbacher: @WenzDaniel do you mean to add a keyword argument to the `resource://` protocol (or even the `straxen.get_resource` function) to return a path instead of bytes? I would be on board with that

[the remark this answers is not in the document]
WenzDaniel: I meant `get_resources` itself, but resource protocol is fine too in case we do not want to do it in get_resources.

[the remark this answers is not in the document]
JoranAngevaare: @jmosbacher @WenzDaniel I don't think `get_resource` is the right place to be honest.

Is it okay for you if we can add it to `resource://`? This I think is a good idea.

straxen/plugins/position_reconstruction.py
```diff
@@ -9,6 +9,36 @@
...
 
+
+from straxen import URLConfig
+
+# TODO
+#  #############
+#  Temporary placed here, I don't want to mess with #817
+@URLConfig.register('tf')
+def open_neural_net(model_path: str, **kwargs):
+    # Nested import to reduce loading time of import straxen
```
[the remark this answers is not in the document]
JoranAngevaare: True!

straxen/plugins/position_reconstruction.py
```diff
@@ -7,10 +7,40 @@
...
 import straxen
 from warnings import warn
+from straxen import URLConfig
+
 export, __all__ = strax.exporter()
 
+
+# TODO
+#  #############
+#  Temporary placed here, I don't want to mess with #817
```
WenzDaniel, line 17: Time to be fixed @JoranAngevaare

tests/test_posrec_plugins.py
```diff
@@ -0,0 +1,69 @@
...
+        cls.field = 'x_mlp'
+        cls.run_id = straxen.test_utils.nt_test_run_id
+        cls.st = straxen.test_utils.nt_test_context()
+        cls.st.make(cls.run_id, cls.target)
+
+    def test_set_path(self):
+        """Test that we can reconstruct even if we set a hardcoded path"""
+        # Manually do a similar thing as the URL config does behind the
+        # scenes
+        # TODO this only works after #817
```
WenzDaniel, line 27: One more to go :D

WenzDaniel, line 27:
```suggestion
```

straxen/plugins/position_reconstruction.py
```diff
@@ -87,105 +77,105 @@ def compute(self, peaks):
...
     provides = "peak_positions_mlp"
     algorithm = "mlp"
 
+    tf_model_mlp = straxen.URLConfig(
+        default=f'tf://'
+                f'resource://'
+                f'cmt://{algorithm}_model'
+                f'?version=ONLINE'
+                f'&run_id=plugin.run_id'
+                f'&fmt=abs_path',
```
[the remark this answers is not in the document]
JoranAngevaare: 👍

PR #822, opened 2021-12-15
Move URLConfig cache to global dictionary

jmosbacher:
Impementation of the URLConfig was saving a non-picklable object (lru_cache) to the URLConfig instance. This PR moves the cache to a global dictionary so that instances using cache remain picklable. This is important for e.g. multi-processing.

Also added a test for the caching and picklability of the plugin when caching is enabled.


straxen/misc.py
```diff
@@ -276,3 +277,31 @@ def convert_array_to_df(array: np.ndarray) -> pd.DataFrame:
...
+
+        while len(self) > self.cache_len:
+            oldkey = next(iter(self))
+            super().__delitem__(oldkey)
+
+    def __getitem__(self, key):
+        val = super().__getitem__(key)
+        super().move_to_end(key)
+
+        return val
```
JoranAngevaare, line 307:
```suggestion
        return val

```

straxen/url_config.py
```diff
@@ -47,8 +47,13 @@ def __init__(self, cache=0, **kwargs):
             self.final_type = self.type
             self.type = OMITTED # do not enforce type on the URL
         if cache:
-            maxsize = cache if isinstance(cache, int) else None
-            self.dispatch = lru_cache(maxsize)(self.dispatch)
+            cache_len = cache if isinstance(cache, int) else 1_000_000
+            cache = straxen.CacheDict(cache_len=cache_len)
+            _CACHES[id(self)] = cache
```
JoranAngevaare, line 52: what is the unit of this? Because 1e6 items/config seems huge

tests/test_url_config.py
```diff
@@ -60,5 +71,24 @@ def test_chained(self):
...
+        self.assertEqual(p.cached_config, 666)
+        self.st.set_config({'cached_config': 'random://abc'})
+        p = self.st.get_single_plugin(nt_test_run_id, 'test_data')
+        cached_value = p.cached_config
+        self.assertEqual(cached_value, p.cached_config)
+
+        # test if previous value is evicted, since cache size is 1
+        self.st.set_config({'cached_config': 'random://dfg'})
+        p = self.st.get_single_plugin(nt_test_run_id, 'test_data')
+        self.assertEqual(p.cached_config, p.cached_config)
```
JoranAngevaare, line 85: can this ever be not true?

straxen/url_config.py
```diff
@@ -8,11 +8,11 @@
 from urllib.parse import urlparse, parse_qs
 from ast import literal_eval
 from functools import lru_cache
-
 from strax.config import OMITTED
 
 export, __all__ = strax.exporter()
 
+_CACHES = {}
```
JoranAngevaare, line 15: is there a way of preventing this thing from becoming enormous, we don't do garbage collection right? I might be missing some key features, but it seems to me that only the values that are stored in _CACHES are checked for not becoming too long, whereas I would have expected we would also have to check that _CACHES itself doesn't become too large

straxen/url_config.py
jmosbacher: yeah its number of items. its supposed to be effectively unbounded when you set it to True. But if people tend to just set the cache to True instead of specifying a specific size, then maybe we can set a smaller default value.

tests/test_url_config.py
jmosbacher: it can be if the caching isnt working ;) I'm doing a sneaky trick using a protocol that returns a random number so if the value is evaluated twice it will get a different number but if it returns the cached value it will be the same :) I guess i could change it to a counter if you are afraid of the random function returning the same value twice in a row...

straxen/url_config.py
JoranAngevaare: òkay, I actually thought `True` would mean `int(True)->1` 😂 .

Just to be sure, each run gets it's own instance right and thereby it's own configs (id)? So while the number here basically reflects how many runs (assuming you don't change your config) you can cache for this given option?

straxen/url_config.py
jmosbacher: well the size of _CACHES is limited by the number of plugin classes, each plugin class will have one instance of the URLConfig for each of its configs. So while its trivial to swtich _CACHES to a CachingDict instead of a regular dict, I didn't think it would make any difference, how many cached configs can you already have?

tests/test_url_config.py
JoranAngevaare: ah okay, I see, thanks, mind adding this as a comment, just to be extra clear?

straxen/url_config.py
JoranAngevaare: good question, but I honestly don't know how many we (are going) to have. 

Maybe we can add a warning to where we append new things to _CACHES to check that `sys.sizeof` `_CACHE` and warns that someone might want to free some memory by doing a `_CACHE={}`  or something if the size is bigger than 1GB.

Maybe this warning will never come, but memory leaks are such a messy business to fix especially if they show up in a year from now when we have some crazy map we need to open or something.

straxen/url_config.py
jmosbacher: ah you are right! I forgot True is an instance of int, so yes it will give you a cache with size 1. fixed that.

> Just to be sure, each run gets it's own instance right and thereby it's own configs (id)? So while the number here basically reflects how many runs (assuming you don't change your config) you can cache for this given option?

each plugin class gets its own instance of URLConfig, each instance of URLConfig instance gets its own cache and each unique URL gets its own cache entry. 
So if the URL depends on the run_id then yes, its how many runs are cached, if the URL doesn't depend on the run_id then you can cache much more. If the URL depends on some plugin attribute that is set to a different value for each chunk then that's the maximum number of chunk values kept in memory at any given time etc. your URL can depend on arbitrary plugin attributes that can change during computation. But this feature probably wont be used in practice, most likely it will just depend on run_id.

Its important to mention that its an LRU cache, so its not that if the cache fills up you get an error or something. The least recently used value is just evicted and replaced by the most recently accessed value.

tests/test_url_config.py
jmosbacher: done.

straxen/url_config.py
jmosbacher: True, maybe i should just replace the LRU cache with a TTL cache?

JoranAngevaare: Sounds good, although I don't oppose this LRU implementation either. It's a first and useful version and we might have learn to learn what works best for us.

I just want to prevent making a mistake like I did here: https://github.com/AxFoundation/strax/pull/561 that pops up silently in ways I did not image people using straxen (which was a clearly a limitation of my imagination, not of the people I should add)

straxen/url_config.py
JoranAngevaare: > If the URL depends on some plugin attribute that is set to a different value for each chunk then that's the maximum number of chunk values kept in memory at any given time etc. your URL can depend on arbitrary plugin attributes that can change during computation. But this feature probably wont be used in practice, most likely it will just depend on run_id.

Right, then we should really make sure people don't set a cache at all, there is almost never any reason to assume you need a value more than once for a same chunk within the same plugin 😉

straxen/url_config.py
jmosbacher: OK so for now i added  `config_cache_size_mb` and  `clear_config_caches` functions. If we want to switch to a TTL cache in the future that would be pretty easy.

straxen/misc.py
```diff
@@ -276,3 +283,73 @@ def convert_array_to_df(array: np.ndarray) -> pd.DataFrame:
...
+        if verbose:
+            print(s, type(o), repr(o), file=stderr)
+
+        for typ, handler in all_handlers.items():
+            if isinstance(o, typ):
+                s += sum(map(sizeof, handler(o)))
+                break
+        return s
+
+    return sizeof(o)
```
WenzDaniel, line 355:
```suggestion
    return sizeof(o)

```

straxen/url_config.py
```diff
@@ -47,8 +57,13 @@ def __init__(self, cache=0, **kwargs):
             self.final_type = self.type
             self.type = OMITTED # do not enforce type on the URL
         if cache:
-            maxsize = cache if isinstance(cache, int) else None
-            self.dispatch = lru_cache(maxsize)(self.dispatch)
+            cache_len = 1_000_000 if cache is True else int(cache)
```
WenzDaniel, line 60: Sorry, I am a bit confused right now. Is not the default size a bit long? Can you have more than one entry per URL per run? Even the gains should be just one entry (1 x 494) is not it?

straxen/url_config.py
```diff
@@ -163,7 +178,18 @@ def fetch(self, plugin):
...
                 kwargs[k] = v
 
-        return self.dispatch(url, **kwargs)
+        # construct a deterministic hash key
+        key = strax.deterministic_hash((url, kwargs))
+
+        # fetch from cache if exists
+        value = self.cache.get(key, None)
+
+        # not in cache, letch fetch it
```
WenzDaniel, line 187:
```suggestion
        # not in cache, lets fetch it
```

straxen/url_config.py
jmosbacher: changed it to 100

WenzDaniel:
I do not think so, but let me quickly check one more time.

straxen/misc.py
```diff
@@ -276,3 +283,73 @@ def convert_array_to_df(array: np.ndarray) -> pd.DataFrame:
...
+            super().__delitem__(oldkey)
+
+    def __getitem__(self, key):
+        val = super().__getitem__(key)
+        super().move_to_end(key)
+
+        return val
+
+@export
+def total_size(o, handlers={}, verbose=False):
```
JoranAngevaare, line 316: for good practice, can you not pass mutables as defaults?

https://florimond.dev/en/posts/2018/08/python-mutable-defaults-are-the-source-of-all-evil/

tests/test_url_config.py
```diff
@@ -1,86 +1,125 @@
-import json
```
JoranAngevaare, line 1: can you fix this messed up github history of this entire file? It's quite annoying to have to resolve merge conflicts later

JoranAngevaare (commented):
Two comments on the code, we still need to see this in action with full chain but it's a bit hard to test without any configs that use it 😉 

Also can you add some tests?

straxen/misc.py
```diff
@@ -7,10 +7,17 @@
...
+from itertools import chain
+from collections import OrderedDict, deque
 from importlib import import_module
 from git import Repo, InvalidGitRepositoryError
 from configparser import NoSectionError
 import typing as ty
+try:
+    from reprlib import repr
+except ImportError:
+    pass
```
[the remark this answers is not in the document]
jmosbacher: This is actually not needed, its just an optimization. `repr` is a bultin so if you dont have reprlib it wont raise an error when you try to use `repr`, it will just use the builtin implementation.
I dont think its important, i just copied this snippet from python recipes, perhaps the person who wrote the snippet also worked on reprlib so they included the attempt to import it if it exists.
This is a similar pattern to the 
```python
try:
    import cPickle as pickle
except ImportError:
    pass

that people would always add add the top of their scripts/packages before cPickle became a builtin and the default.

tests/test_url_config.py
jmosbacher: i can try...

jmosbacher:
> Also can you add some tests?

Do you mean beyond the tests i already added? is there a specific functionality you noticed is not tested?

WenzDaniel:
Your cache size function

straxen/misc.py
jmosbacher: done.

tests/test_url_config.py
jmosbacher: hmm i tried reordering to minimize the changed rows but i cant git diff to identify only the changed regions independently. any ideas?

jmosbacher:
lucky i revisted the tests since i had very buggy tests...

tests/test_url_config.py
JoranAngevaare: sometimes it's the end character that messes these things up, but if I'm not able to find that, I usually think re-making the PR saves me the time from going over all of the lines.

Is it your IDE that does something funny or so?

jmosbacher: Indeed when i switched the line endings to CLRF it fixed it. Not sure how windows line endings made it into the file but ok.

jmosbacher:
> Looks still good to me. I am just wondering if we should not cache stuff by default. I think we discussed it before but I cannot find it back.

I had recommended not to cache things by default because the individual protocols have more information on whether to invalidate a cache so i would recommend always implementing caching at the protocol level except in cases where you have a chained protocol where one is passing a non-hashable value to the other so caching would not work at the protocol level.

WenzDaniel:
> individual protocols have more information on whether to invalidate a cache so i would recommend always implementing caching at the protocol level 

Can you add in this case some information into straxen's documentation? I guess it is anyhow on your agenda. At least I think we should at some explanation and examples how the new URL configs and chaining of protocols works. You can just use a notebook if you like see for example: https://github.com/XENONnT/straxen/tree/master/notebooks/tutorials and add it to the documentation (see source in docs). Can be done at any time in an additional PR.

PR #823, opened 2021-12-15
Filter InterpolatingMap kwargs

jmosbacher:
The InterpolatingMap class passes any kwargs given to it directly to the various interpolating functions. This can cause errors when passing kwargs from URLConfigs since the selected interpolation method may not accept a given kwarg even though the `InterpolatingMap.__init__` function does.

This PR introduces a filter step that filters the kwargs based on the signature of each interpolating class.


ershockley:
Thanks @jmosbacher! Should we maybe rebase this into #817 ?

jmosbacher:
@ershockley sure, just wanted to give it first a clean PR so that its easier to review the changes. If it looks good to you then you can pull the changes into your branch

straxen/itp_map.py
```diff
@@ -206,14 +207,16 @@ def _regular_grid_interpolator(csys, map_data, array_valued, **kwargs):
...
         config.update(kwargs)
+        
         return RegularGridInterpolator(tuple(grid), map_data, **config)
 
     @staticmethod
     def _weighted_nearest_neighbors(csys, map_data, array_valued, **kwargs):
         if array_valued:
             map_data = map_data.reshape((-1, map_data.shape[-1]))
-
+        kwargs = straxen.filter_kwargs(InterpolateAndExtrapolate, kwargs)
```
WenzDaniel, line 219: This looks strongly like a decorator to me. Any reasons to not have it as one?

straxen/misc.py
```diff
@@ -276,3 +277,15 @@ def convert_array_to_df(array: np.ndarray) -> pd.DataFrame:
...
+@export
+def filter_kwargs(func, kwargs):
+    """Filter out keyword arguments that
+        are not in the call signature of func
+        and return filtered kwargs dictionary
+    """
+    params = inspect.signature(func).parameters
+    if any([str(p).startswith('**') for p in params.values()]):
+        # if func accepts wildcard kwargs, return all
+        return kwargs
```
WenzDaniel, line 290: Also here is not this, this typical it is easier to ask for forgiveness than permission example:

```
try: 
    f(**keargs):
except:
     # extract kwargs
```
I am fine with both ways.

straxen/itp_map.py
jmosbacher: that is a great idea!

straxen/misc.py
jmosbacher: In general i agree but there are two reasons not to use that pattern here:

1. I prefer not to make any assumptions as to the existence of side effects (the function may be writing to disk/changing some shared object/releasing the Kraken etc.)
2. I prefer not to make assumptions as to the run-time of the function, it may be a very expensive operation/spin indefinitely when it gets the wrong kwargs.

WenzDaniel: I do not fully follow. Why would this be important? You are not really calling the function are you? You directly raise: 

```
def test():
    print('Do nothing yo')

test(bla='a')
```
returns:
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
/tmp/jobs/15223754/ipykernel_112/2655484829.py in <module>
----> 1 test(bla='a')

TypeError: test() got an unexpected keyword argument 'bla'
```
But as I said for me it is fine either way.

jmosbacher: The problem isnt when the try block fails, its when it succeeds ;) i.e. in cases where the function does accept all the kwargs passed to it.

WenzDaniel: Ah okay I see.

JoranAngevaare (approved):

JoranAngevaare:
closing in favor of #817

PR #824, opened 2021-12-16
Add new peaklet classificaiton dtype

ahiguera-mx:
This is a very rough draft 
Add new peaklet classification method based on a Bayes classifier 

The idea is to add 
```
+ [('type_bayse', np.int8, 'Bayes peak classification type')]
             + [('s1_prob', np.float32, 'S1 probability' )]
             + [('s2_prob', np.float32, 'S2 probability' )]
```
Then at `event_processing` user can add if it builds events based on p['type'] or p[''type_bayse'] or p['s1/s2_prob']
TO DO:

- modify `peak` class
- modify `peak_basics` class

`st.make(run_ID, 'peaklet_classification')`

### _Notes on testing_

### DIFF
--- a/straxen/plugins/peaklet_classification.py
+++ b/straxen/plugins/peaklet_classification.py
@@ -144,7 +144,7 @@ class PeakClassificationBayes():
 
 
     def compute(self, peaklets):
-        bayse_ptype = np.zeros(len(peaklets), dtype=np.int8) 
+        bayes_ptype = np.zeros(len(peaklets), dtype=np.int8) 
         s1_prob = np.zeros(len(peaklets), dtype=np.float32) 
         s2_prob = np.zeros(len(peaklets), dtype=np.float32)
 
--- a/straxen/plugins/peaklet_classification.py
+++ b/straxen/plugins/peaklet_classification.py
@@ -1,0 +1,216 @@
+import numpy as np
+from scipy.special import logsumexp
+
+import strax
+import straxen
+export, __all__ = strax.exporter()
+
+@export
+@strax.takes_config(
+    strax.Option('s1_risetime_area_parameters', default=(50, 80, 12), type=(list, tuple),
+                 help="norm, const, tau in the empirical boundary in the risetime-area plot"),
+    strax.Option('s1_risetime_aft_parameters', default=(-1, 2.6), type=(list, tuple),
+                 help=("Slope and offset in exponential of emperical boundary in the rise time-AFT "
+                      "plot. Specified as (slope, offset)")),
+    strax.Option('s1_flatten_threshold_aft', default=(0.6, 100), type=(tuple, list),
+                 help=("Threshold for AFT, above which we use a flatted boundary for rise time" 
+                       "Specified values: (AFT boundary, constant rise time).")),
+    strax.Option('n_top_pmts', default=straxen.n_top_pmts, type=int,
+                 help="Number of top PMTs"),
+    strax.Option('s1_max_rise_time_post100', default=200, type=(int, float),
+                 help="Maximum S1 rise time for > 100 PE [ns]"),
+    strax.Option('s1_min_coincidence', default=2, type=int,
+                 help="Minimum tight coincidence necessary to make an S1"),
+    strax.Option('s2_min_pmts', default=4, type=int,
+                 help="Minimum number of PMTs contributing to an S2"),
+    strax.Option('do_bayes', default=True,
+                 help="run bayes classification method"),
+    strax.Option('bayes_CPT_config', default='/home/ahiguera-mx/test2/rap-ml-group/peak_classification/conditional_probabilities.npy',     
+                 help="Bayes condition proability tables file"),
+    strax.Option('bayes_bins_config', default='/home/ahiguera-mx/test2/rap-ml-group/peak_classification/discrete_parameter_bins.npy',
+                 help="Bayes bins"),
+    strax.Option('s2_prob_threshold', default=-27,
+                 help="S2 prob value, above this value type=2"),
+)
+class PeakletClassification(strax.Plugin):
+    """Classify peaklets as unknown, S1, or S2."""
+    provides = 'peaklet_classification'
+    depends_on = ('peaklets',)
+    parallel = True
+    dtype = (strax.peak_interval_dtype
+             + [('type', np.int8, 'Classification of the peak(let)')]
+             + [('type_bayse', np.int8, 'Bayes peak classification type')]
+             + [('s1_prob', np.float32, 'S1 probability' )]
+             + [('s2_prob', np.float32, 'S2 probability' )]
+             )
+
+    __version__ = '4.0.0'
+
+    @staticmethod
+    def upper_rise_time_area_boundary(area, norm, const, tau):
+        """
+        Function which determines the upper boundary for the rise-time
+        for a given area.
+        """
+        return norm*np.exp(-area/tau) + const
+
+    @staticmethod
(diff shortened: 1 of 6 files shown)


WenzDaniel:
Hej, Aaron can we discuss a bit the purpose of the second classification plugin a bit? In the sense that we discuss how we want to compare it to our current classification approach? I think to fully test it within the entire processing framework we also have to include event-building later on etc. 

Hence I think your plugin should do the following:

- Provide the same data_type as the current classification plugin: `peaklet_classification`
- Replace `type` with the `type`-classification of your approach (you can add additional fields, but you have to overwrite `type` so we can check event-building)
- Remove all the code which belongs to the old classification from your plugin. Having this duplicated code is dangerous as changes may not be forwarded and the comparison can become faulty. 
- When comparing the two methods you can use two contexts one using the default one and one context using the Rice approach. 

Please let me know if you agree. We can also have a short meeting this week if you like.

ahiguera-mx:
Hi Daniel,

Thanks for looking into this very quickly! This still a rough draft and mostly to start the discussion, here is what I have in mind so hopefully this clarify things 
The idea that I have in mind is that `peaklet_classification` class return two types, one is going to be the current method and another called type_bayes, propagate this new dtypes to `peaks`, then at the event processing the user can decide whether wants to use the current type method or the new, I thought about this since doing event_processing is very quick to process.
In that sense, there is not duplication I just put the   `peaklet_classification` class on its own. And yes probably having a meeting to discuss this would be better

WenzDaniel:
But this is not what we want right? An analyst should not decide which algorithm to use. There should be just a single classification everyone is using. So we can save us here a lot of trouble and work by just having two plugins as long as the ML validation is going on. After that we register the ML-classification plugin by default.

ahiguera-mx:
The way that I see it is that we are adding more information. In my opinion having this flexibility is better, at the end when we define a given context we can set which peak classification method we want.

WenzDaniel:
But why would you keep a classification which is worse? I do not understand.

JoranAngevaare:
@ahiguera-mx thanks for the PR. I'm sorry but what you describe is best achieved by what Daniel is proposing.

What you are doing here will require a new reprocessing without the chance to first validate and compare.

straxen/plugins/peaklet_classification.py
```diff
@@ -0,0 +1,216 @@
...
+    def __init__(self, num_nodes=50, bins=None, CPT=None ):
+        self.num_nodes = num_nodes
+        self.classes = 2
+        self.class_prior = np.array([1. /self.classes for j in range(self.classes)])
+        self.bins = np.load(bins)
+        self.CPT = np.load(CPT)
+
+
+    def compute(self, peaklets):
+        bayse_ptype = np.zeros(len(peaklets), dtype=np.int8)
```
sophiafarrell, line 147: Can we correct this typo?

straxen/plugins/peaklet_processing.py
```diff
@@ -680,8 +588,17 @@ def setup(self):
                                                      self.config['gain_model'])
 
     def infer_dtype(self):
-        return strax.unpack_dtype(self.deps['peaklets'].dtype_for('peaklets'))
-
+        # wrong order
+        dtype = (strax.unpack_dtype(self.deps['peaklets'].dtype_for('peaklets'))
+                 +[('type_bayse', np.int8, 'Bayes peak classification type')]
```
sophiafarrell, line 593: Also can we correct this typo? Causing mailbox errors.

straxen/plugins/peaklet_classification.py
ahiguera-mx: actually we don't need this

straxen/plugins/peaklet_processing.py
ahiguera-mx: good catch, thanks should be fixed now

straxen/plugins/peaklet_processing.py
```diff
@@ -680,8 +588,17 @@ def setup(self):
                                                      self.config['gain_model'])
 
     def infer_dtype(self):
-        return strax.unpack_dtype(self.deps['peaklets'].dtype_for('peaklets'))
-
+        # wrong order
+        dtype = (strax.unpack_dtype(self.deps['peaklets'].dtype_for('peaklets'))
+                 +[('type_bayes', np.int8, 'Bayes peak classification type')]
+                 +[('s1_prob', np.float32, 'S1 probability' )]
```
sophiafarrell, line 594: need to add ln to doc, otherwise it throws mailbox error (same with line below)

sophiafarrell:
Hey Daniel, Joran, thanks for your fast feedback. 
* Regarding `'type'`, I think we do agree with your suggestion: we need to propagate this classifier through -- otherwise, it dies at the peaklet level.  For now, I've made some edits to Aaron's code such that this processes data for us, to do the proper checks. However, I would like to restructure things as you both suggest.
* Given that above point, and the timing of this PR, I hope it is okay for us to keep this PR draft up so we can continue to tune/implement this. There are some hacks in here to get us to test/debug on data, but we plan to make these more legitimate once we implement your type suggestion.

WenzDaniel:
Hej Sophia, sure no problem. If you have any questions or need some advice just let us know.

ahiguera-mx:
Hi @JoranAngevaare @WenzDaniel 
Do you guys have an example on how to overwrite a given dtype in straxen. For instance, my new plug-in see below
how can I overwrite `type` so I can check at the event level. My first guess is to use `strax.MergeOnlyPlugin` however that would only work if the `dtypes` are identical, notice that in my plug-in I want to add two extra dtypes `s1_prob` and `s2_prob`

```
provides = 'bayes_peaklet_classification' 
depends_on = ('peaklets',) 
parallel = True 
 __version__ = '0.0.1' 
dtype = (strax.peak_interval_dtype 
                 + [('type', np.int8, 'Bayes peak classification type')] 
                 + [('s1_prob', np.float32, 'S1 probability' )] 
                 + [('s2_prob', np.float32, 'S2 probability' )] 
                ) 
```

sophiafarrell:
Hey @JoranAngevaare @WenzDaniel! Re-tagging about the above question Aaron had. 

**More info:**
Our current idea is to just create this bayes classification plugin that calculates the Bayes classisfication and peak class probabilities. We'd then need to override the `type` of peaks, if some argument `use_bayes_classifier=True` (we'd default this to False, for now, until I can do more data-driven checks.) 
I'll link Aaron's preliminary plugin [here](https://github.com/XENONnT/straxen/compare/master...RiceAstroparticleLab:master), we greatly appreciate any review or feedback you can give in addition to info on overwriting `type`, as Aaron asked. Thx in advance!

WenzDaniel:
I am afraid, I am a bit lost here, sorry. Why do you need to overwrite the peak-data_type? You make a new plugin which also provides peaklet-classification with some additional fields. Lets say its name is `PeakltClassifcation2`. Is not it sufficient to only update event_basics accordingly? If you use your context and register `PeakletClassification` the additional fields in event_basics are just zeros. If you use `PeakletClassification2` it is filled with the corresponding values.  You can also add an additional event-level plugin which loads the corresponding information. Might be even a bit cleaner and easier to handle while comparing old and new stuff.

tunnell:
This is public so I encourage just a call just to be effective with everybody's time.  There were specific concerns raised from others that this aims to address, where we can discuss details on Slack.

WenzDaniel:
Sure I am usually available.

ahiguera-mx:
Thanks for all the useful conversation, we are closing this as a new and complete PR is coming.

PR #825, opened 2021-12-16
Rename tight coincidence 

WenzDaniel:
Rename tight_coincidence_channel into tight coincidence removes tight_coincidence_channel field from data.


JoranAngevaare (approved):
Can we remove the computation from get_tight coincidence as well?

WenzDaniel:
Sure I do not see why not.

PR #826, opened 2021-12-16
URLConfig take protocol for nested keys

jmosbacher:
Allows passing multiple keys to the `take://` protocol for nested access

Example config:

`take://json://{"a":[1,2,3]}?take=a&take=0`

the take protocol will get the two keys as a list and apply them in order to extract the nested value (1)


JoranAngevaare (approved):
Thanks Yossi, I pushed a few semantics and an extra test just for my own bookkeeping. It looks great!

PR #827, opened 2021-12-16
URLConfig preprocessor dispatch

jmosbacher:
This PR allows you to register preprocessors for existing protocols that will be called on the config value (with access to the argument and kwargs) __before__ the strax hashing of the value occurs and return overrides for the URL/argument values.
This allows you to make changes to the kwargs at plugin initialization, e.g. to replace a global version with a local version fetched from CMT.

This is mostly motivated by CMT global versions but can be useful for other things eg if you want to evaluate the values before it is hashed so the hash depends on the value and not the URL.

A preprocessor should accept a subset of the kwargs taken by the protocol and return a dictionary of overrides or None

A reference implementation for CMT global versions is already registered:

```python 
@URLConfig.register_preprocessor('cmt')
def replace_global_version(correction, version=''):
    if version.startswith('global'):
        local_versions = get_cmt_local_versions(version)
        v = local_versions.get(correction, version)
        return dict(version=v)

```

Alternatively, you can return a new URL entirely eg:
```python 
@URLConfig.register_preprocessor('cmt')
def replace_global_version(correction, version=''):
    if version.startswith('global'):
        local_versions = get_cmt_local_versions(version)
        v = local_versions.get(correction, version)
        return f'cmt://{correction}?version={v}'

```

then in the config you would set the url to eg:
```python
context_config = {
    'elife': 'cmt://elife?version=global_v6&run_id=plugin.run_id',
   ...
# or
context_config = {
    'global_version': 'global_v6',
    'elife': 'cmt://elife?version=config.global_version&run_id=plugin.run_id',
   ...

```

And the version value would be replaced on plugin initialization when the context sets the plugin config dict, before the lineage hash is computed. The lineage hash would then be determined by the local version.
The config passed to the plugin (and hashed) would be:
`cmt://elife?run_id=plugin.run_id&version=v4` (kwargs sorted alphabetically) 

An added benefit of adding a preprocessing stage on the URLs is that we can now sort all keywords so that equivalent URLs result in the same hash. 
Since the equality of two URLs may need to be checked by user code, I added a utility function `straxen.URLConfig.are_equal` which will return `True` iff the two URLs result in the same abstract syntax tree, ie will be evaluated identically at runtime.
 
Working Example:
```python
class ExamplePlugin(strax.Plugin):
    depends_on = ()
    dtype = strax.time_fields
    provides = ('test_data',)
    test_config = straxen.URLConfig(default=42,)
    
    def compute(self):
        pass

st = nt_test_context()
st.register(ExamplePlugin)

st.set_config({
            'test_config':
                'cmt://elife?run_id=plugin.run_id&version=global_v5',
    

})
p = st.get_single_plugin(nt_test_run_id, 'test_data')
p.config['test_config']
>> 'cmt://elife?run_id=plugin.run_id&version=v3'
```


jmosbacher:
@ershockley this is meant to be a more robust solution to the (very beautiful)  hack in #817

jmosbacher:
@JoranAngevaare @ershockley @WenzDaniel I think im pretty much done implementing what I had in mind. Any requests for specific tests, features etc?

JoranAngevaare:
> Just a reminder to anyone thinking of merging this: this PR will start sorting the URL parameters alphabetically before hashing. this will prevent the hashes from depending on the order of the parameters (as it should be) but will cause a one time change of hashes for any current URLs that are not already sorted alphabetically. So only merge when there will be some changes to these hashes anyways.

Alright, let me change it to a draft PR in that case. Is it ready for review though?

JoranAngevaare:
Also, do we really want:
> this will prevent the hashes from depending on the order of the parameters (as it should be)

If we chain several protocols, will the hashed config still be correct?

jmosbacher:
@JoranAngevaare Yes this PR has been ready for review for a long time now :)

> If we chain several protocols, will the hashed config still be correct?

Im not sure I understand the question, are you asking about the effect of the preprocessor on chained protocols or of the sorting of the URL parameters? 

For the former, nested protocols are recursively evaluated before being passed to the preprocessor, so the preprocessor has access to the same level in the nested hierarchy as the protocol it is acting on. The preprocessor can then override the down-stream chain or partially overrride kwargs or leave the URL as it is. Once all pre-processors are run, the final URL is constructed, with sorted kwargs and then hashed.
 
For the latter, the URL parameters are always hashed, regardless of which nested protocols use them, that wont change.

straxen/url_config.py
```diff
@@ -65,16 +75,21 @@ def __init__(self, cache=0, **kwargs):
...
-            cache_len = 100 if cache is True else int(cache) 
+            cache_len = 100 if cache is True else int(cache)
             cache = straxen.CacheDict(cache_len=cache_len)
             _CACHES[id(self)] = cache
 
     @property
     def cache(self):
         return _CACHES.get(id(self), {})
 
+    def __set_name__(self, owner, name):
```
JoranAngevaare, line 89: Could you add a docstring?

straxen/url_config.py
```diff
@@ -303,6 +540,20 @@ def format_arg(arg: str, **kwargs):
     return arg.format(**kwargs)
 
 
+@URLConfig.register_preprocessor('cmt')
+def replace_global_version(correction, version=''):
+    if isinstance(version, str) and version.startswith('global'):
+        local_versions = get_cmt_local_versions(version)
+        v = local_versions.get(correction, version)
+        return dict(version=v)
```
JoranAngevaare, line 554: You don't actually use this right? Apart from the tests? We would have to modify the `apply_cmt_version` function to modify global versions instead of local versions right?

JoranAngevaare (commented):
Thank Yossi, it looks like black magic but I can see it could do really cool stuff.

Generally, it took me some time to understand how it ties together honestly. The main methods/class could have a bit more elaborate docstring in that sense, see e.g. `dispatch protocol with argument and kwargs`. 

By the way, maybe it's not an issue, but do you have any feeling for what extra overhead is added by the pre processor to some simple operations like `st.key_for`?

The cmt pre-processor is currently not used right and would require another piece of logic in the apply_cmt_version?

I fixed some issues with the docs, see e.g. the older build:
https://straxen--827.org.readthedocs.build/en/827/reference/straxen.html#module-straxen.url_config

jmosbacher:
@JoranAngevaare thanks for taking the time to review this, I know its not the easiest code to review. 

> Generally, it took me some time to understand how it ties together honestly. The main methods/class could have a bit more elaborate docstring in that sense, see e.g. `dispatch protocol with argument and kwargs`.

Good point, ill add some more detailed docstrings.

> By the way, maybe it's not an issue, but do you have any feeling for what extra overhead is added by the pre processor to some simple operations like `st.key_for`?

Good question, this would depend entirely on the preprocessor being used. In any case if the plugins are cached, the preprocessor will only be called once for each URL.

> The cmt pre-processor is currently not used right and would require another piece of logic in the apply_cmt_version?

The main goal of having this functionality is so that you dont need the logic in apply_cmt_version, you can set the version to config.global_version and then you just set the global version in the context config and each URL does the lookup of the local version automatically on plugin initialization.
So after merging this we can remove all URLConfig specific logic in the apply_cmt_version function and just leave it as it used to be, only handling the old style cmt configs.

> I fixed some issues with the docs, see e.g. the older build: https://straxen--827.org.readthedocs.build/en/827/reference/straxen.html#module-straxen.url_config

Thanks!

straxen/url_config.py
jmosbacher: This is to replace the functionality of apply_cmt_version

straxen/url_config.py
```diff
@@ -151,6 +254,25 @@ def kwarg_from_url(cls, url: str, key: str):
...
+            return [URLConfig.evaluate(v, **namespace) for v in value]
+
+        if isinstance(value, str) and '.' in value:
+            name, _, key = value.partition('.')
+            if name in namespace:
+                obj = namespace[name]
+                if isinstance(obj, Mapping):
+                    value = obj.get(key, value)
+                else:
+                    value = getattr(obj, key, value)
```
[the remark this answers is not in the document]
jmosbacher: hmmm sure, why not :)

straxen/url_config.py
```diff
@@ -89,44 +104,132 @@ def wrapper(func):
...
+        '''
+
+        if arg is None:
+            protocol, arg, kwargs = cls.url_to_ast(protocol)
+            
+        if kwargs is None:
+            kwargs = {}
+
+        if protocol is None:
+            return arg
```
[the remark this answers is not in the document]
jmosbacher: protocol of an ast is None when there is a literal value

straxen/url_config.py
JoranAngevaare: but you need to specify somewhere which global_vx you are using right?

JoranAngevaare:
> Good question, this would depend entirely on the preprocessor being used. In any case if the plugins are cached, the preprocessor will only be called once for each URL.

okay, I was just thinking about the CMT preprocessor currently added. I don't immediately see other use-cases for this powerful tool, also because it increases the requirement of understanding of the protocols and their preprocessors. It's wonderful that you made it so general of course just so that my limited imagination can be proven wrong 😉

jmosbacher:
I know that making it so general makes it a bit more complex and harder to follow, but I also have to have some incentive to write the code... just implementing the exact logic needed without any generalizations or cool tricks would be so boring i would never get around to actually doing it...

straxen/url_config.py
jmosbacher: either is fine, in any case we can remove the logic in apply_cmt_version that evaluates the URL to get the cmt name.
If we go for option 1 then apply_cmt_version would just set the version keyword in all cmt URLs to the global version so it would be a one-liner

JoranAngevaare:
@jmosbacher should we close this in favor of xedocs?

PR #828, opened 2021-12-17
Unittest for DAQreader

JoranAngevaare:
 - This PR adds a test for the DAQReader. We should have added this _much_ earlier but now it's finally here.
 - Remove the `erase` option. It's one horribly bad idea to ever, ever do this.

Download some test data from [Nikhefs XAMS ](https://github.com/XAMS-nikhef/amstrax)experiment also using redax from https://github.com/XAMS-nikhef/amstrax_files

This data is ideal as we only have a few chunks and only a few channels

Please see `tests/test_daq_reader.py` 

## Time info
The entire test takes about 10 s.

## Coverage
afbeelding

### DIFF
--- a/tests/test_daq_reader.py
+++ b/tests/test_daq_reader.py
@@ -35,6 +35,16 @@ class TestDAQReader(unittest.TestCase):
     """
     Test DAQReader with a few chunks of amstrax data:
     https://github.com/XAMS-nikhef/amstrax
+
+
+    This class is structured with three parts:
+      - A. The test(s) where we execute some tests to make sure the
+      DAQ-reader works well;
+      - B. Setup and teardown logic which downloads/removes test data if
+      we run this test so that we get a fresh sample of data every time
+      we run this test;
+      - C. Some utility functions for part A and B (like setting the
+      context etc).
     """
 
     run_id = '999999'
@@ -43,8 +53,7 @@ class TestDAQReader(unittest.TestCase):
     rundoc_file = 'https://raw.githubusercontent.com/XAMS-nikhef/amstrax_files/73681f112d748f6cd0e95045970dd29c44e983b0/data/rundoc_999999.json'  # noqa
     data_file = 'https://raw.githubusercontent.com/XAMS-nikhef/amstrax_files/73681f112d748f6cd0e95045970dd29c44e983b0/data/999999.tar'  # noqa
 
-    # Tests
-    # -----
+    # # Part A. the actual tests
     def test_make(self) -> None:
         """
         Test if we can run the daq-reader without chrashing and if we
@@ -87,8 +96,7 @@ class TestDAQReader(unittest.TestCase):
         with self.assertRaises(ValueError):
             st.make(self.run_id, 'raw_records')
 
-    # Setup of tests like getting data and removing it after the test
-    # ---------------------------------------------------------------
+    # # Part B. data-download and cleanup
     @classmethod
     def setUpClass(cls) -> None:
         st = strax.Context()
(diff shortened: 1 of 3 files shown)


tests/test_daq_reader.py
```diff
@@ -0,0 +1,161 @@
...
+                    nveto=(1, 2),
+                    aqmon=(ARTIFICIAL_DEADTIME_CHANNEL, ARTIFICIAL_DEADTIME_CHANNEL + 1),
+                )})
+        update_config = {
+            'readout_threads': daq_config['processing_threads'],
+            'record_length': daq_config['strax_fragment_payload_bytes'] // 2,
+            'max_digitizer_sampling_time': 10,
+            'run_start_time': run_doc['start'].replace(tzinfo=timezone.utc).timestamp(),
+            'daq_chunk_duration': int(daq_config['strax_chunk_length'] * 1e9),
+            'daq_overlap_chunk_duration': int(daq_config['strax_chunk_overlap'] * 1e9),
```
darrylmasson, line 156: For most of these fields we directly reference `daq_config[field]` with one exception where we do `daq_config.get(field)`. Is there a guarantee that the test rundoc (and any future replacements) includes all the referenced fields, or do we want to protect ourselves and use `daq_config.get(field, redax_default)`?

darrylmasson (approved):

tests/test_daq_reader.py
```diff
@@ -0,0 +1,161 @@
...
+        st.set_config({'safe_break_in_pulses': int(3600e9),
+                       'dummy_version': 'test_invalid_setting',
+                       })
+        with self.assertRaises(ValueError):
+            st.make(self.run_id, 'raw_records')
+
+    # Setup of tests like getting data and removing it after the test
+    # ---------------------------------------------------------------
+    @classmethod
+    def setUpClass(cls) -> None:
```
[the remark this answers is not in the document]
JoranAngevaare: Absolutely, however, this is a Unittest convention which I have no power of changing

tests/test_daq_reader.py
JoranAngevaare: Right, this doc shouldn't change, especially as we are referencing a permalink, even if we update the github doc, we would first have to change the permalink first before it could hurt the test.

So generally you are right, but since we would be manually changing the URL to the doc, we'd also know if we have to change some of the fields so I suggest to leave it like this

JoranAngevaare:
Thank you Darryl for the review 👍

tests/test_daq_reader.py
```diff
@@ -0,0 +1,161 @@
...
+        """The safe break in pulses cannot be longer than the chunk size"""
+        st = self.st.new_context()
+        st.set_config({'safe_break_in_pulses': int(3600e9),
+                       'dummy_version': 'test_invalid_setting',
+                       })
+        with self.assertRaises(ValueError):
+            st.make(self.run_id, 'raw_records')
+
+    # Setup of tests like getting data and removing it after the test
+    # ---------------------------------------------------------------
```
WenzDaniel, line 91: Hmm, I think this is not needed.

WenzDaniel (approved):
Looks very nice. Neat usage of the class and method SetUp implementation.

tests/test_daq_reader.py
JoranAngevaare: sure can delete, just wanted to illustrate that there are three separate stages to this test:
 - testing
 - setups
 - tools

JoranAngevaare: 👍 I've removed this over-verbosity to the docstring

Issue #835, opened 2021-12-21
filter_kwargs needs some tests


WenzDaniel:
A tiny test will do

JoranAngevaare:
See #837

PR #837, opened 2021-12-21
Add test  for filter_kwargs

jmosbacher:
Add a test for the URLConfig.filter_kwargs function
The function filters a dictionary of arguments to a function based on the functions call signature so that the function is only called with kwargs it can accept. 
This test verifies that the the missing kwargs from the call signature are indeed filtered and that no kwargs are filtered if the function accepts wildcard `**kwargs`

Addreses issue #835


JoranAngevaare (approved):
Looks good, thanks!

WenzDaniel (approved):
Thanks a lot that was fast.

PR #855, opened 2021-12-22
Tests for WFSim contexts

JoranAngevaare:
Add a test for WFSim context that at least the basic code runs.

Additionally global_v5 is not compatible with this straxen version at the time of writing, maybe we should change it to ONLINE instead (as I did in this PR)

Coverage
afbeelding
to
afbeelding


ramirezdiego (approved):
Thanks!

Issue #834, opened 2021-12-21
apply_cmt_version must be tested more throughly.


WenzDaniel:
Currently we are testing `apply_cmt_version` only very superficially when initializing the online context in straxen. However, I think this function is so fundamental in ensuring that we are applying the correct corrections that I think it must be tested more carefully either in straxen or cutax. We should at least check if all CMT options are updated correctly. A mistake here would not be directly obvious in data.

WenzDaniel:
In the same PR we should also add tests for `correction_name_from_urlstring` and `replace_url_version`.

WenzDaniel:
and extend the test for `is_cmt_option` to accommodate the URL config change.

ershockley:
In #860 I add a few of the tests you mention: 

- I remove `correction_name_from_urlstring` 
- I add a test for `replace_url_version`
- I add a case for `is_cmt_option` 

The big one is definitely `apply_cmt_version` though, I agree. I think we need to put this into cutax though. That is where we actually define the contexts with specific versions, so it is easier.

PR #860, opened 2021-12-22
Add cmt tests and fix bug in apply_cmt_version

ershockley:
- Adds more testing to address #834
- Fixes a bug where URL configs were not evaluated correctly when determining the correction name in `apply_cmt_version`

I removed the function `correction_name_from_urlstring` and implemented the same logic into `get_cmt_options`, which was @jmosbacher's initial suggestion in #817(sorry! should have listened then). This allows us to evaluate e.g. `plugin.attribute` when determining the correction name, which is a change I implemented later on in #817 but clearly did not test thoroughly enough. 

This changes slightly what `get_cmt_options` returns, and thus could be a breaking change. I think I have addressed the change in the 2 places where `get_cmt_options` is called in straxen. Now it returns a dictionary that looks something like this: 

```
{
...,
 'electron_drift_time_gate': {'correction': 'electron_drift_time_gate',
  'version': 'ONLINE',
  'strax_option': ('electron_drift_time_gate', 'ONLINE', True)},
'elife': {'correction': 'elife',
  'version': 'ONLINE',
  'strax_option': 'cmt://elife?version=ONLINE&run_id=plugin.run_id'},
...
}
```

where the `'correction'` and `'version'` fields of the dict are the CMT name and version of the correction, respectively, and `'strax_option'` is what actually gets passed to strax. The aim of this is to try to help deal with issues we are seeing with mixture of URL configs and CMT tuple configs.

### DIFF
--- a/straxen/get_corrections.py
+++ b/straxen/get_corrections.py
@@ -1,7 +1,7 @@
 import numpy as np
 import strax
 import straxen
-from warnings import warn
+import typing as ty
 from functools import wraps
 from straxen.corrections_services import corrections_w_file
 from straxen.corrections_services import single_value_corrections
@@ -153,7 +153,7 @@ def _is_cmt_option(run_id, config):
     return is_cmt
 
 
-def get_cmt_options(context):
+def get_cmt_options(context: strax.Context) -> ty.Dict[str, ty.Dict[str, tuple]]:
     """
     Function which loops over all plugin configs and returns dictionary
     with option name as key and a nested dict of CMT correction name and strax option as values.
@@ -166,6 +166,10 @@ def get_cmt_options(context):
 
     for data_type, plugin in context._plugin_class_registry.items():
         for option_key, option in plugin.takes_config.items():
+            if option_key in cmt_options:
+                # let's not do work twice if needed by > 1 plugin
+                continue
+
             if (option_key in context.config and
                     is_cmt_option(context.config[option_key])):
                 opt = context.config[option_key]
--- a/straxen/get_corrections.py
+++ b/straxen/get_corrections.py
@@ -156,20 +156,46 @@ def _is_cmt_option(run_id, config):
 def get_cmt_options(context):
     """
     Function which loops over all plugin configs and returns dictionary
-    with option name as key and current settings as values.
+    with option name as key and a nested dict of CMT correction name and version as values.
 
     :param context: Context with registered plugins.
     """
+
     cmt_options = {}
+    runid_test_str = 'norunids!'
+
     for data_type, plugin in context._plugin_class_registry.items():
         for option_key, option in plugin.takes_config.items():
             if (option_key in context.config and
-                is_cmt_option(context.config[option_key])
-                ):
-                cmt_options[option_key] = context.config[option_key]
-            elif is_cmt_option(option.default):
-                cmt_options[option_key] = option.default
+                    is_cmt_option(context.config[option_key])
+            ):
+                opt = context.config[option_key]
 
+            elif is_cmt_option(option.default):
+                opt = option.default
+
+            else:
+                continue
+
+            # check if it's a URLConfig
+            if isinstance(opt, str) and 'cmt://' in opt:
+                before_cmt, cmt, after_cmt = opt.partition('cmt://')
+                p = context.get_single_plugin(runid_test_str, data_type)
+                url_config = p.takes_config[option_key]
+                url, url_kwargs = url_config.split_url_kwargs(after_cmt)
+                kwargs = {k: url_config.fetch_attribute(p, v) for k, v in url_kwargs.items()}
+                correction_name = url_config.dispatch(url, **kwargs)
+                version = kwargs.get('version', 'ONLINE')
+                cmt_options[option_key] = {'correction': correction_name,
+                                           'version': version,
+                                           'strax_option': opt
+                                           }
+
+            else:
+                cmt_options[option_key] = {'correction': opt[0],
+                                           'version': opt[1],
+                                           'strax_option':  opt
+                                           }
     return cmt_options
 
 
--- a/straxen/get_corrections.py
+++ b/straxen/get_corrections.py
@@ -178,19 +178,23 @@ def get_cmt_options(context):
             if isinstance(opt, str) and 'cmt://' in opt:
                 before_cmt, cmt, after_cmt = opt.partition('cmt://')
                 p = context.get_single_plugin(runid_test_str, data_type)
-                url_config = p.takes_config[option_key]
-                url, url_kwargs = url_config.split_url_kwargs(after_cmt)
-                kwargs = {k: url_config.fetch_attribute(p, v) for k, v in url_kwargs.items()}
-                correction_name = url_config.dispatch(url, **kwargs)
-                version = kwargs.get('version', 'ONLINE')
+                p.config[option_key] = after_cmt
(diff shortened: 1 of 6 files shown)


straxen/get_corrections.py
```diff
@@ -156,20 +156,43 @@ def _is_cmt_option(run_id, config):
...
+                cmt_options[option_key] = {'correction': correction_name,
+                                           'version': version,
+                                           'strax_option': opt,
+                                           }
 
+            else:
+                cmt_options[option_key] = {'correction': opt[0],
+                                           'version': opt[1],
+                                           'strax_option': opt,
+                                           }
```
jmosbacher, line 195: Just a matter of personal taste but i would probably have return a tuple of length 3 here that's compatible with the cmt structure. But I'm fine with an explicit dict too (obviously if i was building the format from scratch i would have done it the way you did, but its already a tuple so not sure its worth the risk of changing)

jmosbacher (commented):
Thanks Evan, looks good. I just added two suggestions

JoranAngevaare (changes requested):
Thanks Evan, I agree with Yossis review so I'll not repeat things already pointed out 😉 

However, this PR doesn't cover `apply_cmt_version` (L333-L395):

```
(base) joran@DESKTOP-F4PI41P:~/google_drive/ubuntu-windows/software/straxen$ coverage run --source=sen $(which pytest) tests/test_cmt.py
========================================= test session starts ==========================================
platform linux -- Python 3.9.5, pytest-6.2.5, py-1.10.0, pluggy-1.0.0
rootdir: /mnt/c/google_drive/PhD-master/ubuntu-storage/ubuntu-windows/software/straxen
plugins: anyio-3.3.4, hypothesis-6.21.1, cov-2.12.1
collected 7 items

tests/test_cmt.py .......                                                                        [100%]

========================================== 7 passed in 12.15s ==========================================
(base) joran@DESKTOP-F4PI41P:~/google_drive/ubuntu-windows/software/straxen$ coverage report -m straxen/corrections_services.py
Name                              Stmts   Miss  Cover   Missing
---------------------------------------------------------------
straxen/corrections_services.py     184     83    55%   79, 82, 94, 104, 128-142, 
                                                        146, 152, 155-158, 193-194, 197, 
                                                        201-206, 215-216, 226, 240, 246, 
                                                        265, 273-298, 302, 308-311, 314, 
                                                        323-324, *338-395**
---------------------------------------------------------------
TOTAL                               184     83    55%
```
In `**` you see that these lines aren't tested.

I think at least one offline version should be compatible with the code and we should have a test for that.

Otherwise we are again going to make releases with no compatibility in CMT. If a PR is not compatible with any CMT global version I don't see any reason to merge it to straxens master as it delays using other new PRs with an offline version.

straxen/get_corrections.py
ershockley: I did that first but then that changes the lineage hash. For example running this function with `global_ONLINE` will overwrite the default options and make them tuples. I don't think this is what we want right?

ershockley:
@JoranAngevaare i see what you're saying but how do we choose which offline version to test? When offline versions become incompatible with straxen it means we need to update CMT, not straxen. I think this test should happen in cutax where we actually define the offline contexts.

JoranAngevaare:
> @JoranAngevaare i see what you're saying but how do we choose which offline version to test? When offline versions become incompatible with straxen it means we need to update CMT, not straxen. I think this test should happen in cutax where we actually define the offline contexts.

Yeah I also see this.

It should be compatible with at least one global version (I don't really care which one). I agree that a stricter test can be run on cutax (keep in mind the limited testing time due to the though).

How about a test function that tests for a given global version if it works? Here in straxen, we could just loop over `set(CorrectionsManagementServices().global_versions[::-1])-{'global_ONLINE'}` and only fail if all of the global versions would raise an `CMTVersionError`? 

In cutax we can run the test for a fixed global version?

ershockley:
Ok I re-added a test we previously removed with the 'latest' offline context. This is the one most likely to be compatible with master straxen. This test will fail if we add new corrections again, unless we agree to modify past offline versions like we discussed on slack. 

Let's try to nail down this last point right after the new year. 

Do you think this works for now?

straxen/get_corrections.py
jmosbacher: I dont follow, how would the structure of what you return here affect the lineage? its just the inputs to your `apply_cmt_version` function right? so why would the structure of the data you pass between the two functions have any affect on anything? its just the structure, a tuple vs a dict, the data is the same

jmosbacher: to be clear, im not talking about replacing the value of the config with a cmt compatible tuple. Im just talking about keeping the return value of this specific function consistent with what it was in the past. so e.g. `(cmt_name, cmt_version/url, True)`
Not super critical but would be backward compatible.

ershockley: Ah right we had this discussion before. It bothers me to have it return something like `(cmt_name, url, True)` for URLConfigs as this is basically nonsensical. We then would need to add another `if` statement to check what kind of config it is when we actualy do the `st.set_config`. I feel like either way we are breaking something, might as well make it something that can be more universal. 

If you insist though we can change. I was wrong before... 😉

straxen/get_corrections.py
```diff
@@ -156,20 +156,43 @@ def _is_cmt_option(run_id, config):
...
             elif is_cmt_option(option.default):
-                cmt_options[option_key] = option.default
+                opt = option.default
+            else:
+                continue
+
+            # check if it's a URLConfig
+            if isinstance(opt, str) and 'cmt://' in opt:
+                before_cmt, cmt, after_cmt = opt.partition('cmt://')
+                p = context.get_single_plugin(runid_test_str, data_type)
```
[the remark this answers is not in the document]
ershockley: Thanks @jmosbacher! I have to admit I still don't fully understand the magic of these URLConfigs -- this is neat. It is implemented now.

straxen/get_corrections.py
jmosbacher: No no, I dont insist at all on the tuple. I think the chances of some current script somewhere using this function is negligible so probably wont break anything. But i would definitely like lean heavily towards not fetching the kwargs for some arbitrary run_id. If you want to return the kwargs then I would probably have this function accept run_id as an input so if anyone else decides to use this function at least its clear that the return value is run_id dependent.

ershockley: Thanks @jmosbacher. I dropped the kwargs part so now it only returns a dict of correction name and the strax option. I still prefer to return the dict -- it should be more future safe as we (hopefully) get rid of these tuple configs

jmosbacher (approved):

JoranAngevaare (approved):
Looks good as a patch for this straxen version, thanks. 

I think it looks good for now, but as some of your comments are also hinting at, it would be nice to work with `cmt2` and/or get rid of the tuples.

When we add those, I hope we can get rid of some of this scaffolding indeed, especially the posrec naming confusion (it became a lot more complex then when I came up with the idea and it is indeed more complex and cumbersome than I would have anticipated).

I just checked, but I couldn't find any place where the CMT tuple would be required by `get_cmt_options`. Doesn't of course mean there is none 😉. 

Finally, maybe we can get the `cmt://` separator from the URLConfig class rather than hardcoding it on several places. I hope it's not a think we ever want to change but with the URLConfigs I'd guess this should be possible.

I'll merge it now since it's not worth stalling this PR for.

Issue #864, opened 2021-12-27
`correction_options`  doesn't work for `URLConfigs`


JoranAngevaare:
`correction_options` is not working for `URLConfigs` because it assumes the tuple config type and otherwise doesn't really understand it's a CMT option

As there are no adequate tests for it, these kind of bugs aren't solved before releases

jmosbacher:
hmm, i thought Evans PR was supposed to fix this? Ok, in any case a more long term solution is already prepared in PR #827. I just need to add some more testing and it will be ready for review.

JoranAngevaare:
See #865

PR #865, opened 2021-12-27
Another WFSim `<->` CMT patch 

JoranAngevaare:
Some patches for wfsim don't work well with URL configs #864 and in combination with #860, we now have some other issues in the WFSim context.

It's again another patch, and I think we should really go away from the entire `cmt_run_id`-hack from used for WFSim and just track the corrections used during processing entirely separate from the instructions as discussed during the Heidelberg workshop (@terliuk @jmosbacher @WenzDaniel @ramirezdiego ).

For a failing example please see: [this fail](https://github.com/XENONnT/pema/runs/4644533396?check_suite_focus=true) from PR https://github.com/XENONnT/pema/pull/141

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.2.0
+current_version = 1.2.1
 files = setup.py straxen/__init__.py
 commit = True
 tag = True
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.2.1
+current_version = 1.2.3
 files = setup.py straxen/__init__.py
 commit = True
 tag = True
--- a/.github/scripts/update-context-collection.py
+++ b/.github/scripts/update-context-collection.py
@@ -1,0 +1,42 @@
+import importlib
+import strax
+import straxen
+from utilix import DB
+import datetime
+
+db = DB()
+
+# list of contexts that gets tracked in runDB context collection
+# needs to be maintained for each straxen release
+context_list = ['xenonnt_led',
+                'xenonnt_online',
+                ]
+
+
+# returns the list of dtype, hashes for a given strax context
+def get_hashes(st):
+    return set([(d, st.key_for('0', d).lineage_hash)
+                for p in st._plugin_class_registry.values()
+                for d in p.provides])
+
+
+def main():
+    for context in context_list:
+        context_module = importlib.import_module('straxen.contexts')
+        st = getattr(context_module, context)()
+        hashes = get_hashes(st)
+        hash_dict = {dtype: h for dtype, h in hashes}
+
+        doc = dict(name=context,
+                   date_added=datetime.datetime.utcnow(),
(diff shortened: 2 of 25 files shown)


jmosbacher:
So is `electron_lifetime_liquid` just not used? When we switched `elife` to a URLConfig we didnt realize it was still being used in the wfsim as a regular cmt config right? so switching the electron_lifetime_liquid to a URLConfig in the wfsim plugin i guess would be the cleanest but i cant find where its actually used? so i guess its used inside wfsim and not as a plugin config?

JoranAngevaare:
@jmosbacher no, the issue is that that config shouldn't need to use a URL config but just fetch a number.

The entire `correction_options` wrapper doesn't work anymore as far as I can tell, because with an URL config we should have replaced `?run_id=plugin.run_id` with `?run_id=config.cmt_run_id` or something. But as I said in the PR description as well as in the Heidelberg workshop, I think we should really drop this. The fax config can be based on numbers from CMT sure, but they should not be tracked with a lenght-5 tuple.

jmosbacher:
Ok, yeah that makes sense. Im wondering if there is a way to make the transition period a bit easier. Maybe I can add support for basic URLs in the `straxen.get_correction_from_cmt` function. Of course I cant support complex URLs that depend on arbitrary plugin attributes, but the simple ones like `cmt://name?version=vx` should be trivial to support for this transition period

jmosbacher:
@terliuk what do you mean by divergent? do you mean setting it to a constant value? since its a URLConfig, it can just be set to any value you want directly in the context config. If you mean taking it from cmt but from some other run_id then you need to set it to `cmt://elife?version=VERSION&run_id=ALT_RUN_ID`. Since the plugin itself is no longer calling the `get_correction_from_cmt` function there is no need to use these hacks with the length 5 tuples.

jmosbacher:
@JoranAngevaare I think its important to keep the promise of not breaking any working hacks ugly as they may be during this transition period. Im considering adding some extra logic here to just edit the URLs automatically for the "processing" part (the elife URLConfig) and constructing tuples for the "simulation" side which match what the hacks already expect. Should I make a separate PR for this or push directly to here?

jmosbacher:
I added some changes to the context creation script that will hopefully handle all use cases during the transition period, the goal is to make the transition as transparent as possible to users.

These are the cases I could find reading the existing code and the extra logic I added for each:

- WFSim using CMT configs from the processing context - replace the URL with a tuple that WFSim expects
- Overriding processing CMT configs with constants - just set the config to the constant
- Overriding run_id in CMT URL configs for processing  - replace run_id value in the URL with a constant instead of `plugin.run_id`
- Overriding simulation CMT configs with constants - replace the URL with a tuple that WFSim expects inside the `fax_config_override_from_cmt`

Anything I missed?

JoranAngevaare (commented):
Thanks @jmosbacher . Seems to work for me too 👍 

I'm with Diego that I hope final patch since we hopefully soon move to thinking about the options for RawRecordsFromWFsim and the options for straxen separately, making it all much more transparent as we discussed at the workshop.

PR #866, opened 2021-12-27
(rucio) storage reorganization

JoranAngevaare:
- **Split the remote and local rucio frontends**. The Local and Remote backends did not share any common logic, as such, there is no need to combine them. Additionally, the Local Backend is only useful when doing a fuzzy operation or when the runsdatabase is down/out of sync. As such it doesn't have to be included by default.
- **Add the remoteness attribute** from https://github.com/AxFoundation/strax/pull/617 to get better loading times
- **kwargs handling in contexts**
- **move the storage frontends** and config storage into a subfolder (do the same for tests).
- **Add tests for the local rucio frontend**
- **Add test that only works in containers for RucioRemote**
- Small refactor on RunDB against complex code

## Why split the remote and local frontend?
There is no logic shared between the two, resulting in many `if`s for determining which of the two can do what. With the remoteness attribute added in https://github.com/AxFoundation/strax/pull/617, there is no need anymore to ensure that the `RucioRemoteFrontend` comes last, it's ensured by this attribute now.

##  kwargs handling in contexts
Some things were underscored whereas they don't really have to be and make it unclear what is allowed for users to touch and what is not


straxen/test_utils.py
```diff
@@ -107,14 +107,31 @@ def _is_on_pytest():
...
 def nt_test_context(target_context='xenonnt_online',
                     deregister=('peak_veto_tags', 'events_tagged'),
                     **kwargs):
+    if not straxen.utilix_is_configured(warning_message=False):
+        kwargs.setdefault('_database_init', False)
+
     st = getattr(straxen.contexts, target_context)(**kwargs)
     st.set_config({'diagnose_sorting': True})
-    st._plugin_class_registry['raw_records'].__version__ = "MOCKTESTDATA"  # noqa
+    st.register(_get_fake_daq_reader())
```
JoranAngevaare, line 128: I noticed that `st._plugin_class_registry['raw_records'].__version__ = "MOCKTESTDATA"  # noqa` caused lineage changes in some later tests too. This way, we don't change the lineage of the real DAQReader

straxen/contexts.py
```diff
@@ -111,58 +113,89 @@ def xenonnt(cmt_version='global_ONLINE', **kwargs):
...
+                   download_heavy: bool = False,
+                   _rucio_path: str = '/dali/lgrandi/rucio/',
+                   _rucio_local_path: ty.Optional[str] = None,
+                   _raw_path: ty.Optional[str] = '/dali/lgrandi/xenonnt/raw',
+                   _processed_path: ty.Optional[str] = '/dali/lgrandi/xenonnt/processed',
+
+                   # Testing options
+                   _context_config_overwrite: ty.Optional[dict] = None,
+                   _database_init: bool = True,
+                   _forbid_creation_of: ty.Optional[dict] = None,
```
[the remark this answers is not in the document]
jmosbacher: Maybe use 
```python
from pandas.util._decorators import deprecate_kwarg

@deprecate_kwarg('_minimum_run_number', 'minimum_run_number')
@deprecate_kwarg('_add_online_monitor_frontend', 'include_online_monitor')
...
def xenonnt_online(output_folder: str = './strax_data',
    ...
```
Not super important but would clean things up a bit...

straxen/storage/rundb.py
```diff
@@ -235,18 +239,27 @@ def find_several(self, keys: typing.List[strax.DataKey], **kwargs):
...
+        results_dict = self._parse_documents(
+            self.collection.find(
+                {**run_query, **dq},
+                projection=projection),
+        )
 
+        return [results_dict.get(k.run_id, False)
+                for k in keys]
+
+    def _parse_documents(self, find: iter):
```
jmosbacher, line 260: I would rename find --> docs/documents but that's just personal taste.

jmosbacher (commented):
Looks good, do I understand correctly that this is just splitting up code that already has been shown to work right? or do you need a detailed review on the frontend/backend logic itself?

JoranAngevaare:
> Looks good, do I understand correctly that this is just splitting up code that already has been shown to work right? or do you need a detailed review on the frontend/backend logic itself?

Yeah, I purposely tried to not change the main logic of the back/frontend because that would be a nightmare to review. I did a few changes/tweaks as I listed in the PR description, which are mostly complementary to the existing logic.

straxen/contexts.py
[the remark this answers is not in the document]
JoranAngevaare: That's super slick, thanks!

jmosbacher (approved):

straxen/storage/rundb.py
JoranAngevaare: Thanks, as a reviewer, you are certainly entitled to expressing your personal tastes 👍 I think you are right.

PR #873, opened 2022-01-06
Bump peaklets version

ershockley:
Bump the peaklets version. The hash for peaklets did not change between v1.1.3 and v1.2.1, but the dtype did, leading to errors when loading previously-processed data. This should fix those errors by forcing a new hash.


jmosbacher (approved):
I think this is fine but probably best to get an OK from @WenzDaniel its his changes

JoranAngevaare (approved):

JoranAngevaare:
Thanks Evan, this is really an oversight on our side. I'll make a new release ASAP.

PR #877, opened 2022-01-10
Tests for scada interface

WenzDaniel:
Adds some very simple tests to some small functions and raise-exceptions in the scada interface.


JoranAngevaare (approved):
Thanks!

PR #887, opened 2022-01-24
Naive Bayes classification 

ahiguera-mx:

Naive Bayes classification 
Add a new class to do peak classification, this requires[ PR 637](https://github.com/AxFoundation/strax/pull/637) 
Using the peaklets waveforms to inference peaklets type, currently we override the current classification method with the Bayes. At the event level user have access to posteriors probabilities. 

### _Notes on testing_


ahiguera-mx:
Currently the test are failing because the URL configs points to a local file on dali, any suggestion here?

jmosbacher:
> Currently the test are failing because the URL configs points to a local file on dali, any suggestion here?

The default value for configs should not be set to values that only work on a specific site.
If you want the defaults to be those specific files, you need to add the files to [private_nt_aux_files](https://github.com/XENONnT/private_nt_aux_files) and change the path the the file name so the resource:// protocol can find the file in the mongo database. 
Alternatively, you can set the defaults to dummy values (arrays) and set the config to the local paths in dali when you run on dali. But if these files are meant to be used in most processing contexts then i would suggest using the former solution.

WenzDaniel:
I converted it to a draft for the moment, until things are verified. Just ping me if we have to discuss things.

straxen/plugins/peaklet_processing.py
```diff
@@ -717,6 +723,8 @@ def compute(self, peaklets, lone_hits):
                 max_buffer=int(self.config['s2_merge_max_duration']//np.gcd.reduce(peaklets['dt'])),
             )
             merged_s2s['type'] = 2
+            merged_s2s['s2_prob'] = 0 # remember this is ln prob so 0 means 100%
```
jmosbacher, line 726: Any reason you keep the ln of the probability instead of converting it to actual probability? Obviously for computing it with the classifier you prefer to use the ln which is fine but why keep it as ln when you are naming everything without the ln? this can produce unnecessary confusion so do you gain anything later down in the chain from not taking the exponent after its calculated?

jmosbacher (commented):
Nice work. Just two small questions:
1) Do you gain anything by editing the MergedS2s plugin? why not just add a small plugin that adds the classification to the merged_s2s so you keep backward compatibility? strax is built for this kind of thing, its the reason there is a data_kind and data_type, where you can have plugins that just add new columns to existing data_kinds. The extra columns are merged automatically before being passed to any plugin that depends on them.
2) Is there any benefit to saving the ln of the probability instead of the probability itself? this can cause some confusion. I would probably either take the exponent of the calculated value or change the naming to include ln so it doesn't confuse people.

straxen/plugins/peaklet_processing.py
ahiguera-mx: We hose to return these ln probabilities rather than the probabilities because of floating point errors / computer rounding.

jmosbacher: A double precision floating point is accurate to ~15 decimal points. Naive bayes uses the product of the empiric probability of features ie the relative frequency in the labeled dataset. Are you telling me that you expect to have a labeled dataset with 10^15 waveforms to reach the level of confidence in your probability estimate that you are worried about the 15th decimal being rounded?

ahiguera-mx:
See this [PR](https://github.com/XENONnT/straxen/pull/924)

PR #889, opened 2022-01-27
Fix leading zeros error

jmosbacher:

Passing a literal URL parameter e.g. run_id with leading zeros e.g. 'cmt://elife?version=v1&run_id=026000' can raise a syntax error when its parsed by URLConfig. Currently you would have to drop the leading zeros or explicitly mark the value as a string using quotes. The PR changes the parsing behavior so that zero-padded numbers will be parsed as strings.

Ignores SyntaxError exceptions when parsing config values using pythons builtin literal_eval


WenzDaniel:
Any idea why this leads to a coverage decrease of 11%?

JoranAngevaare:
> Any idea why this leads to a coverage decrease of 11%?

Yeah it's because this PR is from a remote fork

jmosbacher:
Oh right, I forgot about the request to use origin branches instead of forked ones...

JoranAngevaare (commented):
maybe good to add a few comments to `parse_val`. I'm not entirely sure what will happen with your example here, maybe we can add a test for it? Because dropping the leading 0 silently from a run_id is an easy way to get confusing errors when loading data etc.

jmosbacher:
@JoranAngevaare good point, i should add a test. 
Just FYI, this PR is meant to avoid needing to drop the leading zeros. Currently, if there is a leading zero in one of the URL params, you will get a SyntaxError from python literal_eval to avoid this error you would have to pass the run_id in the URL without the leading zero. This PR ignores this SyntaxError so the e.g run_id will be parsed as a string and include the leading zeros.

JoranAngevaare (approved):
Nice!

PR #890, opened 2022-01-27
Change alt_s2 identification criterion

shenyangshi:
Straxen identifies alt_s2 as the second-largest s2, but this criterion decreases the s2 single scatter cut efficiency by a few percent. S2 before the S1 or after the max-drift-time, if large enough by accident, can become an alt_s2 and be rejected by s2 single scatter cut, e.g. [(waveform).](https://xe1t-wiki.lngs.infn.it/doku.php?id=shenyang:s2_single_scatter_sr0:waveform5), [(waveform)](https://xe1t-wiki.lngs.infn.it/doku.php?id=shenyang:s2_single_scatter_sr0:waveform4) This code aims to select alt_s2 from s2 in (s1_time. s1_time+max_drift_time), making the candidate alt_S2 as physical as possible/

1.  For s2, get_largest_sx_peaks returns all available s2 rather than only 2.
2.  Apply selection mask in set_alt_s2_properties, and chop it back to size 2.

The example event becomes [(waveform),](https://xe1t-wiki.lngs.infn.it/doku.php?id=shenyang:s2_single_scatter_sr0:waveform5_2) after the change.


JoranAngevaare:
@shenyangshi thanks for the PR. I think that in the context of the s2 single scatter cut it makes sense what you are proposing. 

I am wondering though if this redefinition doesn't hurt other analyses, like kr selections. Here the alt S2 is related to the alt S1 albeit with only a small expected delay of course so it might be fine. Have you looked if this affects other analyses/cuts?

shenyangshi:
Thanks for the comments @JoranAngevaare! Yes, I considered events including dual interactions, like BiPo events, so I extended the boundary to 1.1*max_drift_time. An extra ~200mus should suffice for Kr double S1S2, according to [Naxime's note](https://xe1t-wiki.lngs.infn.it/lib/exe/detail.php?id=xenon%3Axenonnt%3Ampierre%3Akr83m_events_selection_xenonnt_commissioning_phase&media=xenon:xenonnt:mpierre:delay_time_s1_preli_selection.png) the interval is below 3 mus. I don't know about other analyses using alt_S2 information though, but if the analysis involves immediate multiple S1 or multiple S2 following the same S1, this change shouldn't affect them.

JoranAngevaare:
I see, I just checked using https://github.com/AxFoundation/strax/pull/638 where `*alt_s2*` or `*s2_b*`is used, but I think there are no other places:
https://gist.github.com/JoranAngevaare/859c924967db2067fad067e8e2f1d8d1

I'll checkout the PR in a bit more detail 👍

WenzDaniel:
I am wondering if we do not want to cut also events in which we have an alt. S2 right before an S1. It would not be a multi-scatter event but anyhow for data-quality I think we should cut those events.

shenyangshi:
> I see, I just checked using [AxFoundation/strax#638](https://github.com/AxFoundation/strax/pull/638) where `*alt_s2*` or `*s2_b*`is used, but I think there are no other places: https://gist.github.com/JoranAngevaare/859c924967db2067fad067e8e2f1d8d1
> 
> I'll checkout the PR in a bit more detail 👍

Thanks！

shenyangshi:
> I am wondering if we do not want to cut also events in which we have an alt. S2 right before an S1. It would not be a multi-scatter event but anyhow for data-quality I think we should cut those events.

Thanks for the reply @WenzDaniel! I think the situation you mentioned can be worked out from the S2 shadow cut, while it operates on the peak level, assigning which events as alt_S2 won't affect it.

WenzDaniel:
>  I think the situation you mentioned can be worked out from the S2 shadow cut, while it operates on the peak level, assigning which events as alt_S2 won't affect it.

Ah yes, this is correct. I forgot that it acts already at the peak level.

straxen/plugins/event_processing.py
```diff
@@ -305,7 +320,7 @@ def fill_events(self, result_buffer, events, split_peaks):
     def fill_result_i(self, event, peaks):
         """For a single event with the result_buffer"""
         # Consider S2s first, then S1s (to enable allow_posts2_s1s = False)
-        largest_s2s, s2_idx = self.get_largest_sx_peaks(peaks, s_i=2)
+        largest_s2s, s2_idx = self.get_largest_sx_peaks(peaks, s_i=2, number_of_peaks=0)
```
WenzDaniel, line 323: This is super misleading. You should put at least a comment that putting `number_of_peaks=0` means you return now all peaks not zero peaks.

straxen/plugins/event_processing.py
```diff
@@ -341,6 +357,31 @@ def fill_result_i(self, event, peaks):
...
+
+            if (~mask).all():
+                # If no other s2 survives, drop the alt s2 field
+                s2_idx = s2_idx[0:1]
+                largest_s2s = largest_s2s[0:1]
+            else:
+                # Take the first valid largest other s2 as alt s2
+                new_alt_s2_index = np.arange(0, len(largest_alt_s2s))[mask][0] + 1
+                s2_idx[1], s2_idx[new_alt_s2_index] = s2_idx[new_alt_s2_index], s2_idx[1]
+                largest_s2s[1], largest_s2s[new_alt_s2_index] = largest_s2s[new_alt_s2_index], largest_s2s[1]
```
WenzDaniel, line 384: Sorry, but here I do not get why you have to move the the index/S2 which is at position 1 to the position of your new alt S2. I mean in the end you only keep the first two values in the array anyhow,

WenzDaniel (changes requested):
Thanks, I think from a logic and physics point of view there is nothing to object. But the current change will break the `force_main_before_alt` option. 

Since this code is a bit more complicated and took me a bit of time to review, I would like to ask you also to add some test for `fill_result_i`. This test can be fairly simple and should check  whether all options are treated correctly.  As an example you can have a look at this [test](https://github.com/XENONnT/straxen/blob/2fc89e8138163a89196d03234983ec320747b40b/tests/test_peaklet_processing.py#L33) which defines a fixed set of dummy records to check whether peaklets will have the correct number of hits. In you case you would define a dummy event and different scenarios of main/alt S2s distrusted before after your main S1.

DonNabla:
From what I see, it will not impact Kr83m selection.

straxen/plugins/event_processing.py
shenyangshi: Thanks! I'll modify it accordingly.

shenyangshi:
Hi @WenzDaniel, thanks for the comments! I made the following change:

1. Add a config that lets users choose whether to use this new alt-s2 criterion, e.g., for ac people or people working on data-driven background modeling.
2. Move force_main_before_alt after this set_alt_s2_properties. As alt_s2 is selected between s1 and max_drift_time, a swap of main_s2 and alt_s2 won't affect the event's topology.
3. Use np.take to select [0] and [new_alt_s2_index]

straxen/plugins/event_processing.py
```diff
@@ -142,7 +154,7 @@ class EventBasics(strax.Plugin):
     The main S2 and alternative S2 are given by the largest two S2-Peaks
     within the event. By default this is also true for S1.
     """
-    __version__ = '1.2.1'
+    __version__ = '1.2.2'
```
WenzDaniel, line 157:
```suggestion
    __version__ = '1.3.0'
```

WenzDaniel: Because it is a logic change rather than some debugging.

WenzDaniel (approved):
Okay for me it looks fine, though I would be still in favor of having some tests here.

shenyangshi:
> Okay for me it looks fine, though I would be still in favor of having some tests here.

Sure! I'll check the logic and output and send the result here.

WenzDaniel:
With tests I meant unittests. Just to be sure :D

jingqiangye:
Actually it might have some impact for BiPo214 coincidence tag (t_0.5 = 164 us)... but it should only affect deep Z events so maybe still fine? @cfuselli

jingqiangye:
Thanks @cfuselli! Then I don't see an issue either :D

WenzDaniel (changes requested):
Sorry, but I just noticed that adding the if statement did not solve the problem is not it? If you set `force_alt2_in_max_drift_time` to false. The code is identical with the previous version except for the `number_of_peaks=0` in `get_largest_sx_peaks`. Hence one has still the issue I mentioned before is not it?

shenyangshi:
Hi @WenzDaniel, sorry for my late reply. I add a slice in force_main_before_alt, so only the largest two S2s get sorted by time. I think this can do the trick.

WenzDaniel (approved):
Thanks a lot sorry for being so pedantic :D But now it looks fine.

WenzDaniel:
I will merge it right before the release as it changes the lineage.

shenyangshi:
Thanks a lot!

straxen/plugins/event_processing.py
```diff
@@ -128,11 +128,23 @@ def compute(self, peaks, start, end):
...
+        help="Make sure alt_s2 is in max drift time starting from main S1"),
     strax.Option(
         name='event_s1_min_coincidence',
         default=2, infer_type=False,
         help="Event level S1 min coincidence. Should be >= s1_min_coincidence "
              "in the peaklet classification"),
+    strax.Option(
+        name='electron_drift_velocity', infer_type=False,
+        default=("electron_drift_velocity", "ONLINE", True),
+        help='Vertical electron drift velocity in cm/ns (1e4 m/ms)',),
```
JoranAngevaare, line 142: could we replace this with an URL config? We are now changing the hash anyway.

JoranAngevaare (commented):
Sorry for being so late on the review. Could you also update things like docstrings in the class since this changes the definition of the alt-s2 but not alt-s1.

I also had another question which we might better discuss on [slack](https://xenonnt.slack.com/archives/C0177FXBEHX/p1645692370453429).

straxen/plugins/event_processing.py
shenyangshi: Sure! I'm wondering if you can provide an example of URL config? Here I'm just copying from class Events' config.

JoranAngevaare: you can set it as a class attribute (see e.g. https://github.com/XENONnT/straxen/blob/92877a183f20aaf425625a5f00af30ac1f0fec31/straxen/plugins/event_processing.py#L577) 
```python
    electron_drift_velocity = straxen.URLConfig(
        default='cmt://electron_drift_velocity?version=ONLINE&run_id=plugin.run_id',
        help='Vertical electron drift velocity in cm/ns (1e4 m/ms)')
``` 
you can remove this from the `setup()` then.

Let me know if you want me to push this commit / fix it when straxen starts complaining (I think `Events` should be similarly updated).

PR #911, opened 2022-02-03
nestpy on py3.10

JoranAngevaare:
Let's retry this since nestpy seems to support py3.10, based on their testing suite https://github.com/NESTCollaboration/nestpy/pull/80. The [first try](https://github.com/XENONnT/straxen/pull/878) failed, but installing nestpy from source seems to fix the issue.

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -33,7 +33,7 @@ jobs:
     strategy:
       fail-fast: False
       matrix:
-        python-version: ["3.10" ]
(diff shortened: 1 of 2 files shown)


sophiafarrell:
Just an idea @JoranAngevaare , do you think the answer on why the `pip install nestpy==1.5.0` isn't working is simply because we didn't build a package for python3.10 for that version, so it's pointing to a non-existent distribution? In that case, we'd have to include 3.10 in the builds suite, then it _should_ work (assuming I'm right.)

PR #912, opened 2022-02-03
Update Shadow plugins and add Ambience plugins

dachengx:

Update shadow plugins to calculate position correlated shadow. Add Ambience cut' to future remove AC-like events. 

Ambience plugins count the number of lone hits, S0, S1, S2 in a time window before S1 peaks and count S2 numbers near S2s in (x,y) space. Shadow cuts calculate shadow with a factor describing position correlation to show the correlation between isolated S2 by SE pile-up and the primary large S2. 

[Note](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:ac:prediction:shadow_ambience): xenon:xenonnt:ac:prediction:shadow_ambience


straxen/plugins/event_processing.py
```diff
@@ -733,58 +733,175 @@ def cs2_to_e(self, x):
...
+@strax.takes_config(
+    strax.Option(name='pre_s1_area_threshold', default=1e3,
+                 help='Only take S1s larger than this into account '
+                      'when calculating PeakShadow [PE]'),
+    strax.Option(name='pre_s2_area_threshold', default=1e4,
+                 help='Only take S2s larger than this into account '
+                      'when calculating PeakShadow [PE]'),
+    strax.Option(name='deltatime_exponent', default=-1.0,
+                 help='The exponent of delta t when calculating shadow'),
+    strax.Option(name='time_window_backward', default=int(1e9),
```
skazama, line 745: sorry @dachengx, in [this note](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:ac:prediction:shadow_refactor), shadow is defined in 3sec time_window, while here 1sec time_window is used. Which is correct?

straxen/plugins/event_processing.py
```diff
@@ -733,58 +733,175 @@ def cs2_to_e(self, x):
...
+                 help='Search for peaks casting shadow in this time window [ns]')
+)
 class EventShadow(strax.Plugin):
     """
     This plugin can calculate shadow at event level.
     It depends on peak-level shadow.
-    The event-level shadow is its first S2 peak's shadow.
-    If no S2 peaks, the event shadow will be nan.
-    It also gives the position infomation of the previous S2s
+    The event-level shadow is its first peak peak's shadow.
```
skazama, line 752: sorry, it's not clear to me the definition of "event-level shadow". Can you please elaborate a bit on this explanation?

straxen/plugins/event_processing.py
dachengx: That note is outdated and I linked the latest version near the title, which is [this](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:ac:prediction:shadow_ambience) one. The time window we are using is 1sec.

straxen/plugins/event_processing.py
```diff
@@ -733,58 +733,175 @@ def cs2_to_e(self, x):
...
+            dtype.append((('previous ' + r + ' largest s2 shadow with position correlation PDF [PE/ns]', sa + 'shadow_s2_corr'), np.float32))
+            dtype.append((('previous ' + r + ' largest s2 shadow position correlation PDF', sa + 'shadow_s2_prob'), np.float32))
+        dtype.append((('index of the peak defining the event shadow', 'shadow_index'), np.int32))
+        for s in ['s1', 's2']:
+            for p, si in zip(['s1', 's2', 's2_re'], ['s1', 's2', 'reordered s2']):
+                dtype.append((('Nearest ' + si + ' delta t to ' + s, s + '_near_dt_' + p), np.int64))
+                dtype.append((('Sum of ' + si + ' shadow in time window casting on ' + s, s + '_shadow_sum_' + p), np.float32))
+        for p, si in zip(['s1', 's2', 's2_re'], ['s1', 's2', 'reordered s2']):
+            dtype.append((('Nearest ' + si + ' delta t to event', 'near_dt_' + p), np.int64))
+            dtype.append((('Sum of ' + si + ' shadow in time window casting on event', 'shadow_sum_' + p), np.float32))
```
skazama, line 790: sorry, but to me it's quite confusing what is the difference between s1/s2, main_s1/s2, previous_s1/s2, reordered_s2, etc. Can you maybe make a plot which illustrates each variables such that people can easily judge?

straxen/plugins/event_processing.py
skazama: What is the reason to change the time_window?

straxen/plugins/event_processing.py
skazama: for example, I cannot judge what "s1_alt_pre_area_s1" means by just looking at the corresponding variable...

straxen/plugins/event_processing.py
dachengx: I just updated the docstring. An event basically has three shadows cast by the primary S2 peak. They are: 

1. Main S1's shadow
2. Main S2's shadow
3. The first S2's shadow

The third one is a little bit confusing. But given the max drift time is ~2ms and we select a 1 sec time window when calculating shadow, there is a probability that the main S2's shadow is from the first S2 if the first S2 is larger than 1e4 PE(the threshold for 'large' peak). Also considering some events have no S1, an event's S2 shadow is its first S2's shadow.

straxen/plugins/event_processing.py
dachengx: The 'large' S2 rate is ~8.7 Hz in bkg runs. The possibility to find no large S2 peaks in 1 sec time window is very low(less than 0.0002). And when using 1 sec time window, PureAC validation is acceptable, which means 1 sec is adequate to describe the correlation.

straxen/plugins/peak_processing.py
```diff
@@ -332,97 +334,283 @@ def find_n_competing(peaks, windows, fraction):
...
+                    res['shadow_dt'][p_i] = p_a['center_time'] - s_a['center_time']
+
+
+@export
+@strax.takes_config(
+    strax.Option('ambience_divide_time', default=False,
+                 help='Whether to divide area by time'),
+    strax.Option('ambience_divide_space', default=False,
+                 help='Whether to divide area by space'),
+    strax.Option(name='time_window_ambience', default=int(2e6),
```
skazama, line 506: how this 2-msec value is optimized? Do you have any studies with different value?

dachengx: It is optimized for signal acceptance ~99% in ambience cut. I have not checked other values.

straxen/plugins/peak_processing.py
```diff
@@ -332,97 +334,283 @@ def find_n_competing(peaks, windows, fraction):
...
-        res['endtime'] = strax.endtime(peaks)
+        return self.compute_s1_s2(peaks, peaks)
+
+    def compute_s1_s2(self, peaks, current):
+        roi_shadow = np.zeros(len(current), dtype=strax.time_fields)
+        roi_shadow['time'] = current['center_time'] - self.time_window_backward
+        roi_shadow['endtime'] = current['center_time']
+
+        resdict = dict()
+        for stype, key, reorder in zip([2, 2, 1], ['s2_re', 's2', 's1'], [True, False, False]):
```
WenzDaniel, line 425: instead of using stype you could also check: `if 's2' in key:`. Not sure if this make things easier to read though.

straxen/plugins/peak_processing.py
```diff
@@ -332,97 +334,283 @@ def find_n_competing(peaks, windows, fraction):
...
+            mask_pre = (peaks['type'] == stype) & (peaks['area'] > self.threshold[key])
+            split_peaks = strax.touching_windows(peaks[mask_pre], roi_shadow)
+            array = np.zeros(len(current), np.dtype(self.sdtype + self.ddtype))
+            for sa in ['', 'alt_']:
+                for x in ['x', 'y']:
+                    array[sa + 'pre_' + x] = np.nan
+                array[sa + 'shadow_dt'] = self.time_window_backward
+                array[sa + 'pre_area'] = self.threshold[key]
+                array[sa + 'shadow'] = array[sa + 'pre_area'] * array[sa + 'shadow_dt'] ** self.exponent
```
WenzDaniel, line 434: Since you seem to define here a default value too, why do we have to repeat it at the event level? Can you not just copy over all values?

straxen/plugins/peak_processing.py
```diff
@@ -332,97 +334,283 @@ def find_n_competing(peaks, windows, fraction):
...
+            array['near_dt'] = self.time_window_backward
+            array['shadow_sum'] = 0
+            if len(current):
```
WenzDaniel, line 437: Do you really need this if statement? If current is empty also peaks is empty hence you function gets two empty inputs why bothering in that case with the function at all. You could move this if to the start of your function and just return an empty array of the correct dtype in that case.

straxen/plugins/peak_processing.py
```diff
@@ -332,97 +334,283 @@ def find_n_competing(peaks, windows, fraction):
...
+                               np.dtype(self.sdtype + self.ddtype).names]):
+            for name in names:
+                res[name + '_' + key] = resdict[key][name]
+        for sa in ['', 'alt_']:
+            distance = np.sqrt((res[sa + 'pre_x_s2_re'] - current['x']) ** 2 + (res[sa + 'pre_y_s2_re'] - current['y']) ** 2)
+            distance = np.where(np.isnan(distance), 2 * straxen.tpc_r, distance)
+            res[sa + 'shadow_s2_prob'] = halfcauchy.pdf(distance, scale=self.getsigma(self.sigma, self.baseline, current['area']))
+            res[sa + 'shadow_s2_corr'] = res[sa + 'shadow_s2_re'] * res[sa + 'shadow_s2_prob']
+        res['time'] = current['time']
+        res['endtime'] = strax.endtime(current)
```
WenzDaniel, line 453: Here you lost me completely there are too many nested things going on.

straxen/plugins/peak_processing.py
```diff
@@ -332,97 +334,283 @@ def find_n_competing(peaks, windows, fraction):
...
         return res
 
+    @staticmethod
+    def getsigma(sigma, baseline, s2):
+        return sigma / np.sqrt(s2) + baseline
```
WenzDaniel, line 458: A comment would be good.

straxen/plugins/peak_processing.py
```diff
@@ -332,97 +334,283 @@ def find_n_competing(peaks, windows, fraction):
...
+                near_dt = p_a['center_time'] - s_a['center_time']
+                if near_dt <= 0:
                     continue
-                new_shadow = s2_a['area'] * (
-                        p_a['center_time'] - s2_a['center_time'])**exponent
-                if new_shadow > res['shadow'][p_i]:
+                new_shadow = s_a['area'] * near_dt**exponent
+                if pos_corr:
+                    distance = np.sqrt((p_a['x'] - s_a['x']) ** 2 + (p_a['y'] - s_a['y']) ** 2)
+                    distance = np.where(np.isnan(distance), 2 * straxen.tpc_r, distance)
```
WenzDaniel, line 479: Why can the distance just not be defined in this case?

straxen/plugins/peak_processing.py
```diff
@@ -332,97 +334,283 @@ def find_n_competing(peaks, windows, fraction):
...
+                    shadow_ref = new_shadow * 2 / (np.pi * sigma * (1 + (distance / sigma) ** 2))
+                else:
+                    shadow_ref = new_shadow
+                res['shadow_sum'][p_i] += shadow_ref
```
WenzDaniel, line 483: This should be equivalent to:

```
    new_shadow *= 2 / (np.pi * sigma * (1 + (distance / sigma) ** 2))
# (skipping else and shadow_ref)
res['shadow_sum'][p_i] += new_shadow 
```

straxen/plugins/peak_processing.py
```diff
@@ -332,97 +334,283 @@ def find_n_competing(peaks, windows, fraction):
...
+    provides = 'peak_ambience'
+    data_kind = 'peaks'
+    save_when = strax.SaveWhen.EXPLICIT
+
+    def setup(self):
+        self.ambience_divide_time = self.config['ambience_divide_time']
+        self.ambience_divide_space = self.config['ambience_divide_space']
+        self.time_window_ambience = self.config['time_window_ambience']
+        self.ambience_area_parameters = self.config['ambience_area_parameters']
+        self.ambience_radius = self.config['ambience_radius']
```
WenzDaniel, line 531: Maybe you want to check out the new URL config by Yossi, Then you can skip these lines.

WenzDaniel (changes requested):
Hej Dacheng thank you very much for your work and nice study, but I would like to ask you to rework the plugins. I spent now 1 h to read it and even now I do not really understand whats going on. You have very many nested for loops and very short names which make it super hard to follow. In addition the way you format your data_types is not very readable either. 

Hence, I am afraid that in the current state the code is not maintainable and therefore cannot be merged as it is. Secondly, due to all your nested for loops I am very concerned about the performance of your plugins. Have you tested how long it takes to process some background and some calibration data. I put already quite a number of suggestions how you can increase readability of your code, but I am afraid that we will have to iterate 2 or 3 times.

straxen/plugins/event_processing.py
```diff
@@ -733,58 +733,173 @@ def cs2_to_e(self, x):
...
-                 ('pre_s2_x', np.float32, 'x of previous s2 peak causing shadow [cm]'),
-                 ('pre_s2_y', np.float32, 'y of previous s2 peak causing shadow [cm]'),
-                 ('shadow_distance', np.float32, 'distance to the s2 peak with max shadow [cm]')]
+        dtype = []
+        for sa, r in zip(['', 'alt_'], ['1st', '2nd']):
+            for s in ['s1', 's2']:
+                for p, si in zip(['s1', 's2', 's2_re'], ['s1', 's2', 'reordered s2']):
+                    dtype.append((('main ' + s + ' shadow from ' + si + ' casting ' + r + ' largest shadow [PE/ns]', s + '_' + sa + 'shadow_' + p), np.float32))
+                    dtype.append((('previous ' + si + ' area casting ' + r + ' largest shadow on ' + s + ' [PE]', s + '_' + sa + 'pre_area_' + p), np.float32))
+                    dtype.append((('time difference from ' + s + ' to the previous ' + si + ' casting ' + r + ' largest shadow [ns]', s + '_' + sa + 'shadow_dt_' + p), np.int64))
```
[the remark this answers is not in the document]
dachengx: You're entirely right. Let me delete some useless variables and rewrite the plugin. :)

straxen/plugins/peak_processing.py
dachengx: Would you please link me to an example?

dachengx:
Hi, @WenzDaniel ! I made changes and also tested the codes. And I also updated comments in between lines. Hope it is now more readable and these updates will make review easier. The main changes are:

1. Delete all features related to `alt_`
2. Delete all features related to `shadow_distance`
3. Classify features in 2 categories: `time` shadow and `position` shadow, corresponding to the latest [note](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:ac:prediction:shadow_ambience). 
4. Only calculate shadow for the main S1 and the 1st S2 peak in `EventShadow`

There are still some issues that remain:

1. I am still using `strax.touching_windows` to find the primary peaks in the time window. Related [comment](https://github.com/XENONnT/straxen/pull/912#discussion_r811053151).  
2. I have not checked out the new URL config. Related [comment](https://github.com/XENONnT/straxen/pull/912#discussion_r811072427). 
3. In initialization, all values are not set to be NaN. Related [comment](https://github.com/XENONnT/straxen/pull/912#discussion_r811032235). 
4. Maybe I should open another PR to store `n_hits` in `PeakBasics` and `EventBasics`? Related [comment](https://github.com/XENONnT/straxen/pull/912#discussion_r811052848).

straxen/plugins/event_processing.py
```diff
@@ -733,58 +733,173 @@ def cs2_to_e(self, x):
...
+            for re in ['', '_re']:
+                for x in ['x', 'y']:
+                    res[sa + 'pre_' + x + '_s2' + re] = np.nan
+                res[sa + 'shadow_distance' + re] = np.nan
+            res[sa + 'shadow_s2_prob'] = np.nan
+            res[sa + 'shadow_s2_corr'] = np.nan
+        for key in ['s1_', 's2_', '']:
+            for s, ty in zip(['s1', 's2', 's2_re'], ['s1', 's2', 's2_re']):
+                res[key + 'near_dt_' + s] = self.time_window_backward
+                res[key + 'shadow_sum_' + s] = self.threshold[ty] * self.time_window_backward ** self.exponent
```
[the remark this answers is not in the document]
dachengx: As features related to shadow is describing roughly the status of the TPC(the status is more related to isolated peak). I think I need to set the default features corresponding to the stablest, i.e., set the feature as we can not find the primary peak(which means the TPC is stable, at least for isolated peaks).

straxen/plugins/event_processing.py
```diff
@@ -733,58 +734,153 @@ def cs2_to_e(self, x):
...
+            for key in ['s1_time', 's2_time', 's2_position']:
+                type_str = key.split('_')[0]
+                tp_desc = key.split('_')[1]
+                dtype.append(((f'{tp_desc} shadow casting from {type_str} to {main_peak_desc} [PE/ns]', f'{main_peak}shadow_{key}'), np.float32))
+                dtype.append(((f'area of primary {type_str} casting {tp_desc} shadow on {main_peak_desc} [PE]', f'{main_peak}pre_area_{key}'), np.float32))
+                dtype.append(((f'time difference from the primary {type_str} casting {tp_desc} shadow to {main_peak_desc} [ns]', f'{main_peak}shadow_dt_{key}'), np.int64))
+                if 'time' in key:
+                    dtype.append(((f'Time difference from the nearest primary {type_str} to {main_peak_desc} [ns]', f'{main_peak}nearest_dt_{type_str}'), np.int64))
+        for x in ['x', 'y']:
+            dtype.append(((f'{x} of primary s2 peak casting time shadow on the 1st S2 peak [cm]', f'pre_{x}_s2_time'), np.float32))
```
skazama, line 775: What is the difference between primary S2, main S2 and 1st S2?

skazama: For example, I don't understand yet the description of following dtype:

- "time shadow casting from s1 to main s1 [PE/ns]"  (btw, what is time shadow?)
- "area of primary s1 casting time shadow on main s1 [PE]"
-

dachengx:
I will mark this PR as a draft and explain more on the questions from reviewers. I will also revise the code more. Just to avoid wasting the reviewers' time.

dachengx:
In the last commit, I further delete features unnecessary. I also made an illustration of features in plugins in [note](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:ac:prediction:shadow_ambience#shadow_cuts_summary).

straxen/plugins/event_processing.py
dachengx: @skazama I updated the dtype descriptions. Hope you can get the points now. :)

straxen/plugins/event_processing.py
```diff
@@ -733,58 +734,137 @@ def cs2_to_e(self, x):
...
-                 ('pre_s2_x', np.float32, 'x of previous s2 peak causing shadow [cm]'),
-                 ('pre_s2_y', np.float32, 'y of previous s2 peak causing shadow [cm]'),
-                 ('shadow_distance', np.float32, 'distance to the s2 peak with max shadow [cm]')]
+        dtype = []
+        for main_peak, main_peak_desc in zip(['s1_', 's2_'], ['main S1', 'main S2']):
+            # previous S1 can only cast time shadow, previous S2 can cast both time & position shadow
+            for key in ['s1_time', 's2_time', 's2_position']:
+                type_str = key.split('_')[0]
+                tp_desc = key.split('_')[1]
+                dtype.append(((f'largest {tp_desc} shadow casting from previous {type_str} to {main_peak_desc} [PE/ns]', f'{main_peak}shadow_{key}'), np.float32))
```
skazama, line 769: still I have a question about` nearest previous large s1`.

- what do you mean by large? If area_size is above pre_s1_area_threshold , then such s1s are called as large s1s?
- what do you mean by nearest? in which time_window? maybe time_window_backward(1sec)? Maybe nearest mean minimum time_difference w.r.t main S1, right?

straxen/plugins/event_processing.py
```diff
@@ -733,58 +734,137 @@ def cs2_to_e(self, x):
...
+                if 'time' in key:
+                    dtype.append(((f'time difference from the nearest previous large {type_str} to {main_peak_desc} [ns]', f'{main_peak}nearest_dt_{type_str}'), np.int64))
+            # Also record the PDF of HalfCauchy when calculating S2 position shadow
+            dtype.append(((f'PDF describing correlation between previous s2 and {main_peak_desc}', f'{main_peak}shadow_s2_position_pdf'), np.float32))
         dtype += strax.time_fields
         return dtype
 
+    def setup(self):
+        self.time_window_backward = self.config['time_window_backward']
+        self.threshold = dict(s1_time=self.config['pre_s1_area_threshold'], s2_time=self.config['pre_s2_area_threshold'], s2_position=self.config['pre_s2_area_threshold'])
```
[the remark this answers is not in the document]
dachengx: In all codes in this PR. `s1_time` is S1 time shadow, `s2_time` is S2 time shadow, `s2_position` is S2 position shadow.

straxen/plugins/event_processing.py
dachengx: - The S1 larger than `pre_s1_area_threshold` is referred to as large S1 in codes. 
- Yes!

straxen/plugins/event_processing.py
```diff
@@ -733,58 +734,137 @@ def cs2_to_e(self, x):
...
+        * v0.1.2 reference: xenon:xenonnt:ac:prediction:shadow_ambience
     """
-    __version__ = '0.0.8'
+    __version__ = '0.1.2'
     depends_on = ('event_basics', 'peak_basics', 'peak_shadow')
     provides = 'event_shadow'
+    data_kind = 'events'
     save_when = strax.SaveWhen.EXPLICIT
 
     def infer_dtype(self):
```
skazama, line 762: I still think that naming is bit confusing. What about naming like this?

`xx_yy_zz_ww`

xx -> variable (dt, ts, ps, pdf, x, y)
yy -> ms1 or ms2
zz -> ps1, ps2, ns1, or ns2
ww -> ts, ps

, where 

ms1/2 -> main s1/2
ps1/2 -> previous s1/2
ts -> largest time_shadow 
ps -> largest position_shadow 
dt -> time_difference 


Then, my proposal is like this:

- s1_pre_dt_s1_time -> dt_ms1_ps1_ts
- s1_pre_dt_s2_time -> dt_ms1_ps2_ts
- s2_pre_dt_s1_time -> dt_ms2_ps1_ts
- s2_pre_dt_s2_time -> dt_ms2_ps2_ts

- s1_pre_x_s2_time -> x_ms1_ps2_ts
- s1_pre_y_s2_time -> y_ms1_ps2_ts
- s2_pre_x_s2_time -> x_ms2_ps2_ts
- s2_pre_y_s2_time -> y_ms2_ps2_ts

- s1_nearest_dt_s1 -> dt_ms1_ns1
- s1_nearest_dt_s2 -> dt_ms1_ns2
- s2_nearest_dt_s1 -> dt_ms2_ns1
- s2_nearest_dt_s2 -> dt_ms2_ns2

- s1_shadow_s1_time -> ts_ms1_ps1_ts
- s1_shadow_s2_time-> ts_ms1_ps2_ts
- s2_shadow_s1_time -> ts_ms2_ps1_ts
- s2_shadow_s2_time-> ts_ms2_ps2_ts

- s1_pre_dt_s2_position -> dt_ms1_ps2_ps
- s2_pre_dt_s2_position -> dt_ms2_ps2_ps

- s1_pre_x_s2_position -> x_ms1_ps2_ps
- s1_pre_y_s2_position -> y_ms1_ps2_ps

- s1_shadow_s2_position -> ps_ms1_ps2_ps
- s2_shadow_s2_position -> ps_ms2_ps2_ps

- s2_pre_x_s2_position -> x_ms2_ps2_ps
- s2_pre_y_s2_position -> y_ms2_ps2_ps

- s1_shadow_s2_position_pdf -> pdf_ms1_ps2_ps (not sure if this is proper naming)
- s2_shadow_s2_position_pdf -> pdf_ms2_ps2_ps (not sure if this is proper naming)

skazama: @WenzDaniel do you have any suggestions?

straxen/plugins/peak_processing.py
dachengx: I've found examples among codes in the master branch. I will try to use these new features.

straxen/plugins/event_processing.py
dachengx: The purpose to set the names like this is to stay consistent with other names in straxen, for example in `event_info`([doc](https://straxen.readthedocs.io/en/latest/reference/datastructure_nT.html#event-info)). Name of the main S1's features begin with `s1_`, and the name of the main S2's features begin with `s2_`. I keep using `s1_time` & `s2_time` & `s2_position` as suffix because I would call them S1 time shadow, S2 time shadow, S2 position shadow.

dachengx: Maybe rename `s1_shadow_s2_position_pdf` to `s1_pdf_s2_position` would be better.

skazama: ok, but for example `s1_pre_dt_s1_time` is really confusing to me because `pre_` and `s1_` are separated, while it should be `pre_s1`, so at least something like `s1_dt_ps1_ts` makes sense to me.

WenzDaniel: From my personal point of view I would also prefer a more explicit naming like `s1_dt_preS1_time_shadow`. It adds to readability of the code and I mean during analysis you can always press `tab` to get naming suggestions and select from that.

straxen/plugins/peak_processing.py
```diff
@@ -332,97 +334,316 @@ def find_n_competing(peaks, windows, fraction):
...
+    def peaks_shadow(peaks, pre_peaks, touching_windows, exponent, result, pos_corr, sigmas=None):
         """
-        For each peak in peaks, check if there is a shadow-casting S2 peak
+        For each peak in peaks, check if there is a shadow-casting peak
         and check if it casts the largest shadow
         """
+        for p_i, (p_a, sigma) in enumerate(zip(peaks, sigmas)):
+            indices = touching_windows[p_i]
+            for idx in range(indices[0], indices[1]):
+                s_a = pre_peaks[idx]
```
WenzDaniel, line 490: Can we rename s_a in something more readable like `shadow_peak` or `shadow_casting_peak`?

WenzDaniel: Same for p_a ? I do not know what this `a` stands for.

straxen/plugins/peak_processing.py
```diff
@@ -332,97 +334,316 @@ def find_n_competing(peaks, windows, fraction):
...
     @numba.njit
-    def compute_shadow(peaks, pre_s2_peaks, touching_windows, exponent, res):
+    def peaks_shadow(peaks, pre_peaks, touching_windows, exponent, result, pos_corr, sigmas=None):
         """
-        For each peak in peaks, check if there is a shadow-casting S2 peak
+        For each peak in peaks, check if there is a shadow-casting peak
         and check if it casts the largest shadow
         """
+        for p_i, (p_a, sigma) in enumerate(zip(peaks, sigmas)):
+            indices = touching_windows[p_i]
```
WenzDaniel, line 488: Same comment as above you can do:

```
for p_i, (p_a, sigma) in enumerate(zip(peaks, sigmas)):
    res_i = result[p_i]
```
Reduces work and complexity later,

straxen/plugins/peak_processing.py
```diff
@@ -332,97 +334,316 @@ def find_n_competing(peaks, windows, fraction):
...
+            # Shadow is set to be the lowest possible value
+            if 'time' in key:
+                array['shadow'] = self.threshold[key] * array['pre_dt'] ** self.exponent
+            else:
+                array['shadow'] = 0
+            array['nearest_dt'] = self.time_window_backward
+
+            # Calculating shadow, the Major of the plugin. Only record the previous peak casting the largest shadow
+            if len(current_peak):
+                self.peaks_shadow(current_peak, peaks[mask_pre], split_peaks, self.exponent, array, is_position, self.getsigma(self.sigma, self.baseline, current_peak['area']))
```
WenzDaniel, line 450: Can you use line breaks after each argument? I have already a very large screen and it still does not fit :D In case you are using pycharm you can set a max line number. If I recall correctly we settled on something like 130. Maybe also check the rest f your code you have some functions with very long arguments.

straxen/plugins/peak_processing.py
```diff
@@ -332,97 +334,316 @@ def find_n_competing(peaks, windows, fraction):
...
+
+    @property
+    def origin_dtype(self):
+        return ['lh_before', 's0_before', 's1_before', 's2_before', 's2_near']
+
+    def infer_dtype(self):
+        dtype = []
+        for ambience in self.origin_dtype:
+            dtype += [(('Number of small ' + ' '.join(ambience.split('_')) + ' a peak', f'n_{ambience}'), np.int16), 
+                      (('Area sum of small ' + ' '.join(ambience.split('_')) + ' a peak', f's_{ambience}'), np.float32)]
```
WenzDaniel, line 557: See comment above

straxen/plugins/peak_processing.py
```diff
@@ -332,97 +334,316 @@ def find_n_competing(peaks, windows, fraction):
...
+        touching_windows = strax.touching_windows(lone_hits, roi)
+        # Calculating ambience
+        self.lonehits_ambience(current_peak, lone_hits, touching_windows, num_array, sum_array, self.ambience_divide_time)
+        result['n_lh_before'] = num_array
+        result['s_lh_before'] = sum_array
+
+        # 4. Calculate number and area sum of small S0, S1, S2 before a peak
+        radius = -1
+        for stype, area, type_str in zip([0, 1, 2], self.ambience_area_parameters, ['s0_before', 's1_before', 's2_before']):
+            num_array = np.zeros(len(current_peak), dtype=np.int16)
```
WenzDaniel, line 585: You do not need the type_str do you? You can also do in the end `result[f'n_s{stype}_before']` etc.

straxen/plugins/peak_processing.py
```diff
@@ -332,97 +334,316 @@ def find_n_competing(peaks, windows, fraction):
...
-            # reset for every peak
-            new_shadow = 0
-            s2_indices = touching_windows[p_i]
-            for s2_idx in range(s2_indices[0], s2_indices[1]):
-                s2_a = pre_s2_peaks[s2_idx]
-                if p_a['center_time'] - s2_a['center_time'] <= 0:
+            indices = touching_windows[p_i]
+            for idx in range(indices[0], indices[1]):
+                s_a = pre_peaks[idx]
+                r = ((p_a['x'] - s_a['x'])**2 + (p_a['y'] - s_a['y'])**2)**0.5
```
WenzDaniel, line 636: You compute the radius quite a lot at different positions in your code. I think it could be useful to define a small function for it.

straxen/plugins/peak_processing.py
```diff
@@ -332,97 +334,316 @@ def find_n_competing(peaks, windows, fraction):
...
                     continue
-                new_shadow = s2_a['area'] * (
-                        p_a['center_time'] - s2_a['center_time'])**exponent
-                if new_shadow > res['shadow'][p_i]:
-                    res['shadow'][p_i] = new_shadow
-                    res['pre_s2_area'][p_i] = s2_a['area']
-                    res['shadow_dt'][p_i] = p_a['center_time'] - s2_a['center_time']
-                    res['pre_s2_x'][p_i] = s2_a['x']
-                    res['pre_s2_y'][p_i] = s2_a['y']
+                if (dt != 0) and ((radius < 0) or (r <= radius)):
```
WenzDaniel, line 640: `radius` is a bit general. Maybe call it max_ambient_radius?

WenzDaniel: You can drop the test for `dt` here. Given the `dt <= 0` above you know that all dts must be positive here.

WenzDaniel (commented):
Thanks Dacheng it is already much more readable. Although I am still not sure if I am understanding everything 100%. In your lat version you also computed the number of hits for every main/alt S1/S2 and I suggested to move it to event_basics. However, I have not seen this change in the current version. 

Additionally, I put another iteration of comments.

straxen/plugins/event_processing.py
```diff
@@ -733,58 +734,137 @@ def cs2_to_e(self, x):
 
 
 @export
+@strax.takes_config(
+    strax.Option(name='pre_s1_area_threshold', default=1e3,
+                 help='Only take S1s larger than this into account '
+                      'when calculating PeakShadow [PE]'),
+    strax.Option(name='pre_s2_area_threshold', default=1e4,
+                 help='Only take S2s larger than this into account '
+                      'when calculating PeakShadow [PE]'),
```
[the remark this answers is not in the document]
dachengx: Thanks for the comment. Actually, S1 and S2 have different light distributions in TPC(x,y,z) so the thresholds can be different. This is actually the AC analysts are not very confident with. The threshold is a little bit arbitrary here. I think if the threshold is enough for AC suppression for at least the current stage, we can stay with those values.

straxen/plugins/event_processing.py
```diff
@@ -733,58 +734,137 @@ def cs2_to_e(self, x):
...
+                result[f'{main_peak}pre_dt_{key}'] = self.time_window_backward
+                if 's2' in key:
+                    result[f'{main_peak}pre_x_{key}'] = np.nan
+                    result[f'{main_peak}pre_y_{key}'] = np.nan
+                if 'time' in key:
+                    result[f'{main_peak}nearest_dt_{type_str}'] = self.time_window_backward
+                    result[f'{main_peak}shadow_{key}'] = self.threshold[key] * result[f'{main_peak}pre_dt_{key}'] ** self.exponent
+                else:
+                    result[f'{main_peak}shadow_{key}'] = 0
+            result[f'{main_peak}shadow_s2_position_pdf'] = np.nan
```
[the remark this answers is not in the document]
dachengx: Oh sorry, I missed that comment. I will delete these.

straxen/plugins/peak_processing.py
dachengx: I will check them. :)

straxen/plugins/peak_processing.py
```diff
@@ -332,97 +334,316 @@ def find_n_competing(peaks, windows, fraction):
...
+    strax.Option(name='ambience_divide_time', default=False,
+                 help='Whether to divide area by time'),
+    strax.Option(name='ambience_divide_space', default=False,
+                 help='Whether to divide area by space'),
+    strax.Option(name='time_window_ambience', default=int(2e6),
+                 help='Search for ambience in this time window [ns]'),
+    strax.Option(name='ambience_area_parameters', default=(5, 60, 60), type=(list, tuple),
+                 help='The upper limit of S0, S1, S2 area to be counted'),
+    strax.Option(name='ambience_radius', default=6.7,
+                 help='Search for ambience in this radius [cm]'),
```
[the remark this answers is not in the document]
dachengx: Not like that. When calculating `n_sx_before`, `radius` is set as -1. In that case we do not require `r <= radius` because `r <= radius` is for `n_s2_near`.

dachengx:
@WenzDaniel I just add `n_hits` in `PeakBasics` & `EventBasics`.

straxen/plugins/event_processing.py
dachengx: I updated the features' names following reviewers' suggestions. :)

straxen/plugins/peak_processing.py
```diff
@@ -331,98 +336,393 @@ def find_n_competing(peaks, windows, fraction):
...
+    shadow_sigma_and_baseline = straxen.URLConfig(
+        default=[15.220, 0.036],
+        type=list,
+        track=True,
+        help='Fitted position correlation sigma[cm*PE^0.5] and baseline[cm] using in position shadow'
+    )
 
     def get_window_size(self):
-        return 3 * self.config['time_window_backward']
+        return 10 * self.config['shadow_time_window_backward']
```
skazama, line 384: sorry @dachengx, did you change time window from 3sec to 10sec? Do you have any notes for this change (sorry I maybe missed your updates)?

dachengx: Hey @skazama . This function is related to [this](https://strax.readthedocs.io/en/latest/developer/overlaps.html) doc and it is fully technical. The time window is still 1 sec. 3 & 10 will give the same results here.

straxen/plugins/peak_processing.py
```diff
@@ -331,98 +336,393 @@ def find_n_competing(peaks, windows, fraction):
...
+    def lonehits_ambience(peaks, pre_hits, 
+                          touching_windows, num_array, 
+                          sum_array, ambience_divide_t):
+        # Function to find lonehits before a peak
+        # creating_hit is the lonehit creating ambience
+        # suspicious_peak is the suspicious peak in the ambience created by creating_hit
+        for p_i, suspicious_peak in enumerate(peaks):
+            indices = touching_windows[p_i]
+            for idx in range(indices[0], indices[1]):
+                creating_hit = pre_hits[idx]
```
skazama, line 684: According to @GiovanniVolta's study, sometimes there are lone_hits with negative area (See [this plot ](https://xe1t-wiki.lngs.infn.it/lib/exe/fetch.php?cache=&media=xenon:giovo:led_vs_lh_spe_area.png) or [this note](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:giovo:led_calibration_vs_lone_hits)), so probably it might be better to just use lone_hits which have area > 0

straxen/plugins/peak_processing.py
```diff
@@ -331,98 +336,393 @@ def find_n_competing(peaks, windows, fraction):
...
+
+    ambience_divide_r = straxen.URLConfig(
+        default=False,
+        type=bool,
+        track=True,
+        help='Whether to divide area by radius'
+    )
+
+    ambient_radius = straxen.URLConfig(
+        default=6.7,
```
skazama, line 576: sorry @dachengx, where can I see how you optimized this parameter?  In [this note](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:ac:prediction:shadow_ambience), I could not find quantitative estimation.

straxen/plugins/peak_processing.py
```diff
@@ -331,98 +336,393 @@ def find_n_competing(peaks, windows, fraction):
...
+        type=float,
+        track=True,
+        help='Search for ambience in this radius [cm]'
+    )
+
+    ambience_area_parameters = straxen.URLConfig(
+        default=(5, 60, 60),
```
skazama, line 583: Also for this, where can I see how these parameters are  optimized?

straxen/plugins/peak_processing.py
```diff
@@ -331,98 +336,393 @@ def find_n_competing(peaks, windows, fraction):
...
+        * v0.0.5 reference: xenon:xenonnt:ac:prediction:shadow_ambience
+    """
+    __version__ = '0.0.5'
+    depends_on = ('lone_hits', 'peak_basics', 'peak_positions')
+    provides = 'peak_ambience'
+    data_kind = 'peaks'
+    save_when = strax.SaveWhen.EXPLICIT
+
+    ambience_time_window_backward = straxen.URLConfig(
+        default=int(2e6),
```
skazama, line 555: Also for this, where can I see how these parameters are optimized?

straxen/plugins/peak_processing.py
dachengx: They are just naive selections of lonehits and SE. I have not optimized these values.

straxen/plugins/peak_processing.py
dachengx: It is in the [definition](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:ac:prediction:shadow_ambience#definition) of Ambience.

straxen/plugins/peak_processing.py
dachengx: It is in the [definition](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:ac:prediction:shadow_ambience#definition) of Ambience.

straxen/plugins/peak_processing.py
skazama: sorry @dachengx, which is the right value, 2us or 2ms? In your note, it is mentioned as 2ms, no?

straxen/plugins/peak_processing.py
skazama: Which plots should I look at to understand how you optimized this 6.7 cm cut value?

straxen/plugins/peak_processing.py
skazama: Which plots should I look at to understand how you optimized this 2 milliseconds cut value?

straxen/plugins/peak_processing.py
```diff
@@ -331,98 +336,393 @@ def find_n_competing(peaks, windows, fraction):
...
+    def compute_shadow(self, peaks, current_peak):
+        # 1. Define time window for each peak, we will find previous peaks within these time windows
+        roi_shadow = np.zeros(len(current_peak), dtype=strax.time_fields)
+        roi_shadow['time'] = current_peak['center_time'] - self.config['shadow_time_window_backward']
+        roi_shadow['endtime'] = current_peak['center_time']
+
+        # 2. Calculate S2 position shadow, S2 time shadow, and S1 time shadow
+        result = np.zeros(len(current_peak), self.dtype)
+        for key, is_position in zip(['s2_position_shadow', 's2_time_shadow', 's1_time_shadow'], [True, False, False]):
+            type_str = key.split('_')[0]
```
WenzDaniel, line 446: You could replace your second loop parameter by:
for key in 
    is_position =  'position' in key:

WenzDaniel: I am also wondering if this cannot be simplified further into two cases only. Case A. it is an S1 so you only have time_shadow, case B. it is an S2 so you have both. Maybe think a bit about it for the future. It is still hard to read but for the moment I think it is fine.

straxen/plugins/peak_processing.py
```diff
@@ -331,98 +336,393 @@ def find_n_competing(peaks, windows, fraction):
...
+            stype = 2 if 's2' in key else 1
+            mask_pre = (peaks['type'] == stype) & (peaks['area'] > self.config['shadow_threshold'][key])
+            split_peaks = strax.touching_windows(peaks[mask_pre], roi_shadow)
+            array = np.zeros(len(current_peak), np.dtype(self.shadowdtype))
+
+            # Initialization
+            array['x'] = np.nan
+            array['y'] = np.nan
+            array['dt'] = self.config['shadow_time_window_backward']
+            # Shadow is set to be the lowest possible value
```
WenzDaniel, line 456: `# Shadow is set to be the lowest possible value -> # The default value for shadow...`
 is not it?

straxen/plugins/peak_processing.py
```diff
@@ -331,98 +336,393 @@ def find_n_competing(peaks, windows, fraction):
...
-                if p_a['center_time'] - s2_a['center_time'] <= 0:
+        for p_i, (suspicious_peak, sigma) in enumerate(zip(peaks, sigmas)):
+            # casting_peak is the previous large peak casting shadow
+            # suspicious_peak is the suspicious peak which in shadow from casting_peak
+            indices = touching_windows[p_i]
+            for idx in range(indices[0], indices[1]):
+                casting_peak = pre_peaks[idx]
+                dt = suspicious_peak['center_time'] - casting_peak['center_time']
+                if dt <= 0:
+                    continue
```
WenzDaniel, line 520: I think you can even `break` here is not it?. Peaks should be sorted by time. Although I think the difference is small.

straxen/plugins/peak_processing.py
```diff
@@ -331,98 +336,393 @@ def find_n_competing(peaks, windows, fraction):
...
+                # First we record the time difference to the nearest previous peak
+                if dt < result['nearest_dt'][p_i]:
+                    result['nearest_dt'][p_i] = dt
```
WenzDaniel, line 523: FYI,  `result['nearest_dt'][p_i] = min(result['nearest_dt'][p_i], dt)` works too and is a bit shorter.

straxen/plugins/peak_processing.py
```diff
@@ -331,98 +336,393 @@ def find_n_competing(peaks, windows, fraction):
...
+                    # Calculate position shadow which is time shadow with a HalfCauchy PDF multiplier
+                    distance = distance_in_xy(suspicious_peak, casting_peak)
+                    distance = np.where(np.isnan(distance), 2 * straxen.tpc_r, distance)
+                    new_shadow *= 2 / (np.pi * sigma * (1 + (distance / sigma) ** 2))
+                # Only the previous peak with largest shadow is recorded
+                if new_shadow > result['shadow'][p_i]:
+                    result['shadow'][p_i] = new_shadow
+                    result['x'][p_i] = casting_peak['x']
+                    result['y'][p_i] = casting_peak['y']
+                    result['dt'][p_i] = suspicious_peak['center_time'] - casting_peak['center_time']
```
WenzDaniel, line 536: Here `dt` refers to the time difference to the largest shadowing peak right? Just to understand the difference to `nearest_dt`

straxen/plugins/peak_processing.py
```diff
@@ -331,98 +336,393 @@ def find_n_competing(peaks, windows, fraction):
...
+        type=bool,
+        track=True,
+        help='Whether to divide area by time'
+    )
+
+    ambience_divide_r = straxen.URLConfig(
+        default=False,
+        type=bool,
+        track=True,
+        help='Whether to divide area by radius'
```
WenzDaniel, line 572: I think here you have to be a bit more explicit what you mean by radius. I think you mean radial distance of shadow peak to current peak is not it?

straxen/plugins/peak_processing.py
```diff
@@ -331,98 +336,393 @@ def find_n_competing(peaks, windows, fraction):
...
+        touching_windows = strax.touching_windows(lone_hits, roi)
+        # Calculating ambience
+        self.lonehits_ambience(current_peak, 
+                               lone_hits, 
+                               touching_windows, 
+                               num_array, 
+                               sum_array, 
+                               self.config['ambience_divide_t'])
+        result['n_lh_before'] = num_array
+        result['s_lh_before'] = sum_array
```
WenzDaniel, line 630: Can you not pass also directly 
```
result['n_lh_before'] ,
result['s_lh_before'] ,
```
instead of 
```
num_array, 
sum_array, 
```
to your function? Then you do not have to define them and copy the values over.

straxen/plugins/peak_processing.py
```diff
@@ -331,98 +336,393 @@ def find_n_competing(peaks, windows, fraction):
...
+            self.peaks_ambience(current_peak, 
+                                peaks[mask_pre], 
+                                touching_windows, 
+                                radius, 
+                                num_array, 
+                                sum_array, 
+                                self.config['ambience_divide_t'], 
+                                self.config['ambience_divide_r'])
+            result[f'n_s{stype}_before'] = num_array
+            result[f's_s{stype}_before'] = sum_array
```
WenzDaniel, line 650: I think for a future version it would be better to absorb your stype into your `peak_ambience` function. Such that in your peak ambience function you have three buffers for S0, S1, S2 and you are adding the values depending on the type. E.g. sum_array would look like:

```
sum_array = np.zeros((len(peaks), 3))
# adding sum result:
sum_array[p_i, creating_peak['type']] = your_result

```

 In this way you only have to loop over all peaks once instead of 3 times.

WenzDaniel: We would have to pay a bit attention to your different thresholds though.

WenzDaniel (approved):
Thanks a lot @dachengx it looks much better now. I just left some smaller comments and considerations for the future. From my side we can merge.

WenzDaniel:
Ah I just notice that there is a merge conflict. This one has to be resolved first. In addition can you drop maybe some number how long it takes now to compute peak_shadow and ambient for bkg and calibration runs?

straxen/plugins/event_processing.py
```diff
@@ -714,15 +716,15 @@ def compute(self, events):
...
 
             # corrected s2 with s2 xy map only, i.e. no elife correction
             # this is for s2-only events which don't have drift time info
             cs2_top_xycorr = (events[f'{peak_type}s2_area'] * events[f'{peak_type}s2_area_fraction_top'] /
-                                    self.s2_xy_map(s2_positions, map_name=s2_top_map_name))
+                                    self.s2_xy_map(s2_position_shadows, map_name=s2_top_map_name))
             cs2_bottom_xycorr = (events[f'{peak_type}s2_area'] *
                                        (1 - events[f'{peak_type}s2_area_fraction_top']) /
-                                       self.s2_xy_map(s2_positions, map_name=s2_bottom_map_name))
+                                       self.s2_xy_map(s2_position_shadows, map_name=s2_bottom_map_name))
```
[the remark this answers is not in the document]
dachengx: That is my mistake. I will revert it.

straxen/plugins/peak_processing.py
dachengx: There are peaks too small to have a position([codes](https://github.com/XENONnT/straxen/blob/master/straxen/plugins/position_reconstruction.py#L17)).

straxen/plugins/peak_processing.py
dachengx: Yes.

straxen/plugins/peak_processing.py
dachengx: Yes. I will add more descriptions.

straxen/plugins/peak_processing.py
dachengx: The mean signal acceptance of Ambience cut as a function of time window(default 2ms) and radial distance(default 6.7cm) is at [plot](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:ac:prediction:shadow_ambience#img_ambience_time_window_r_220303).

dachengx:
> Ah I just notice that there is a merge conflict. This one has to be resolved first. In addition can you drop maybe some number how long it takes now to compute peak_shadow and ambient for bkg and calibration runs?

straxen/plugins/peak_processing.py
skazama: Thanks Dacheng, looks fine to me now!

skazama (approved):

WenzDaniel:
Thanks @dachengx this sounds reasonable.

PR #915, opened 2022-02-04
reduce warnings during testing

JoranAngevaare:
Currently we have many warnings (267 for py38). This PR tries to reduce those to get somewhat more readable tests.
afbeelding

After
afbeelding


JoranAngevaare:
@WenzDaniel , I'm merging this. I cannot see any issue with it

PR #916, opened 2022-02-09
S1 aft map & CMT tweaks

GiovanniVolta:

Update s1 aft map using CMT global v6. It is time to let CMT to handle this, instead of making a PR every time. Next time I will change it. I promise.
I have also updated the version of the plugin.
The code for making map is available in [nton](https://github.com/XENONnT/nton/tree/s1aft_map_data_driven/nton/analyses/maps) folder

See [here](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:giovo:dd_s1aftmap_straxen_cmt_v6_vs_cmt_v6&do=edit)

  - [] _Tests to check the (new) code is working as desired._

This PR is linked to the [private_nt_aux_files PR](https://github.com/XENONnT/private_nt_aux_files/pull/140).


l-althueser:
Hey maintainer, I would start the process on moving the map information also to CMT. Should we already implement that here or what is the current timescale for the PR?

WenzDaniel:
I am happy with moving things to CMT first. See also slack.

JoranAngevaare:
we could also use a straxen.URLconfig instead of an strax.Option to make the transition trivial between tracking the file directly or in CMT entirely trivial

l-althueser:
We are very close to having the maps in CMT. So we can prepare already to update here.

WenzDaniel:
@GiovanniVolta will you make the update?

jingqiangye:
We need to include this PR in the next straxen release, otherwise the S1AFT related studies won't be updated :D

l-althueser:
Isn't it "just" a few lines of extra code? I have no experience but at least from my feeling .. hm

GiovanniVolta:
Sorry guys, I have been distracted by other business. I have no idea what it is required, but I will discover it and do it before the end of the day.

GiovanniVolta:
Off-topic respect CMT discussion.
@toriaankel, @jpienaar13 please have a look at the last commit. There was a bug in the alt-s1aft cut. This should be affect your study, especially the portion of events with p-value = 0 you observed.

WenzDaniel:
@jmosbacher  could explain @GiovanniVolta how to replace the s1_aft_map config by your new URL config? I am not so familiar with the new syntax myself. Can we use your protocols to replace  L114 following? 
```
self.s1_aft_map = straxen.InterpolatingMap(
            straxen.get_resource(
                self.config['s1_aft_map'],
fmt=self._infer_map_format(self.config['s1_aft_map'])))
```

WenzDaniel:
In case there is not a protocol for the interpolation yet you can use the following syntax as an example: https://github.com/XENONnT/straxen/blob/db53b787b81533558741b25f3897dac0b04050ba/straxen/plugins/event_processing.py#L634-L638 @GiovanniVolta. You would drop the `itp_map://`.

jmosbacher:
@GiovanniVolta , as @WenzDaniel mentioned, the s1_xyz_map config is a good example. So in your case, it would be something like:

```python
s1_aft_map = straxen.URLConfig( 
     default='itp_map://resource://cmt://' 
             's1_aft_xyz_map' 
             '?version=ONLINE&run_id=plugin.run_id&fmt=json', 
     cache=True) 
```

or without using CMT

```python
s1_aft_map = straxen.URLConfig( 
     default='itp_map://resource://' 
             's1_aft_dd_xyz_XENONnT_kr-83m_08Feb2022.json' 
             '?version=ONLINE&run_id=plugin.run_id&fmt=json', 
     cache=True) 
```
then when you want to get the actual map, you just access `self.s1_aft_map` and it will be loaded at runtime and cached. So you can remove the loading of the map from the `setup()` function.

Notice that moving to and from CMT requires no changing of code, you just change the config value when loading the context.. The difference between the two options i gave is just in the default value, not the code itself.

straxen/plugins/event_patternfit.py
```diff
@@ -53,7 +53,7 @@ class EventPatternFit(strax.Plugin):
     
     depends_on = ('event_area_per_channel', 'event_basics', 'event_positions')
     provides = 'event_pattern_fit'
-    __version__ = '0.1.1'
+    __version__ = '0.1.2'
```
jmosbacher, line 57:
```suggestion
    s1_aft_map = straxen.URLConfig( 
           default='itp_map://resource://cmt://' 
                   's1_aft_xyz_map' 
                   '?version=ONLINE&run_id=plugin.run_id&fmt=json', 
           cache=True) 
```

straxen/plugins/event_patternfit.py
```diff
@@ -19,7 +19,7 @@
     strax.Option('s2_tf_model', help='S2 (x, y) optical data-driven model', infer_type=False,
                  default='XENONnT_s2_optical_map_data_driven_ML_v0_2021_11_25.tar.gz'),
     strax.Option('s1_aft_map', help='Date drive S1 area fraction top map.', infer_type=False,
-                 default='s1_aft_dd_xyz_XENONnT_Kr83m_41500eV_31Oct2021.json'),
+                 default='s1_aft_dd_xyz_XENONnT_kr-83m_08Feb2022.json'),
```
jmosbacher, line 22:
```suggestion
```

straxen/plugins/event_patternfit.py
```diff
@@ -19,7 +19,7 @@
     strax.Option('s2_tf_model', help='S2 (x, y) optical data-driven model', infer_type=False,
                  default='XENONnT_s2_optical_map_data_driven_ML_v0_2021_11_25.tar.gz'),
     strax.Option('s1_aft_map', help='Date drive S1 area fraction top map.', infer_type=False,
```
jmosbacher, line 21:
```suggestion
```

straxen/plugins/event_patternfit.py
WenzDaniel: Thanks Yossi, but this suggestions confuses me now :D Since we want to use CMT I think your upper example is correct and therefore we drop this option completely right?

jmosbacher: yeah, the entire "Option()" needs to be removed and replaced by a URLConfig class attribute.  I had to split my suggested change into two suggestions due to githubs limitations on suggesting changes on existing changes.
The map loading also needs to be removed from the `setup()` method.

jmosbacher:
@WenzDaniel did you change the other configs to a URLConfig just for good measure or is there still an issue with missing the old and new styles that needs to be fixed?

jmosbacher:
@WenzDaniel Looks like the `load_corrected_positions` mini analysis also assumes some of the configs are old style CMT tuples. Are you updating that too?

WenzDaniel:
> is there still an issue with mixing the old and new styles that needs to be fixed?

I think there is an issue with mixing old/new style when fetching something with CMT. But lets see first if the tests pass

WenzDaniel:
FFS now it becomes stupid `d: connect ETIMEDOUT 13.107.42.16:443` ....

tests/test_contexts.py
```diff
@@ -139,6 +139,7 @@ def context(*args, **kwargs):
         kwargs.setdefault('cmt_version', 'global_ONLINE')
         return straxen.contexts.xenonnt_simulation(*args, **kwargs)
 
+    @unittest.skipIf(not straxen.utilix_is_configured(), "No db access, cannot test!")
```
JoranAngevaare, line 142: looks good

JoranAngevaare (approved):

WenzDaniel:
Finally... thanks a lot everyone.

PR #917, opened 2022-02-10
Add plugin to compute time offsets per chunk

WenzDaniel:
Plugin which computes time offsets between detectors.


WenzDaniel:
I opened the draft as a reference. I think this plugin should cover all current and future edge-cases for our offsets. It is not nice but should work.

darrylmasson (commented):
Overall looks good, a couple of points worth discussion.

straxen/plugins/acqmon_processing.py
```diff
@@ -216,3 +216,125 @@ def compute(self, events, aqmon_hits):
...
+
+        offsets = []
+        count_valid = 0
+        prev_time = 0
+        for i in range(len(hits_det0)):
+            offset = self.find_offset_nearest(hits_det1['time'], hits_det0['time'][i])
+            time_to_prev = hits_det0['time'][i] - prev_time
+
+            # Additional check to avoid spurious signals
+            if (abs(offset) < coinc_time) & (time_to_prev > 1e10):
```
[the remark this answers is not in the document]
molinario: Yes, I agree with this. I would do like Darryl suggests.

straxen/plugins/acqmon_processing.py
```diff
@@ -216,3 +216,125 @@ def compute(self, events, aqmon_hits):
...
+                offsets.append(offset)
+                count_valid += 1
```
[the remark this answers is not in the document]
molinario: Do you mean the count_valid? We should check if it's >0. If it's not the case (which points to problems in recording the synchronization signal) we should issue an error (it should never happens, but better to be safe).

straxen/plugins/acqmon_processing.py
```diff
@@ -216,3 +216,125 @@ def compute(self, events, aqmon_hits):
         result['time'] = events['time']
         result['endtime'] = events['endtime']
         return result
+
+
+DIGITIZER_DELAY_CHANGE_RUNID = '021286'
+TO_BE_EXCPECTED_MAX_DELAY = 11e3
+
+@strax.takes_config(
+    strax.Option('tpc_internal_delay', default=[4817, 10137], type=int, track=True,
```
JoranAngevaare, line 225: probably worth making an immutabledict?

JoranAngevaare: just to remove the hardcoding required

JoranAngevaare: also worth making a straxen config instead of an option so that we can change it easily to a CMT option if we ever wanted to

straxen/plugins/acqmon_processing.py
```diff
@@ -216,3 +216,125 @@ def compute(self, events, aqmon_hits):
...
+                  'raw_records_aqmon_nv',
+                  'raw_records_aux_mv')
+    provides = 'detector_time_offsets'
+    data_kind = 'detector_time_offsets'
+
+    def infer_dtype(self):
+        dtype = []
+        dtype += strax.time_fields
+        dtype += [((('Time offset for nV to synchornize with TPC'), 'time_offset_nv'), np.int64),
+                  ((('Time offset for mV to synchornize with TPC'), 'time_offset_mv'), np.int64)
```
JoranAngevaare, line 255: `+= [ns]` units in help are always nice

straxen/plugins/acqmon_processing.py
```diff
@@ -216,3 +216,125 @@ def compute(self, events, aqmon_hits):
...
+        rr_nv = raw_records_aqmon_nv
+        rr_mv = raw_records_aux_mv
+
+        extra_offset = 0
+        _mask_tpc = (rr_tpc['channel'] == 798)
+        if not np.any(_mask_tpc):
+            # For some runs in the beginning no signal has been acquired here.
+            # In that case we have to add the internal DAQ delay as an extra offset later.
+            _mask_tpc = (rr_tpc['channel'] == 801)
+            extra_offset = self.config['tpc_internal_delay'][self.run_id >= DIGITIZER_DELAY_CHANGE_RUNID]
```
JoranAngevaare, line 273:
```suggestion
            extra_offset = self.config['tpc_internal_delay'][int(self.run_id) >= int(DIGITIZER_DELAY_CHANGE_RUNID)]
```

JoranAngevaare (commented):
Thanks Daniel and Andrea!

straxen/plugins/acqmon_processing.py
```diff
@@ -216,3 +216,125 @@ def compute(self, events, aqmon_hits):
...
+        result['time'] = hits_tpc['time']
+        result['endtime'] = strax.endtime(hits_tpc)
+        result['time_offset_nv'] = offsets_nv + extra_offset + nveto_extra_offset
+        result['time_offset_mv'] = offsets_mv + extra_offset
+
+        return result
+
+    @staticmethod
+    def get_nim_edge(raw_records, threshold=500):
+        records = strax.raw_to_records(raw_records)
```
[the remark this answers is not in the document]
WenzDaniel: You cannot use the hitfinder otherwise. The data size is so small that computation time is not a problem at all here. Hence, I prefer to use the hitfinder over additional custom code which you would have to maintain.

straxen/plugins/acqmon_processing.py
```diff
@@ -216,3 +216,125 @@ def compute(self, events, aqmon_hits):
...
+
+        offsets_mv = self.estimate_delay(hits_tpc, hits_mv)
+        offsets_nv = self.estimate_delay(hits_tpc, hits_nv)
+        assert len(offsets_mv) == len(offsets_nv), 'Unequal number of sync signals!'
+
+        result = np.zeros(len(offsets_mv), dtype=self.dtype)
+        result['time'] = hits_tpc['time']
+        result['endtime'] = strax.endtime(hits_tpc)
+        result['time_offset_nv'] = offsets_nv + extra_offset + nveto_extra_offset
+        result['time_offset_mv'] = offsets_mv + extra_offset
```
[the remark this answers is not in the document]
WenzDaniel: Will do in https://github.com/XENONnT/straxen/blob/21f053b35910f7764bb1693b012c9e0e61fa46ae/straxen/plugins/veto_events.py#L470 just not yet updated.

straxen/plugins/acqmon_processing.py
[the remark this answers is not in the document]
WenzDaniel: Ah yes sorry, I changed this a bit Andrea. The plugin will compute the offsets for all Clock signals in a run. The averaging can be then done later. Hence, the counter is not needed anymore. In case we do not have any valid signal I just add the error value.

straxen/plugins/acqmon_processing.py
```diff
@@ -216,3 +216,125 @@ def compute(self, events, aqmon_hits):
         result['time'] = events['time']
         result['endtime'] = events['endtime']
         return result
+
+
+DIGITIZER_DELAY_CHANGE_RUNID = '021286'
+TO_BE_EXCPECTED_MAX_DELAY = 11e3
```
darrylmasson, line 222: Also maybe for the sake of those who come after us, this isn't the correct spelling of 'expected'.

straxen/plugins/acqmon_processing.py
WenzDaniel: Unfortunately immutabledicts are not json serializable for what ever reason.  So we can either untrack the option or have to use something different. Maybe not tracking it is fine. What do you think?

straxen/plugins/acqmon_processing.py
```diff
@@ -216,3 +217,138 @@ def compute(self, events, aqmon_hits):
...
+        if not len(hits_nv):
+            # During SR0 sync signal was not recorded properly for the
+            # neutron-veto, hence take waveform itself as "hits".
+            _mask_nveto &= rr_nv['record_i'] == 0
+            nveto_extra_offset = self.config['epsilon_offset']
+            hits_nv = rr_nv[_mask_nveto]
+
+        offsets_mv = self.estimate_delay(hits_tpc, hits_mv)
+        offsets_nv = self.estimate_delay(hits_tpc, hits_nv)
+        assert len(offsets_mv) == len(offsets_nv), 'Unequal number of sync signals!'
```
molinario, line 296: They should be the same length, but how do we manage if this condition is not met? In principle we can calculate the offsets even if mv or nv is missing one of the sync signals for the run.

WenzDaniel: I changed the logic a bit. In case we are missing one signal we fill the corresponding row with the error value. Hence, the two offsets should always have the same length unless there is bug in the code.

WenzDaniel:
Okay I checked it with different runs and things seem to work properly. I only had to exclude the VetoSync plugins from testing since it requires synced test data.

molinario (approved):

PR #919, opened 2022-02-12
Bump version of merged S2s

WenzDaniel:

Fixes bug in merged S2s, see https://github.com/AxFoundation/strax/pull/649.


zhut19 (approved):

tests/test_peaklet_processing.py
```diff
@@ -79,3 +83,45 @@ def test_tight_coincidence(hits, channel):
...
+    Test for different peaklet properties.
+    """
+    highest_data_type = 'peaks'
+
+    def test_area_fraction_top(self):
+        assert self.st.is_stored(nt_test_run_id, 'merged_s2s')
+
+        merged_s2s = self.st.get_array(nt_test_run_id, 'merged_s2s')
+        _area_close_to_area_per_channel = np.isclose(merged_s2s['area']/np.sum(merged_s2s['area_per_channel'], axis=1), 1)
+        assert np.all(_area_close_to_area_per_channel)
```
WenzDaniel, line 127: @JoranAngevaare I added now this base TestDataClass which we can inherit and use in different tests. This should unify things a bit. Since it depends on unittest I was not sure where to put it. Best place might be test_utils, any different opinion?

JoranAngevaare: Right, but this is going to be way to slow if we remove the data each time to be useful since we build it from raw-records

WenzDaniel: But how do we intend to catch any bugs in processing if we are just loading already stored data all the time? Those would be only caught on github.

JoranAngevaare: I don't oppose this, however, it simply would destroy the testing times.

 We could very easily make the distinction on github actions by setting an environment variable on github
`STORE_TEST_OUTCOMES` that would disable the `tearDownClass` method you wrote here.

I did similar things before:
https://github.com/XENONnT/straxen/blob/master/.github/workflows/pytest.yml#L104-L105

PR #923, opened 2022-02-15
Remove data after testing

WenzDaniel:

My achievement for today....

Clears test-data after test.


JoranAngevaare:
although I don't oppose it, we did agree on this method when we discussed this PR:
https://github.com/XENONnT/straxen/pull/703

> Will make github (where a fresh fetch of the data is always done) more reliable than doing tests locally. Locally, you might edit plugin Y, test, try again and fail because you have something stored while working on this that causes the fail in Y. You would have to remove ./strax_test_data in order to prevent another fail. This might be a hurdle for most analysists to recognize and I do foresee someone someday loosing their mind over this. However, I think there are only 2 developers which I know that regularly run pytest, and these to persons will probably recognize.

WenzDaniel:
Sure no problem, but given the issues I had the other day I am now much more in favor of more transparency rather than speed. Especially, also taking into account future generations.

JoranAngevaare:
sure sure 👍

PR #924, opened 2022-02-15
Add Bayes peak probability

ahiguera-mx:
Add Bayes peak probability
Bayes peak probability computation
Bayes peak probability computation and propagate this information up to event level 

### _Notes on testing_


WenzDaniel (commented):
Here are already a few comments. It would be helpful if you could break down your monolithic computes code into a few smaller functions. This will increase readability and makes the code more transparent. In addition I think it would be good to get some unittests for these smaller functions.

straxen/plugins/bayes_peaklet_classification.py
```diff
@@ -0,0 +1,132 @@
...
+from scipy.special import logsumexp
+import strax
+import straxen
+
+class BayesPeakletClassification(strax.Plugin):
+    """
+    Bayes Peaklet classification
+    """
+    provides = 'peaklet_classification_bayes'
+    depends_on = ('peaklets',)
```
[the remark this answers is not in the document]
ahiguera-mx: This would imply retraining, and for most of the cases we focus on s1_prob

straxen/plugins/bayes_peaklet_classification.py
```diff
@@ -0,0 +1,132 @@
...
+        help='Number of nodes'
+    )
+    classes = straxen.URLConfig(
+        default=2,
+        help='Number of classes'
+    )
+
+    def setup(self):
+
+        self.class_prior = np.array([1. / self.classes for j in range(self.classes)])
```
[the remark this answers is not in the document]
ahiguera-mx: fixed

straxen/plugins/bayes_peaklet_classification.py
```diff
@@ -0,0 +1,132 @@
...
+        default=np.log(0.5), #best-fit value from optimization of ROC curve
+        help='S2 log prob value threshold, above this value type=2'
+    )
+    num_nodes = straxen.URLConfig(
+        default=50,
+        help='Number of nodes'
+    )
+    classes = straxen.URLConfig(
+        default=2,
+        help='Number of classes'
```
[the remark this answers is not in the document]
ahiguera-mx: fixed

straxen/plugins/bayes_peaklet_classification.py
```diff
@@ -0,0 +1,132 @@
...
+        bayes_ptype = class_assignments
+        '''
+
+        return dict(time=peaklets['time'],
+                    dt=peaklets['dt'],
+                    channel=-1,
+                    length=peaklets['length'],
+                    s1_ln_prob=s1_ln_prob,
+                    s2_ln_prob=s2_ln_prob
+                    )
```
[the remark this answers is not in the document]
ahiguera-mx: we are not doing class assignments in the plugin in, good catch, fixed

straxen/plugins/bayes_peaklet_classification.py
```diff
@@ -0,0 +1,132 @@
...
+        
+        class_assignments = np.zeros(len(peaklets))
+        C_S1 = s2_ln_prob < self.s2_prob_threshold
+        C_S2 = s2_ln_prob >= self.s2_prob_threshold
+
+        class_assignments[C_S1] = 1
+        class_assignments[C_S2] = 2
+        bayes_ptype = np.zeros(len(peaklets), dtype=np.int8)
+        bayes_ptype = class_assignments
+        '''
```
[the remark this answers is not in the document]
ahiguera-mx: fixed

ahiguera-mx:
@WenzDaniel @sophiafarrell 
Thanks for the very useful comments, I've implemented all of them. More importantly, I've added unit test for the plugin. 
Hopefully this can go into the next release so analysts can start use it. 
As you may notices from the pytest this PR is from a fork... yes I should have done a branch from master... sorry about that

straxen/plugins/bayes_peaklet_classification.py
```diff
@@ -0,0 +1,141 @@
...
+        help='Bayes configuration file, conditional probabilities tables and Bayes discrete bins'
+    )
+    num_nodes = straxen.URLConfig(
+        default=50,
+        help='Number of attributes(features) per waveform and quantile'
+    )
+    classes = straxen.URLConfig(
+        default=2,
+        help='Number of label classes S1(1)/S2(2)'
+    )
```
JoranAngevaare, line 40: Maybe keep in mind that these config names like `'classes'` are visible to all other plugins, as such, I would advice making it a bit more explicit: `classes->n_bayes_classes / bayes_peak_types`, `bayes_nodes` or something like it?

straxen/plugins/bayes_peaklet_classification.py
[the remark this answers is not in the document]
JoranAngevaare: Ah I was about to make the same suggestion as Daniel. There is one extra benefit and that would be that it would be much faster since we are running on much fewer Sx since the S2s are split over many peaklets by design

straxen/plugins/bayes_peaklet_classification.py
```diff
@@ -0,0 +1,141 @@
...
+
+    :param peaklets: peaklets
+    :param waveforms: peaklets waveforms in PE/ns
+    :param quantiles: quantiles in ns
+    :returns: the ln probability of a each peaklet belonging to S1 and S2 class
+    """
+    provides = 'peaklet_classification_bayes'
+    depends_on = ('peaklets',)
+    __version__ = '0.0.1'
+    dtype = (strax.peak_interval_dtype
```
JoranAngevaare, line 22: maybe the strax.time_fields is more natural here?

straxen/plugins/bayes_peaklet_classification.py
```diff
@@ -0,0 +1,141 @@
...
+
+    data = peaklets['data'].copy()
+    dts = peaklets['dt'].copy()
+    data[data < 0.0] = 0.0
+    for i, p in enumerate(peaklets):
+        sample_number = np.arange(0, num_samples+1, 1)*p['dt']
+        frac_of_cumsum = np.append([0.0], np.cumsum(data[i, :]) / np.sum(data[i, :]))
+        cumsum_steps = np.interp(np.linspace(0., 1., num_nodes, endpoint=False), frac_of_cumsum, sample_number)
+        cumsum_steps = np.append(cumsum_steps, sample_number[-1])
+        quantiles[i, :] = cumsum_steps[1:] - cumsum_steps[:-1]
```
JoranAngevaare, line 87: This looks like a very intense operation with a for loop which is potentially quite slow, have you checked the performance of this plugin?

straxen/plugins/event_w_bayes_class.py
```diff
@@ -0,0 +1,55 @@
...
+        s1 = np.zeros(len(events), strax.time_fields)
+        s2 = np.zeros(len(events), strax.time_fields)
+        s1_alt = np.zeros(len(events), strax.time_fields)
+        s2_alt = np.zeros(len(events), strax.time_fields)
+
+        # For S1 and S2, match to peak posteriors
+        for peaks, name in zip([s1, s2, s1_alt, s2_alt], ['s1', 's2', 'alt_s1', 'alt_s2']):
+            peaks['time'] = events[f'{name}_time']
+            peaks['endtime'] = events[f'{name}_endtime']
+            fci = strax.fully_contained_in(peaklets, peaks)
```
JoranAngevaare, line 39: If I remember correctly, you can't use this function as peaklets aren't sorted by time since an S1 can happen during an S2

straxen/plugins/event_w_bayes_class.py
```diff
@@ -0,0 +1,55 @@
...
+            # -1 if no contained, see https://github.com/AxFoundation/strax/blob/109a9da0194dcf5b99992e16ba1bf598874c7e7c/strax/processing/general.py#L105
+            mask = fci >= 0
+            fci = fci[mask]
+            result[f'{name}_s1_ln_prob'][fci] = peaklets[mask]['s1_ln_prob']
+            result[f'{name}_s2_ln_prob'][fci] = peaklets[mask]['s2_ln_prob']
```
JoranAngevaare, line 44: I'm confused here, multiple peaklets an constitute a single peak. I don't understand how this is supposed to work

straxen/plugins/event_w_bayes_class.py
```diff
@@ -0,0 +1,55 @@
...
+            result['time'] = events['time']
+            result['endtime'] = events['endtime']
+
+        # events can be made out of one peak, this is to enusure user should not look at prob
+        for name in (['s1', 'alt_s1', 'alt_s2']):
+            no_s_peak = np.where(events[f'{name}_index'] == -1)
+            if no_s_peak:
+                result[f'{name}_s1_ln_prob'][no_s_peak] = np.nan
+                result[f'{name}_s2_ln_prob'][no_s_peak] = np.nan
```
JoranAngevaare, line 53: might be better to set everything to nan beforehand (top of this function) so you don't have to do an extra search where things are filled

straxen/plugins/event_w_bayes_class.py
```diff
@@ -0,0 +1,55 @@
...
+
+        dtype = []
+        dtype += strax.time_fields
+        for name in ['s1', 's2', 'alt_s1', 'alt_s2']:
+            dtype += [(f'{name}_s1_ln_prob', np.float32, f'Given an {name}, s1 ln probability')]
+            dtype += [(f'{name}_s2_ln_prob', np.float32, f'Given an {name}, s2 ln probability')]
+
+        return dtype
+
+    def compute(self, peaklets, events):
```
JoranAngevaare, line 26: If you were to do this on peaks (see earlier comment), it would be much faster since you could just select peaks on their start and endtime

JoranAngevaare (commented):
Thanks guys, the code looks carefully written and nicely documented etc.

Some things confuse me a bit as I will explain below. Additionally, I am wondering about the performance of this plugin. Did you check if there are any bottlenecks?

straxen/plugins/bayes_peaklet_classification.py
ahiguera-mx: good point, fixed

straxen/plugins/bayes_peaklet_classification.py
ahiguera-mx: fixed

straxen/plugins/bayes_peaklet_classification.py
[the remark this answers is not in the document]
ahiguera-mx: Right, since we are doing classification we decided to switch to peak as you guys suggested

straxen/plugins/event_w_bayes_class.py
ahiguera-mx: n/a

straxen/plugins/event_w_bayes_class.py
ahiguera-mx: done

straxen/plugins/bayes_peaklet_classification.py
ahiguera-mx: Here are a few numbers on Dali
for an  average background run `CPU times: user 1min 15s, sys: 15.6 s, total: 1min 31s` 
for an average Rd run `CPU times: user 4min 59s, sys: 57.6 s, total: 5min 57s`
for an average Kr run `CPU times: user 8min 22s, sys: 1min 36s, total: 9min 59s`
which includes loading peaks

straxen/plugins/event_w_bayes_class.py
ahiguera-mx: I tried but since results has a predefined type 
`result = np.empty(len(events), dtype=self.dtype)`
if I set to nan I got `ValueError: cannot convert float NaN to integer` 
any suggestion here?

straxen/plugins/event_w_bayes_class.py
ahiguera-mx: n/a we switch to peaks

ahiguera-mx:
Thanks @JoranAngevaare for very useful comments, I've addressed.

straxen/plugins/event_w_bayes_class.py
JoranAngevaare: Ah, no so what I meant was setting this all the way at the top of this `compute` method:

```
for name in (['s1', 'alt_s1', 'alt_s2']):
    result[f'{name}_s1_ln_prob'] = np.nan
    result[f'{name}_s2_ln_prob'] = np.nan
``` 

The ones where you will find a peak for will get filled by your logic.

straxen/plugins/bayes_peaklet_classification.py
JoranAngevaare: comparing to run 12236, this seems much longer than any other plugin we had at that time I wrote this notebook:
https://github.com/XENONnT/workshop_zoom_2021_04_19/blob/master/1_xenonnt_data/1_Data_Structure_and_processing_tutorial.ipynb
afbeelding

This being said, the times may differ a lot from run to run, so I don't know if the max ~350 s (total time) compares to the 600s wall-time you report here.

My suspicion is that this for-loop is probably a bottle neck. I  personally very much like [line-profiling ](https://pypi.org/project/line-profiler/) a lot in order to find bottlenecks and focus optimization on the parts of the code that are actually worth optimizing.

Let me know if you need some examples of how I would try and optimize for performance. A few general hints would concerning this block of code would be:
 - Figure out which part of the code takes the longest, if you spend a lot of time speeding something up that wasn't slow, you don't gain much (only because I've done this often myself, and it's a very tempting pitfall 😉 )
 - avoid for loops (unless you can do it in a numbafied function) or the numer of iterations is small
 - avoid `np.append`, every time you call that, you instruct numpy to allocate an entirely new block of memory which you then have to fill with the old stuff array

I can write you an example if you want, probably also useful for other people if @WenzDaniel or me are asking again and again if something is "fast" or "slow"

straxen/plugins/event_w_bayes_class.py
```diff
@@ -0,0 +1,48 @@
...
+    def compute(self, peaks, events):
+
+        result = np.empty(len(events), dtype=self.dtype) 
+
+        for name in ['s1', 's2', 'alt_s1', 'alt_s2']:
+            container = np.zeros(len(events), strax.time_fields)
+            container['time'] = events[f'{name}_time']
+            container['endtime'] = events[f'{name}_endtime']
+
+            mask = strax.touching_windows(peaks, container)
```
JoranAngevaare, line 35:
```suggestion
        # Select peaks based on their start time
        mask = np.in1d(peaks['time'], events[f'{name}_time'])
```

straxen/plugins/event_w_bayes_class.py
```diff
@@ -0,0 +1,48 @@
...
+            result[f'{name}_ln_prob_s1'] = peaks['ln_prob_s1'][mask[:,0]]
+            result[f'{name}_ln_prob_s2'] = peaks['ln_prob_s2'][mask[:,0]]
+            result['time'] =  peaks[mask[:,0]]['time']
+            result['endtime'] =  peaks[mask[:,0]]['endtime']
```
JoranAngevaare, line 39:
```suggestion
            result[f'{name}_ln_prob_s2'] = peaks['ln_prob_s2'][mask[:,0]]
        result['time'] =  events['time']
        result['endtime'] =  events['endtime']
```

straxen/plugins/bayes_peaklet_classification.py
ahiguera-mx: Hi @JoranAngevaare 
Good point, at the moment this plugin is at the level of `event_pattern_fit` in terms of slowness. We already have an idea on how to remove `np.append` for the reason you mentioned, but we would need to retrain. 
If the priority is to have this ready so analyst can use it, I think we can go ahead and meanwhile we work on the optimization. Also I would like to get other people opinion @tunnell @tinapeters 

Btw, once I tried to run sneakviz  to check this but I end up lost looking at the profile because all straxen calls , do you have an example for dummies on how to do this for a single straxen plugin

ahiguera-mx: When running this plugin ~20% of the time is just loading peaks, I will try to debug more and see if can speed up 
```
Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
    17                                           @profile
    18                                           def _compute_bayes(st):
    19                                           
    20         1      45984.0  45984.0      0.0      p = st.get_single_plugin('031711', 'peak_classification_bayes')
    21                                           
    22         1   21889439.0 21889439.0     21.0      peaks = st.get_array('031711', 'peaks')
    23                                               
    24         1   82340242.0 82340242.0     79.0      result =p.compute(peaks)

```

tunnell:
@JoranAngevaare will the speed things block you merging it?  Tina will work on it, but we really need to get it merged if people are going to test it more broadly.  Can we parallelize the work such that analysts can test this while Tina looks at the speed such that we don't make some last minute optimization mistake?

Also, what blocks the merge at this point?  Discussion hard to follow.  Do we know when the next release is?

l-althueser:
One thing I just noticed and missed before. We added the config files to CMT and latest global. We might want to consider using it here, right? If that's the case the default should be a cmt option to avoid confusion. If not then we should make it clear once more for me :)

ahiguera-mx:
> One thing I just noticed and missed before. We added the config files to CMT and latest global. We might want to consider using it here, right? If that's the case the default should be a cmt option to avoid confusion. If not then we should make it clear once more for me :)

Done, I change the config option to CMT, once is ready to merge I will add Bayes model to ONLINE

WenzDaniel:
> Can we parallelize the work such that analysts can test this while Tina looks at the speed such that we don't make some last minute optimization mistake?

Not a fan of it, but in the end it wont be a hard requirement, I think. The problem is, that we heard this promise already several times from different people, and in the end there are always the same 3-4 for people making the work and required changes. So we made a lot of bad experience here. 

> Also, what blocks the merge at this point? Discussion hard to follow. Do we know when the next release is?

I think the first requirement is that all tests pass which is not the case at the moment. Then I still have to review the PR. So far I have not had any time yet. At the moment we are preparing a release in which wont include this PR yet. There is another patch release planned by the end of this week, but I am not sure if this is a realistic time line. My feeling it will be mid next week, We are currently, fighting with CMT and the implementation of one of our new corrections. I spent yesterday the entire afternoon and evening to make it work in straxen.

straxen/plugins/bayes_peaklet_classification.py
```diff
@@ -0,0 +1,138 @@
...
+class BayesPeakletClassification(strax.Plugin):
+    """
+    Bayes Peaklet classification
+    Returns the ln probability of a each event belonging to the S1 and S2 class.
+    Uses conditional probabilities and data parameterization learned from wfsim data.
+
+    :param peaks: peaks
+    :param waveforms: peaks waveforms in PE/ns
+    :param quantiles: quantiles in ns
+    :returns: the ln probability of a each peaklet belonging to S1 and S2 class
```
WenzDaniel, line 17: This looks a bit out of place as your class does not take those arguments.

straxen/plugins/bayes_peaklet_classification.py
```diff
@@ -0,0 +1,138 @@
...
+    n_bayes_classes = straxen.URLConfig(
+        default=2,
+        help='Number of label classes S1(1)/S2(2)'
+    )
+
+    def setup(self):
+
+        self.class_prior = np.ones(self.n_bayes_classes)/self.n_bayes_classes
+        self.bins = self.bayes_config_file['bins']
+        self.cpt = self.bayes_config_file['cprob']
```
WenzDaniel, line 47: A comment about those would be good.

straxen/plugins/bayes_peaklet_classification.py
```diff
@@ -0,0 +1,138 @@
...
+import strax
+import straxen
+
+
+class BayesPeakletClassification(strax.Plugin):
+    """
+    Bayes Peaklet classification
+    Returns the ln probability of a each event belonging to the S1 and S2 class.
+    Uses conditional probabilities and data parameterization learned from wfsim data.
+
```
WenzDaniel, line 13: Can you maybe add also the reference to a note here since this is something new.

straxen/plugins/bayes_peaklet_classification.py
```diff
@@ -0,0 +1,138 @@
...
+                                                   self.n_bayes_classes, self.class_prior,
+                                                   waveforms, quantiles)
+
+        return dict(time=peaks['time'],
+                    endtime=peaks['time'] + peaks['dt'] * peaks['length'],
+                    ln_prob_s1=ln_prob_s1,
+                    ln_prob_s2=ln_prob_s2
+                    )
+
+def compute_wf_and_quantiles(peaks: np.ndarray, bayes_n_nodes: int):
```
WenzDaniel, line 63: I think this function can be numbafied easily. Within your for-loop your can use a fixed length buffer instead of np.append as the length of your buffer is determined by the number of quartiles  and peaks is not it? The other thing which your have to change is `np.linspace(0., 1., bayes_n_nodes, endpoint=False)` into `np.linspace(0., 1., bayes_n_nodes)[:-1]` (assuming this is what endpoint false does).

straxen/plugins/bayes_peaklet_classification.py
```diff
@@ -0,0 +1,138 @@
...
+    """
+    waveforms = np.zeros((len(peaks), bayes_n_nodes))
+    quantiles = np.zeros((len(peaks), bayes_n_nodes))
+
+    num_samples = peaks['data'].shape[1]
+    step_size = int(num_samples/bayes_n_nodes)
+    steps = np.arange(0, num_samples+1, step_size)
+
+    data = peaks['data'].copy()
+    dts = peaks['dt'].copy()
```
WenzDaniel, line 78: This line is not necessary is it?

straxen/plugins/bayes_peaklet_classification.py
```diff
@@ -0,0 +1,138 @@
...
+    data[data < 0.0] = 0.0
+    for i, p in enumerate(peaks):
+        sample_number = np.arange(0, num_samples+1, 1)*p['dt']
+        frac_of_cumsum = np.append([0.0], np.cumsum(data[i, :]) / np.sum(data[i, :]))
```
WenzDaniel, line 82: Why do you need this leading zero?

straxen/plugins/bayes_peaklet_classification.py
```diff
@@ -0,0 +1,138 @@
...
+        cumsum_steps = np.interp(np.linspace(0., 1., bayes_n_nodes, endpoint=False), frac_of_cumsum, sample_number)
+        cumsum_steps = np.append(cumsum_steps, sample_number[-1])
+        quantiles[i, :] = cumsum_steps[1:] - cumsum_steps[:-1]
+
+    for j in range(bayes_n_nodes):
+        waveforms[:, j] = np.sum(data[:, steps[j]:steps[j+1]], axis=1)
+    waveforms = waveforms/(dts*step_size)[:, np.newaxis]
```
WenzDaniel, line 89: Can you explain and maybe add a comment what you are doing here. It is not immediately obvious to me.

straxen/plugins/bayes_peaklet_classification.py
```diff
@@ -0,0 +1,138 @@
...
+    """
+    Bin the waveforms and quantiles according to Bayes bins and compute inference
+    :param bins: Bayes bins
+    :param bayes_n_nodes: number of nodes or atributes
+    :param cpt: conditioanl probability tables
+    :param n_bayes_classes: number of classes
+    :param class_prior: class_prior
+    :param waveforms: waveforms
+    :param quantiles: quantiles
+    :return: ln probability per class S1/S2
```
WenzDaniel, line 105: Can you please add more information. Explaining e.g. a `car` with the word `car` is not really helpful.

straxen/plugins/bayes_peaklet_classification.py
```diff
@@ -0,0 +1,138 @@
...
+                    ln_prob_s1=ln_prob_s1,
+                    ln_prob_s2=ln_prob_s2
+                    )
+
+def compute_wf_and_quantiles(peaks: np.ndarray, bayes_n_nodes: int):
+    """
+    Compute waveforms and quantiles for a given number of nodes(atributes)
+    :param peaks:
+    :param bayes_n_nodes: number of nodes or atributes
+    :return: waveforms and quantiles
```
WenzDaniel, line 68: The waveforms are the down sampled versions of the original waveform right?

straxen/plugins/bayes_peaklet_classification.py
```diff
@@ -0,0 +1,138 @@
...
+    quantile_bin_edges = bins[1, :][bins[1, :] > -1]
+    quantile_num_bin_edges = len(quantile_bin_edges)
+
+    waveform_values = np.digitize(waveforms, bins=waveform_bin_edges)-1
+    waveform_values[waveform_values < 0] = int(0)
+    waveform_values[waveform_values > int(waveform_num_bin_edges - 2)] = int(waveform_num_bin_edges - 2)
+
+    quantile_values = np.digitize(quantiles, bins=quantile_bin_edges)-1
+    quantile_values[quantile_values < 0] = int(0)
+    quantile_values[quantile_values > int(quantile_num_bin_edges - 2)] = int(quantile_num_bin_edges - 2)
```
WenzDaniel, line 119: Can you explain what you are doing here I cannot follow. Also having the suffix `_values` is a bit miss leading. If I understand np.digitize correctly it will return indices not the waveform values.

straxen/plugins/event_info.py
```diff
@@ -23,11 +23,12 @@ class EventInfo(strax.MergeOnlyPlugin):
                   'event_positions',
                   'corrected_areas',
                   'energy_estimates',
+                  'event_w_bayes_class',
                   # 'event_pattern_fit', <- this will be added soon
                   ]
     save_when = strax.SaveWhen.ALWAYS
     provides = 'event_info'
-    __version__ = '0.0.2'
+    __version__ = '0.0.3'
```
WenzDaniel, line 31: Can we exclude it from event_info like we do for `event_pattern_fit`. Then we can process it fairly independently from the main data and hence performance is less of an issue if things arrive late.

straxen/plugins/event_w_bayes_class.py
```diff
@@ -0,0 +1,42 @@
...
+
+        for name in ['s1', 's2', 'alt_s1', 'alt_s2']:
+        
+            result[f'{name}_ln_prob_s1'] = np.nan
+            result[f'{name}_ln_prob_s2'] = np.nan
+            # Select peaks based on their start time
+            mask = np.in1d(peaks['time'], events[f'{name}_time'])
+            mask_ev = np.in1d(events[f'{name}_time'], peaks['time'])
+            result[f'{name}_ln_prob_s1'][mask_ev] = peaks['ln_prob_s1'][mask]
+            result[f'{name}_ln_prob_s2'][mask_ev] = peaks['ln_prob_s2'][mask]
```
WenzDaniel, line 38: You could use also our strax-methods like `strax.fully_contained_in` and `s1_index` etc. Not sure if it will be faster,

tests/test_bayes_plugin.py
```diff
@@ -0,0 +1,37 @@
...
+
+    def test_inference(self):
+        """
+        Test inference, ln prob <= 0
+        """
+        self.st.make(self.run_id, 'peaks')
+        self.st.make(self.run_id, self.target)
+        bayes = self.st.get_array(self.run_id, self.target)
+        prob = np.where(bayes['ln_prob_s1'] > 0.0)
+        assert bayes[prob].size == 0, "ln prob gratter than zero, impossible"
```
WenzDaniel, line 37: `assert np.any(bayes['ln_prob_s1'] > 0)` does the same and is shorter.

WenzDaniel (commented):
I think it looks mostly good. Here some comments.

ahiguera-mx:
move this to https://github.com/XENONnT/straxen/pull/954

tests/test_bayes_plugin.py
ahiguera-mx: done

straxen/plugins/event_w_bayes_class.py
ahiguera-mx: `strax.fully_contained_in` is not an option since and S1 can be contained within an S2

straxen/plugins/event_info.py
ahiguera-mx: I will leave this up to you

straxen/plugins/bayes_peaklet_classification.py
ahiguera-mx: yes

straxen/plugins/bayes_peaklet_classification.py
ahiguera-mx: fixed

straxen/plugins/bayes_peaklet_classification.py
ahiguera-mx: those parameters are explained in the URL config help

straxen/plugins/bayes_peaklet_classification.py
ahiguera-mx: I did an attempt to get rid of the np.append following your suggestion 
```
# create buffer to avoid np.append
frac_of_cumsum = np.empty(num_samples+1)
frac_of_cumsum[0] = 0.0
frac_of_cumsum[1:] = np.cumsum(data[i, :]) / np.sum(data[i, :])
# create buffer to avoid np.append
cumsum_steps = np.empty(bayes_n_nodes+1)
cumsum_steps[0:bayes_n_nodes] = np.interp(np.linspace(0., 1., bayes_n_nodes, endpoint=False), frac_of_cumsum, sample_number)
 cumsum_steps[-1:] = sample_number[-1]
quantiles[i, :] = cumsum_steps[1:] - cumsum_steps[:-1]
```

Locally passed all tests, however did not pass test on GitHub

WenzDaniel: You mean with njit decorator? I think it cannot be numbafied due to `, endpoint=False` in linespace. You can check numbas very excellent [reference manual](https://numba.readthedocs.io/en/stable/reference/numpysupported.html).

ahiguera-mx: My first attempt was to remove the np.append(), once that is working I can try to numbafied

PR #925, opened 2022-02-16
Test nveto alt baseline

JoranAngevaare:
Not a substitute for a real test but better than nothing


WenzDaniel:
This wont be necessary. I will remove this option in the near future. We only implemented as we had some issues with our digitizer in the beginning and it was not clear if we can fix those.

JoranAngevaare:
doesn't hurt either though, I think you have been wanting to remove it for quite a while right

WenzDaniel:
See https://github.com/XENONnT/straxen/pull/927

PR #927, opened 2022-02-16
Remove median baseline from nveto processing

WenzDaniel:

Removes median baseline from nveto processing. This will change the lineage. Should not be a problem since we are getting a new gain model anyhow.


JoranAngevaare:
This does change the lineage though

WenzDaniel:
Yes, but we get a new gain model anyhow. So we have to reprocess anyhow.

JoranAngevaare (approved):

Issue #62, opened 2020-03-02
Parse artificial deadtime in channel 799


JelleAalbers:
See #60

jorana:
@AlexElykov , would this easy to add in the plugins from https://github.com/XENONnT/straxen/pull/207?

AlexElykov:
@jorana I'll check.

JoranAngevaare:
This issue is fixed at last 🎉 #929

PR #929, opened 2022-02-18
Rewrite aqmon processing

JoranAngevaare:
- Fix issues in the processing of aqmon data.
  - Veto intervals weren't used but should be used for veto proximity
  - Hitfinder thresholds were not set correctly for logic signals
  - Don't allow random matching. I've made the logic strict to only allow* an even number of starts and stops, preventing us from making the same mistake again. *With the exception of us loosing a veto ON before starting the run, or an OFF after completing the run.
- Fix artificial deadtime parsing #62

https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:j_angevaare:busy_logic_straxen

If we ever want to consider the left/right side of an event w.r.t a veto as something other than the start/endtime of an event, we can do so easily:
```python
st.set_config({'event_window_fields': ('s1_time', 's2_endtime')})
```

### DIFF
--- a/straxen/plugins/acqmon_processing.py
+++ b/straxen/plugins/acqmon_processing.py
@@ -49,7 +49,8 @@ class AqmonHits(strax.Plugin):
                 AqmonChannels.HE_START,
                 AqmonChannels.BUSY_STOP,
                 AqmonChannels.BUSY_START,)),
-            # Fake signals
+            # Fake signals, 0 meaning that we won't find hits using
+            # strax but just look for starts and stops
             (0, (AqmonChannels.ARTIFICIAL_DEADTIME,)),
 
         ),
--- a/straxen/plugins/acqmon_processing.py
+++ b/straxen/plugins/acqmon_processing.py
@@ -466,7 +466,7 @@ class DetectorSynchronization(strax.Plugin):
         track=True,
         help='Internal delay between aqmon and regular TPC channels ins [ns]'
     )
-    adc_threshold_nim_signal= straxen.URLConfig(
+    adc_threshold_nim_signal = straxen.URLConfig(
         default=500,
         type=int,
         track=True,
--- a/straxen/plugins/acqmon_processing.py
+++ b/straxen/plugins/acqmon_processing.py
@@ -1,10 +1,11 @@
-import strax
-import straxen
+from enum import IntEnum
+
 import numba
 import numpy as np
-from enum import IntEnum
-from .daqreader import ARTIFICIAL_DEADTIME_CHANNEL
+import strax
+import straxen
 
+from .daqreader import ARTIFICIAL_DEADTIME_CHANNEL
 
 export, __all__ = strax.exporter()
 
@@ -141,7 +142,7 @@ class VetoIntervals(strax.OverlapWindowPlugin):
         straxen_deadtime <= special case of deadtime introduced by the
             DAQReader-plugin
     """
-    __version__ = 'test_2'
+    __version__ = '1.0.0'
     depends_on = 'aqmon_hits'
     provides = 'veto_intervals'
     data_kind = 'veto_intervals'
@@ -293,7 +294,7 @@ class VetoProximity(strax.OverlapWindowPlugin):
         result_buffer[f'time_to_previous_{veto_name}'] = T_NO_VETO_FOUND
         result_buffer[f'time_to_next_{veto_name}'] = T_NO_VETO_FOUND
 
-        selected_intervals = veto_intervals[veto_intervals['veto_type']==f'{veto_name}_veto']
+        selected_intervals = veto_intervals[veto_intervals['veto_type'] == f'{veto_name}_veto']
         print(len(selected_intervals), veto_name)
         if not len(selected_intervals):
             return
--- a/straxen/plugins/acqmon_processing.py
+++ b/straxen/plugins/acqmon_processing.py
@@ -473,7 +473,6 @@ class DetectorSynchronization(strax.Plugin):
         help='Threshold in [adc] to search for the NIM signal'
     )
     epsilon_offset = straxen.URLConfig(
-        'epsilon_offset',
         default=0,
         type=(int, float),
         track=True,
--- a/straxen/plugins/acqmon_processing.py
+++ b/straxen/plugins/acqmon_processing.py
@@ -146,7 +146,7 @@ class VetoIntervals(strax.OverlapWindowPlugin):
         straxen_deadtime <= special case of deadtime introduced by the
             DAQReader-plugin
     """
-    __version__ = '1.0.1'
+    __version__ = '1.1.0'
     depends_on = 'aqmon_hits'
     provides = 'veto_intervals'
     data_kind = 'veto_intervals'
--- a/straxen/plugins/acqmon_processing.py
+++ b/straxen/plugins/acqmon_processing.py
@@ -321,8 +321,9 @@ class VetoProximity(strax.OverlapWindowPlugin):
         default=int(3.6e+12),
         track=True,
         type=int,
-        help='If we don\'t find a veto close to the event, say that '
-             'the closest event is this many ns removed from it.'
+        help='If no next/previous veto is found, we will fill the fields '
+             'time_to_previous_XX with this time. Set to a large number '
+             'such that one will never cut events that are < YY ns.'
     )
 
     veto_names = ['busy', 'busy_he', 'hev', 'straxen_deadtime']
--- a/straxen/plugins/acqmon_processing.py
+++ b/straxen/plugins/acqmon_processing.py
@@ -261,18 +261,14 @@ class VetoIntervals(strax.OverlapWindowPlugin):
             veto_hits_start = np.concatenate([extra_start, veto_hits_start])
 
         something_is_wrong = len(veto_hits_start) != len(veto_hits_stop)
+
+        message = (f'Got inconsistent number of {veto_name} starts '
+                   f'{len(veto_hits_start)}) / stops ({len(veto_hits_stop)}).')
+        if len(extra_start):
+            message += ' Despite the fact that we inserted one extra start at the beginning of the run.'  # noqa
+        elif len(extra_stop):
+            message += ' Despite the fact that we inserted one extra stop at the end of the run.'  # noqa
         if something_is_wrong:
-            message = (f'Got inconsistent number of {veto_name} '
-                       f'starts ({len(veto_hits_start)}) / '
-                       f'stops ({len(veto_hits_stop)}).')
-            if len(extra_start):
-                message += (' Despite the fact that we inserted one extra '
-                            'start at the beginning of the run.')
-            elif len(extra_stop):
-                message += (' Despite the fact that we inserted one extra '
-                            'stop at the end of the run.')
-            if len(extra_start) and len(extra_stop):
-                raise RuntimeError('Someone broke my code, this cannot happen?!')
             raise ValueError(message)
 
         if np.any(veto_hits_start['time'] > veto_hits_stop['time']):
--- a/straxen/plugins/acqmon_processing.py
+++ b/straxen/plugins/acqmon_processing.py
@@ -34,7 +34,7 @@ class AqmonHits(strax.Plugin):
     GPS SYNC analysis, etc.
     """
     save_when = strax.SaveWhen.TARGET
-    __version__ = '1.0.0'
+    __version__ = '1.0.1'
     hit_min_amplitude_aqmon = straxen.URLConfig(
         default=(
             # Analogue signals
@@ -71,14 +71,10 @@ class AqmonHits(strax.Plugin):
     dtype = strax.hit_dtype
 
     def compute(self, raw_records_aqmon):
-        allowed_channels = [channel for hit_and_channel_list
-                            in self.hit_min_amplitude_aqmon for
-                            channel in hit_and_channel_list[1]]
-
         not_allowed_channels = (set(np.unique(raw_records_aqmon['channel']))
-                                - set(allowed_channels))
+                                - set(self.aqmon_channels))
         if not_allowed_channels:
-            raise ValueError(f'Unknown channel {not_allowed_channels} {allowed_channels}')
+            raise ValueError(f'Unknown channel {not_allowed_channels}. Only know {self.aqmon_channels}')
         records = strax.raw_to_records(raw_records_aqmon)
         strax.sort_by_time(records)
         strax.zero_out_of_bounds(records)
@@ -86,21 +82,26 @@ class AqmonHits(strax.Plugin):
         aqmon_hits = self.find_aqmon_hits_per_channel(records)
         return aqmon_hits
 
+    @property
+    def aqmon_channels(self):
+        return [channel for hit_and_channel_list in self.hit_min_amplitude_aqmon
+                for channel in hit_and_channel_list[1]]
+
     def find_aqmon_hits_per_channel(self, records):
         """Allow different thresholds to be applied to different channels"""
-        aqmon_hits = [
-            strax.find_hits(
-                records[np.in1d(records['channel'], channels)],
-                min_amplitude=hit_threshold
-            )
-            for hit_threshold, channels in self.hit_min_amplitude_aqmon
-            if hit_threshold
-        ]
-        artificial_deadtime = records[records['channel'] == AqmonChannels.ARTIFICIAL_DEADTIME]
-        if len(artificial_deadtime):
-            aqmon_hits += [self.get_deadtime_hits(artificial_deadtime)]
-        aqmon_hits = np.concatenate(aqmon_hits)
-        strax.sort_by_time(aqmon_hits)
+        aqmon_thresholds = np.zeros(np.max(self.aqmon_channels))
+        for hit_threshold, channels in self.hit_min_amplitude_aqmon:
+            aqmon_thresholds[channels] = hit_threshold
+
+        # Split the artificial deadtime ones and do those separately if there are any
+        is_artificial = records['channel'] == AqmonChannels.ARTIFICIAL_DEADTIME
+        aqmon_hits = strax.find_hits(records[~is_artificial],
+                                     min_amplitude=aqmon_thresholds)
+
+        if np.sum(is_artificial):
+            aqmon_hits = np.concatenate([
+                aqmon_hits, self.get_deadtime_hits(records[is_artificial])])
+            strax.sort_by_time(aqmon_hits)
         return aqmon_hits
 
     def get_deadtime_hits(self, artificial_deadtime):
@@ -241,7 +242,7 @@ class VetoProximity(strax.OverlapWindowPlugin):
         default=('time', 'endtime'),
         help='Fields to determine where to look for overlaps for using '
              'this plugin in the events. The default uses start and endtime '
-             'of an event, but this can also be the S1 or S2 start/endtime'git 
+             'of an event, but this can also be the S1 or S2 start/endtime'
     )
 
     veto_proximity_window = straxen.URLConfig(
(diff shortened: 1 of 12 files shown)


straxen/plugins/acqmon_processing.py
```diff
@@ -130,89 +220,145 @@ def channel_select_(rr, ch):
...
-    veto_names = ['busy', 'he', 'hev']
+    __version__ = '2.0.0'
+    # Strictly speaking, we could depend on 'events', but then you couldn't
+    # change the event_window_fields to e.g. s1_time and s2_endtime.
+    depends_on = ('event_basics', 'veto_intervals')
+    provides = 'veto_proximity'
+    data_kind = 'events'
+    save_when = strax.SaveWhen.NEVER
+
+    event_window_fields = straxen.URLConfig(
```
JoranAngevaare, line 237: at the special request of @WenzDaniel

WenzDaniel: I must admit I was confused for 30s what exactly you mean :D

straxen/plugins/acqmon_processing.py
```diff
@@ -13,35 +17,108 @@
...
+    def find_aqmon_hits_per_channel(self, records):
+        """Allow different thresholds to be applied to different channels"""
+        aqmon_hits = [
+            strax.find_hits(
+                records[np.in1d(records['channel'], channels)],
+                min_amplitude=hit_threshold
+            )
+            for hit_threshold, channels in self.hit_min_amplitude_aqmon
+            if hit_threshold
+        ]
```
WenzDaniel, line 102: This looks fairly complicated. Why are you not just passing the different thresholds as an array like we do for the TPC?

straxen/plugins/acqmon_processing.py
```diff
@@ -13,35 +17,108 @@
 # More info about the acquisition monitor can be found here:
 # https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenon1t:alexelykov:acquisition_monitor
 
+class AqmonChannels(IntEnum):
+    GPS_SYNC = 798
+    ARTIFICIAL_DEADTIME = ARTIFICIAL_DEADTIME_CHANNEL
+    SUM_WF = 800
+    M_VETO_SYNC = 801
```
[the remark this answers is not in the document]
JoranAngevaare: good suggestion, thanks

straxen/plugins/acqmon_processing.py
```diff
@@ -109,15 +188,26 @@ def compute(self, aqmon_hits, start, end):
                     if inx == len(veto_hits_stop['time']):
                         continue
                     else:
-                        result['veto_interval'][vetos_seen] = veto_hits_stop['time'][inx] - time
                         result["time"][vetos_seen] = time
                         assert time < veto_hits_stop['time'][inx]
                         result["endtime"][vetos_seen] = veto_hits_stop['time'][inx]
-                        result["veto_type"][vetos_seen] = name + 'veto'
+                        result["veto_type"][vetos_seen] = veto_name + 'veto'
```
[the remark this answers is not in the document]
JoranAngevaare: this is correct, it also confused me, maybe it's good to change indeed, do you have a suggestion in mind?

straxen/plugins/acqmon_processing.py
```diff
@@ -56,48 +133,50 @@ def compute(self, raw_records_aqmon):
...
 
-        for name in self.veto_names:
-            veto_hits_start = channel_select_(hits, self.channel_map[name + 'start'])
-            veto_hits_stop = channel_select_(hits, self.channel_map[name + 'stop'])
+        for veto_name in self.veto_names:
+            veto_hits_start = channel_select_(aqmon_hits, self.channel_map[veto_name + 'start'])
+            veto_hits_stop = channel_select_(aqmon_hits, self.channel_map[veto_name + 'stop'])
 
             # Here we rely on the fact that for each start, there is a single stop that
             # follows it in time. If this is not true, our hardware does not work.
```
WenzDaniel, line 182: I do not get the part which follows here. If there should only for each start a single stop why not just subtracting them from each other as they are?

straxen/plugins/acqmon_processing.py
```diff
@@ -109,15 +188,26 @@ def compute(self, aqmon_hits, start, end):
                     if inx == len(veto_hits_stop['time']):
                         continue
                     else:
-                        result['veto_interval'][vetos_seen] = veto_hits_stop['time'][inx] - time
                         result["time"][vetos_seen] = time
                         assert time < veto_hits_stop['time'][inx]
                         result["endtime"][vetos_seen] = veto_hits_stop['time'][inx]
-                        result["veto_type"][vetos_seen] = name + 'veto'
+                        result["veto_type"][vetos_seen] = veto_name + 'veto'
                         vetos_seen += 1
```
WenzDaniel, line 195: So Basically up to here I do not get why it is needed.

straxen/plugins/acqmon_processing.py
```diff
@@ -109,15 +188,26 @@ def compute(self, aqmon_hits, start, end):
...
+        if n_artificial:
+            result[vetos_seen:n_artificial]['time'] = artificial_deadtime['time']
+            result[vetos_seen:n_artificial]['endtime'] = strax.endtime(artificial_deadtime)
+            result[vetos_seen:n_artificial]['veto_type'] = 'straxen_deadtime_veto'
+            vetos_seen += n_artificial
+
         result = result[:vetos_seen]
-        result['time'] = np.clip(result['time'], start, end)
-        result['endtime'] = np.clip(strax.endtime(result), 0, end)
+        result['veto_interval'] = result['endtime'] - result['time']
```
WenzDaniel, line 210: I think it does not really matter, but this is somewhat redundant is not it?

straxen/plugins/acqmon_processing.py
```diff
@@ -130,89 +220,145 @@ def channel_select_(rr, ch):
...
+
+    event_window_fields = straxen.URLConfig(
+        default=('time', 'endtime'),
+        help='Fields to determine where to look for overlaps for using '
+             'this plugin in the events. The default uses start and endtime '
+             'of an event, but this can also be the S1 or S2 start/endtime'
+    )
+
+    veto_proximity_window = straxen.URLConfig(
+        default=int(1e9),
```
WenzDaniel, line 245: I think before it was `10e9`.

WenzDaniel (commented):
Sorry, but I have to do the rest of `VetoProximity` either tomorrow or on Monday. I must go now.

straxen/plugins/acqmon_processing.py
JoranAngevaare: yeah, let me look into this indeed, there were edge cases due to the bug I fixed above, but there are also edge cases here at the start/stop of a run as @RoBGlaBe found where we miss one start or one stop, but let me add that logic here

straxen/plugins/acqmon_processing.py
JoranAngevaare: fully redundant, but since it's such a light weight dtype, I would just leave it because it's the field that is mostly used for this dtype

straxen/plugins/acqmon_processing.py
JoranAngevaare: yeah it should be even bigger I found out actually. I think the longest busys are order minutes

straxen/plugins/acqmon_processing.py
JoranAngevaare: good catch forgot about that posibility

tests/test_aqmon.py
```diff
@@ -0,0 +1,267 @@
...
+    We are interested in the deadtime, which is the time difference
+    between ON and OFF singals. This plugin computes that deadtime, even
+    if we have missed the first ON (so we start with OFF) or missed the
+    last OFF (such that the last signal is an ON). If we miss one ON at
+    the start or one OFF at the end, we should just consider all time
+    before the first OFF or all the time after the last ON as deadtime.
+    """
+    vetos_per_chunk = strax.Config(
+        # Strax does not like numpy, let's give it a list of integers
+        default=[int(n) for n in np.arange(1, 10)],
```
WenzDaniel, line 26: FYI: `list(range(10))` does the same.

tests/test_aqmon.py
```diff
@@ -0,0 +1,267 @@
...
+
+class DummyAqmonHits(strax.Plugin):
+    """
+    Dummy plugin to make some aqmon hits that may span ON and OFF
+    signals over chunks
+
+    There are two channels, that signify ON or OFF of a logic signal
+
+    We are interested in the deadtime, which is the time difference
+    between ON and OFF singals. This plugin computes that deadtime, even
```
WenzDaniel, line 18: Okay, but I think this is only the deadtime of the digitizer not the "true" dead time of the detector. For the latter one has to include the daq-veto cut boundaries.

tests/test_aqmon.py
```diff
@@ -0,0 +1,267 @@
...
+            res['channel'][::2] = self.channel_on
+            res['channel'][1::2] = self.channel_off
+            starts = res[::2]['time']
+            stops = res[1::2]['time']
+
+            # Sum the time differences between starts and stops to get the deadtime
+            self.TOTAL_DEADTIME += [np.sum(stops - starts[:len(stops)])]
+
+        else:
+            # Previous chunk, we ended with an off, so now we start with an OFF
```
WenzDaniel, line 96: start with on?

tests/test_aqmon.py
```diff
@@ -0,0 +1,267 @@
...
+
+        return self.chunk(start=self.chunk_start(chunk_i),
+                          end=self.get_chunk_end(chunk_i),
+                          data=events)
+
+
+class TestAqmonProcessing(TestCase):
+    def setUp(self) -> None:
+        st = straxen.test_utils.nt_test_context().new_context()
+        st.set_context_config({'free_options': list(st.config.keys())})
```
WenzDaniel, line 194: Why do you need this?

straxen/plugins/acqmon_processing.py
```diff
@@ -56,163 +137,307 @@ def compute(self, raw_records_aqmon):
...
+    veto_proximity_window = straxen.URLConfig(
+        default=int(300e9),
+        help='Maximum separation between veto stop and start pulses [ns]'
+    )
+    time_no_aqmon_veto_found = straxen.URLConfig(
+        default=int(3.6e+12),
+        track=True,
+        type=int,
+        help='If we don\'t find a veto close to the event, say that '
+             'the closest event is this many ns removed from it.'
```
WenzDaniel, line 325: Sorry, but I do not get that sentence :D

straxen/plugins/acqmon_processing.py
```diff
@@ -130,89 +220,145 @@ def channel_select_(rr, ch):
...
-                ]
         dtype += strax.time_fields
-        return dtype
+        start_field, stop_field = self.event_window_fields
+        for name in self.veto_names:
+            dtype += [
+                ((f'Duration of event overlapping with "{name}"-veto [ns]',
+                  f'veto_{name}_overlap'),
+                 np.int64),
+                ((f'Time (absolute) to previous "{name}"-veto '
```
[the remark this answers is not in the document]
WenzDaniel: You could also join time to prev/next into a single quantity using positive/negative times. I think in the end you are just interested in the time difference to the next closest veto. But I think it is also fine as it is.

straxen/plugins/acqmon_processing.py
```diff
@@ -56,163 +137,307 @@ def compute(self, raw_records_aqmon):
...
+                starts = np.clip(vetos_in_window['time'],
+                                 event_window[event_i]['time'],
+                                 event_window[event_i]['endtime'])
+                stops = np.clip(vetos_in_window['endtime'],
+                                event_window[event_i]['time'],
+                                event_window[event_i]['endtime'])
+                # Now sum over all the stops-starts that are clipped
+                # within the duration of the event
+                result_buffer[event_i][f'veto_{veto_name}_overlap'] = np.sum(stops -
+                                                                             starts)
```
WenzDaniel, line 397: So this refers to the fraction of time an event overlaps with a veto_interval is not it? Which should vary somewhere between 0 and left/right event extension per event?

straxen/plugins/acqmon_processing.py
```diff
@@ -56,163 +137,307 @@ def compute(self, raw_records_aqmon):
...
+                time_to_prev = np.abs(ev_wind['time'] - prev_intervals['endtime'])
+                prev_idx = np.argmin(time_to_prev)
+                times_to_prev[event_i] = time_to_prev[prev_idx]
+
+            if np.sum(interval_after):
+                next_intervals = selected_intervals[interval_after]
+                time_to_next = np.abs(next_intervals['endtime'] - ev_wind['endtime'])
+                next_idx = np.argmin(time_to_next)
+                times_to_next[event_i] = time_to_next[next_idx]
+        return times_to_prev, times_to_next
```
WenzDaniel, line 429: Not the best optimized code, but for the busy logic it will suffice.

WenzDaniel (approved):
Looks good. Just have to think about the deadtime of the detector a bit as it is not very well defined right now.

tests/test_aqmon.py
JoranAngevaare: right, as discussed, it's the other way around. This is the true deadtime, but the effective livetime loss may be greater and depends on the length of the signals we expect

darrylmasson: We're getting into semantics here. This plugin computes DAQ deadtime, which is closely related but not identical to detector deadtime.

tests/test_aqmon.py
darrylmasson: `list(range(1,10))` if we're being pedantic

JoranAngevaare: yeah, you can probably imagine why I wrote it like this instead of `list(range(1,10))`

tests/test_aqmon.py
JoranAngevaare: yes thanks!

tests/test_aqmon.py
JoranAngevaare: otherwise I get a lot of warnings since I deregister a lot of plugins. I could also have started with a bare context instead of the nT testing one

JoranAngevaare: ill comment

straxen/plugins/acqmon_processing.py
JoranAngevaare: no it can be the entire event, why would it be limited to the extension of the vents?

straxen/plugins/acqmon_processing.py
JoranAngevaare: ok I'll change it

straxen/plugins/acqmon_processing.py
JoranAngevaare: Sorry Daniel, this is not a very nice nor justified comment, it's `njitted` and very fast to compute. I could unroll a couple of numpy vector operations but I doubt it will make any difference.

Just for a high rate run (018469) the entire `compute` part takes 220 ms of which only 30% is in this bit of logic.

straxen/plugins/acqmon_processing.py
WenzDaniel: Ah sorry I should have been more precise. I meant if you assume no pile-up and that S2 is your triggering peak.

straxen/plugins/acqmon_processing.py
WenzDaniel: My apologize I did not mean it in an offending way. What I wanted to say is that I am aware that you can do better if it would be necessary, but it is not in this particular case as the data type is small. 

In general I was referring to this logic:

You first always take all intervals before and after 
```
 for event_i, ev_wind in enumerate(event_window):
            # Two cases left, either veto's are before or after the event window
            interval_before = selected_intervals['endtime'] < ev_wind['time']
            interval_after = selected_intervals['time'] > ev_wind['endtime']
```
and take afterwards the one with the smallest time difference. 
```
 if np.sum(interval_before):
                prev_intervals = selected_intervals[interval_before]
                time_to_prev = np.abs(ev_wind['time'] - prev_intervals['endtime'])
                prev_idx = np.argmin(time_to_prev)
                times_to_prev[event_i] = time_to_prev[prev_idx]
```
One could have the code more performant if one exploits the fact that the events are sorted in time and non-overlapping. E.g. it does not make sense to compute still the time difference to the first veto interval if you are close to the end of your run. 

But as I said making it another factor X  faster does not matter here.

straxen/plugins/acqmon_processing.py
JoranAngevaare: ah yeah, this cannot be longer than your event and is usually 0

straxen/plugins/acqmon_processing.py
JoranAngevaare: Yeah this is what I meant with unrolling the the vector operations which numba should be able to already handle at the speed of lightning. Event if I were to reduce this entire block with harder to read numba optimized code, the compute argument of this plugin would be 30% faster, so I'm definitely not changing it 😝

straxen/plugins/acqmon_processing.py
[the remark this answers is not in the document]
darrylmasson: "Absolute" might indicate to some people that it's an actual unix timestamp rather than a timestamp difference. This was my initial interpretation.

straxen/plugins/acqmon_processing.py
darrylmasson: I've made similar points before - if the acquisition monitor is ever producing so much data that the efficiency of its plugins becomes important, we have much larger issues. The complexity and nuance of especially this plugin is such that we should emphasize the readability much more than its performance.

darrylmasson (approved):
Eventually we'll get this working properly

straxen/plugins/acqmon_processing.py
```diff
@@ -56,163 +137,308 @@ def compute(self, raw_records_aqmon):
...
+            veto_hits_stop = np.concatenate([veto_hits_stop, extra_stop])
+        if len(veto_hits_stop) - len(veto_hits_start) == 1:
+            # There is one *stop* of the //beginning// of the run
+            # -> the **start** is missing (because it's from before
+            # starting the run), # let's add one **start** at the
+            # //beginning// of this chunk
+            extra_start = self.fake_hit(chunk_start)
+            veto_hits_start = np.concatenate([extra_start, veto_hits_start])
+
+        something_is_wrong = len(veto_hits_start) != len(veto_hits_stop)
```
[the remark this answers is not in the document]
JoranAngevaare: we don't see them anymore, thats the greatest fix of this PR https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:j_angevaare:busy_logic_straxen

Otherwise, we will crash here https://github.com/XENONnT/straxen/blob/a3da24e9f88c3a9e190efc5afe0fbadf109fbf3d/straxen/plugins/acqmon_processing.py#L274

straxen/plugins/acqmon_processing.py
```diff
@@ -56,163 +137,308 @@ def compute(self, raw_records_aqmon):
...
+                or (len(veto_hits_start) and not len(veto_hits_stop)
+                    )
+        )
+        if missing_a_final_stop:
+            # There is one *start* of the //end// of the run -> the
+            # **stop** is missing (because it's outside of the run),
+            # let's add one **stop** at the //end// of this chunk
+            extra_stop = self.fake_hit(chunk_end)
+            veto_hits_stop = np.concatenate([veto_hits_stop, extra_stop])
+        if len(veto_hits_stop) - len(veto_hits_start) == 1:
```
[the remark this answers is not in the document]
JoranAngevaare: ah right, I actually do this for all vetos:
https://github.com/XENONnT/straxen/blob/a3da24e9f88c3a9e190efc5afe0fbadf109fbf3d/straxen/plugins/acqmon_processing.py#L274

straxen/plugins/acqmon_processing.py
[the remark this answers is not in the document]
JoranAngevaare: right I understand the confusion, let me change the wording a bit

RoBGlaBe (approved):
Addresses all the issues that I could find so far. Nice!

straxen/plugins/acqmon_processing.py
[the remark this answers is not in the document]
RoBGlaBe: It's probably not necessary, as this didn't happen before, but we could throw an error if a start is missing for something else than the HEV. This would mean a digitizer went busy *before* run start.

PR #931, opened 2022-02-21
Update corrected areas

ershockley:
This adds functionality for a potential time-dependent S1 correction, like we already have for S2s. It assumes a relative correction, similar to that of the extraction efficiency.

### DIFF
--- a/straxen/plugins/event_processing.py
+++ b/straxen/plugins/event_processing.py
@@ -655,9 +655,8 @@ class CorrectedAreas(strax.Plugin):
         help='Actual SE gain for a given run (allows for time dependence)')
 
     # relative extraction efficiency which can change with time and modeled by CMT.
-    # defaults to no correction
     rel_extraction_eff = straxen.URLConfig(
-        default=1.0,
+        default='cmt://rel_extraction_eff?version=ONLINE&run_id=plugin.run_id',
         help='Relative extraction efficiency for this run (allows for time dependence)')
 
     # relative light yield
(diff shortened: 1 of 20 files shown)


skazama:
@ershockley maybe we also need to add this correction to `EventInfoDouble`? @DonNabla ?

l-althueser:
Good point, we should add the special case to be consistent.

l-althueser:
And add also the missing cs2 without time correction.

straxen/plugins/event_processing.py
```diff
@@ -659,12 +659,21 @@ class CorrectedAreas(strax.Plugin):
         default='cmt://rel_extraction_eff?version=ONLINE&run_id=plugin.run_id',
         help='Relative extraction efficiency for this run (allows for time dependence)')
 
+    # relative light yield
+    # defaults to no correction
+    rel_light_yield = straxen.URLConfig(
+        default='cmt://relative_light_yield?version=ONLINE&run_id=plugin.run_id',
```
skazama, line 665: sorry @l-althueser and @a1exndr, what's the value for the future data? 1?

l-althueser: Yes: https://github.com/XENONnT/private_nt_aux_files/pull/152/files#diff-c19121527752e93c5048c5104740d1470ffeb8b4152ce77c836ea3a8a4868e3fR354

l-althueser: image

skazama (approved):

PR #932, opened 2022-02-22
Bumpy pymongo

JoranAngevaare:
Maybe we should really do this, it's causing issues with python3.10:
https://pymongo.readthedocs.io/en/stable/examples/tls.html#python-3-10-incompatibilities-with-tls-ssl-on-mongodb-4-0


JoranAngevaare:
probably need to merge strax first

jmosbacher:
while im all for making the effort to get with the times and switch to the latest version, I'm not sure this would affect us. our deployments dont have tls enabled as far as i know

JoranAngevaare:
I'm sure there are places, especially in bootstrax / ajax that will be affected since the tz-info is now off by default so I have to test and find the issues

jmosbacher:
Yeah definitely, I was referring to the issue with python 3.10 you linked to, I'm not sure it would cause any problems for us to use 3.10 with the old pymongo since we dont use tls as far as i know

bin/bootstrax
```diff
@@ -959,14 +979,17 @@ def clear_shm():
...
         try:
-            run_db.command('ping')
-            daq_db.command('ping')
+            print(run_coll)
+            print(run_coll.find_one())
+            run_coll.find_one(projection={'_id':1})
+            bs_coll.find_one(projection={'_id':1})
             break
         except Exception as ping_error:
+            raise
```
JoranAngevaare, line 988: Still getting issues here (despite following https://pymongo.readthedocs.io/en/stable/faq.html#using-pymongo-with-multiprocessing) 

```
  File "/home/xedaq/miniconda/envs/py38/lib/python3.8/multiprocessing/process.py", line 315, in _bootstrap
    self.run()
  File "/home/xedaq/miniconda/envs/py38/lib/python3.8/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/home/xedaq/software/straxen/bin/bootstrax", line 1279, in run_strax
    ping_dbs()
  File "/home/xedaq/software/straxen/bin/bootstrax", line 983, in ping_dbs
    print(run_coll.find_one())
  File "/home/xedaq/miniconda/envs/py38/lib/python3.8/site-packages/pymongo/collection.py", line 1114, in find_one
    for result in cursor.limit(-1):
  File "/home/xedaq/miniconda/envs/py38/lib/python3.8/site-packages/pymongo/cursor.py", line 1159, in next
    if len(self.__data) or self._refresh():
  File "/home/xedaq/miniconda/envs/py38/lib/python3.8/site-packages/pymongo/cursor.py", line 1080, in _refresh
    self.__send_message(q)
  File "/home/xedaq/miniconda/envs/py38/lib/python3.8/site-packages/pymongo/cursor.py", line 971, in __send_message
    response = client._run_operation(
  File "/home/xedaq/miniconda/envs/py38/lib/python3.8/site-packages/pymongo/mongo_client.py", line 1215, in _run_operation
    return self._retryable_read(
  File "/home/xedaq/miniconda/envs/py38/lib/python3.8/site-packages/pymongo/mongo_client.py", line 1307, in _retryable_read
    with self._secondaryok_for_server(read_pref, server, session) as (
  File "/home/xedaq/miniconda/envs/py38/lib/python3.8/contextlib.py", line 113, in __enter__
    return next(self.gen)
  File "/home/xedaq/miniconda/envs/py38/lib/python3.8/site-packages/pymongo/mongo_client.py", line 1162, in _secondaryok_for_server
    with self._get_socket(server, session) as sock_info:
  File "/home/xedaq/miniconda/envs/py38/lib/python3.8/contextlib.py", line 113, in __enter__
    return next(self.gen)
  File "/home/xedaq/miniconda/envs/py38/lib/python3.8/site-packages/pymongo/mongo_client.py", line 1099, in _get_socket
    with server.get_socket(
  File "/home/xedaq/miniconda/envs/py38/lib/python3.8/contextlib.py", line 113, in __enter__
    return next(self.gen)
  File "/home/xedaq/miniconda/envs/py38/lib/python3.8/site-packages/pymongo/pool.py", line 1371, in get_socket
    sock_info = self._get_socket(all_credentials)
  File "/home/xedaq/miniconda/envs/py38/lib/python3.8/site-packages/pymongo/pool.py", line 1436, in _get_socket
    self._raise_if_not_ready(emit_event=True)
  File "/home/xedaq/miniconda/envs/py38/lib/python3.8/site-packages/pymongo/pool.py", line 1407, in _raise_if_not_ready
    _raise_connection_failure(
  File "/home/xedaq/miniconda/envs/py38/lib/python3.8/site-packages/pymongo/pool.py", line 250, in _raise_connection_failure
    raise AutoReconnect(msg) from error
pymongo.errors.AutoReconnect: xenon1t-daq.lngs.infn.it:27017: connection pool paused
```

JoranAngevaare: will have to look later

darrylmasson: Why do we still have things pointing at `xenon1t-daq.lngs.infn.it`? I thought we updated all of those things a long time ago.

jmosbacher: Just want to mention here that I think its still defined as xenon1t-daq.lngs.infn.it on the MongoDB server side. This is important because Mongodb has a protocol for a client to negotiate which node to send a request to on a given replica set, the first step of the protocol is a request to the first node in the list of nodes to send back a list of available nodes. So when connecting to a replicaset, the address you give to the client is ignored except for the initial list-servers request and the addresses defined in the replica set are the ones used for the actual query requests.
If the old address is not resolvable from the DAQ (eg the DNS resolves to the public IP instead of the internal LNGS one) and the client is using the replica set protocol then the connection would fail. 
No idea if this is what happened here but @darrylmasson s comment triggered this possibility in my head.
The solution would be to force pymongo to use the single node protocol instead of the replica set protocol.

darrylmasson: So a bit of digging into the MongoDB interface we use seems to indicate that changing hostnames isn't possible, so we're stuck with `xenon1t-daq` until the DNS name expires and everything burns down. There are still a few references to the old alias in `/etc/hosts` on the DAQ gateway, I don't particularly want to remove them because I don't want to spend a day fixing the fallout.

PR #933, opened 2022-02-24
Updated nveto resolving time to 200 ns. Change to URL configs

WenzDaniel:
Changes resolving time for the neutron-veto event building from 300 ns to 200 ns [see this note](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:wenz:nveto:comissioning:detector_analysis:center_time). Also changes option to URL based config as lineage changes.

### DIFF
--- a/straxen/plugins/veto_events.py
+++ b/straxen/plugins/veto_events.py
@@ -11,17 +11,6 @@ from immutabledict import immutabledict
 export, __all__ = strax.exporter()
 
 
-@strax.takes_config(
-    strax.Option('event_left_extension_nv', default=0, infer_type=False,
-                 help="Extends events this many ns to the left"),
-    strax.Option('event_resolving_time_nv', default=300, infer_type=False,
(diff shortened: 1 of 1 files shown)


straxen/plugins/veto_events.py
```diff
@@ -31,9 +20,34 @@ class nVETOEvents(strax.OverlapWindowPlugin):
...
+        help='Resolving time for window coincidence [ns].'
+    )
+    event_min_hits_nv = straxen.URLConfig(
+        default=3,
+        track=True,
+        type=int,
+        help='Minimum number of fully confined hitlets to define an event.'
+    )
+    channel_map = straxen.URLConfig(
+        default=3,
```
jmosbacher, line 45: Is this a typo? should be a immutabledict right?

WenzDaniel: Ah yes good catch. Never just copy paste things. Thanks. Processing worked and setting and lineage used the correct channel map. Will leave the default empty.

PR #939, opened 2022-02-28
set  runid_test_str=0000

### LINKED ISSUE #938: get_cmt_options() fails when some CMT options are evaluated at plugin initialization
**Describe the bug**
Looks like this is an issue with the "get_cmt_options" function for url configs, using the string 'norunids!' as a test to verify that the name of the correction doesnt depend on the run_id. 
This works fine when all configs are were evaluated at runtime, eg when all configs are URLConfigs, but fails when some of the configs are CMT corrections evaluated at plugin initialization in the setup() method because ' norunids!' is not a valid run id. 

I think just changing this test runid to eg '0000' would probably fix this. 

**To Reproduce**
see [PR916](https://github.com/XENONnT/straxen/runs/5360518617?check_suite_focus=true#step:11:235)

**Expected behavior**
Mixing URLConfig and old style CMT options should work without any issues.

jmosbacher:
This should fix #938.
set runid_test_str = '0000'
in the get_cmt_options function

This will prevent failures in cases where the run id is actually used to fetch a correction from CMT in the plugin setup() method

### DIFF
--- a/straxen/get_corrections.py
+++ b/straxen/get_corrections.py
@@ -162,7 +162,7 @@ def get_cmt_options(context: strax.Context) -> ty.Dict[str, ty.Dict[str, tuple]]
     """
 
     cmt_options = {}
-    runid_test_str = 'norunids!'
+    runid_test_str = '0000'
 
     for data_type, plugin in context._plugin_class_registry.items():
         for option_key, option in plugin.takes_config.items():


WenzDaniel (approved):

WenzDaniel:
Directly merged it with https://github.com/XENONnT/straxen/pull/916 to see whether it does the trick.

PR #954, opened 2022-03-01
Bayes plugin

ahiguera-mx:
Bayes peak probability computation

Bayes peak probability computation and propagate this information up to event level

### _Notes on testing_

### DIFF
--- a/straxen/contexts.py
+++ b/straxen/contexts.py
@@ -20,6 +20,7 @@ common_opts = dict(
         straxen.PulseProcessing,
         straxen.Peaklets,
         straxen.PeakletClassification,
+        straxen.BayesPeakletClassification,
         straxen.MergedS2s,
         straxen.Peaks,
         straxen.PeakBasics,
@@ -29,6 +30,7 @@ common_opts = dict(
         straxen.EventPositions,
         straxen.CorrectedAreas,
         straxen.EnergyEstimates,
+        straxen.EventwBayesClass,
     ],
     check_available=('raw_records', 'peak_basics'),
     store_run_fields=(
--- a/straxen/plugins/bayes_peaklet_classification.py
+++ b/straxen/plugins/bayes_peaklet_classification.py
@@ -1,0 +1,138 @@
+import numpy as np
+import numba
+from scipy.special import logsumexp
+import strax
+import straxen
+
+
+class BayesPeakletClassification(strax.Plugin):
+    """
+    Bayes Peaklet classification
+    Returns the ln probability of a each event belonging to the S1 and S2 class.
+    Uses conditional probabilities and data parameterization learned from wfsim data.
+
+    :param peaks: peaks
+    :param waveforms: peaks waveforms in PE/ns
+    :param quantiles: quantiles in ns
+    :returns: the ln probability of a each peaklet belonging to S1 and S2 class
+    """
+    provides = 'peak_classification_bayes'
+    depends_on = ('peaks',)
+    __version__ = '0.0.1'
+    dtype = (strax.time_fields
+             + [('ln_prob_s1', np.float32, 'S1 ln probability')]
+             + [('ln_prob_s2', np.float32, 'S2 ln probability')]
+            )
+
+    # Descriptor configs
+    bayes_config_file = straxen.URLConfig(
+        default='resource://cmt://'
+                'bayes_model'
+                '?version=ONLINE&run_id=plugin.run_id&fmt=npy',
+        help='Bayes model, conditional probabilities tables and Bayes discrete bins'
+    )
+    bayes_n_nodes = straxen.URLConfig(
+        default=50,
+        help='Number of attributes(features) per waveform and quantile'
+    )
+    n_bayes_classes = straxen.URLConfig(
+        default=2,
+        help='Number of label classes S1(1)/S2(2)'
+    )
+
+    def setup(self):
+
+        self.class_prior = np.ones(self.n_bayes_classes)/self.n_bayes_classes
+        self.bins = self.bayes_config_file['bins']
+        self.cpt = self.bayes_config_file['cprob']
+
+    def compute(self, peaks):
+
+        waveforms, quantiles = compute_wf_and_quantiles(peaks, self.bayes_n_nodes)
+
+        ln_prob_s1, ln_prob_s2 = compute_inference(self.bins, self.bayes_n_nodes, self.cpt,
+                                                   self.n_bayes_classes, self.class_prior,
+                                                   waveforms, quantiles)
+
+        return dict(time=peaks['time'],
+                    endtime=peaks['time'] + peaks['dt'] * peaks['length'],
+                    ln_prob_s1=ln_prob_s1,
+                    ln_prob_s2=ln_prob_s2
+                    )
+
+def compute_wf_and_quantiles(peaks: np.ndarray, bayes_n_nodes: int):
+    """
+    Compute waveforms and quantiles for a given number of nodes(atributes)
+    :param peaks:
+    :param bayes_n_nodes: number of nodes or atributes
+    :return: waveforms and quantiles
+    """
+    waveforms = np.zeros((len(peaks), bayes_n_nodes))
+    quantiles = np.zeros((len(peaks), bayes_n_nodes))
+
+    num_samples = peaks['data'].shape[1]
+    step_size = int(num_samples/bayes_n_nodes)
+    steps = np.arange(0, num_samples+1, step_size)
+
+    data = peaks['data'].copy()
+    dts = peaks['dt'].copy()
+    data[data < 0.0] = 0.0
+    for i, p in enumerate(peaks):
+        sample_number = np.arange(0, num_samples+1, 1)*p['dt']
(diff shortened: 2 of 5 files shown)


straxen/contexts.py
```diff
@@ -20,6 +20,7 @@
         straxen.PulseProcessing,
         straxen.Peaklets,
         straxen.PeakletClassification,
+        straxen.BayesPeakletClassification,
```
JoranAngevaare, line 23: you probably don't want to add this for 1T but for nT only in `xnt_common_opts`

ahiguera-mx: done

WenzDaniel:
@ahiguera-mx thanks for porting the PR. Please do not forget my comments in the closed one. https://github.com/XENONnT/straxen/pull/924

ahiguera-mx:
Thanks for the comments @WenzDaniel, see my replies on [#924](https://github.com/XENONnT/straxen/pull/924)

WenzDaniel:
Can you elaborate a bit more 

> strax.fully_contained_in is not an option since and S1 can be contained within an S2

I do not see why this would be an issue here. Are you worried that the S1 actually belongs to the S2 and changes therefore its waveform? But especially in this case you would like to know if the quality of the S2 suffers from that is not it? 

Because, since your are using peaks the waveform of the "S1" and the "S2" are not different from any other peak.

JoranAngevaare:
> Can you elaborate a bit more
> 
> > strax.fully_contained_in is not an option since and S1 can be contained within an S2
> 
> I do not see why this would be an issue here. Are you worried that the S1 actually belongs to the S2 and changes therefore its waveform? But especially in this case you would like to know if the quality of the S2 suffers from that is not it?
> 
> Because, since your are using peaks the waveform of the "S1" and the "S2" are not different from any other peak.

@WenzDaniel I would suspect `np.in1d(peaks['time'], events[f'{name}_time'])` is going to be faster than `strax.fully_contained_in` in because you don't need to check the endtimes, you know the timestamps of the peaks are unique. In that sense you don't need a more elaborate check as in `strax.fully_contained_in`. I also don't think this is going to be the place were performance could be lost (if any). [Hence I actually advice doing it like this](https://github.com/XENONnT/straxen/pull/924#discussion_r814564866), I think it's quite a nice approach for code clarity and performance.

WenzDaniel:
Just pointed it out as an option, but I also mentioned that it probably wont be faster. I am just surprised by Aarons reply as I do not understand why having S1s inside of S2s would be a problem. I would like to clarify before merging.

@ahiguera-mx  could you also please address my other comments. At least leaving a short reason why you do not consider a suggestion this would help me to understand better why certain aspects are needed. 

In addition could you explain me once more in detail what these quantiles features  represent. I looked again at your notes and presentation but I do not fully catch it. Are those the inner n-quantiles around the area weighted mean? Otherwise I would have expected something monotonically increasing from the left to the right.

straxen/plugins/bayes_peaklet_classification.py
```diff
@@ -0,0 +1,138 @@
+import numpy as np
+import numba
+from scipy.special import logsumexp
+import strax
+import straxen
+
+
+class BayesPeakletClassification(strax.Plugin):
```
skazama, line 8: is this based on peaklets or peaks? seems BayesPeakClassification is the right class name here?

ahiguera-mx:
> Just pointed it out as an option, but I also mentioned that it probably wont be faster. I am just surprised by Aarons reply as I do not understand why having S1s inside of S2s would be a problem. I would like to clarify before merging.
@WenzDaniel 
As Joran mentioned [here](https://github.com/XENONnT/straxen/pull/924#discussion_r813700446), that is why my comment, although I most said that I'm not an expert so are relying on others

straxen/plugins/bayes_peaklet_classification.py
skazama: then, I think it's better to change it to `BayesPeakClassification` to avoid confusion.

straxen/plugins/bayes_peaklet_classification.py
```diff
@@ -0,0 +1,138 @@
...
+    """
+    Bayes Peaklet classification
```
skazama, line 10: Also here

straxen/plugins/bayes_peaklet_classification.py
ahiguera-mx: Ok, so then I will have to remove/add the file since is named `bayes_peaklet_classificaiton.py`

straxen/plugins/bayes_peaklet_classification.py
ahiguera-mx: ok

straxen/plugins/bayes_peaklet_classification.py
ahiguera-mx: @WenzDaniel @JoranAngevaare @skazama 
I guess we all agree to change the name?

straxen/plugins/bayes_peaklet_classification.py
```diff
@@ -0,0 +1,138 @@
...
+    """
+    waveforms = np.zeros((len(peaks), bayes_n_nodes))
+    quantiles = np.zeros((len(peaks), bayes_n_nodes))
+
+    num_samples = peaks['data'].shape[1]
+    step_size = int(num_samples/bayes_n_nodes)
+    steps = np.arange(0, num_samples+1, step_size)
+
+    data = peaks['data'].copy()
+    dts = peaks['dt'].copy()
```
WenzDaniel, line 78: I still think you do not have to copy `dt`.

WenzDaniel (approved):
Okay thanks a lot. I think it looks good. Maybe extend a bit the doc-strings here and there. After resolving the conflict we can merge.

skazama (commented):
I tested the plugin yesterday and it worked well (though it took hours for Rn220 calibration data), so once the conflicts are solved, I can also approve the PR.

WenzDaniel:
> (though it took hours for Rn220 calibration data)

You mean for the entire data or just a single run?

skazama:
> You mean for the entire data or just a single run?
for just a single run of Rn220 data with HE-veto. I don't remember the exact time, but it's like 2-3 hours

WenzDaniel:
Oh wow no in that case we cannot approve.

straxen/plugins/bayes_peaklet_classification.py
JoranAngevaare: sounds good!

ahiguera-mx:
> You mean for the entire data or just a single run?
> for just a single run of Rn220 data with HE-veto. I don't remember the exact time, but it's like 2-3 hours

@skazama this contradict with what I have found [here](https://github.com/XENONnT/straxen/pull/924#discussion_r814448839). Also Daniel check at the performance

straxen/plugins/bayes_peaklet_classification.py
tinapeters: Yes, I agree.

skazama:
@ahiguera-mx @WenzDaniel can you try run `023982`? Maybe the way I tested on my notebook was not good.

straxen/plugins/bayes_peaklet_classification.py
ahiguera-mx: fixed

straxen/plugins/bayes_peaklet_classification.py
ahiguera-mx: done

straxen/plugins/bayes_peak_classification.py
```diff
@@ -0,0 +1,140 @@
...
+                      from zero to the total area with normalized cumulative sum to determine the time
+    :returns: the ln probability of a each peak belonging to S1 and S2 class
+    """
+
+    provides = 'peak_classification_bayes'
+    depends_on = ('peaks',)
+    __version__ = '0.0.1'
+    dtype = (strax.time_fields
+             + [('ln_prob_s1', np.float32, 'S1 ln probability')]
+             + [('ln_prob_s2', np.float32, 'S2 ln probability')]
```
sophiafarrell, line 27: My only comment is whether we want to write this as the description or if it would be more straightforward for us to use: `ln probability_S1` and `ln probability_S2` or something to illustrate that we mean the log-probability peak is an S1 and S2 respectively.

sophiafarrell (commented):
I've left one comment but it is just a minor doc comment. This looks good to me and thank you for all the speed-up checks.

straxen/plugins/bayes_peak_classification.py
ahiguera-mx: @sophiafarrell if we write `s1_ln_prob`, then when running `event_w_bayes_class` it would add `s1_s1_ln_prob` wich is confusing

sophiafarrell: I just meant the description of the variable! That's all. e.g. proposed line change: 
`+ [('ln_prob_s1', np.float32, 'ln probability S1')]`
so it's minor and not huge if it doesn't get included. Just a suggestion

PR #960, opened 2022-03-04
Aqmon hits ordering fix

RoBGlaBe:
# What does the code in this PR do / what does it improve?
aqmon_hits are now ordered by time, as expected. Also, we check if the raw_records_aqmon are assigned wrong times so that we get overlapping records.

# Can you briefly describe how it works?
We thought we'd order aqmon_hits by time, already. As the strax.sort_by_time method returns the result instead of an in place change, so far we didn't. [This is fixed](https://github.com/XENONnT/straxen/blob/9d7539a517c306460380a6cb284ddb6db733bf0f/straxen/plugins/acqmon_processing.py#L45).
As done with raw_records already, we also check here that there are no overlapping records, which can lead to time-ordering issues again. (as done in [the other case](https://github.com/XENONnT/straxen/blob/e341819beb9b44eba0b734397c02777e8131f9b4/straxen/plugins/pulse_processing.py#L178), this check can be turned off via a strax-option)

# Comment
Time ordering would currently only lead to issues if, within one channel, time-ordering is not given. However, if records from two different channels are swapped, this doesn't do anything.
As we always expect our data to be ordered by time in strax and to prevent funny things from happening, it doesn't hurt to make sure everything is correctly ordered.

### DIFF
--- a/straxen/plugins/acqmon_processing.py
+++ b/straxen/plugins/acqmon_processing.py
@@ -18,7 +18,10 @@ T_NO_VETO_FOUND = int(3.6e+12)
 @strax.takes_config(strax.Option('hit_min_amplitude_aqmon', default=50, track=True, infer_type=False,
                                  help='Minimum hit threshold in ADC*counts above baseline'),
                     strax.Option('baseline_samples_aqmon', default=10, track=True, infer_type=False,
-                                 help='Number of samples to use at the start of the pulse to determine the baseline'))
+                                 help='Number of samples to use at the start of the pulse to determine the baseline'),
+                    strax.Option('check_raw_record_aqmon_overlaps', default=True, track=False, infer_type=False,
+                                 help='Crash if any of the pulses in raw_records_aqmon overlap with others '
+                                      'in the same channel'))
 class AqmonHits(strax.Plugin):
     """ Find hits in acquisition monitor data. These hits could be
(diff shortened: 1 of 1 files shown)


JoranAngevaare:
Thanks Robin for spotting this issue! 

Your PR seems to be from an out of date branch? Could you fix the merge conflict or open a new PR starting from current master?

Three small additional questions:
 - could you increment the version number?
 - Could you use an URLConfig (such as here https://github.com/XENONnT/straxen/blob/e341819beb9b44eba0b734397c02777e8131f9b4/straxen/plugins/acqmon_processing.py#L68) it's a new - fancier version of the strax.Option
 - Perhaps we should move this statement https://github.com/XENONnT/straxen/blob/e341819beb9b44eba0b734397c02777e8131f9b4/straxen/plugins/acqmon_processing.py#L112  to here https://github.com/XENONnT/straxen/blob/e341819beb9b44eba0b734397c02777e8131f9b4/straxen/plugins/acqmon_processing.py#L91 such that we are certain the hits are also always sorted by time.
 
Thanks again!

RoBGlaBe:
I thought I did `git pull` before creating the new branch... well, turns out I didn't. Sorry for that.
I created a new branch from the latest straxen version with the changes you pointed out above, @JoranAngevaare. PR follows.

PR #961, opened 2022-03-05
fix hit sorting, add raw-record-aqm overlap check

RoBGlaBe:
# What does the code in this PR do / what does it improve?
aqmon_hits are now ordered by time, as expected. Also, we check if the raw_records_aqmon are assigned wrong times so that we get overlapping records.

# Can you briefly describe how it works?
We thought we'd order aqmon_hits by time, already. As the strax.sort_by_time method returns the result instead of an in place change, so far we didn't. [This is fixed](https://github.com/XENONnT/straxen/blob/3efde7a0459025d8cd3bf02f4f6c85b831f3c6fc/straxen/plugins/acqmon_processing.py#L100).
As done with raw_records already, we also check here that there are no overlapping records, which can lead to time-ordering issues again. (as done in [the other case](https://github.com/XENONnT/straxen/blob/e341819beb9b44eba0b734397c02777e8131f9b4/straxen/plugins/pulse_processing.py#L178-L179), this check can be turned off via a strax-option)

# Comment
Time ordering would currently only lead to issues if, within one channel, time-ordering is not given. However, if records from two different channels are swapped, this doesn't do anything.
As we always expect our data to be ordered by time in strax and to prevent funny things from happening, it doesn't hurt to make sure everything is correctly ordered.


RoBGlaBe:
For reference, the first PR on outdated straxen version: [outdated PR](https://github.com/XENONnT/straxen/pull/960#issuecomment-1059494789)

darrylmasson (approved):
Looks good to me

JoranAngevaare (commented):
Nice work!

Could you delete this line:
https://github.com/XENONnT/straxen/blob/0fee29e0ca868c80a1c2f28eb4a12bdb6f21c4a8/straxen/plugins/acqmon_processing.py#L124

And set the version to 1.1.1?
https://github.com/XENONnT/straxen/blob/0fee29e0ca868c80a1c2f28eb4a12bdb6f21c4a8/straxen/plugins/acqmon_processing.py#L44

RoBGlaBe:
> Nice work!
> 
> Could you delete this line:
> 
> https://github.com/XENONnT/straxen/blob/0fee29e0ca868c80a1c2f28eb4a12bdb6f21c4a8/straxen/plugins/acqmon_processing.py#L124
> 
> And set the version to 1.1.1?
> 
> https://github.com/XENONnT/straxen/blob/0fee29e0ca868c80a1c2f28eb4a12bdb6f21c4a8/straxen/plugins/acqmon_processing.py#L44

Ah, right. I just did. Thanks!

JoranAngevaare (approved):
👍

PR #972, opened 2022-03-14
Replace print with logger in straxer

JoranAngevaare:
Use `logging.info`  instead of `print`


JoranAngevaare:
superseded by #973

Issue #966, opened 2022-03-08
Many warnings peakShadow


JoranAngevaare:
**Describe the bug**
This bit of logic in the peak shadow plugin gives many numpy warning (probably s2 is negative or 0)

```python
    def getsigma(sigma_and_baseline, s2):
        # The parameter of HalfCauchy, which is a function of S2 area
        return sigma_and_baseline[0] / np.sqrt(s2) + sigma_and_baseline[1]
```

The warnings:
```
/home/xedaq/software/straxen/straxen/plugins/peak_processing.py:502: RuntimeWarning: invalid value encountered in sqrt
  return sigma_and_baseline[0] / np.sqrt(s2) + sigma_and_baseline[1]
```

**To Reproduce**
Insert the MWE of how to reproduce the error
process run 37952

**Expected behavior**
Less warnings

**Software**

````
(py38) xedaq@eb0:~$ python -c "import straxen ; straxen.print_versions()"
cutax is not installed
Host eb0.xenon.local
 module version                                       path                     git
 python  3.8.12 /home/xedaq/miniconda/envs/py38/bin/python                    None
  strax   1.1.8           /home/xedaq/software/strax/strax branch:master | c99957e
straxen   1.5.0       /home/xedaq/software/straxen/straxen branch:master | ddc3d9c
```

JoranAngevaare:
@dachengx would you mind having a look?

dachengx:
@JoranAngevaare I tried to solve it in [PR](https://github.com/XENONnT/straxen/pull/980).

JoranAngevaare:
Thanks @dachengx

PR #980, opened 2022-03-15
Position shadow sigma set to nan when S2 not positive

dachengx:

Set Position shadow sigma to nan when S2 is not positive to suppress the warning in processing. 

Solving issue: https://github.com/XENONnT/straxen/issues/966

-

### DIFF
--- a/straxen/plugins/peak_processing.py
+++ b/straxen/plugins/peak_processing.py
@@ -497,9 +497,10 @@ class PeakShadow(strax.OverlapWindowPlugin):
         return result
 
     @staticmethod
+    @np.errstate(invalid='ignore')
     def getsigma(sigma_and_baseline, s2):
         # The parameter of HalfCauchy, which is a function of S2 area
-        return sigma_and_baseline[0] / np.sqrt(np.where(s2 > 0, s2, np.nan)) + sigma_and_baseline[1]
+        return sigma_and_baseline[0] / np.sqrt(s2) + sigma_and_baseline[1]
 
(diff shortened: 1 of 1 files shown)


straxen/plugins/peak_processing.py
```diff
@@ -497,6 +497,7 @@ def compute_shadow(self, peaks, current_peak):
         return result
 
     @staticmethod
+    @np.errstate(invalid='ignore')
```
JoranAngevaare, line 500: Hi @dachengx , thanks, this will definitely solve the error. In case we use `@np.errstate(invalid='ignore')` I don't think a version bump (and implied reprocessing) is really necessary right: the code will produce exactly the same data. Do you agree?

JoranAngevaare (commented):
Thanks @dachengx, so the issue was really negative `np.sqrt(s2)` in the end? I'm (I think not for the first time) a bit surprised we actually have negative area S2s, that sounds extremely weird. I was going to suggest that it might be worth checking that the results you are getting for those peaks make sense but that probably isn't needed since no outcome makes sense for negative S2s.

straxen/plugins/peak_processing.py
dachengx: I agree. The `@np.errstate` does not influence any results.

dachengx: I just revert the version bump of `PeakShadow`.

JoranAngevaare (approved):

PR #982, opened 2022-03-17
Remove Shadow&Ambience plugin SaveWhen.EXPLICIT

dachengx:

Remove `save_when` attribute of `PeakShadow`, `EventShadow`, `PeakAmbience`, `EventAmbience` plugins. 

Motivation: When analysts are loading `event_shadow` of calibration data, the `save_when = strax.SaveWhen.EXPLICIT` `PeakShadow` plugins will be stored in RAM and RAM can not hold that large amount of data. 

### DIFF
--- a/straxen/plugins/event_processing.py
+++ b/straxen/plugins/event_processing.py
@@ -803,7 +803,6 @@ class EventShadow(strax.Plugin):
     __version__ = '0.1.4'
     depends_on = ('event_basics', 'peak_basics', 'peak_shadow')
     provides = 'event_shadow'
-    save_when = strax.SaveWhen.EXPLICIT
 
     def infer_dtype(self):
         dtype = []
@@ -883,7 +882,6 @@ class EventAmbience(strax.Plugin):
     __version__ = '0.0.4'
     depends_on = ('event_basics', 'peak_basics', 'peak_ambience')
     provides = 'event_ambience'
-    save_when = strax.SaveWhen.EXPLICIT
 
     @property
     def origin_dtype(self):
--- a/straxen/plugins/event_processing.py
+++ b/straxen/plugins/event_processing.py
@@ -803,6 +803,7 @@ class EventShadow(strax.Plugin):
     __version__ = '0.1.4'
     depends_on = ('event_basics', 'peak_basics', 'peak_shadow')
     provides = 'event_shadow'
+    save_when = strax.SaveWhen.EXPLICIT
 
     def infer_dtype(self):
         dtype = []
--- a/straxen/plugins/event_processing.py
+++ b/straxen/plugins/event_processing.py
@@ -803,7 +803,6 @@ class EventShadow(strax.Plugin):
     __version__ = '0.1.4'
     depends_on = ('event_basics', 'peak_basics', 'peak_shadow')
     provides = 'event_shadow'
-    save_when = strax.SaveWhen.EXPLICIT
 
     def infer_dtype(self):
         dtype = []
--- a/straxen/plugins/peak_processing.py
+++ b/straxen/plugins/peak_processing.py
@@ -497,6 +497,7 @@ class PeakShadow(strax.OverlapWindowPlugin):
         return result
 
     @staticmethod
+    @np.errstate(invalid='ignore')
     def getsigma(sigma_and_baseline, s2):
         # The parameter of HalfCauchy, which is a function of S2 area
         return sigma_and_baseline[0] / np.sqrt(s2) + sigma_and_baseline[1]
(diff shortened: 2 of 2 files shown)


WenzDaniel:
Can you please be a bit more specific and give an example? Because, if you load all data it does not make any difference if it is first stored to disk. Further, I am wondering how you can run into any RAM issues. All the data_types you are using are small. Could it  be that for one of the runs `peak_basics` were not available so you started making `peaks` first?

ershockley:
If you check e.g. run 24075, it was going to require O(50-100) GB to load a single run. Peak basics is available, it's just calculating the peak_shadow I guess.

WenzDaniel:
I doubt that unless there is a bug. Is `peak_positions` available too? At least for v6 it is not. I just looked at `peak_basics` which is 2.2 GB and `peak_positions` should be smaller I think. So there is no way that these two data_types require 50 - 100 GB memory if they are both available.

ershockley:
I guess it's during the computation of peak_shadow that we hit extreme memory consumption (not during loading of the lower level data)? Did you try yourself? Both @skazama and I hit this problem

WenzDaniel:
But do you agree that if you load two data_types which are in total less than 10 GB it should not occupy 100 GB unless there is a bug in code? I am not arguing that I do not believe you that the memory consumption is so high. All I want to say that making it a save_when.ALWAYS plugin does not solve the issue. It sounds like a bug to me.

WenzDaniel:
I am sorry, but I cannot confirm to observe the high memory consumption for 24075. I processed `peak_shadow` for `024075`, `peak_basics` was stored and `peak_positions` was not (but the individual pos rec data_types were stored.) `peak_basics` and `peak_positions` have a size of 2219 MB and 1638 MB respectively. See attached image:

example

I also did some additional checks. I compared things with the previous straxen v1.2.8
for which my memory consumption was slightly smaller around 2 GB - 3 GB during processing. I compared the output data_types of the old and new plugin version and its byte-size doubled which explains the lower amount of memory before.

ershockley:
We were doing `st.get_df(runid, 'event_shadow')`. I think this loads all of peak_shadow into memory, whereas `st.make` handles memory better? Or am I misunderstanding?

What we are trying to do is have `peak_shadow` be written to disk always, rather than explicitly as you are doing, so that you don't have to go out of your way to call `st.make` before `st.get_df`. But maybe i'm wrong.

WenzDaniel:
Okay let me check with get_df still sounds like a bug to me. I checked the number of bytes of the peak_shadow data_type which is 88 per row and compared it to peaklets which was something like 3100ish. Peaklets for  `24075` was about 118 GB. So peakshaodws should be actually below 10 GB. But, I must admit I am not working a lot with DataRames I prefer the numpy.structured arryas. So might be that some funky is going on like converting everything to 64 bit for the dataframe.

skazama:
sorry @WenzDaniel, do you think we can merge this PR or should wait a bit to investigate potential dataframe bugs?

JoranAngevaare:
@skazama @ershockley @WenzDaniel I must admit that I think what you observed is simply because you tried loading the data when much lower level processing was still ongoing. Specifically the neural nets can take a large amount of memory.
afbeelding


By the way this is not what is happening (you clear peak_shadow from your memory as soon as it's processed up to `event_shadow`)
> We were doing st.get_df(runid, 'event_shadow'). I think this loads all of peak_shadow into memory, whereas st.make handles memory better? Or am I misunderstanding?

I don't think that loading in a dataframe can cause a memory usage that is that high (you only buffer each of the chunks output of `event_shadow`). When I made 8k runs last friday, it was super easy to compute.

JoranAngevaare (commented):
@dachengx I think the explicit saving of the peaks info is fine, could you revert that? I would definitely approve on removing the explicit saving at the events level, there is a clear speed benefit to be gained from not having to load so much data at that level

WenzDaniel:
@JoranAngevaare can you reproduce the error? I simply cannot and I real would like to understand whats going wrong here.

WenzDaniel:
>  I must admit that I think what you observed is simply because you tried loading the data when much lower level processing was still ongoing

This is/was my suspicion too. But according to the others the low level data was already stored.

JoranAngevaare:
So for this specific - very high rate run - I can definitely imagine that if peak_basics is stored but the pos-rec algo's are not, it's going to take a lot of resources. 
afbeelding

I'll double check the memory, but the plugin does take some time to compute (since it's a very high rate run)

JoranAngevaare (approved):

dachengx:
> @dachengx I think the explicit saving of the peaks info is fine, could you revert that? I would definitely approve on removing the explicit saving at the events level, there is a clear speed benefit to be gained from not having to load so much data at that level

Hey. I just reverted it.

JoranAngevaare:
@WenzDaniel , I've checked but the max usage for this run was ~6.5 GB:
afbeelding

ershockley:
Hmm okay something weird is going on then, because both Shingo and I could load the other event-level data for this run just fine, it was just the event_shadow giving problems. But good if it is fine now

Issue #977, opened 2022-03-15
Out of bounds error for BayesPeakClassification


JoranAngevaare:
**Describe the bug**
Getting errors like
```
strax.mailbox.MailboxKilled: (<class 'numpy.AxisError'>, AxisError('axis 1 is out of bounds for array of dimension 0'), <traceback object at 0x7f7cf5c94200>)
    raise MailboxKilled(self.killed_because)
strax.mailbox.MailboxKilled: (<class 'numpy.AxisError'>, AxisError('axis 1 is out of bounds for array of dimension 0'), <traceback object at 0x7f7cf5c94200>)
    result = self.compute(**kwargs)
  File "/opt/XENONnT/anaconda/envs/XENONnT_2022.03.5/lib/python3.8/site-packages/straxen/plugins/bayes_peak_classification.py", line 56, in compute
```

**To Reproduce**
Insert the MWE of how to reproduce the error
Run straxer for `030129` or `027319`

**Versions**
```
 module version                                                                             path  git
 python  3.8.12                          /opt/XENONnT/anaconda/envs/XENONnT_2022.03.5/bin/python None
straxen   1.6.0 /opt/XENONnT/anaconda/envs/XENONnT_2022.03.5/lib/python3.8/site-packages/straxen None
  cutax   1.6.0                                /dali/lgrandi/xenonnt/software/cutax/latest/cutax None
  strax   1.2.0   /opt/XENONnT/anaconda/envs/XENONnT_2022.03.5/lib/python3.8/site-packages/strax None
```

ahiguera-mx:
The issue was due to last chunk having only one peak, then [this line](https://github.com/XENONnT/straxen/blob/ffcf06ad86471caf11cc831f2ff68d70b59464af/straxen/plugins/bayes_peak_classification.py#L137) did not know how to do the sum (over wich axis), fixed by specify `axis=0`

ahiguera-mx:
solved in [this PR ](https://github.com/XENONnT/straxen/pull/984)

PR #984, opened 2022-03-21
fix issue 977  

ahiguera-mx:
This PR fixes [issue 977](https://github.com/XENONnT/straxen/issues/977)
n/A
n/a

  - [x ] _Does it solve one of the open issues on github?_

### _Notes on testing_

### DIFF
--- a/straxen/plugins/bayes_peak_classification.py
+++ b/straxen/plugins/bayes_peak_classification.py
@@ -62,7 +62,7 @@ class BayesPeakClassification(strax.Plugin):
                     ln_prob_s1=ln_prob_s1,
                     ln_prob_s2=ln_prob_s2
                     )
-
+        
 def compute_wf_and_quantiles(peaks: np.ndarray, bayes_n_nodes: int):
     """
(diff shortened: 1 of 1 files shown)


straxen/plugins/bayes_peak_classification.py
```diff
@@ -62,7 +62,7 @@ def compute(self, peaks):
                     ln_prob_s1=ln_prob_s1,
                     ln_prob_s2=ln_prob_s2
                     )
-
+
```
JoranAngevaare, line 65:
```suggestion

```

JoranAngevaare (commented):
Thanks @ahiguera-mx (I did not retry this specific run but your explanation makes a lot of sense), should we version bump the plugin to be sure? Or will it only be relevant for runs that failed before this change? I guess the latter.

ahiguera-mx:
@JoranAngevaare There is no need to bump version, it did not affect anything that has been processed, is only relevant for those runs with one peak on a given chunk

Issue #930, opened 2022-02-21
Save number of contributing PMTs from individual arrays


GiovanniVolta:
Dear @JoranAngevaare and @WenzDaniel,
It will be nice to have the number of contributing PMTs for individual arrays, We can put it on **event_area_per_channel** such that it will not be saved. Hence, no memory on dali is needed. What do you think? I am only interested on S1 btw.

I removed the "assignees". I didn't mean to ask you o do it.
```
@strax.takes_config(
    strax.Option('n_tpc_pmts', type=int,
                 help='Number of TPC PMTs'),
)
class EventAreaPerChannel(strax.LoopPlugin):
    """
    Simple plugin that provides area per channel for main 
    and alternative S1/S2 in the event. 
    """
    depends_on = ('event_basics', 'peaks')
    provides = "event_area_per_channel"
    __version__ = '0.0.1'

    compressor = 'zstd'
    save_when = strax.SaveWhen.EXPLICIT

    def infer_dtype(self):
        dtype = [(("Area per channel for main S2", "s2_area_per_channel"), 
                  np.float32, (self.config['n_tpc_pmts'],)),
                 (("Area per channel for alternative S2", "alt_s2_area_per_channel"),
                  np.float32, (self.config['n_tpc_pmts'],)),
                 (("Area per channel for main S1", "s1_area_per_channel"), 
                  np.float32, (self.config['n_tpc_pmts'],)),
                 (("Main S1 count of contributing PMTs", "s1_n_channels"), 
                  np.int16),
                 (("Main S1 top count of contributing PMTs", "s1_top_n_channels"), 
                  np.int16),
                 (("Main S1 bottom count of contributing PMTs", "s1_bottom_n_channels"), 
                  np.int16),
                 (("Area per channel for alternative S1", "alt_s1_area_per_channel"), 
                  np.float32, (self.config['n_tpc_pmts'],)),
                ]
        dtype += strax.time_fields
        return dtype
    
    def compute_loop(self, event, peaks):
        result = dict()
        result['time'] = event['time']
        result['endtime'] = strax.endtime(event)
        
        for type_ in ['s1', 's2', 'alt_s1', 'alt_s2']:
            type_index = event[f'{type_}_index']                                  
            if type_index != -1:
                type_area_per_channel = peaks['area_per_channel'][type_index]
                result[f'{type_}_area_per_channel'] = type_area_per_channel
                if type_index == 1:
[...]

PR #991, opened 2022-03-31
Fixing binomial

GiovanniVolta:

Fixing p-value calculation and add the chance to save S1 contributing PMTs for top and bottom (addressing #930 ).

Three problems were occurring before. They are listed here below.
1. For one-side left test, namely when the p-value is given by the sf(k;n,p) = 1 - cdf(k;n,p), we were not accounting for the probability of the observation k. Then we were doing p-value = P(x > k) instead of P(x >= k).
2. If j, the point with same probability of the observation k, didn't exist, we were assigning to it 0 or n. Now we assign nan. This allows to write easily the p-value calculation.
3. The interval of the secant method into which look for j were suboptimal. This was leading to wrong j.

The origin of acceptance saw by @skazama at low S1 was due to this.

Last but not least, for the `discrete` test, the scipy one is much more reliable. This would require rounding to the closet integer the k (S1 top) and n (S1). Unfortunately, numba heats me and this is not yet ready - maybe @WenzDaniel knows how to numbafied it.

In the following plots are shown a few self-explanatory examples.

binom_0 95
binom_0 05
binom_0 20
binom_0 40
binom_0 60


straxen/plugins/event_patternfit.py
```diff
@@ -523,6 +532,8 @@ def s1_area_fraction_top_probability(aft_prob, area_tot, area_fraction_top, mode
     if do_test:
         if mode == 'discrete':
             binomial_test = binom_pmf(area_top, area_tot, aft_prob)
+            # TODO:
+            #binomial_test = binomtest(k=round(area_top), n=round(area_tot), p=aft_prob, alternative='two-sided').pvalue
```
kdund, line 536: ?

kdund, line 536: ?

straxen/plugins/event_patternfit.py
```diff
@@ -478,22 +479,30 @@ def binom_test(k, n, p):
...
-        j_min, j_max = 0, mode
-
+    distance = abs(mode - k)
     target = lbinom_pmf(k, n, p)
-    j = lbinom_pmf_inverse(j_min, j_max, target, (n, p))
-
+   
+    if k < mode: 
+        j_min = mode
+        j_max = min(mode + 1.5 * distance, n)
```
kdund, line 487: why is j_max not just n? are there no cases where one crossing is more than 1.5 times the distance from the mode than the other one?

kdund (commented):
Hi, @GiovanniVolta! thanks a lot for the checks. 
I think in general this is good-- see the one test that fails though, is there a way we can check if this is approx OK?

straxen/plugins/event_patternfit.py
GiovanniVolta: I wanted to implement the discrete binomial test implemented in scipy instead of the one there is right now, but numba does like it

kdund: gotcha, lets have that wait the next round then

straxen/plugins/event_patternfit.py
```diff
@@ -478,22 +479,30 @@ def binom_test(k, n, p):
...
+        j = lbinom_pmf_inverse(j_min, j_max, target, (n, p))
+        ls, rs = k, j
+    elif k > mode:
```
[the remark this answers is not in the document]
kdund: (come to think of it if k == mode, you should just return 1. to begin with, I guess)

straxen/plugins/event_patternfit.py
GiovanniVolta: Approaching high S1, the secant method has hard time to find the right j if we ask it to look between [0, mode] or [mode, n] since the interval is very big. Generally the relation:

- for k < mode: P(mode; n, p) < P(k; n, p) < P(mode + 1.5 * distance; n, p) 
- for k > mode: P(mode - 1.5 * distance; n, p) < P(k; n, p) < P(mode; n, p) 

is not true. Anyway, it is reasonably true with a tolerance of 1e-6. However, we could relax it a little, for instance to 2.
The only problem with finding j is when we are extremely close to the mode - especially for high S1, above 1e4. Here, no matter the range, the tolerance is worst. But I don't think it is a problem since we are far from the "bad" p-values.

Here a snip code for testing it with the current binomial test:
```
def valide_distance(f, n, p, step=0.001):
    print(f'n: {n:.6f}\tp: {p:.6f}')
    mode = lbinom_pmf_mode(0, n, 0, (n, p))
    print(f'mode: {mode}')
    k_check = np.arange(0, n, step)
    for k in tqdm.tqdm(k_check):
        target = lbinom_pmf(k, n, p)
        
        distance = abs(mode - k)
        if k != 0 and k != n and binom_pmf(k, n, p) > 1e-06:
            if k < mode: 
                    j_min = mode
                    j_max = min(mode + f * distance, n)
                    j = lbinom_pmf_inverse(j_min, j_max, target, (n, p))
                    if not np.isnan(j) and j != n:
                        if not binom_pmf(j_min, n, p) > binom_pmf(k, n, p) > binom_pmf(j_max, n, p):
                            print(j_min, k, j_max)
                            print(binom_pmf(j_min, n, p), binom_pmf(k, n, p), binom_pmf(j_max, n, p))
                            break
                        if not np.isclose(binom_pmf(j, n, p), binom_pmf(k, n, p), rtol=0, atol=1e-06, equal_nan=False):
                            print(j, k)
                            print(binom_pmf(j, n, p), binom_pmf(k, n, p))
                            break
            elif k > mode:
                    j_min = max(mode - f * distance, 0)
                    j_max = mode
                    j = lbinom_pmf_inverse(j_min, j_max, target, (n, p))
                    if not np.isnan(j) and j != n:
                        if not binom_pmf(j_min, n, p) < binom_pmf(k, n, p) < binom_pmf(j_max, n, p):
                            print(j_min, k, j_max)
                            print(binom_pmf(j_min, n, p), binom_pmf(k, n, p), binom_pmf(j_max, n, p))
                            break
                        if not np.isclose(binom_pmf(j, n, p), binom_pmf(k, n, p), rtol=0, atol=1e-06, equal_nan=False):
                            print(j, k)
                            print(binom_pmf(j, n, p), binom_pmf(k, n, p))
                            break
            else:
                pass
```

straxen/plugins/event_patternfit.py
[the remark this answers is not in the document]
GiovanniVolta: Yes it is intended. The p-vlaue is 1 and if `ls, rs = k, k`. So either I can leave it like this or I can just `return 1`

straxen/plugins/event_patternfit.py
```diff
@@ -453,22 +448,28 @@ def lbinom_pmf_mode(x_min, x_max, target, args, err=1e-7, max_iter=50):
...
+        if x0 == x1 == args[0] and y0 - target < 0:
+            x1 = np.nan
+    else:
+        x1 = np.nan
+        
     return x1
 
-
 @numba.njit
 def binom_test(k, n, p):
```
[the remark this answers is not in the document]
GiovanniVolta: Yeah, the precision depends on S1. For low S1 we have high tolerance, but this decrease with the increasing of the area. Can we maybe make it as a warning instead of error ? I will then be very clear in the summary note.

straxen/plugins/event_patternfit.py
kdund: For that high S1s, the gaussian approximation is already excellent-- should we use that, perhaps?

straxen/plugins/event_patternfit.py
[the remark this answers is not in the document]
kdund: I think the latter is cleaner for readability

straxen/plugins/event_patternfit.py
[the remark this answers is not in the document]
kdund: Yeah, that sounds good (how to turn test into warning?)

straxen/plugins/event_patternfit.py
GiovanniVolta: Yes sure, do you want to me to do it or can we leave it for the next version? I can open an issue to remember that

GiovanniVolta:
@kdund the tests are working now. In the case that k < mode and j doesn't exist, the algorithm set j = n. This doesn't make any problem for the p-value but for the test yes. Now it should be fine. Let me know what you think. 
Cheers!

straxen/plugins/event_patternfit.py
kdund: open issue is fine

kdund (approved):
Thanks for sticking with it, @GiovanniVolta !

GiovanniVolta:
Do we need another approval?

kdund:
> Do we need another approval?

No, but for straxen I believe we've held off on merging before until we were ready for a new version, or am I misremembering ( @ershockley ?)

ershockley:
Yeah usually we merge PRs that will change lineage close to a new release. This is a high level plugin though so it's probably fine. I leave decision to @WenzDaniel and @JoranAngevaare

JoranAngevaare:
Thanks all, we will merge it when making the new release. We could even consider making one specifically for this one, depending on how urgent this PR is?

>This is a high level plugin though so it's probably fine.

It's indeed high level but one of the computationally more intense plugins

PR #1015, opened 2022-04-13
Numbafy function

WenzDaniel:

In this PR we are speeding up the peak bayes classifer using numba. The basics logic of the code is still the same. I just moved `np.linspace` in front of the loop and used numba friendly buffers (np.zeros) instead of append.

### DIFF
--- a/straxen/plugins/bayes_peak_classification.py
+++ b/straxen/plugins/bayes_peak_classification.py
@@ -1,8 +1,8 @@
 import numpy as np
 import numba
-from scipy.special import logsumexp
 import strax
 import straxen
+from scipy.special import logsumexp
 
 
 class BayesPeakClassification(strax.Plugin):
@@ -21,11 +21,11 @@ class BayesPeakClassification(strax.Plugin):
 
     provides = 'peak_classification_bayes'
     depends_on = ('peaks',)
-    __version__ = '0.0.2'
+    __version__ = '0.0.3'
     dtype = (strax.time_fields
              + [('ln_prob_s1', np.float32, 'S1 ln probability')]
              + [('ln_prob_s2', np.float32, 'S2 ln probability')]
-            )
+             )
 
     # Descriptor configs
     bayes_config_file = straxen.URLConfig(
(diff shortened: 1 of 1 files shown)


WenzDaniel:
Just for completeness below is the code profiling for 5 min of calibration data before/after the change:
bayes_classifier_vanilla
bayes_classifier_numba

I also checked whether we can do something for the `compute_inference` function, but the slowest part of the code is given by those two lines https://github.com/XENONnT/straxen/blob/84dc673495441bb7b308e25e9f3493351ed6a0aa/straxen/plugins/bayes_peak_classification.py#L132-L133

which cannot really be optimized any further.

WenzDaniel:
One more thing for completeness there seems to be a small numerical difference in the computed quantiles between the two versions of the code. I do not know exactly where it is coming from though. For the computed waveforms it is fine.

quantiels_numeric_issue

JoranAngevaare:
Nice work, seems like a 6x speedup?


One question:
>which cannot really be optimized any further.

Isn't there something we can do for the 100-iterations for loop in the line you mentioned ([straxen/straxen/plugins/bayes_peak_classification.py](https://github.com/XENONnT/straxen/blob/84dc673495441bb7b308e25e9f3493351ed6a0aa/straxen/plugins/bayes_peak_classification.py#L132-L133))

I can imagine the many slicing-steps can be written in a more effective way?

ahiguera-mx:
Hi @WenzDaniel 
Thanks a lot for looking at this, I don't have any issue with the changes, the small difference is coming from `np.cumsum()` maybe numba an numpy use different precision (dtypes).
Also for my own education I remove the nested loop and put it in a separate function and got the same performance time-wise, so I guess numba does all the magic.
 @tinapeters Do you have any comment?

ahiguera-mx (approved):

WenzDaniel:
> I can imagine the many slicing-steps can be written in a more effective way?

If you have an idea please let me know. I tried a few things to numbafy this loop, but there was not any performance increase. My impression was that compute the log over so many values simply takes time and there is not anything we can do.

JoranAngevaare:
> > I can imagine the many slicing-steps can be written in a more effective way?
> 
> If you have an idea please let me know. I tried a few things to numbafy this loop, but there was not any performance increase. My impression was that compute the log over so many values simply takes time and there is not anything we can do.

Sure, I could have a try but consider it unlikely that I will fix it, could I re-use your notebook? Is it e.g. on analysiscode?

PR #1017, opened 2022-04-19
Change FDC z offset and add alternative interaction

ftoschi:

Two main changes:
- the field map is now sampled using the z position excluding the gate transit time to be consistent with the FDC map scaled only by the drift speed;
- the `event_positions` plugin returns the corrected values also the alternative S1 and S2. This is changed also for the minianalysis `load_corrected_positions` where there is also the possibility to have both alternative peaks.

It is an easy update of the `EventPositions` plugin by looping over alternative S1 and S2 and splitting the `z_obs` evaluation first including the drift speed, then the FDC map is sampled and eventually the transit gate time is added back to `z_obs` to return the final value to be used.

No new way of using the `event_positions` plugin, the only change is that `load_corrected_positions` minianalysis now takes two new arguments `alt_s1` and `alt_s2` (False by default): if True the alternative respective peak is used as main.


ftoschi:
@ershockley now the plugin `event_positions` will take three times the memory, do you see this as a problem from computing point of view?

straxen/plugins/event_processing.py
```diff
@@ -522,27 +522,32 @@ class EventPositions(strax.Plugin):
...
+    for j in 'r z'.split():
+        dtype += [('_'.join([j, 'field_distortion_correction']), np.float32, f'Correction added to {j}_naive for field distortion (cm)')]
+        for s_i in [1, 2]:
+            dtype += [(f'alt_s{s_i}_{j}_field_distortion_correction', np.float32, f'Correction added to alt_s{s_i}_{j}_naive for field distortion (cm)')]
+            
+    dtype += [('theta', np.float32, f'Main interaction angular position (radians)')]
+    for s_i in [1, 2]:
+        dtype += [(f'alt_s{s_i}_theta', np.float32, f'Alternative S{s_i} interaction angular position (radians)')]
+    
+    dtype = dtype + strax.time_fields
```
JoranAngevaare, line 550:
```suggestion
def infer_dtype(self):
        dtype = []
        for j in 'x y r'.split():
            comment = f'Main interaction {j}-position, field-distortion corrected (cm)'
            dtype += [(j, np.float32, comment)]
            for s_i in [1, 2]:
                comment = f'Alternative S{s_i} interaction {j}-position, field-distortion corrected (cm)'
                field = f'alt_s{s_i}_{j}_fdc'
                dtype += [(field, np.float32, comment)]

        for j in ['z']:
            comment = 'Interaction z-position, using mean drift velocity only (cm)'
            dtype += [(j, np.float32, comment)]
            for s_i in [1, 2]:
                comment = f'Alternative S{s_i} z-position, using mean drift velocity only (cm)'
                field = f'alt_s{s_i}_z'
                dtype += [(field, np.float32, comment)]

        naive_pos = []
        fdc_pos = []
        for j in 'r z'.split():
            naive_pos += [(f'{j}_naive',
                           np.float32,
                           f'Main interaction {j}-position with observed position (cm)')]
            fdc_pos += [(f'{j}_field_distortion_correction',
                         np.float32,
                         f'Correction added to {j}_naive for field distortion (cm)')]
            for s_i in [1, 2]:
                naive_pos += [(
                    f'alt_s{s_i}_{j}_naive',
                    np.float32,
                    f'Alternative S{s_i} interaction {j}-position with observed position (cm)')]
                fdc_pos += [(f'alt_s{s_i}_{j}_field_distortion_correction',
                             np.float32,
                             f'Correction added to alt_s{s_i}_{j}_naive for field distortion (cm)')]
        dtype += naive_pos + fdc_pos
        for s_i in [1, 2]:
            dtype += [(f'alt_s{s_i}_theta',
                       np.float32,
                       f'Alternative S{s_i} interaction angular position (radians)')]

        dtype += [('theta', np.float32, f'Main interaction angular position (radians)')]
        return dtype + strax.time_fields
```
I had a hard time following this, perhaps something like this makes it a bit easier to read?

straxen/plugins/event_processing.py
```diff
@@ -560,45 +565,53 @@ def setup(self):
...
-                       'r_naive': r_obs,
-                       'r_field_distortion_correction': delta_r,
-                       'theta': np.arctan2(orig_pos[:, 1], orig_pos[:, 0]),
-                       'z_naive': z_obs,
-                       # using z_obs in agreement with the dtype description
-                       # the FDC for z (z_cor) is found to be not reliable (see #527)
-                       'z': z_obs,
-                       'z_field_distortion_correction': delta_z
-                       })
+        for s_i in [0, 1, 2]:
```
JoranAngevaare, line 574: I was confused for a bit, maybe we could make this a bit more explicit by doing something like 
```python
for field, xy_pos in zip(
   ['drift_time', 'alt_s1_interaction_drift_time', 'alt_s2_interaction_drift_time'],
   ['s2_', 'alt_s2_', 'alt_s2_']
):
or something similar. Otherwise, it certainty would help if `s_i` -> `alt_s_i`  there would be a comment that `alt_s_i == 0` is the main pair.

**Review by JoranAngevaare — COMMENTED** [review]
Hi Francesco, 
Just to be sure, where you would use this PR for does not clash with the previous re-definition of alt-S2 https://github.com/XENONnT/straxen/pull/890?

>now the plugin event_positions will take three times the memory, do you see this as a problem from computing point of view?

This was to be expected since this PR ~triples the number of fields for event_positions. These fields are also all added to `event_info` by the way, which will have a total of 135 fields after this PR. You might see that people will request more memory on dali to accommodate for these ~20 extra fields in `event_info`.

From a technical point of view, the PR looks good, thanks! I'd have two request for reformatting it a bit for readability and using `infer_dtype` to avoid unnecessary computation on `import straxen`.

File: straxen/plugins/event_processing.py (hunk 1 above)
Comment (ftoschi) [reply straxen/plugins/event_processing.py:550]: I think this is nicer,  I am thinking that there are some fields that do not really make sense (e.g., `alt_s1_theta` and `alt_s1_r_naive` are identical to `theta` or `r_naive`). Would it be better not to include them or is it worth to have more space but having more homogeneity in the field names?

File: straxen/plugins/event_processing.py (hunk 2 above)
Comment (JoranAngevaare) [reply straxen/plugins/event_processing.py:592]: yeah it would be only that, 👍

File: straxen/plugins/event_processing.py [hunk 3]
```diff
@@ -560,45 +583,54 @@ def setup(self):
...
+                invalid = np.abs(z_obs) < np.abs(delta_r)
+                # do not apply z correction above gate
+                invalid |= z_obs >= 0
+            z_cor[invalid] = z_obs[invalid]
+            delta_z = z_cor - z_obs
+            
+            pre_field = '' if s_i == 0  else f'alt_s{s_i}_'
+            post_field = '' if s_i == 0  else '_fdc'
+            result.update({f'{pre_field}x{post_field}': orig_pos[:, 0] * scale,
+                           f'{pre_field}y{post_field}': orig_pos[:, 1] * scale,
```
WenzDaniel, line 623: I think you have an edge case here. In case x/y is 0 and dr is zero too. You will get here x/y = np.nan. Because youe scale will be 0/0. Might be unlikely but I think it is better to avoid this edge case.

straxen/plugins/event_processing.py
```diff
@@ -560,45 +583,54 @@ def setup(self):
...
+                           f'{pre_field}r{post_field}': r_cor,
+                           f'{pre_field}r_naive': r_obs,
+                           f'{pre_field}r_field_distortion_correction': delta_r,
+                           f'{pre_field}theta': np.arctan2(orig_pos[:, 1], orig_pos[:, 0]),
```
WenzDaniel, line 627: In the neutron-veto we use angles from 0 to 360  starting at y=0 counter clockwise. But we can switch to [-pi, pi) too.  I do not have any strong opinion about it.

WenzDaniel (approved):
Thanks @ftoschi looks good. A few small comments and one edge case I think we should fix.

l-althueser:
As mentioned today in the meeting. Would be good to have a minor straxen release with and without this change if we are going to merge soon. Just to be sure that we can go one step back again in case that the change has an influence on the current corrections. I would not expect something significant but this would be a good fallback point. 
All assuming that we make other merges that soonish that are needed to move forward.

straxen/plugins/event_processing.py
```diff
@@ -522,27 +522,50 @@ class EventPositions(strax.Plugin):
                 '?version=ONLINE&run_id=plugin.run_id',
         help='Electron drift time from the gate in ns',
         cache=True)
+    
+    def infer_dtype(self):
+        dtype = []
+        for j in 'x y r'.split():
```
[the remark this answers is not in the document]
ftoschi: It is indeed trivial, but on the other hand this is rather used and it is something I guess most of the people would define by themselves if it was not available, leading anyway to the same amount of used RAM. From the storage point of view this could help, but being anyway just at the even level I do not think it is a real issue

straxen/plugins/event_processing.py
```diff
@@ -522,27 +522,50 @@ class EventPositions(strax.Plugin):
...
+            for s_i in [1, 2]:
+                comment = f'Alternative S{s_i} interaction {j}-position, field-distortion corrected (cm)'
+                field = f'alt_s{s_i}_{j}_fdc'
+                dtype += [(field, np.float32, comment)]
+
+        for j in ['z']:
+            comment = 'Interaction z-position, using mean drift velocity only (cm)'
+            dtype += [(j, np.float32, comment)]
+            for s_i in [1, 2]:
+                comment = f'Alternative S{s_i} z-position, using mean drift velocity only (cm)'
```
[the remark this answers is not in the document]
ftoschi: Yeah, that can be added

straxen/plugins/event_processing.py
```diff
@@ -560,45 +583,54 @@ def setup(self):
...
-                       'z_field_distortion_correction': delta_z
-                       })
+       
+	# s_i == 0 indicates the main event, while s_i != 0 means alternative S1 or S2 is used based on s_i value 
+        for s_i in [0, 1, 2]:
+            
+            drift_time = events['drift_time']  if not s_i else events[f'alt_s{s_i}_interaction_drift_time']
+            
+            z_obs = - self.electron_drift_velocity * drift_time
+            xy_pos = 's2_' if s_i != 2 else 'alt_s2_'
```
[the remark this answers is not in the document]
ftoschi: The alternative drift time is computed in another plugin, so I am not sure if this is the place where this should be explained as here it is simply used for one of its original purpose (to just call it without having to re-evaluate it). I can implement the comment at the beginning of the loop saying that one peak is alternative and the other is main

straxen/plugins/event_processing.py
```diff
@@ -560,45 +583,54 @@ def setup(self):
...
+            # apply radial correction
+            with np.errstate(invalid='ignore', divide='ignore'):
+                r_cor = r_obs + delta_r
+                scale = r_cor / r_obs
+
+            # z correction due to longer drift time for distortion
+            # (geometrical reasoning not valid if |delta_r| > |z_obs|,
+            #  as cathetus cannot be longer than hypothenuse)
+            with np.errstate(invalid='ignore'):
+                z_cor = -(z_obs ** 2 - delta_r ** 2) ** 0.5
```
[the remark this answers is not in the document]
ftoschi: There is already a comment before the calculation: is that not enough? It is really just the Pythagorean theorem assuming a straight field line

straxen/plugins/event_processing.py
ftoschi: True, on the other hand this is true for numerical zeros, very unlike for floating points. I can take care of it forcing it to zero, but I do really think this edge case might actually never happen

straxen/plugins/event_processing.py
ftoschi: As `arctan2` is a rather straightforward way to evaluate the angles, I feel like it would be better to keep it this way. Also I do not think it is a problem if there is a mismatch between the two detectors, but I can understand it can be a little annoying

straxen/plugins/event_processing.py
WenzDaniel: Yes, but I would prefer to fix it. These things usually tend to appear at some point. And it is not difficult to fix.

straxen/plugins/event_processing.py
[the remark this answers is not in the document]
WenzDaniel: I had also something very simplistic in mind like:
```
# alt drift times are always computed between main Sx and alt Sy
``` 
I agree that this is not the place for a full explanation. Just a small reminder

straxen/plugins/event_processing.py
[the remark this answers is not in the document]
WenzDaniel: Okay fine with.

straxen/plugins/event_processing.py
WenzDaniel: Okay

straxen/plugins/event_processing.py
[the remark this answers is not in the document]
WenzDaniel: Okay together with 
> |delta_r| > |z_obs| as cathetus cannot be longer than hypothenuse

it is indeed more clear. Before, I naively flipped z_corr and z_obs and I did not understand your formula.

ftoschi:
Hi everyone, what is the status of this PR? I implemented the changes that were required, should I change something else or is there any other constraint to merge?

GiovanniVolta:
Hey Francesco,
do you perhaps have processed some data, maybe Krypton, so we can play a bit and check the difference ?

JoranAngevaare:
@ftoschi could you let us know what the status is from this PR?

ftoschi:
@JoranAngevaare CodeFactor is not complaining anymore, so if the rest of the tests are happy as they were before, then we can merge

JoranAngevaare:
okay, I'm fine with merging 👍

PR #1019, opened 2022-04-20
modified corrected_area

Jianyu010:

Modified CorrectedArea to enable partitioned TPC (AB and CD).

Added `partitions_event.py` under `plugins` directory. Then it is called in `event_processing.py` and used in `CorrectedArea`

### _Notes on testing_

### DIFF
--- a/straxen/plugins/event_processing.py
+++ b/straxen/plugins/event_processing.py
@@ -7,8 +7,9 @@ from .position_reconstruction import DEFAULT_POSREC_ALGO
 from straxen.common import pax_file, get_resource, first_sr1_run
 from straxen.get_corrections import get_correction_from_cmt, get_cmt_resource, is_cmt_option
 from straxen.itp_map import InterpolatingMap
-export, __all__ = strax.exporter()
 
+from partitions_event import A_and_B_selection_event, C_and_D_selection_event
+export, __all__ = strax.exporter()
 
 @export
 @strax.takes_config(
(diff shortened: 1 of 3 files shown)


Jianyu010:
I am not sure what the output of `straxen.URLConfig` is... Please checkout [here](https://github.com/XENONnT/straxen/blob/3171cb38c44726843456c6c0481bf19ffcac31b5/straxen/plugins/event_processing.py#L719). The use is pretty straightforward if my understanding of this function is correct.

JoranAngevaare:
Hi Jianyu, thanks for the PR! If I understand the gist correctly, you want to add the functionality to allow a position dependent correction? Would it in that case perhaps make sense [to make a map ](https://github.com/XENONnT/straxen/blob/partitioned_TPC/straxen/itp_map.py#L59) out of it (@l-althueser ?). This would require less hard-coding and make it simpler to update and maintain?

l-althueser:
I guess we need something in between. We want to have like a map but with per run time dependence. So we might need a file with arbitrary volume definitions that works well with per run numbers or such.

Jianyu010:
@JoranAngevaare This is more treating two partitions of the TPC differently in dealing with time dependent corrections. The note can be found [here](https://xe1t-wiki.lngs.infn.it/doku.php?id=jlong:sr0_2_region_se_correction). A hard coded partitioning is needed for the moment.

PR #1022, opened 2022-04-20
Start testing with PluginTestingSuite, fix #881

### LINKED ISSUE #881: Event level S1 threshold must be reflected in triggering peaks
Currently, we only set the [S1 PMT threshold](https://github.com/XENONnT/straxen/blob/4e288fd1283f99564b990c687d982cb8c9162cbc/straxen/plugins/event_processing.py#L132) during the event-building step but not during the search for triggering peaks. If we allow S1 as triggering peaks and increase the threshold to be larger than the peak-level threshold this will lead to ambiguous events.

JoranAngevaare:
This PR is a fix #881, it also starts refactoring the plugin tests with a simple example for #881, I will open a separate PR for doing a bit of refactoring to use this PR efficiently.

The documentation follows, [see this file](https://github.com/XENONnT/straxen/blob/excl_tc2_trigg_peaks/tests/plugins/README.md):

-------------

# Test for plugins

In this directory, we accumulate tests for the plugins we are using. 
This makes data-manipulation and plugin configuration tests development easy.

## Writing new tests
Write a new test by adding a `.py` file and write a function that should be run
in the testing suite by adding this decorator:
```python
from _core import PluginTestAccumulator


@PluginTestAccumulator.register('test_example')
def test_example(self, # You should always accept self as an argument!
                ):
    raise ValueError('Test failed')
```

Make sure that the new `.py`-file is imported in the `test_plugins.py` file 
(see explanation below).

## Example
See for example this event_building.py-file. Here we add 
a test that works on event_basics (by changing a few configurations and 
checking that the output is roughly what we expect.

## Why this organization (technical - only for developers)
At the time of writing, there are ~132 different dependencies. Testing them 
does require quite some computational power if we all want to compute them, 
especially if we want to start computation from scratch (to avoid trying to a 
test that turns out to rely on data that was scrambled).

A few design consideration were taken into account:
 - We want a single class with a `tearDownClass`-method to clean up the data 
   after running. If we would have many classes doing this, you would end up 
   producing the same data many times. If you don't have a `tearDownClass` with
   storage cleanup, you could end up working on cached data which can cause a 
   lot of frustration  https://github.com/XENONnT/straxen/pull/923.
 - Add a modular approach where new tests can be added (in a for loop). 
   While [subtests](https://docs.python.org/3/library/unittest.html#distinguishing-test-iterations-using-subtests) 
 - should be a nice tool for doing this, it turns out to be not as transparent 
   as it claims. Furthermore, you somehow still need to have the entire class 
   in one file, whereas we'd lile to split these tests into small, targeted 
   files.
 - In order to accumulate the test functions, we made a dedicated class 
   `PluginTestAccumulator` that just is a dump for the tests that we write in
   various places. We can't very cleanly re-import `unittest.TestCase`-classes,
   it seems to break the `pytest` (not doing proper setup- and teardowns).

With this setup, we can very easily write tests at any level of data without 
doing a lot of re-computes of already stored data while providing a simple 
infrastructure for developers.

### DIFF
--- a/straxen/plugins/event_processing.py
+++ b/straxen/plugins/event_processing.py
@@ -36,6 +36,11 @@ export, __all__ = strax.exporter()
                  default=True, type=bool,
                  help='If true exclude S1s as triggering peaks.',
                  ),
+    strax.Option(
+        name='event_s1_min_coincidence',
+        default=2, infer_type=False,
+        help="Event level S1 min coincidence. Should be >= s1_min_coincidence "
+             "in the peaklet classification"),
 )
 class Events(strax.OverlapWindowPlugin):
     """
@@ -56,7 +61,7 @@ class Events(strax.OverlapWindowPlugin):
     depends_on = ['peak_basics', 'peak_proximity']
     provides = 'events'
     data_kind = 'events'
-    __version__ = '0.1.0'
+    __version__ = '0.1.1'
     save_when = strax.SaveWhen.NEVER
 
     electron_drift_velocity = straxen.URLConfig(
@@ -92,6 +97,11 @@ class Events(strax.OverlapWindowPlugin):
         _is_triggering &= (peaks['n_competing'] <= self.config['trigger_max_competing'])
         if self.config['exclude_s1_as_triggering_peaks']:
             _is_triggering &= peaks['type'] == 2
+        else:
+            is_not_s1 = peaks['type'] != 1
+            has_tc_large_enough = (peaks['tight_coincidence']
+                                   >= self.config['event_s1_min_coincidence'])
+            _is_triggering &= (is_not_s1 | has_tc_large_enough)
 
         triggers = peaks[_is_triggering]
 
--- a/straxen/plugins/event_processing.py
+++ b/straxen/plugins/event_processing.py
@@ -1,11 +1,10 @@
-from re import X
 import strax
 import numpy as np
 import numba
 import straxen
 from .position_reconstruction import DEFAULT_POSREC_ALGO
 from straxen.common import pax_file, get_resource, first_sr1_run
-from straxen.get_corrections import get_correction_from_cmt, get_cmt_resource, is_cmt_option
+from straxen.get_corrections import get_cmt_resource, is_cmt_option
 from straxen.itp_map import InterpolatingMap
 export, __all__ = strax.exporter()
 
--- a/straxen/plugins/event_processing.py
+++ b/straxen/plugins/event_processing.py
@@ -83,11 +83,11 @@ class Events(strax.OverlapWindowPlugin):
         if self.config['s1_min_coincidence'] > self.config['event_s1_min_coincidence']:
             raise ValueError('Peak s1 coincidence requirement should be smaller '
                              'or equal to event_s1_min_coincidence')
-        self.right_extension = self.config['right_event_extension']
         self.drift_time_max = int(self.config['max_drift_length'] / self.electron_drift_velocity)
         # Left_extension and right_extension should be computed in setup to be
         # reflected in cutax too.
         self.left_extension = self.config['left_event_extension'] + self.drift_time_max
+        self.right_extension = self.config['right_event_extension']
 
     def get_window_size(self):
         # Take a large window for safety, events can have long tails
--- a/straxen/plugins/event_processing.py
+++ b/straxen/plugins/event_processing.py
@@ -35,11 +35,12 @@ export, __all__ = strax.exporter()
                  default=True, type=bool,
                  help='If true exclude S1s as triggering peaks.',
                  ),
-    strax.Option(
-        name='event_s1_min_coincidence',
-        default=2, infer_type=False,
-        help="Event level S1 min coincidence. Should be >= s1_min_coincidence "
-             "in the peaklet classification"),
+    strax.Option(name='event_s1_min_coincidence',
+                 default=2, infer_type=False,
+                 help="Event level S1 min coincidence. Should be >= "
+                      "s1_min_coincidence in the peaklet classification"),
+    strax.Option(name='s1_min_coincidence', default=2, type=int,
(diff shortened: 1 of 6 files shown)


jmosbacher:
I like this, Ill review the actual code as soon as i have time.
@JoranAngevaare Just wondering if we could take it a step further and use the strax plugin framework to manage the plugin tests? So instead of registering methods on the TestCase, you inherit from a TestPlugin which then specifies what the dependencies are and performs the test in the compute method and produces booleans that indicate whether the tests passed or failed. The upside would be that the logic of whether to raise an exception or warning would be independent of the test logic so we could imagine maybe calling these with warnings-only when processing and raise exceptions when doing the unit tests.
Creating a test context would be as simple as just registering all the subclasses of the TestPlugin.

JoranAngevaare:
> Just wondering if we could take it a step further and use the strax plugin framework to manage the plugin tests? So instead of registering methods on the TestCase, you inherit from a TestPlugin which then specifies what the dependencies are and

Intriguing idea, I will think about it. I actually designed it the other way around, having `pytest` actually managing all the tests, which really helps understanding trackbacks, building coverage files et cetera. The goal was to have at least one test per plugin, (being recognized as such by pytest) and allowing it to be easily extended to include tests like in the `event_building.py`-file.

jmosbacher:
> which really helps understanding trackbacks, building coverage files et cetera

Right that's a big upside for your approach.  Was brainstorming, not claiming i know which approach would be better overall.

JoranAngevaare:
> > which really helps understanding trackbacks, building coverage files et cetera
> 
> Right that's a big upside for your approach. Was brainstorming, not claiming i know which approach would be better overall.

No definitely worth considering - I'll think about it. I'm sure that there is also a way of combining the two, would just take a bit of reading up on `pytest` and `unittest`.

straxen/plugins/event_processing.py
```diff
@@ -36,6 +35,11 @@
                  default=True, type=bool,
                  help='If true exclude S1s as triggering peaks.',
                  ),
+    strax.Option(
+        name='event_s1_min_coincidence',
+        default=2, infer_type=False,
+        help="Event level S1 min coincidence. Should be >= s1_min_coincidence "
```
WenzDaniel, line 41: I am wondering if we should not add the condition as an assert statement.

tests/plugins/test_plugins.py
```diff
@@ -0,0 +1,48 @@
...
+        cls.run_id = nt_test_run_id
+
+    @classmethod
+    def tearDownClass(cls) -> None:
+        """
+        Removes test data after tests are done.
+        """
+        path = os.path.abspath(cls.st.storage[-1].path)
+        for file in os.listdir(path):
+            shutil.rmtree(os.path.join(path, file))
```
WenzDaniel, line 35: I think we should follow https://stackoverflow.com/a/50649804

tests/plugins/event_building.py
```diff
@@ -0,0 +1,35 @@
...
+    events = st.get_array(self.run_id, 'event_basics')
+    new_min_coincidence = int(np.max(events['s1_tight_coincidence']) - 1)
+
+    # Create an alternative config, but with less
+    st_alt = st.new_context()
+    st_alt.set_config(dict(event_s1_min_coincidence=new_min_coincidence))
+    events_alt = st_alt.get_array(self.run_id, 'event_basics')
+
+    # There should be more "alt" events since we don't concatenate as
+    # many peaks into one event
```
WenzDaniel, line 25: I think this statement is a bit fragile. You could also have less events because you can trigger on below cathode events etc. I think it would be more robust to really check the triggering conditions like the s1 coincidence and peak area to be larger than the specified settings.

WenzDaniel (approved):
Looks cool. Not sure how and when we should migrate our tests. I guess a separate PRs would make sense.

straxen/plugins/event_processing.py
JoranAngevaare: sure can add - not strictly the scope of this PR though

tests/plugins/event_building.py
JoranAngevaare: Right, a bit of the issue is that it's not super clear which peaks are triggering at the event-basics stage. Any suggestions to this effect? Could probably find some at least based on the extensions the last and first triggering peaks.

WenzDaniel: I think you can ask if the event has either an "s2 larger peak area and smaller n_competing" or an "S1 larger coincidence, peak area and smaller n_competing.

JoranAngevaare: but you can still have those peaks - they only won't be triggering. We actually designed the S1 to be incorporated even when only the S2 is triggering - so I think I don't see your point. 

I'll just use the fact that the event is extended left and right to see if a peak is triggering.

JoranAngevaare: okay, I just implemented it in [63a0b07](https://github.com/XENONnT/straxen/pull/1022/commits/63a0b0765d2330cb9ff1643dedec5f7761007c2a), I think it's good enough like this

JoranAngevaare:
@jmosbacher I thought about it a bit more and would prefer to keep pursuing the current approach. For me the easy interfacing with `pytest` is the most important. Maybe we will revisit sometime, luckily we can't really break things when working on our testing suite, so we can always further develop it if we want to.

tests/plugins/event_building.py
WenzDaniel: > but you can still have those peaks - they only won't be triggering. We actually designed the S1 to be incorporated even when only the S2 is triggering - so I think I don't see your point.

Yes what I meant is that you are basically checking the trigger condition itself. All events must have either a triggering S1 **or** (not and) a triggering S2 with the given conditions.

JoranAngevaare:
Thanks for the feedback both 😄

PR #1025, opened 2022-04-21
Add 'electron_diffusion_cte' variable from CMT

ricmperes:
The longitudinal diffusion constant value was added to the CMT but not to the `single_value_corrections` list. Without it, `straxen.get_correction_from_cmt(run_id', ('electron_diffusion_cte', version, detector))` leads to an error.
Issue in code found by @l-althueser.

This PR adds support for `straxen.get_correction_from_cmt` to fetch the diffusion constant from CMT.

Main code stays unaltered, `'electron_diffusion_cte'` added to the  `single_value_corrections` list.

`straxen.get_correction_from_cmt('17895', ('electron_diffusion_cte', 'ONLINE', 'nT'))`

### DIFF
--- a/straxen/corrections_services.py
+++ b/straxen/corrections_services.py
@@ -21,7 +21,7 @@ corrections_w_file = ['mlp_model', 'cnn_model', 'gcn_model',
 single_value_corrections = ['elife_xenon1t', 'elife', 'baseline_samples_nv',
                             'electron_drift_velocity', 'electron_drift_time_gate',
                             'se_gain', 'rel_extraction_eff', 'relative_light_yield',
-                            'avg_se_gain']
+                            'avg_se_gain', 'electron_diffusion_cte']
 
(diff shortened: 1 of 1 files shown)


ahiguera-mx:
As far as I know this is not use anywhere in straxen, right?  but we retrieve the information via CMT therefore this patch.
The electron diffusion cte is use on cutax, right?

ricmperes:
Right. It's only used in cutax for the S2Width cut to make it time-dependent and/or directly dependent to CMT instead of having hardcoded values ([here ](https://github.com/XENONnT/cutax/blob/caa0493a2d36057c74b4ad57ec3b7e6d5e7b4e08/cutax/cuts/s2_width.py#L72)in PR).

WenzDaniel:
Tagging https://github.com/XENONnT/cutax/pull/166/files for completeness.

PR #1027, opened 2022-04-27
Partitioned tpc

Jianyu010:
This is a straxen enabling regional corrections to be applied. Please check out the dedicated notes [1](https://xe1t-wiki.lngs.infn.it/doku.php?id=jlong:sr0_2_region_se_correction) and [2](https://xe1t-wiki.lngs.infn.it/doku.php?id=jlong:sr0_extraction_efficiency_correction_v3)

## Can you give a minimal working example (or illustrate with a figure)?

_Please include the following if applicable:_
  - [ ] _Update the docstring(s)_
  - [ ] _Update the documentation_
  - [ ] _Tests to check the (new) code is working as desired._
  - [ ] _Does it solve one of the open issues on github?_

### _Notes on testing_
 - _Until the automated tests pass, please mark the PR as a draft._
 - _On the XENONnT fork we test with database access, on private forks there is no database access for security considerations._

All _italic_ comments can be removed from this template.

PR #1031, opened 2022-05-03
tweaks to #1027 - single more general protocol

JoranAngevaare:
This PR just tweaks the protocols in #1027 to make it a bit more general. Core idea is still @ershockley 's 

## Differences
 - Only difference is that now the protocol would return the keys as they are requested from the json`'gain_ab'` instead of `'ab'` - depending on what the columns are in the file.
 - Call the time column `'time'` instead of `'times'` or `'timestamps'`

## Updates
- I've also added a few words of documentation and also expanded the test to work on CSV files 
- testing code on the CorrectedAreas-plugin

## Further generalization options
 - Not sure if the comma separation `keys = strax.to_str_tuple(itp_dict_keys.split(','))` could be handled by the URLConfig

### DIFF
--- a/straxen/plugins/event_processing.py
+++ b/straxen/plugins/event_processing.py
@@ -785,10 +785,12 @@ class CorrectedAreas(strax.Plugin):
             # corrected s2 with s2 xy map only, i.e. no elife correction
             # this is for s2-only events which don't have drift time info
             
-            cs2_top_xycorr = (events[f'{peak_type}s2_area'] * events[f'{peak_type}s2_area_fraction_top'] / self.s2_xy_map(s2_positions, map_name=s2_top_map_name))
-            cs2_bottom_xycorr = (events[f'{peak_type}s2_area'] *
-                                       (1 - events[f'{peak_type}s2_area_fraction_top']) /
-                                       self.s2_xy_map(s2_positions, map_name=s2_bottom_map_name))
+            cs2_top_xycorr = (events[f'{peak_type}s2_area']
+                              * events[f'{peak_type}s2_area_fraction_top']
+                              / self.s2_xy_map(s2_positions, map_name=s2_top_map_name))
+            cs2_bottom_xycorr = (events[f'{peak_type}s2_area']
+                                 * (1 - events[f'{peak_type}s2_area_fraction_top'])
+                                 / self.s2_xy_map(s2_positions, map_name=s2_bottom_map_name))
             
             # For electron lifetime corrections to the S2s,
             # use drift time computed using the main S1.
--- a/straxen/url_config.py
+++ b/straxen/url_config.py
@@ -345,7 +345,7 @@ def get_itp_dict(loaded_json,
     Interpolate the dictionary at the start time that is queried from
     a run-id.
 
-    The loaded JSON (dictionary), should have at least one collum `times`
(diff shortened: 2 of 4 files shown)


straxen/plugins/event_processing.py
```diff
@@ -785,10 +785,12 @@ def compute(self, events):
...
-            cs2_top_xycorr = (events[f'{peak_type}s2_area'] * events[f'{peak_type}s2_area_fraction_top'] / self.s2_xy_map(s2_positions, map_name=s2_top_map_name))
-            cs2_bottom_xycorr = (events[f'{peak_type}s2_area'] *
-                                       (1 - events[f'{peak_type}s2_area_fraction_top']) /
-                                       self.s2_xy_map(s2_positions, map_name=s2_bottom_map_name))
+            cs2_top_xycorr = (events[f'{peak_type}s2_area']
+                              * events[f'{peak_type}s2_area_fraction_top']
+                              / self.s2_xy_map(s2_positions, map_name=s2_top_map_name))
+            cs2_bottom_xycorr = (events[f'{peak_type}s2_area']
+                                 * (1 - events[f'{peak_type}s2_area_fraction_top'])
+                                 / self.s2_xy_map(s2_positions, map_name=s2_bottom_map_name))
```
JoranAngevaare, line 793: Hmm, this wasn't intended to go into this PR, nevertheless, we try not to make lines as long as this one so probably doesn't hurt adding this change. The other tweak is fixing https://peps.python.org/pep-0008/#should-a-line-break-before-or-after-a-binary-operator

> `cs2_top_xycorr = (events[f'{peak_type}s2_area'] * events[f'{peak_type}s2_area_fraction_top'] / self.s2_xy_map(s2_positions, map_name=s2_top_map_name))`

straxen/url_config.py
```diff
@@ -335,57 +336,45 @@ def open_neural_net(model_path: str, **kwargs):
...
-        if is_partition:
-            interp_ab = interp1d(times, loaded_json['ee_ab'], bounds_error=True)
-            interp_cd = interp1d(times, loaded_json['ee_cd'], bounds_error=True)
-            gain_ab = interp_ab(start)
-            gain_cd = interp_cd(start)
-            return {'ab': gain_ab, 'cd': gain_cd}
+        if len(strax.to_str_tuple(keys))>1:
+            return {key:
+                    interp1d(times, loaded_json[key], bounds_error=True)(start)
+                    for key in keys}
```
JoranAngevaare, line 373: The difference: used to return `{'ab': gain_ab, 'cd': gain_cd}`, now returns `{'gain_ab': gain_ab, 'gain_cd': gain_cd}` (depending on how the json looks)

ershockley:
@JoranAngevaare I made a few changes to make the protocol more generalizable (allows for passing time key as a parameter) in order for it to be backwards compatible. I also had to add an additional protocol that reformats the dict to be what the plugin expects, again for backwards compatibility. 

Personally I still feel like the other protocols are a bit more transparent, even if they aren't as general as these ones, as now the urls get crazy long. But if you prefer this way we can go with this. Let me know what you think.

ershockley:
I've changed my mind, let's go with the general protocols but with longer urls. Probably better for longterm.

ershockley (approved):

PR #1040, opened 2022-05-16
allow dry eval of URL configs

JoranAngevaare:
Allow evaluation of URL configs outside of context.

Sometimes, it's very useful to test the output of url-configs outside of a context. Just to sanity check that we are getting the right values. Until this moment, I always first changed the plugin, than checked that the plugin was giving me the correct value I was asking for. However, apart from the .plugin attributes, the URL config can be used in a broader sense than only for plugins. So I wrote a very small class method to do just that.

I had to scratch my head that it was so simple (of course that's @jmosbacher 's work 🎉) but probably still useful to add for it's nice to have since I'm sure people will also be running into the question _"how do I know I've written some URL config that makes sense?"_

Please see the example also written in the docstring/test:
```python
plugin_url = 'cmt://electron_drift_velocity?run_id=plugin.run_id&version=v3'
self.st.set_config({'test_config': plugin_url})
correct_val = p.test_config

# We can also get it from one of these methods
dry_val1 = straxen.URLConfig.evaluate_dry(
    f'cmt://electron_drift_velocity?run_id={nt_test_run_id}&version=v3')
dry_val2 = straxen.URLConfig.evaluate_dry(
    f'cmt://electron_drift_velocity?version=v3', run_id=nt_test_run_id)

# All methods should yield the same
assert correct_val == dry_val1 == dry_val2
```

### DIFF
--- a/tests/test_url_config.py
+++ b/tests/test_url_config.py
@@ -242,3 +242,28 @@ class TestURLConfig(unittest.TestCase):
         filtered2 = straxen.filter_kwargs(func2, all_kwargs)
         self.assertEqual(filtered2, all_kwargs)
         func2(**filtered2)
+
+    @unittest.skipIf(not straxen.utilix_is_configured(), "No db access, cannot test CMT.")
+    def test_dry_evaluation(self):
+        """
+        Check that running a dry evaluation can be done outside of the
+        context of a URL config and yield the same result.
+        """
+        plugin_url = 'cmt://electron_drift_velocity?run_id=plugin.run_id&version=v3'
+        self.st.set_config({'test_config': plugin_url})
+        p = self.st.get_single_plugin(nt_test_run_id, 'test_data')
+        correct_val = p.test_config
+
+        # We can also get it from one of these methods
+        dry_val1 = straxen.URLConfig.evaluate_dry(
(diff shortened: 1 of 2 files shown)


tests/test_url_config.py
```diff
@@ -242,3 +242,28 @@ def func2(**kwargs):
         filtered2 = straxen.filter_kwargs(func2, all_kwargs)
         self.assertEqual(filtered2, all_kwargs)
         func2(**filtered2)
+
+    @unittest.skipIf(not straxen.utilix_is_configured(), "No db access, cannot test CMT.")
+    def test_dry_evaluation(self):
```
JoranAngevaare, line 247: While not being very general, I think the concrete examples are probably more valuable than writing an extensive unit-test

jmosbacher:
Nice! BTW this was actually addressed in #827 with the `URLConfig.eval` class method. But I am completely fine with adding this stop-gap until we decide what parts of #827 to merge post sr0.

jmosbacher (approved):
Code looks fine to me

JoranAngevaare:
@jmosbacher Thanks, should of course have consulted that PR first.

WenzDaniel (approved):
Nice, thanks a lot certainly useful. So far I also always used the work around via get_single_plugin.

PR #1045, opened 2022-05-30
Gps time corrections

ricmperes:
A new data_kind "gps_sync" is created to compute and store the event timestamps in corrected absolute time (GPS-time).

The new data_kind is a single new array "t_gps" to be added to event level data. It gets computed in two steps:
  1. Finds the TTL GPS pulses coming into the AM from the gps module and their pairs in the ASCII file coming from the module for the correspondant run.
  2. Corrects the timestamp of all events by linearly interpolating between the previous and next sync pulses. 
 
`gps_t = st.get_array(run_id, 'gps_sync')`
correction_hist

### _Notes on testing_


straxen/plugins/acqmon_processing.py
```diff
@@ -240,6 +242,160 @@ def compute(self, aqmon_hits, start, end):
...
+        Compute the corrected time given the previous and next sync pulse.
+          * t0: previous pulse (daq or gps)
+          * t1: previous pulse (daq or gps)
+          * t_evt: uncorrected time of event (in DAQ time)
+        Returns a single value, t_evt in gps-time.
+        '''
+        delta_sync_offset = t0_gps-t0_daq
+        delta_t_evt = t_evt - t0_daq
+        
+        return t0_daq + delta_sync_offset + (t1_gps-t0_gps)/(t1_daq-t0_daq)*delta_t_evt
```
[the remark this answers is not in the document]
darrylmasson: Also, does this need to be its own function? I don't see the linear interpolation as "sufficiently complex".

Also, `t1_gps-t0_gps` is always exactly 10 seconds, yes? Why do we calculate it every time?

straxen/plugins/acqmon_processing.py
```diff
@@ -240,6 +242,160 @@ def compute(self, aqmon_hits, start, end):
...
+            run_id = int(run_id)
+
+        query = {"number": run_id}
+
+        doc = rundb.find_one(query, projection={'start': 1, 'end': 1})
+        
+        start_t = doc['start'].replace(tzinfo=datetime.timezone.utc).timestamp()
+        end_t = doc['end'].replace(tzinfo=datetime.timezone.utc).timestamp()
+        
+        query = {"gps_sec": { '$gte': start_t - 1, '$lte': end_t + 1}, 'channel': 0 }
```
darrylmasson, line 287: If, rather than `start-1` and `stop+1` you add/subtract 10 (or even 11), you should be guaranteed* to get the first and last GPS pulses outside the run. I think this might help with L334 and L342.

* At least since the new GPS reader script, this might not be true for the SR0 data.

straxen/plugins/acqmon_processing.py
```diff
@@ -240,6 +242,160 @@ def compute(self, aqmon_hits, start, end):
...
+        l_gps_evt = np.zeros(len(l_daq_evt))
+        
+        #correlate the values of t_daq_evt with the array l_daq_sync
+        l_idx_evt_sync_daq = np.searchsorted(l_daq_sync, l_daq_evt)
+
+        for counter,_t_evt in enumerate(l_daq_evt):
+            index_in_matched_pulses = l_idx_evt_sync_daq[counter]
+            
+            if index_in_matched_pulses == 0: # events before the first pulse
+                _t0_daq = l_daq_sync[0] - 10*1e9
```
darrylmasson, line 333: Are we sure about this exact 10s offset? We know the DAQ clock drifts at 0.2ppm, we should be off by 2us in this interval. (Same for L341)

straxen/plugins/acqmon_processing.py
```diff
@@ -240,6 +242,160 @@ def compute(self, aqmon_hits, start, end):
         result = result[sort]
         return result
 
+@export
+class GPS_sync(strax.Plugin):
+    """
+    Correct the event times to GPS time. 
+      1. Finds the TTL GPS pulses coming into the AM from the gps 
+    module and their pairs in the ASCII file coming from the module
```
darrylmasson, line 250: There are no ASCII files now, correct? Everything's been ported into the database?

straxen/plugins/acqmon_processing.py
```diff
@@ -240,6 +242,160 @@ def compute(self, aqmon_hits, start, end):
...
+    for the correspondant run.
+      2. Corrects the timestamp of all events by linearly interpolating
+    between the previous and next sync pulses. 
+    """
+        
+    __version__ ='0.1.0'
+    depends_on = ('aqmon_hits', 'event_info')
+    provides  = 'gps_sync'
+    data_kind = 'gps_sync'
```
darrylmasson, line 259: Is this actually a new data kind? It seems to me to be an `event`.

darrylmasson (commented):
Overall looks good, a couple of relatively minor questions.

straxen/plugins/acqmon_processing.py
ricmperes: This was actually implemented in an earlier version when reading from files but from the collection I asusmed would not be needed. I'm not against bring it back, it might help recover some runs but will force to check if any pulses need to be cut, using `cut_outside_run` from L291.

straxen/plugins/acqmon_processing.py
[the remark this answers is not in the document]
ricmperes: If there are any missed pulses the difference might be different (~20s) and therefore better to have the function robust on that parameter.

straxen/plugins/acqmon_processing.py
ricmperes: That's right, we do know it's not exactly 10s. This is an approximation to correct the events happening before the first pulse (and the analogous after the last). I could change to use the two next ones and extrapolate and that might be a more reliable and correct way.

straxen/plugins/acqmon_processing.py
```diff
@@ -240,6 +242,160 @@ def compute(self, aqmon_hits, start, end):
...
+          * t1: previous pulse (daq or gps)
+          * t_evt: uncorrected time of event (in DAQ time)
+        Returns a single value, t_evt in gps-time.
+        '''
+        delta_sync_offset = t0_gps-t0_daq
+        delta_t_evt = t_evt - t0_daq
+        
+        return t0_daq + delta_sync_offset + (t1_gps-t0_gps)/(t1_daq-t0_daq)*delta_t_evt
+
+    def compute_time_array(self, l_daq_sync, l_gps_sync, l_daq_evt):
```
jmosbacher, line 323: Did you measure the average time this function takes? is it worth using numba or is it most of the time spent loading the data from the db anyway?

straxen/plugins/acqmon_processing.py
```diff
@@ -240,6 +242,160 @@ def compute(self, aqmon_hits, start, end):
...
+        hits = aqmon_hits
+        evts = event_info
+
+        # Load pulses from aqmon
+        aqmon_array = self.load_aqmon_array(hits)
+        
+        #Load GPS-module pulses
+        gps_array = self.load_gps_array()
+
+        # Make sure first and last pulses match
```
jmosbacher, line 365: dont forget to remove all this commented out code before the PR is merged

straxen/plugins/acqmon_processing.py
```diff
@@ -240,6 +242,160 @@ def compute(self, aqmon_hits, start, end):
...
+
+        rundb = utilix.xent_collection()
+        gps_times = utilix.xent_collection(collection='gps_sync',  database='xenonnt')
+
+        if isinstance(run_id, str):
+            run_id = int(run_id)
+
+        query = {"number": run_id}
+
+        doc = rundb.find_one(query, projection={'start': 1, 'end': 1})
```
jmosbacher, line 282: probably should assert that the doc is not None here (if there is no run that matches the run_id doc will be None)

jmosbacher (commented):
Looks good. I think Darryl already covered all the important stuff in his review so i just added some minor thoughts none of which is crucial.

straxen/plugins/acqmon_processing.py
ricmperes: Yes. I'll update the docstring.

straxen/plugins/acqmon_processing.py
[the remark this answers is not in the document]
darrylmasson: I'm not sure that making `channel` a config is productive. Ch0 is the 0.1Hz signal, it's the only one we store long-term and the only one present in the AM data. If we start changing anything about that it means we made a major change to how the entire GPS subsystem works.

straxen/plugins/acqmon_processing.py
```diff
@@ -610,3 +612,152 @@ def find_offset_nearest(self, array, value):
         array = np.asarray(array)
         idx = (np.abs(array - value)).argmin()
         return value-array[idx]
+
+@export
+class GPS_sync(strax.Plugin):
```
jmosbacher, line 617:
```suggestion
class GpsSync(strax.Plugin):
```

WenzDaniel:
Adding this PR is not straight forward I am afraid. We need it to be based of v1.7.1 instead of 1.8.0...

PR #1054, opened 2022-06-07
Online hsm develop

mflierm:
This plugin is able to provide online data for the HS monitor. This will allow us to look online to SE data and therefore potentially increase the run time.

Because the data is written to the MongoDB, documents should not be too large. To avoid this, the plugin compares the size of the incoming data to a `max_bytes` size. When the data is bigger than that, it randomly selects '`max_bytes`' amount of peaks. It saves the relevant parameters for the HS monitor plots. Since in the end we are interested in the (SE) peak rate, also a weight is added to each peak. We will later on multiply the calculated rate (in the monitor code) by this weight to correct for data we had to throw away.


straxen/plugins/online_monitor.py
```diff
@@ -178,6 +179,82 @@ def area_width_hist(self, data):
...
+        if peaks_size > self.max_bytes:
+            # Calculate fraction of the data that can be kept
+            # to reduce datasize
+            new_len = int(len(peaks) / peaks_size * self.max_bytes)
+            idx = np.random.choice(np.arange(len(peaks)), replace=False, size=new_len)
+            data = peaks[np.sort(idx)]
+
+        elif peaks_size <= self.max_bytes:
+            data = peaks
+
```
JoranAngevaare, line 241:
```suggestion
        else:  # peaks_size <= self.max_bytes:
            data = peaks
```

straxen/plugins/online_monitor.py
```diff
@@ -178,6 +179,82 @@ def area_width_hist(self, data):
...
+            (('Reconstructed mlp peak y-position', 'y_mlp'),
+             np.float32),
+            (('Width (in ns) of the central 50% area of the peak', 'range_50p_area'),
+             np.float32),
+            (('Fraction of original peaks array length that is saved', 'weight'),
+             np.float32),
+            (('End time of the chunk', 'endtime'),
+             np.int64),
+        ]
+        return dtype
```
JoranAngevaare, line 227:
```suggestion
        dtype = [
            (('Peak integral in PE', 'area'),
             np.float32),
            (('Reconstructed mlp peak x-position', 'x_mlp'),
             np.float32),
            (('Reconstructed mlp peak y-position', 'y_mlp'),
             np.float32),
            (('Width (in ns) of the central 50% area of the peak', 'range_50p_area'),
             np.float32),
            (('Fraction of original peaks array length that is saved', 'weight'),
             np.float32),
        ] + strax.time_fields
        return dtype
```

straxen/plugins/online_monitor.py
```diff
@@ -1,6 +1,7 @@
 import strax
 import numpy as np
 from immutabledict import immutabledict
+from numpy import random
```
JoranAngevaare, line 4:
```suggestion
from numpy import random
import straxen
```

straxen/plugins/online_monitor.py
```diff
@@ -178,6 +179,82 @@ def area_width_hist(self, data):
...
+
+    max_bytes = straxen.URLConfig(
+        default=10e6,
+        track=True,
+        help='Maximum amount of bytes of data for MongoDB document'
+    )
+
+    depends_on = ('peak_basics', 'peak_positions_mlp')
+    provides = 'online_se_monitor'
+    data_kind = 'online_se_monitor'
```
JoranAngevaare, line 207: I think you can call it something more general than SE monitor, the plugin does monitor much more than just SE, what do you think

JoranAngevaare (commented):
Very nice, clean work Maricke. Thanks a lot!

I only have some stylistic comments:

JoranAngevaare (approved):

JoranAngevaare:
Great work, let's merge 👍

PR #1062, opened 2022-07-11
Allow long runs

JoranAngevaare:
if we want to allow > 2h long runs, we should not kill bootstrax before that time. Instead - only kill bootstrax if it is still running for a long time despite the run having ended. Additionally, we can also fail if we did not write any new file for more than 30 minutes which would be a better indicator of hangs than the total processing time.

### DIFF
--- a/bin/bootstrax
+++ b/bin/bootstrax
@@ -15,7 +15,7 @@ How to use
 For more info, see the documentation:
 https://straxen.readthedocs.io/en/latest/bootstrax.html
 """
-__version__ = '1.2.1'
+__version__ = '1.3.0'
 
 import argparse
 from datetime import datetime, timedelta, timezone
@@ -31,7 +31,7 @@ import time
 import traceback
 import numpy as np
 import pymongo
-from psutil import pid_exists, disk_usage
+from psutil import pid_exists, disk_usage, Process
 import pytz
 import strax
 import straxen
@@ -121,13 +121,13 @@ timeouts = {
     'retry_run': 60,
     # Maximum time [s] for strax to complete a processing
     # if exceeded, strax will be killed by bootstrax
-    'max_processing_time': 24*3600,
+    'max_processing_time': 24*36000,
     # Max processing time [s] for bootstrax to finish processing after run ended
     'max_processing_time_after_completion': 6*3600,
     # Every this many [s] we should have at least one new file written to disk
-    'max_no_write_time': 15*60,
+    'max_no_write_time': 30*60,
     # Sleep between checking whether a strax process is alive
-    'check_on_strax': 60,
+    'check_on_strax': 10,
     # Maximum time a run is 'busy' without a further update from
     # its responsible bootstrax. Bootstrax normally updates every
     # check_on_strax seconds, so make sure this is substantially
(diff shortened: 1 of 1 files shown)


bin/bootstrax
```diff
@@ -455,24 +460,24 @@ def now(plus=0):
...
+    for child in parent.children(recursive=True):
+        child.kill()
+    parent.kill()
+
+    # Just make it extra dead
+    os.kill(pid, signal.SIGKILL)
+
+    if pid_exists(pid):
+        message = f"Could not kill process {pid}?!"
+        log_warning(message, priority='fatal')
```
JoranAngevaare, line 480: Old version only killed the parent - resulting in many child processes waiting to die

bin/bootstrax
```diff
@@ -1453,16 +1477,32 @@ def process_run(rd, send_heartbeats=args.production):
...
+                # fail(bla) raises RunFailed so no need for elifs. Make sure to
+                # kill main process before raising errors in fail(bla)
+
+                # Fail because we are taking a very long time to process
                 if t0 < now(-timeouts['max_processing_time']):
+                    kill_process(strax_proc.pid)
                     fail(f"Processing took longer than {timeouts['max_processing_time']} sec")
+
+                # Fail because we are taking a very long time to process despite the run having ended
+                if (run_coll.find_one({'_id': rd['_id']}, projection={'end': 1}).get('end', None) is not None
```
darrylmasson, line 1489: This feels like it should be a helper function. `get_end_time(rd['_id'])` or something

bin/bootstrax
```diff
@@ -1340,6 +1345,25 @@ def run_strax(run_id, input_dir, targets, readout_threads, compressor,
...
+    :return: TZ aware timestamp
+    """
+    last_time = datetime.fromtimestamp(0).replace(tzinfo=pytz.utc)
+    matched_folders = glob(match_location)
+    for folder in matched_folders:
+        # check last three files only
+        chunk_files = sorted(glob(os.path.join(folder, '*')))[-3:]
+        for chunk in chunk_files:
+            # Check that we did not rename this file since the glob above
+            if os.path.exists(chunk):
```
darrylmasson, line 1361: Do we care about the edge case where _all_ the files we check were renamed, so that we end up returning the dawn of time, or is that sufficiently unlikely?

darrylmasson (approved):
I'm sure there's an edge case we aren't considering, but otherwise looks good.

bin/bootstrax
```diff
@@ -1453,16 +1477,32 @@ def process_run(rd, send_heartbeats=args.production):
...
+                if (run_coll.find_one({'_id': rd['_id']}, projection={'end': 1}).get('end', None) is not None
+                    and t0 < now(-timeouts['max_processing_time_after_completion'])):
+                        kill_process(strax_proc.pid)
+                        fail(f"After the run ended, processing did not succeed in "
+                             f"{timeouts['max_processing_time_after_completion']} sec.")
+
+                # Fail because for some reason, we are not writing any new files to disk.
+                if (t0 < now(-timeouts['max_no_write_time'])
+                    and (last_write := last_file_write_time(os.path.join(output_folder, f'{run_id}*'))
+                        ) < now(-timeouts['max_no_write_time'])):
```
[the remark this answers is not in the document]
JoranAngevaare: because if we start processing a run like 3 days later, we don't want to fail.

bin/bootstrax
JoranAngevaare: I think the `os.path.exists(chunk)`  check is already quite conservative - I doubt it every actually returns False, the edge case where _all_ things are renamed is somethings I consider sufficiently unlikely. It's an easy fix if my imagination is ever proven to be too limited 😉

PR #1065, opened 2022-07-17
Fix hanging straxer

jmosbacher:
status thread is keeping straxer from exiting when the work is done. This PR is a trivial fix, just adding `daemon=True` when creating the thread so it exits when the main thread exits.


JoranAngevaare (approved):
Thanks Yossi!

PR #1071, opened 2022-08-04
Enforce endtime at readout level

JoranAngevaare:
Simple example of how to trim all the endtime to the endtime in the rundoc

Read an endtime from the rundoc when the run has ended, clip al the raw-records to be within that time interval


WenzDaniel:
Makes a lot of sense and looks good to me. Makes also a few things easier for vetoing all events too close to the run boundaries.

JoranAngevaare:
This approach is fundamentally flawed due to the timestamps in the rundoc being machine timestamps rather than anything else

PR #1077, opened 2022-09-07
Add dynamic event display docs

WenzDaniel:

Adds dynamic event display tutorial to docs. Before it was quite hidden in the analysis code repository.


WenzDaniel:
Added a small fix for the wiki friendly html conversion.

docs/source/tutorials/EventDisplayInteractive.md
```diff
@@ -0,0 +1,284 @@
...
+
+.. code-block:: python
+
+    tool_box = fig.children[1].children[0]
+    tool_box.visible=False
+    bklt.show(fig)
+    
+
+Record matrix: 
+============
```
JoranAngevaare, line 176:
```suggestion
Record matrix: 
-----------------------------------------------------
```

docs/source/tutorials/EventDisplayInteractive.md
```diff
@@ -0,0 +1,284 @@
...
+
+.. code-block:: python
+
+    fig
+
+
+The re-plotting of the record array requires an active python kernel in the background. This means that the interactivity of this plot gets lost as soon as it is exported as an HTML file.
+
+Event selection:
+=============
```
JoranAngevaare, line 208:
```suggestion
Event selection:
-----------------------------------------------------
```

JoranAngevaare (commented):
Hi Daniel, nice to add!

1. Would we need the `.md` and the notebook both? Otherwise probably adding a link to the `.ipynb` in the proper repo would probably be enough?
2. Would you want to include the HTML file into the md file? I'm sure something like that should be possible with something like `.. image:: `
3. Probably worth changing the headers as below, otherwise they make another section 
image

WenzDaniel:
> Would we need the .md and the notebook both? Otherwise probably adding a link to the .ipynb in the proper repo would probably be enough?

I think since it is a tool provided along the package it make sense to put it here. In my personal opinion, the other repository has become a bit useless as it is too heavy I stopped using it already a long time ago.

> Would you want to include the HTML file into the md file?

Nice idea also though about it, but I did not want to put any unnecessary weight to the repository.

> Probably worth changing the headers as below, otherwise they make another section

Ah good catch, I will change it

JoranAngevaare:
@WenzDaniel should we merge as it is or? I hope that we could somehow embed the html doc into the .md file

dachengx:
Hey @WenzDaniel. How did you generate the https://github.com/XENONnT/straxen/blob/eef2a06b500cc14ce2b3222a514ff4570ef7110e/docs/source/tutorials/EventDisplayInteractive.md from https://github.com/XENONnT/straxen/blob/eef2a06b500cc14ce2b3222a514ff4570ef7110e/notebooks/tutorials/EventDisplayInteractive.ipynb? And shall we keep them both?

dachengx:
I think we can only keep the notebooks because users will want to directly run the codes.

notebooks/tutorials/EventDisplayInteractive.ipynb
```diff
@@ -691,6 +691,9 @@
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
    "version": "3.8.16"
+  },
+  "nbsphinx": {
+    "execute": "never"
   }
```
dachengx, line 697: Here
```
"nbsphinx": {
    "execute": "never"
  }
```

is to disable the execution of the notebook when readthedocs building.

PR #1079, opened 2022-09-14
Transition plugins to URLConfig

jmosbacher:
The goal of this PR is to be a minimally viable transition to URLConfigs, i.e. minimal amount of optimizations.

Changes:

- All strax.Option replaced with straxen.URLConfig
- Allow for unhashable kwargs in a URL. This is useful for cases where e.g a function needs to be passed as a kwarg to a protocol
- Allow overriding kwargvalues in dry URL evaluation.
- Adds protocols to support legacy "constant" config values


straxen/plugins/afterpulse_processing.py
```diff
@@ -8,77 +8,72 @@
...
-        # if hitfinder_thresholds config
-        elif isinstance(self.config['hit_min_amplitude'], str):
-            self.hit_thresholds = straxen.hit_min_amplitude(
-                self.config['hit_min_amplitude'])
-        else:  # int or array
-            self.hit_thresholds = self.config['hit_min_amplitude']
-
+        self.to_pe = self.gain_model
+        self.hit_thresholds = self.hit_min_amplitude
+        self.hit_left_extension, self.hit_right_extension = self.save_outside_hits
```
JoranAngevaare, line 75: Isn't this just a beautiful reduction of repeated code 👍

straxen/plugins/afterpulse_processing.py
```diff
@@ -87,21 +82,21 @@ def compute(self, raw_records):
 
         # calculate baseline and baseline rms
         strax.baseline(records,
-                       baseline_samples=self.config['baseline_samples'],
+                       baseline_samples=self.baseline_samples,
```
JoranAngevaare, line 85: This doesn't hurt, but perhaps - if we start doing a full sweep of strax-options, using the `self.config` reference is not an issue? Just to keep the review readable

JoranAngevaare (commented):
Looks good, I agree totally with this transition. I started a [thread](https://xenonnt.slack.com/archives/C016UJZ090B/p1663153992125459) on some more practical questions on this transition.

straxen/plugins/afterpulse_processing.py
jmosbacher: The issue with leaving the `self.config` code is that it can cause confusion if someone assumes the config can be set to a URL but then gets errors because the plugin is reading the URL directly instead of accessing it through the getattr interface where the URL is interpreted.

JoranAngevaare: okay, thanks for reminding me. Probably this shouldn't be impossible to fix on the plugin side. But okay, maybe we shouldn't over-complicate and indeed just use the `getattr` interface

jmosbacher: Right, in theory we could make a subclass of `immutabledict` that does all the magic when you try to access one of its keys and use it as the config dict for plugins. I actually thought about using this approach instead of descriptors when i first made URLConfig but ended up going with descriptors because i wanted to leave the ability to access the raw config value from the plugin in case someone needs to do some hack

jmosbacher:
I think it would probably be best to have the plugin writers for each affected plugin do a review on changed affecting their code

ahiguera-mx:
@jmosbacher do we want to change the configs option on `event_processing` ?
( I can help with that) or that was left out for a reason

jmosbacher:
@ahiguera-mx I just looked and looks like i had switched event_processing and forgot to push. Just pushed, hopefully i had finished working on it :) ill take a deeper look tomorrow. Alternatively feel free to fix any issues you see and push to this branch

straxen/plugins/pulse_processing.py
```diff
@@ -11,98 +11,18 @@
...
-            'or a tuple like (correction=str, version=str, nT=boolean),'
-            'which means we are using cmt.'
-    )])
+HITFINDER_DEFAULTS = {
+    'hit_min_amplitude': 'cmt://hit_thresholds_tpc?version=ONLINE&run_id=plugin.run_id',
+}
+
+HITFINDER_DEFAULTS_he = {
+    'hit_min_amplitude_he': 'cmt://hit_thresholds_he?version=ONLINE&run_id=plugin.run_id',
+}
```
JoranAngevaare, line 20: @jmosbacher can't we just get rid of this? I.e. just change the default to the url?

straxen/plugins/pulse_processing.py
```diff
@@ -263,27 +237,35 @@ class PulseProcessingHighEnergy(PulseProcessing):
...
-        dtype['records_he'] = strax.record_dtype(self.config["record_length"])
-        dtype['pulse_counts_he'] = pulse_count_dtype(self.config['n_he_pmts'])
+        dtype['records_he'] = strax.record_dtype(self.record_length)
+        dtype['pulse_counts_he'] = pulse_count_dtype(self.n_he_pmts)
         return dtype
 
     def setup(self):
         self.hev_enabled = False
+
+        #FIXME: This looks hacky. Maybe find a better way?
```
JoranAngevaare, line 265: I agree that it looks hacky, now that we are changing the lineage anyway, we can just change this to read len(self.hit_thresholds) right?

straxen/url_config.py
```diff
@@ -207,11 +208,15 @@ def fetch(self, plugin):
...
+
         kwargs = {k: self.fetch_attribute(plugin, v)
                   for k, v in url_kwargs.items()}
 
         # construct a deterministic hash key
-        key = strax.deterministic_hash((url, kwargs))
+        try:
+            key = strax.deterministic_hash((url, kwargs))
+        except TypeError:
+            key = uuid4()
```
JoranAngevaare, line 219: I don't like the look of this, can you explain where/why it's needed?

straxen/url_config.py
```diff
@@ -264,14 +269,16 @@ def evaluate_dry(cls, url: str, **kwargs):
...
-                'extra kwargs)')
         url_arg, url_kwarg = cls.split_url_kwargs(url)
-        return cls.dispatch(cls, url_arg, **url_kwarg, **kwargs)
+
+        combined_kwargs = dict(url_kwarg, **kwargs)
+        for k,v in combined_kwargs.items():
+            if isinstance(v, str) and cls.PLUGIN_ATTR_PREFIX in v:
+                raise ValueError(f'The URL parameter {k} depends on the plugin'
+                                'You must specify the value for this parameter'
+                                'for this URL to be evaluated correctly.')
```
JoranAngevaare, line 279: perhaps worth keeping the example? At least for some people might be lost in abstraction here?

straxen/url_config.py
```diff
@@ -264,14 +269,16 @@ def evaluate_dry(cls, url: str, **kwargs):
...
-        if cls.PLUGIN_ATTR_PREFIX in url:
-            raise ValueError(
-                'You specified at least one parameter that depends on the '
-                'plugin. We cannot fetch this in the dry run. Try replacing '
-                'e.g. plugin.run_id=027000 with run_id=027000 (or pass as an '
-                'extra kwargs)')
         url_arg, url_kwarg = cls.split_url_kwargs(url)
-        return cls.dispatch(cls, url_arg, **url_kwarg, **kwargs)
+
+        combined_kwargs = dict(url_kwarg, **kwargs)
```
JoranAngevaare, line 274: Do you mean?
` combined_kwargs = dict(**url_kwarg, **kwargs)`

straxen/url_config.py
```diff
@@ -428,3 +437,18 @@ def rekey_dict(d, replace_keys='', with_keys=''):
...
+    from .get_corrections import FIXED_TO_PE
+
+    return FIXED_TO_PE[name]
+
+
+@URLConfig.register('fixed-thresholds')
+def get_thresholds(model: str):
+    """Return a fixed value for a given model"""
+    
+    return straxen.hit_min_amplitude(model)
```
JoranAngevaare, line 454: perhaps it's worth considering to call this something like `legacy` or something, since we hopefully shouldn't be introducing new ones of this type? I'd prefer less protocols, especially if it's just to keep 1T stuff alive.

JoranAngevaare (commented):
Thanks @jmosbacher for going through the painstaking process of replacing all this. I just pushed a last few.

Can you update the PR description to list the changes you made that are not just an update of `strax.Option`->`straxen.URLConfig` because you also made some changes in the `urlconfig.py`?

This PR changes everything from raw-records upwards, which is fine by me, but something we need to be all on board for. But seeing again how integrally connected everything became, I think it's the better way. In other words, leaving out peaklets/pulse_processing would kind of defeat the purpose at this point

straxen/url_config.py
jmosbacher: This is to allow non-hashable kwargs such as functions.

straxen/url_config.py
jmosbacher: Yeah good idea. feal free to change it, otherwise will probably have to wait until the weekend when i have some time.

straxen/url_config.py
jmosbacher: This is to allow non-hashable kwargs such as functions.

jmosbacher: Sure, i assumed whomever put it there wanted it to be in a single place so changes are synched between the two plugins that use this default value

straxen/url_config.py
JoranAngevaare: ah you are right, I'm confused with {**kwargs, **more_kwargs}
image

this is so much easier in python 3.10
image

jmosbacher: I only know this because i tried overriding a value and got an error :sweat_smile:

straxen/plugins/pulse_processing.py
jmosbacher: Sure, i assumed whomever put it there wanted it to be in a single place so changes are synched between the two plugins that use this default value

straxen/url_config.py
jmosbacher: Sure, not sure why i removed it. Maybe i was too lazy to reformulate it with kwarg substitution allowed...

straxen/plugins/pulse_processing.py
```diff
@@ -150,38 +144,24 @@ def infer_dtype(self):
...
-        # if cmt config
-        if is_cmt_option(self.config['hit_min_amplitude']):
-            self.hit_thresholds = straxen.get_correction_from_cmt(self.run_id,
-                self.config['hit_min_amplitude'])
-        # if hitfinder_thresholds config
-        elif isinstance(self.config['hit_min_amplitude'], str):
-            self.hit_thresholds = straxen.hit_min_amplitude(
-                self.config['hit_min_amplitude'])
-        else: # int or array
-            self.hit_thresholds = self.config['hit_min_amplitude']
```
ahiguera-mx, line 175: Since we are getting rid of these lines, should we remove [this](https://github.com/XENONnT/straxen/blob/55cea4ced3498181aee450357079e6d6f5b7b942/straxen/hitfinder_thresholds.py#L11) one as well?

jmosbacher: I think i still use that code in the protocol. Since its a shared function and not hardcoded into a plugin i saw no reason to remove it

straxen/plugins/event_processing.py
```diff
@@ -534,7 +532,18 @@ class EventPositions(strax.Plugin):
...
+    fdc_map = straxen.URLConfig(
+        infer_type=False,
+        help='3D field distortion correction map path',
+        default_by_run=[
+            (0, pax_file('XENON1T_FDC_SR0_data_driven_3d_correction_tf_nn_v0.json.gz')),  # noqa
+            (first_sr1_run, pax_file('XENON1T_FDC_SR1_data_driven_time_dependent_3d_correction_tf_nn_part1_v1.json.gz')), # noqa
+            (170411_0611, pax_file('XENON1T_FDC_SR1_data_driven_time_dependent_3d_correction_tf_nn_part2_v1.json.gz')), # noqa
+            (170704_0556, pax_file('XENON1T_FDC_SR1_data_driven_time_dependent_3d_correction_tf_nn_part3_v1.json.gz')), # noqa
+            (170925_0622, pax_file('XENON1T_FDC_SR1_data_driven_time_dependent_3d_correction_tf_nn_part4_v1.json.gz'))], # noqa
+    )
```
JoranAngevaare, line 545: maybe we can even make the default_by_run a legacy url config

jmosbacher: I guess i can do that. This would change xenont1t lineages though, not sure its worth the trouble

JoranAngevaare: aren't you doing so already because of the pulse-processing changes?

anyway, 1T is kind of a freak chain now and we've already broken it some time ago with some of the corrected area plugins.


The added benifit is that we could enforce our own "specify options once, at the plugin" policy better, see https://github.com/AxFoundation/strax/issues/400

straxen/url_config.py
jmosbacher: Dpone.

straxen/plugins/event_processing.py
jmosbacher: Right, good point. Ok ill change it

jmosbacher: Done.

straxen/url_config.py
jmosbacher: simplified to be easier to follow logic

straxen/plugins/pulse_processing.py
jmosbacher: Removed.

straxen/plugins/pulse_processing.py
jmosbacher: Leaving this for now, not a priority and scared i may break something.

PR #1080, opened 2022-09-20
Ms events

michaweiss89:

1. This  code provides the sum of all cs2 in the drift length. With this you can evaluate the CES of MS events. 
2. it gives peaks related to each event and correction for MS studies.

1. Assumes that the main S1 is the only physical one in the event -peak level
2. Calculates all s2 drift length from that main s1-peak level
3. Calculate correction to get cs2 in the max drift length-peak level
4. Sum up the cs2.-event level

In the wiki note https://xe1t-wiki.lngs.infn.it/doku.php?id=weiss:analysis:ms_plugin


straxen/plugins/multi_scatter.py
```diff
@@ -0,0 +1,325 @@
...
+export, __all__ = strax.exporter()
+
+@export
+class EventPeaks(strax.Plugin):
+    """
+    Add event number for peaks and drift times of all s2 depending on the largest s1.
+    Link - https://xe1t-wiki.lngs.infn.it/doku.php?id=weiss:analysis:ms_plugin
+    """
+    __version__ = '0.0.57'
+    depends_on = ('event_info', 'peak_basics','peak_positions')
```
JoranAngevaare, line 18: You don't need event_info, right? The lower level, the better especially since `event_info` is overwritten e.g. here:
https://github.com/XENONnT/cutax/blob/9e858f9504e00a7e92531149aef61b6dae0cf4f8/cutax/plugins/unblind_nr_tagged.py

Can you also make it work with 'event_basics' or 'events'?

straxen/plugins/multi_scatter.py
```diff
@@ -0,0 +1,325 @@
...
+    provides = 'peaksevent'
```
JoranAngevaare, line 19:
```suggestion
    provides = 'peaks_per_event'
```

straxen/plugins/multi_scatter.py
```diff
@@ -0,0 +1,325 @@
...
+    data_kind = 'peaks'
+
+
+    
+    def infer_dtype(self):
+        dtype = []
+        dtype += [
+            ((f'event_number'), np.float32),
+            ((f'drift_time'), np.float32),
+            ((f'correction'), np.float32),
```
JoranAngevaare, line 29: Not used?

straxen/plugins/multi_scatter.py
```diff
@@ -0,0 +1,325 @@
...
+             ]
+        dtype += strax.time_fields
+        return dtype
+
+    save_when = strax.SaveWhen.TARGET
+ 
+        
+    def compute(self, events, peaks):
+        EB = straxen.EventBasics()
```
JoranAngevaare, line 38: not used?

straxen/plugins/multi_scatter.py
```diff
@@ -0,0 +1,325 @@
...
+        EB = straxen.EventBasics()      
+        split_peaks = strax.split_by_containment(peaks, events)
+        split_peaks_ind = strax.fully_contained_in(peaks, events)
+        result = np.zeros(len(peaks), self.infer_dtype())
+        result.fill(np.nan)
+        #result['s2_sum'] = np.zeros(len(events))
+   
+         #1. Assign peaks features to main S1 and main S2 in the event
+        for event_i, (event, sp) in enumerate(zip(events, split_peaks)):
+            result[f'drift_time'][(split_peaks_ind==event_i)] = sp[f'center_time']-event[f's1_center_time']
```
JoranAngevaare, line 47: can we vectorize this? Seems a bit slow to loop over over something that we have just split

straxen/plugins/multi_scatter.py
```diff
@@ -0,0 +1,325 @@
...
+        result[f'event_number']= split_peaks_ind
+        result[f'drift_time'][peaks["type"]!=2]= np.nan
```
JoranAngevaare, line 49: can't we pre-select s2 peaks earlier?

straxen/plugins/multi_scatter.py
```diff
@@ -0,0 +1,325 @@
...
+        result['time'] = peaks['time']
+        result['endtime'] = strax.endtime(peaks)
+        return result
+
+
+@export
+class PeakCorrectedAreas(strax.Plugin):
```
JoranAngevaare, line 56: perhaps it's worth inheriting some of the logic from the CorrectedAreas, especially in the setup of e.g. the ab,cd regions etc. than we don't have to track the same code in two places.

straxen/plugins/multi_scatter.py
```diff
@@ -0,0 +1,325 @@
...
+        help='Total length of the TPC from the bottom of gate to the '
+             'top of cathode wires [cm]',),
+)
+class EventInfoMS(strax.Plugin):
+    """
+    Get the sum of s2 and s1,
+    Get the sum of cs2 inside the drift length. 
+    """
+    __version__ = '0.0.9'
+    depends_on = ('event_info', 'peak_basics','peaksevent','peaks_corrections')
```
JoranAngevaare, line 256: can we depend on events?

JoranAngevaare (commented):
Thanks @michaweiss89 , I see that this plugin is useful to have to identify MS events! 

You can see that the code has already made some miles, based for example on the version 0.0.57 😉. I have some questions before going too much in depth (see below) and:
 - can you write a test? See tests/test_plugins for some examples
 - Have you considered limiting this to events-level data only? Or do you really need to have information of _all_ the peaks in an event? I'm just wondering if a single array for the alt drift time of x S2 peaks that come after the main S1 would be enough? Since you don't care for S1s, nor S2s from before the main S1. If you have more than x peaks, is this this plugin still useful? (I don't know what x would be, but you probably do :wink:). On the other hand, this plugin can probably be rather quick to compute?

straxen/plugins/multi_scatter.py
michaweiss89: Much better :)

straxen/plugins/multi_scatter.py
michaweiss89: Old test and I will remove. Thanks

straxen/plugins/multi_scatter.py
michaweiss89: Yes we can

JoranAngevaare:
Hi @michaweiss89 , when you are able to get back to this PR. Could you follow the new plugin structure from https://github.com/XENONnT/straxen/pull/1094

straxen/plugins/multi_scatter.py
```diff
@@ -0,0 +1,325 @@
...
+        result[f"cs2_wo_elifecorr"][peaks["type"]!=2] = np.nan
+        result[f"cs2_area_fraction_top"][peaks["type"]!=2] = np.nan
+        result[f"cs2"][peaks["type"]!=2] = np.nan
+        result[f"cs2_bottom"][peaks["type"]!=2] = np.nan
+        return result
+   
+
+@export
+@strax.takes_config(
+    strax.Option(
```
jmosbacher, line 244: Can you make this a URLConfig?

michaweiss89: Sorry I am not sure what you want to make in to a URL config

straxen/plugins/multi_scatter.py
michaweiss89: Because the function is using URLconfig it is giving me trouble in dooin so.

straxen/plugins/multi_scatter.py
michaweiss89: The problem is that I am using result which is the size of peaks . I f I pre select have trouble putting the drift time in the correct location. If I decide on removing all s1 but the main one this will work. However this will not be peak data anymore.

straxen/plugins/multi_scatter.py
michaweiss89: I want cs1 for ces_sum

JoranAngevaare:
Thanks again Micha and Henning!

PR #1081, opened 2022-09-22
Fix help of peak basics.

WenzDaniel:

Small fix of peak basics help. 

Not sure if necessary as the byte order of the dtype does not change, but maybe it is saver to reprocess.


dachengx:
Looks good. We should be more careful on the definition.

dachengx (approved):

JoranAngevaare (changes requested):
I think this is fine as long as we only merge it if we have a lineage change for some more fundamental reason (like https://github.com/XENONnT/straxen/pull/1079)

JoranAngevaare (approved):

PR #1089, opened 2022-10-04
Pytest-xdist for parallel testing

JoranAngevaare:
Speed up testing with https://pypi.org/project/pytest-xdist/


JoranAngevaare:
Closing as this decreases coverage and therefore usability

PR #1092, opened 2022-10-12
Remove `tight_coincidence_channel` fix #1078

### LINKED ISSUE #1078: Channel and tight_coincidence_channel field can be removed?
* Not sure anymore why we have the channel field in peaklets. 
* I noticed that peaks still has the `tight_coincidence_channel` field which is not used anymore. Should be removed before the next big reprocessing.

JoranAngevaare:
Remove unused field from peaks


WenzDaniel (approved):
Nice thanks.

PR #1097, opened 2022-10-13
Event level S2 posrec

terliuk:
This PR adds possibility to rerun S2 posrec at event level, without requiring reprocessing data from peak level. Also, as number of events much lower than number of peaks, this can significantly improve speed. Currently speed is limited by loading of `event_area_per_channel`

each plugin has own model that can be set to divergent from peak level, corresponding parameter is `tf_event_model_<algorithm>`. Now it is initialized to the same default as at peak level. 

```
event_s2positions = st.get_array(run_id, ["event_s2position_mlp", 
                                          "event_s2position_cnn", 
                                          "event_s2position_gcn"])
```


JoranAngevaare (approved):
looks nice and clean to me, I think it's good to add and test.

One small question, should we call it `event_position_cnn` or `event_s2_position_cnn`?

JoranAngevaare:
as discussed on slack I added one underscore, I'll merge 😄

PR #1101, opened 2022-10-17
Change timeouts

JoranAngevaare:
Given our recent calibration period, it seems we can change some of the time outs a bit of change the verbosity of some of the messages


mflierm (approved):
Looks good!

PR #1103, opened 2022-10-18
Use mongodb v4.4.1 when testing to match real version used in production

jmosbacher:
Current mongodb version being used is 4.4.1 so we should also use this version when testing.


JoranAngevaare (approved):
sure, thanks!

PR #1104, opened 2022-10-18
New URLConfig protocols - list-to-array and list-to-dict

LuisSanchez25:
…gurations

This code enables the user to change the format of the output of the url-protocol, some corrections require dictionaries while others require arrays. In Xedocs all of the data is stored in lists therefore this transformation stage is necessary to enable us to change the format of the output to a dictionary or an array.

It works by taking the output of the URLConfig and depending on the added field changing the output from a list to either a dictionary with keys or an array.

This code reprocesses some XENONnT data using a correction stored in xedocs

import cutax 
import xedocs as xd

run_id = '047493'
st_xd = cutax.contexts.xenonnt_online()
st_xd.set_config({"rel_extraction_eff":"objects-to-dict://xedocs://"
                  "rel_extraction_effs"
                  "?version=v!&context=staging_db&run_id=047122&partition=['ab','cd']&key_attr=partition&value_attr=value"})

kr_ei_xd = st_xd.get_array(run_id, 'event_info')

### _Notes on testing_


jmosbacher:
@LuisSanchez25 thanks for this PR, I think its important to mention that these protocols can be used independently from xedocs and describe them in a generic manner in the comments naming withing the code.
e.g. i would replace `docs` with `objects` and maybe even generalize the `objects-to-dicts` protocol to handle lists of dicts as well as lists of python objects.

jmosbacher:
I made a few changes including comments. All that is missing are tests.

JoranAngevaare:
Tests also fail because of:
>On the XENONnT fork we test with database access, on private forks there is no database access for security considerations.

I'll change the testing to fix this

JoranAngevaare:
github closed this PR by accident but the failing tests should be resolved in #1105

jmosbacher:
Thanks @JoranAngevaare !
@LuisSanchez25 just add tests and then you can mark this ready for review

jmosbacher (approved):
I have made the changes I requested already.

jmosbacher:
@JoranAngevaare any comments before we merge?

straxen/url_config.py
```diff
@@ -448,3 +454,36 @@ def rekey_dict(d, replace_keys='', with_keys=''):
     for old_key, new_key in zip(replace_keys, with_keys):
         new_dict[new_key] = new_dict.pop(old_key)
     return new_dict
+
+
+@URLConfig.register('objects-to-dict')
+def objects_to_dict(objects: list, key_attr=None, value_attr='value'):
```
JoranAngevaare, line 460: Worth adding an immutable dict to these options? Somthing like `immutable=False`

straxen/url_config.py
```diff
@@ -448,3 +454,36 @@ def rekey_dict(d, replace_keys='', with_keys=''):
...
+    key and value from each of the objects/dicts. If key_attr is not provided,
+    the list index is used as the key.
+
+    :param objects: list of objects/dicts that will be converted to a dictionary
+    :param key_attr: key/attribute of the objects that will be used as key in the dictionary
+    :param value_attr: key/attribute of the objects that will be used as value in the dictionary
+    '''
+    if not isinstance(objects, list):
+        raise TypeError(f'The objects-to-dict protocol expects a list '
+                        f'of objects but received {type(objects)} instead.')
```
JoranAngevaare, line 471: why not tuples?

tests/test_url_config.py
```diff
@@ -270,3 +281,15 @@ def test_dry_evaluation(self):
...
+        N = 3
+        self.st.set_config({'test_config': f'objects-to-dict://object-list://{N}?key_attr=a&value_attr=b'})
+        p = self.st.get_single_plugin(nt_test_run_id, 'test_data')
+        self.assertEqual(p.test_config, {i:i+1 for i in range(3)})
+
+    def test_list_to_array(self):
+        N = 3
+        self.st.set_config({'test_config': f'list-to-array://object-list://{N}'})
+        p = self.st.get_single_plugin(nt_test_run_id, 'test_data')
+        self.assertIsInstance(p.test_config, np.ndarray)
```
JoranAngevaare, line 295:
```suggestion
    def test_objects_to_dict(self):
        n = 3
        self.st.set_config({'test_config': f'objects-to-dict://object-list://{n}?key_attr=a&value_attr=b'})
        p = self.st.get_single_plugin(nt_test_run_id, 'test_data')
        self.assertEqual(p.test_config, {i:i+1 for i in range(n)})

    def test_list_to_array(self):
        n = 3
        self.st.set_config({'test_config': f'list-to-array://object-list://{n}'})
        p = self.st.get_single_plugin(nt_test_run_id, 'test_data')
        self.assertIsInstance(p.test_config, np.ndarray)
```

JoranAngevaare (approved):
Looks good. Only smaller suggestions, if you like any of them, you can add them but this state is already good. Thanks!

straxen/url_config.py
jmosbacher: sure, why not

straxen/url_config.py
jmosbacher: done

JoranAngevaare (approved):
Thanks! Looks good!

PR #1108, opened 2022-10-20
Lower the default config value of online_max_bytes

mflierm:
Since we were running into the error of a too large document size multiple times, we decided to lower the max_bytes that the data array can have. We have the strong suspicion that the error must come from conversion from our np.float32 to standard python float64, which doubles the datasize. However, it remains unclear that this 'suddenly' happens, with normal data rates and chunk lengths.


JoranAngevaare:
Better be on the save side

PR #1109, opened 2022-10-21
Single core 1T test

JoranAngevaare:
Run 1T test single core first to prevent numba compilation issues during compilation.

Should only add about 12 s extra to the testing
afbeelding


JoranAngevaare:
seems to be the issue in https://github.com/XENONnT/base_environment/pull/1116

PR #1110, opened 2022-10-23
URLConfig preprocessor

jmosbacher:
A few small improvements to URLConfigs, similar to #827 but a bit simpler implementation.

1. Formalizes the URL parsing as a string to AST and add reverse operation
2. Adds `validate` method that runs before the config is passed to the plugin.
3. Adds the ability to register preprocessors that can alter the URL/config before it is hashed.

Preprocessors are passed the config and return a new/same config. They are applied in order of higher precedence first.

### DIFF
--- a/straxen/url_config.py
+++ b/straxen/url_config.py
@@ -62,6 +62,7 @@ class URLConfig(strax.Config):
 
     SCHEME_SEP = '://'
     QUERY_SEP = '?'
+    NAMESPACE_SEP = '.'
     PLUGIN_ATTR_PREFIX = 'plugin.'
 
     def __init__(self, cache=0, **kwargs):
@@ -209,8 +210,8 @@ class URLConfig(strax.Config):
         if isinstance(value, list):
             return [cls.lookup_value(v, **namespace) for v in value]
 
-        if isinstance(value, str) and '.' in value:
-            name, _, key = value.partition('.')
+        if isinstance(value, str) and cls.NAMESPACE_SEP in value:
+            name, _, key = value.partition(cls.NAMESPACE_SEP)
             if name in namespace:
                 obj = namespace[name]
                 if isinstance(obj, Mapping):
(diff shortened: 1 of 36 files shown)


straxen/url_config.py
```diff
@@ -1,24 +1,30 @@
...
 from ast import literal_eval
 from strax.config import OMITTED
-import os
-import tempfile
-import tarfile
-import pytz
 from utilix import xent_collection
 from scipy.interpolate import interp1d
+from straxen.misc import filter_kwargs
+from urllib.parse import urlparse, parse_qs
```
JoranAngevaare, line 26:
```suggestion
```

straxen/url_config.py
```diff
@@ -1,24 +1,30 @@
...
+import numbers
 import straxen
 import inspect
+import tarfile
+import tempfile
+
+import numpy as np
+import pandas as pd
+
+from uuid import uuid4
```
JoranAngevaare, line 17:
```suggestion
```

straxen/url_config.py
```diff
@@ -1,24 +1,30 @@
+import os
 import json
+import pytz
 import typing
-from typing import Container, Iterable
-from uuid import uuid4
-import numpy as np
 import strax
+import typing
```
JoranAngevaare, line 6:
```suggestion
```

JoranAngevaare (commented):
@jmosbacher , looks good. In https://github.com/XENONnT/straxen/pull/827/ you had a few tests, can you re-use those examples again?

Should we make the `format_url_kwargs` a default someday, or is xedocs already planned to use this feature?

Uhm, there are two `self.format_url_kwargs` methods? L193, L325 🤔

jmosbacher:
@JoranAngevaare hmm you are right, now i realize why i left it as a draft for so long. I never got around to finishing polishing it and adding tests... Ill do that now.

jmosbacher:
@JoranAngevaare I added some tests and documentation

JoranAngevaare:
Nice work Yossi, I added a test to assert I understand it well.

Two final requests that I think would be usefull for dummies like me:
- Add a `preprossor_descr(cls)` like `protocol_descr(cls)` and
- `show(cls)`, that shows the overall state (what is registered, what precedence is associated to each protocol, where is the code that adds it). We can add the output of that to the autodocs [a la plugin tree](https://github.com/XENONnT/straxen/blob/master/docs/source/build_datastructure_doc.py), which would also solve https://github.com/XENONnT/straxen/issues/1111.. The autodocs I can also do and/or make a seperated PR for it

JoranAngevaare (approved):

jmosbacher:
Good idea. I added methods for the preprocessor as well as a `print_summary` method that prints both

jmosbacher:
@JoranAngevaare did you want to already add the autodoc feature to this PR before its merged?

JoranAngevaare:
@jmosbacher let's do it in a seperate PR, nice work Yossi, thanks!

PR #1112, opened 2022-10-26
Adding peak waveforms at event level

terliuk:
This PR adds `data`, `data_top`, `length` and `dt` fields from peak level for main/alternative S1/S2, similar to `area_per_channel` that was added previously. As side product it refactors `infer_dtype` function to make sure that it is always in line with peak dtype. 

By advice of Joran, i keep the same name, but we can consider renaming. 

### DIFF
--- a/straxen/plugins/events/event_area_per_channel.py
+++ b/straxen/plugins/events/event_area_per_channel.py
@@ -13,7 +13,7 @@ class EventAreaPerChannel(strax.LoopPlugin):
     """
     depends_on = ('event_basics', 'peaks')
     provides = "event_area_per_channel"
-    __version__ = '0.0.1'
+    __version__ = '0.0.2'
 
     compressor = 'zstd'
     save_when = strax.SaveWhen.EXPLICIT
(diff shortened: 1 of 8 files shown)


straxen/plugins/events/event_area_per_channel.py
```diff
@@ -22,15 +22,33 @@ class EventAreaPerChannel(strax.LoopPlugin):
...
-                  np.float32, (self.config['n_tpc_pmts'],)),
-                 (("Area per channel for main S1", "s1_area_per_channel"),
-                  np.float32, (self.config['n_tpc_pmts'],)),
-                 (("Area per channel for alternative S1", "alt_s1_area_per_channel"),
-                  np.float32, (self.config['n_tpc_pmts'],)),
-                 (("Main S1 count of contributing PMTs", "s1_n_channels"),
+        # setting data type from peak dtype
+        peak_dtype_dict = {}
+        for t_ in strax.peak_dtype(n_channels=self.config['n_tpc_pmts'], digitize_top=True):
+            peak_dtype_dict[t_[0][1]]=t_
```
JoranAngevaare, line 28: please checkout how this is done in strax: https://github.com/XENONnT/straxen/blob/35ef0abbc3e90664bce2490ae4321722a58bda6b/straxen/plugins/merged_s2s/merged_s2s.py#L67

something like `strax.unpack_dtype(self.deps['peaklets'].dtype_for('peaklets'))`

straxen/plugins/events/event_area_per_channel.py
```diff
@@ -8,12 +8,12 @@
 @export
 class EventAreaPerChannel(strax.LoopPlugin):
     """
-    Simple plugin that provides area per channel for main 
-    and alternative S1/S2 in the event. 
+    Simple plugin that provides area per channel, totla (data) and top (data_top)
```
JoranAngevaare, line 11:
```suggestion
    Simple plugin that provides area per channel, total (data) and top (data_top) 
```

JoranAngevaare (commented):
Seems simple and fine. 

Documentation seems in order too:
image

straxen/plugins/events/event_area_per_channel.py
terliuk: It should be equivalent  for this particular case for peak dtype. Also, if i read the code correctly, it does not really return dictionary of name -> dtype,  but a tuple for type. So one anyway needs to construct a similar structure, and I think it looks a bit cleaner this way.

terliuk: alternatively, one can probably do: 

```python
pfields_= np.zeros(0, dtype = strax.peak_dtype(n_channels=self.config['n_tpc_pmts'], digitize_top=True)).dtype.fields
```
and then call something like
```python
dtype +=[((f'Waveform for {infoline[type_]} [ PE / sample ]', f'{type_}_data'),pfields_['data'][0])] 
```
do you think it will be better?

JoranAngevaare: No the reason I'm asking is more that the dtype of peaks changes (hypothetically) if you provide a different argument for `sum_waveform_top_array`, [see line](https://github.com/XENONnT/straxen/blob/35ef0abbc3e90664bce2490ae4321722a58bda6b/straxen/plugins/peaklets/peaklets.py#L109)


I had not looked at how exactly `strax.unpack_dtype` or `Peaklets.dtype_for` formatted the dtype (that was not the intend of my comment)

PR #1115, opened 2022-11-07
Added backup URL to scada

WenzDaniel:
Add backup url to SCADA. Requires a change of our config file, should be backwards compatible.

### DIFF
--- a/straxen/scada.py
+++ b/straxen/scada.py
@@ -1,4 +1,5 @@
 import urllib
+import posixpath
 import requests
 
 import pandas as pd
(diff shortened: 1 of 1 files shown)


jmosbacher (commented):
I think this implementation is fine but if you feel like making some improvements I would:
1) use the base URL of the current `sclogin_url` as a default value so old configs still work
2) instead of having a `url` and `backup_url` configs, just parse the `url` (or `base_url`) config as a list with `url.split(',')` so that multiple URLs can be passed similar to the mongo url syntax
3) just do a for loop over the URL list and break on the first one that succeeds during login, then set the url to that one.

ahiguera-mx:
@LuisSanchez25

PR #1117, opened 2022-11-08
Event waveform CEvNS

shenyangshi:

It contains the plugins used for the hit/SE level waveform-based AC suppression:

* HitsS1: Perform hit finding for S1s smaller than 100PE, and calculate the hit_max_time; a new data type, as it's uneconomical to store all hit information, while only small S1's are useful.
* PeakWaveformS1: Pile hit level information into the Peak level, and calculate per PMT hit shadow.
* PeakAmbienceS1: Pile ambient lone-hit information into Peak level.
* EventWaveformS1: Pass peak to event
* EventAmbienceS1: Pass peak to event
* EventWaveformS2: Calculate the per single electron time and area in small S2s
* EventAmbienceS2: Pile the ambient single electrons near S2

They serve as extractors from low-level information to peak/event-level data. The data is compiled into NumPy arrays like event_area_per_channel for peak and event-level information.

## Note that:
* Since the HitS1 plugin is computationally expensive, it would be great if it can catch the next reprocessing.
* The code styles need refactoring to reflect the change of URL config and the newest straxen data-structure, please help :)

## Relevant notes:
* https://xe1t-wiki.lngs.infn.it/doku.php?id=shenyang:xenonnt_sr0:electron_statistics
* https://xe1t-wiki.lngs.infn.it/doku.php?id=shenyang:xenonnt_sr0:hit_statistics

### DIFF
--- a/straxen/plugins/peaks/peak_waveform.py
+++ b/straxen/plugins/peaks/peak_waveform.py
@@ -1,0 +1,318 @@
+from straxen.plugins.peak_processing import PeakBasics
+import straxen
+import strax
+import numpy as np
+import numba
+from straxen.plugins.peaklet_processing import hit_max_sample
+from straxen.plugins.pulse_processing import HITFINDER_OPTIONS, HITFINDER_OPTIONS_he, HE_PREAMBLE
+export, __all__ = strax.exporter()
+
+
+
+
+@export
+@strax.takes_config(
+    strax.Option('gain_model', infer_type=False,
+                 help='PMT gain model. Specify as '
+                      '(str(model_config), str(version), nT-->boolean'),
+    strax.Option('peak_waveform_max_s1_area', default=100, infer_type=False,
+                 help='Maximum area of S1 we do hit analysis'),
+    *HITFINDER_OPTIONS,
+)
+class HitsS1(strax.OverlapWindowPlugin):
+    """
+    Return timing and area for single hits in S1s.
+    Reconstruct hits for S1s with area smaller than 100PE
+    """
+    __version__ = "0.0.0"
+    parallel = True
+    depends_on = ('peak_basics', 'records',)
+    provides = 'hits_s1'
+    data_kind = 'hits'
+    save_when = strax.SaveWhen.ALWAYS
+
+    electron_drift_velocity = straxen.URLConfig(
+        default='cmt://'
+                'electron_drift_velocity'
+                '?version=ONLINE&run_id=plugin.run_id',
+        cache=True,
+        help='Vertical electron drift velocity in cm/ns (1e4 m/ms)'
+    )
+
+
+    def setup(self):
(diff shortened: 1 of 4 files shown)


jmosbacher (changes requested):
Thanks for the nice work. Can you please change all of the configs to the newer URLConfig syntax? we have decided to deprecate the old syntax

shenyangshi:
Thanks, Yossi! This is exactly what I want to ask, for some configs that are already being defined in other plugins, such as "hit_min_amplitude", should I define it again, or the URL config can automatically recognize them from other plugins?

straxen/plugins/peaks/peak_waveform.py
```diff
@@ -0,0 +1,326 @@
...
+             'which means we are using cmt.'
+    )
+
+
+    def setup(self):
+        self.drift_time_max = int(straxen.tpc_z / self.electron_drift_velocity)
+        self.to_pe = straxen.get_correction_from_cmt(self.run_id,
+                                                     self.config['gain_model'])
+        self.hit_thresholds = straxen.get_correction_from_cmt(self.run_id,
+                                                              hit_min_amplitude)
```
jmosbacher, line 57:
```suggestion
                                                              self.hit_min_amplitude)
```

straxen/plugins/peaks/peak_waveform.py
```diff
@@ -0,0 +1,326 @@
...
+             'or a tuple like (correction=str, version=str, nT=boolean),'
+             'which means we are using cmt.'
+    )
+
+
+    def setup(self):
+        self.drift_time_max = int(straxen.tpc_z / self.electron_drift_velocity)
+        self.to_pe = straxen.get_correction_from_cmt(self.run_id,
+                                                     self.config['gain_model'])
+        self.hit_thresholds = straxen.get_correction_from_cmt(self.run_id,
```
jmosbacher, line 56: Please use the URLs to get corrections from CMT.

jmosbacher (changes requested):

jmosbacher:
@shenyangshi sorry, i dont understand your question. can you elaborate  bit more?

shenyangshi:
Sorry, what I mean is that, should I define every parameter in the plugin, even if it's also defined elsewhere with the same name in other plugins?

jmosbacher:
Yeah, you need to define all parameters that you want to use in the plugin. For each plugin, strax only passes the configs that are defined in that plugin, all the other configs from other plugins are not accessible at runtime within the plugin.

shenyangshi:
Hi @WenzDaniel, thanks for looking into the code, I almost forgot there is still a PR here. The hit time information is already included in hit time diff https://github.com/AxFoundation/strax/pull/716 for the S1 BDT, so I can close the PR.

PR #1119, opened 2022-11-10
Top and bottom timing parameters at event and peak level

terliuk:

This PR adds plugins computing timing parameters (center time, rise time, 50% and 50% ranges) at peak level as well as event level, depending on needs of an individual analyst. 

Technical remark, in order to compute parameter, it forms "fake peaks" based on `data_top`  for top array or `(data - data_top)` for bottom array. And then reuses the quantile calculation from strax to avoid unnecessary code duplication. 

To get parameters at peak level, one needs to run: 
```
st.get_array(run_id, 'peak_top_bottom_params')
```
which will return fields 
  - `center_time_top`, `center_time_bot` 
  - `rise_time_top` , `rise_time_bot`
  - `range_50p_area_top`, `range_50p_area_bot` 
  - `range_90p_area_top`, `range_90p_area_bot`
with definitions the same as in peak_basics. 

To get it at event level one needs: 
```
st.get_array(run_ids, "event_top_bottom_params")
```
which will provide the similar fields, but for main/alt s1/s2 forming the event, e.g. `s1_rise_time_top`, `alt_s2_center_time` etc. 

Few plots from few Kr83m runs: 

time difference between top and bottom center times as function of Z:

or rise time of bottom array 

### DIFF
--- a/straxen/plugins/events/__init__.py
+++ b/straxen/plugins/events/__init__.py
@@ -13,6 +13,9 @@ from .event_ambience import *
 from . import event_area_per_channel
 from .event_area_per_channel import *
 
+from . import event_top_bottom_params
+from .event_top_bottom_params import *
+
 from . import event_basics
 from .event_basics import *
 
--- a/straxen/plugins/events/event_top_bottom_params.py
+++ b/straxen/plugins/events/event_top_bottom_params.py
@@ -1,0 +1,89 @@
+import numpy as np
+import strax
+import straxen
+from warnings import warn
+export, __all__ = strax.exporter()
+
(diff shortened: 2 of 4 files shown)


FaroutYLq:
Something general idea: do we want to put the parameters in `peaklets` level instead of above `peaks`, so these info can be used in peak classification?

FaroutYLq (approved):
Looks good to me! Check [here](https://xe1t-wiki.lngs.infn.it/doku.php?id=lanqing:s1_wfsim_data_match_with_sep_array#appendix) for plots.

PR #1120, opened 2022-11-13
Add entry points

jmosbacher:
Adds entry point loading. This can be used to register context methods, mini-analyses, URLConfig protocols etc by other packages when straxen is imported.

Defines an entry point called `straxen` that packages can register modules or functions to which will be imported/ called on import of straxen.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -3,6 +3,7 @@
 omit =
         ./straxen/storage/rucio_remote.py
         ./straxen/legacy/plugins_1t/pax_interface.py
+        ./straxen/entry_points.py
 
 # Regexes for lines to exclude from consideration
 exclude_lines =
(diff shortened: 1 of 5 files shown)


JoranAngevaare:
Looks like a nice feature, can we write a very simple example that we can run as a test? Then we can easily point people to how this should be used.

jmosbacher:
@JoranAngevaare Im not sure it makes sense to have an entry point in the package itself (although this is definitely possible), and its not possible to have entry points registered in the tests as this is a setuptools mechanism registered when a package is installed. Instead I added some documentation with an example. Can you maybe read the example and see if it is clear or do i need to add a more concrete example?

JoranAngevaare:
@jmosbacher thanks, very neat. Looks good thanks! I've added it to the autodocs too and excluded it from the coverage report

JoranAngevaare:
Looks good https://straxen--1120.org.readthedocs.build/en/1120/extending_straxen.html!

PR #1133, opened 2023-01-16
Fix bootstrax timeouts

JoranAngevaare:
Small patch that allows supporting > 13 h long runs.


mflierm (commented):
Looks good! Shouldn't there be an extra indent in lines 1506 and 1507?

JoranAngevaare:
@mflierm thanks, no the indents are fine (was one too many for some reason)

PR #1135, opened 2023-02-07
Add rundoc URLConfig protocol

jmosbacher:
Inspired by [this](https://xenonnt.slack.com/archives/C049JJZ2U3A/p1675791764694399) thread.

This PR add a URLConfig protocol that fetches a value from the rundoc.

The protocol accepts the path to the value inside the rundoc and returns only the value if it exists.

e.g.

```
URL = "rundoc://source?run_id=plugin.run_id"
```
will return the source from the rundoc for the run_id set on the plugin.

the path can also be set to access a nested data structure using the standard mongo dot notation.

e.g.

```
URL = "rundoc://daq_config.boards?run_id=plugin.run_id"
```
will return the array  `rundoc["daq_config"]["boards"]`.


jmosbacher:
Still need to add tests, then can mark it ready for review.

JoranAngevaare (commented):
Looks good, although we might want to check with simulations experts @terliuk since this will for sure break MC again since there is no `int('mc_run_aweseome')` 

We can merge it as it is (no defaults are currently changed) but would have to change the simulation context explicitly when we add this feature in the data stream.

jmosbacher:
@JoranAngevaare but doesnt the plugin run get set with the "cmt_run_id" value? which is an int no?

JoranAngevaare:
@jmosbacher not entirely sure anymore, this is not a CMT option, so no I don't think so? I think this would cause errors if used directly.

It by the way doesn't allow setting the thresholds yet based on the rundoc because of the way we organized the rundoc

jmosbacher:
Yeah I am now not so sure this is useful for the thresholds specifically

dachengx:
I am curious about what the 'thresholds' are. Is this PR now in good shape?

jmosbacher:
@dachengx this was created initially because the pmt thresholds are stored by the DAQ in the run doc but the plugin actually reads them from CMT. I think this was setup because people thought that we may want to use different thresholds for processing than the ones used by DAQ. In the end we just ended up with the same values being in both places and in some cases not matching due to an error in the cmt db.

This PR adds a protocol that allows you to use values from the run doc in you plugin configuration. Im not sure its really needed.

dachengx (approved):
Looks good. Let's go with it.

PR #1139, opened 2023-02-14
fix xedocs for testing

JoranAngevaare:
Fix the testing version


JoranAngevaare:
Version requirement was added in https://github.com/XENONnT/straxen/pull/1125, but this should have been done in https://github.com/XENONnT/ax_env/pull/281

PR #1148, opened 2023-03-30
Adding correction of Z position due to non-uniform drift velocity

terliuk:
Adds correction of Z position bias due to non-uniform field and, correspondingly, drift velocity. 

It loads a map produced from COMSOL field simulation. The map gets r_naive, z_naive and returns expected bias value that is then subtracted from z_naive. 

Map is uploaded to database in this PR https://github.com/XENONnT/private_nt_aux_files/pull/251

```
st.get_array(runid, ["event_positions"])
```
the new corrected position is in field `z_dv_corr`

### DIFF
--- a/straxen/contexts.py
+++ b/straxen/contexts.py
@@ -63,6 +63,11 @@ xnt_common_config = dict(
             '&run_id=plugin.run_id'
             '&fmt=binary'
             '&scale_coordinates=plugin.coordinate_scales',
+    z_bias_map='itp_map://'
+               'resource://'
+               'XnT_z_bias_map_chargeup_20230329.json.gz?'
+               'fmt=json.gz'
+               '&method=RegularGridInterpolator',
 )
 # these are placeholders to avoid calling cmt with non integer run_ids. Better solution pending.
 # s1,s2 and fd corrections are still problematic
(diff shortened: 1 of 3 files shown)


jingqiangye (approved):
Hey I did a test and it looks good, see [Fig. 5](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:terliuk:drift_field_z_bias_correction).

jingqiangye:
Thanks for the explanation @terliuk ! I like the current improvement and also the future plan. I have no objection to the current name unless someone has a better idea. However, the automatic tests still failed today complaining about the absence of the file but I think it was uploaded to the database yesterday. Any idea?

dachengx:
> Thanks for the explanation @terliuk ! I like the current improvement and also the future plan. I have no objection to the current name unless someone has a better idea. However, the automatic tests still failed today complaining about the absence of the file but I think it was uploaded to the database yesterday. Any idea?

`EventPositions` is registered in the 1T context when auto-test. But it has no default map for 1T. We should add a legacy placeholder for 1T, like here:
https://github.com/XENONnT/straxen/blob/83ede84027855ba172db124165f61c66042073f1/straxen/plugins/events/event_positions.py#L47

dachengx:
@jingqiangye If the testing codes still exist, would you please test the performance again to make sure I did not write anything wrong?

dachengx (approved):
Looks good to me.

jingqiangye (approved):
@dachengx thanks for fixing this! I just tested with another run and don't see any problem.

jingqiangye:
Thanks @terliuk for the nice PR and @dachengx for fixing the auto test. I will merge this now.

PR #1149, opened 2023-03-30
reforming _raw_path and _processed_path

FaroutYLq:
To handle the directories where we put processed data automatically, when not being able to access `/dali` storage.

Refactored `_raw_path` and `_processed_path` to lists, so that we can have multiple sites. 
On midway2 compute nodes, the following was impossible, because you didn't direct straxen to `/project2/lgrandi/xenonnt/processed`. Lanqing can be pinged less then.


jmosbacher (approved):
Looks good

dachengx (approved):
Looks good to me! Thanks! @FaroutYLq

PR #1150, opened 2023-04-06
AmBe skip heavy computation

cfuselli:
This is the straxen that is currently running on the EBs. This is just meant to be a pull request for reference. 

It's a mixtures of small operations done on the DAQ. 
It includes:

- bug fix from @JoranAngevaare for getting filesize from rundoc
- temporary patch from @mflierm due to failures in uploading veto_intervals to the MongoDB when too many vetos are fired (happened during ambe_hev) -> it needs to be fixed in some other permanent way, so that we don't completely loose one of our nice online monitor panels
- only process raw_records and veto_intervals by default for ambe modes, added by myself 
- bypass restrax operations by default for ambe modes, added by myself

### DIFF
--- a/bin/bootstrax
+++ b/bin/bootstrax
@@ -1449,7 +1449,7 @@ def process_run(rd, send_heartbeats=args.production):
                     if not len(md['chunks']):
                         fail("Processing succeeded, but no chunks were written!")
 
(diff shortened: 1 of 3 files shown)


cfuselli:
Closing in favor of https://github.com/XENONnT/straxen/pull/1483

PR #1180, opened 2023-05-10
Make hitfinder amplitude of software trigger independent

WenzDaniel:
In this PR we change just the naming of the hitfinder amplitude of the software trigger such that it becomes independent of the regular hitfinder. This will allow in the future to set independently the threshold used for the software trigger in case we decided to delete the NV raw data.


WenzDaniel:
Please note, beside changing the naming we also need to change the CMT/xedocs setting such that this threshold will remain static for the rest of the experiment.

WenzDaniel:
Closed in favor for: https://github.com/XENONnT/straxen/pull/1180

PR #1185, opened 2023-05-23
Faster subtyping calculation

dachengx:

The second function of subtyping calculation used `while` loop and the algorithm depends on the time order of peak scanning.

This PR accelerates the calculation of subtyping and makes the definition consistent and independent of the time order of calculation. 

In the first function `mark_peaks_after_s1s`,

1. Identify the S1-pS2 pairs
2. Divide timeline into non-overlapping windows following `loneS1`, `sloneS1`, `S1`, `pS2`
3. Assign `S1PH` following `loneS1`, `slS1PH` following `sloneS1`, `S1PH` & `S1olS2` following `S1` before `pS2`, `S2PH` & `S2olS2` following `pS2`.

In the second function `mark_other_s2s`,

1. Identify fakeS2-pS2 pairs
2. Divide timeline into non-overlapping windows following `lonepS2`, `fakeS2`, `pS2`
3. Assign `S2PH` & `S2olS2` following `lonepS2`, `fakeS2_PH` & `fakeS2_olS2` following `fakeS2` before `pS2`, `S2PH` & `S2olS2` following `pS2`.

### _Notes on testing_


straxen/plugins/peaks/peak_subtyping.py
```diff
@@ -0,0 +1,621 @@
...
+
+        _s1_area = peaks['area'][type_1_mask]
+        _s2_area = peaks['area'][type_2_mask]
+
+        base_times = peaks['center_time']
+        base_areas = peaks['area']
+        base_indices = np.arange(len(peaks))
+
+        # find S1-pS2 pairing
+        _peaks = peaks[type_2_mask].copy()
```
WenzDaniel, line 165: FYI: I personally would call parameters/selections always explicit, e.g., here I would change _peaks into _s2_peaks. Adds readability.

straxen/plugins/peaks/peak_subtyping.py
```diff
@@ -0,0 +1,621 @@
...
+        tw_s1_ps2 = strax.touching_windows(_peaks, _window)
+
+        # Find index of max S2 of each group of S2s following S1
+        _ps2_max_indices = np.ones(len(_window), dtype=np.int32) * -1
+        for i, tw_12 in enumerate(tw_s1_ps2):
+            if tw_12[1] - tw_12[0] == 0:
+                continue
+            else:
+                # do we have pS2 or a bunch of small S2s
+                _ps2_max_indices[i] = np.argmax(_s2_area[tw_12[0]:tw_12[1]]) + tw_12[0]
```
WenzDaniel, line 184: Here too, I would do something like:

```python
index_s2_with_max_area = ...
 _ps2_max_indices[i] = index_s2_with_max_area

File: straxen/plugins/peaks/peak_subtyping.py [hunk 3]
```diff
@@ -0,0 +1,621 @@
...
+            base_times[pS2_indices], after_s2_window_ext, limitation)
+
+        # extra check, will be deleted after debugging
+        for i_a, a in enumerate([loneS1_containers, sloneS1_containers, S1_containers, pS2_containers]):
+            for i_b, b in enumerate([loneS1_containers, sloneS1_containers, S1_containers, pS2_containers]):
+                if i_a == i_b or len(a) == 0 or len(b) == 0:
+                    continue
+                r = strax.touching_windows(a, b)
+                assert (r[:, 1] - r[:, 0]).max() == 0, ''\
+                    + 'triggering containers should not be overlapping'
```
WenzDaniel, line 255: You should write a unitest or even better hypothesis test for that.

straxen/plugins/peaks/peak_subtyping.py
```diff
@@ -0,0 +1,621 @@
...
+
+        # assign loneS1, S1, sloneS2, pS2
+        mask[loneS1_indices] = PeakSubtyping.loneS1
+        mask[sloneS1_indices] = PeakSubtyping.sloneS1
+        mask[S1_indices] = PeakSubtyping.S1
+        mask[pS2_indices] = PeakSubtyping.pS2
+        return mask
+
+    @staticmethod
+    def mark_other_s2s(
```
WenzDaniel, line 306: I have not checked in very much detail, but to me it looks like that 70 % of this code is the same as above. If this is the case can we reduce the copy/past a bit.

straxen/plugins/peaks/peak_subtyping.py
```diff
@@ -0,0 +1,621 @@
...
+        result['subtype'][undefined_mask] = self.mark_other_s2s(
+            peaks[undefined_mask],
+            self.subtype_dtype,
+            self.large_s2_threshold, self.other_large_s2_fac,
+            self.s1_s2_window, self.after_s2_window_ext)
+
+        # check if there is any undefined peaks
+        n_undefined = (result['subtype'] == PeakSubtyping.Undefined).sum()
+        if n_undefined:
+            raise RuntimeError(f'We still find {n_undefined} peaks undefined!')
```
WenzDaniel, line 517: Should make this a unit/hypothesis test

straxen/plugins/peaks/peak_subtyping.py
```diff
@@ -0,0 +1,621 @@
...
+        return result
+
+
+def limited_containers(anchor, length, limitation, forward=True):
+    """
+    Build container given anchor, length and limitation of end time.
+    The retuned containers should not be overlapping
+    :param anchor: sorted array of interval anchor points
+    :param length: length of each interval, len(anchor) == len(length)
+    """
```
WenzDaniel, line 531: Can you guys give me an example and explanation what limitation should be? I do not fully understand the purpose of this function.

straxen/plugins/peaks/peak_subtyping.py
```diff
@@ -0,0 +1,621 @@
...
+    return time
+
+
+def combine_indices(result):
+    """
+    Combine the indices from touching_windows results
+    :param result: touching_windows results, each row is a pair of indices
+    """
+    if len(result) == 0:
+        return np.array([], dtype=np.int)
```
WenzDaniel, line 591: np.empty(0, dtype=np.int64) is the better style to return something empty

straxen/plugins/peaks/peak_subtyping.py
```diff
@@ -0,0 +1,621 @@
...
+    indices = []
+    for r in result:
+        indices.append(np.arange(r[0], r[1]))
```
WenzDaniel, line 594: FYI: you could also concatenate instead of appending. Then you can remove the flattening operation in the next line.

WenzDaniel: However, this might be less efficient. More efficient would be defining an array of the appropriate length first, adding information and then computing unqiue.

straxen/plugins/peaks/peak_subtyping.py
```diff
@@ -0,0 +1,621 @@
...
+    if len(result) == 0:
+        return np.array([], dtype=np.int), np.array([], dtype=np.int)
+    if not hasattr(reference_areas, '__len__'):
+        reference_areas = np.full(len(result), reference_areas)
+    assert len(result) == len(reference_areas), ''\
+        + 'result and reference_areas must have the same length'
+    small_indices = []
+    large_indices = []
+    for r, ref in zip(result, reference_areas):
+        small_indices.append(np.arange(r[0], r[1])[areas[r[0]:r[1]] <= ref])
```
WenzDaniel, line 617: I would also break this down to make it more readable, but also a bit more efficient e.g.: 

```
_area_is_larger_reference = areas[...] > ref
.... np.arange(r[0], r[1])[_area_is_larger_reference]
.... np.arange(r[0], r[1])[~_area_is_larger_reference]
```

WenzDaniel (commented):
I have not checked the code in detail. I just gave some general comments about readability and maintainability.

straxen/plugins/peaks/peak_subtyping.py
```diff
@@ -0,0 +1,621 @@
...
+             "if no S2 is found in s1_s2_window"
+    )
+
+    after_s2_window_ext_fac = straxen.URLConfig(
+        default=2.0, type=(int, float),
+        help="extend scanning window after identified primary S2 by this much full drift time"
+    )
+
+    def infer_dtype(self):
+        dtype = [('subtype', np.int16, 'time subtyping of peaks')] + strax.time_fields
```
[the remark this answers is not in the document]
Jianyu010: The subtyping is mostly based on time correlations so "time subtyping" is not entirely unreasonable. But I would agree with removing time just to reduce confusion

straxen/plugins/peaks/peak_subtyping.py
```diff
@@ -0,0 +1,621 @@
...
+    Note: 
+      In this plugin, the timing of peak is only defined by their center_time.
+      And the length of peaks are zero.
+    """
+
+    __version__ = '0.2.1'
+    provides = 'subtype_mask'
+    depends_on = ('peak_basics')
+    save_when = strax.SaveWhen.NEVER
+    subtype_dtype = np.int16
```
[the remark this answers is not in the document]
dachengx: Yes, you are right, also `np.int` is deprecated now: https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations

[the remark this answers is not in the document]
dachengx: The subtyping can be larger than 255. I will update this information also in the data_type.

straxen/plugins/peaks/peak_subtyping.py
dachengx: For readability, I agree. But will this slow down the calculation?

straxen/plugins/peaks/peak_subtyping.py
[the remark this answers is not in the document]
WenzDaniel: Yes, but it does not have to be right? We do not have 255 different types, do we? But I think 16 bit is also fine

straxen/plugins/peaks/peak_subtyping.py
WenzDaniel: I never tested it, but I do not think so. At least I doubt that it will be noticeable

straxen/plugins/peaks/peak_subtyping.py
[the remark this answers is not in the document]
dachengx: @Jianyu010 Maybe you will have some ideas. : )

WenzDaniel:
@dachengx I think this one was good to go, but we need to fix now conflicts due to pre-commit.

dachengx:
> @dachengx I think this one was good to go, but we need to fix now conflicts due to pre-commit.

I was discussing this plugin with @Jianyu010 , but found some discrepancy between the proposed PR and his result. So I would convert this to a draft. 

Another point I want to raise is that, the `PeaksSubtypes` is actively developed in https://github.com/XENONnT/s2_only: https://github.com/XENONnT/s2_only/blob/0b6833a7af438877ca6269fc9c53432dc0c8c339/processors/SubTyping_Class.py#L21. So maybe we should remove this plugin from straxen, and add the completed version later when really needed for reprocessing.

PR #1201, opened 2023-06-12
Try to make hashing more coinsistent

LuisSanchez25:

It aims to ensure URLConfigs with the same strings but in different order gives you the same hash by alphabetizing the arguments as a preprocess for the URLConfig class.

Before the URL is hashed, a preprocessing function for the URLConfig class is run to alphabetize the order of the queries in the URL

This is also included on the tests:
st1 = xenonnt_online()
st2 = xenonnt_online()
# Same URLs but the queries are in a different order
st1.apply_xedocs_configs(version = 'v_test1', db='development_db')
st2.apply_xedocs_configs(version = 'v_test2', db='development_db')
assert st1.key_for(25000, 'corrected_areas').lineage_hash == st2.key_for(25000, 'corrected_areas').lineage_hash 

### _Notes on testing_


LuisSanchez25:
Hey Yossi, I made the code with the preprocessor decorator (altough I dont think this was actually necessary since there was already inbuilt code to the URLConfig class to handle this). I tried following he code on how context configs are set and at no point that I could find was the class URLConfig ever called. I dealt with the issue in the apply_xedocs_config by including a function that reorders the URLs before calling the set_config function, at which point I believe the hash is made.

tests/test_contexts.py
```diff
@@ -189,3 +189,14 @@ def test_sim_context():
...
     st = xenonnt('latest')
     st.provided_dtypes()
+
+
+@unittest.skipIf(not straxen.utilix_is_configured(), "No db access, cannot test!")
+def test_xedocs_global_version_hash_coinsistency():
+    st1 = xenonnt_online()
+    st2 = xenonnt_online()
+    # Same URLs but the quesries are in different order
+    st1.apply_xedocs_configs(version = 'v_test1', db='development_db')
```
jmosbacher, line 199: Tests should not depend on the state of the development database. 
Please set the context config in the test using two differently ordered URLs.
Move the test to test_url_config.py

tests/test_url_config.py
```diff
@@ -344,6 +344,14 @@ def test_global_version_preprocessor(self):
         self.st.set_config({'test_config': 'fake://url?version=global_v1'})
         p = self.st.get_single_plugin(nt_test_run_id, 'test_data')
         self.assertEqual(p.test_config, 'fake://url?version=v0')
+
+    def alphabetize_url_kwargs(self):
```
jmosbacher, line 348: Why is alphabetize_url_kwargs a method of URLConfig? keep it in the root of the url_config.py file and import it directly from there.

jmosbacher (changes requested):
Next time please clean your PRs up before requesting a review.

WenzDaniel:
@LuisSanchez25 short question how does then the URL config in the lineage look like? 
```
st.key_for(run_id, data_type).lineage
```
Will it be alphabetically sported or as provided as config to the plugin?

jmosbacher:
@WenzDaniel The preprocessors are run by URLConfig at the plugin initialization stage before setting `plugin.config`, so the URL in the lineage will always be sorted.

WenzDaniel:
I am afraid I have not followed the issue enough, but I am worried a bit that this could lead to confusion. What would happen if I pass the sorted config string again as a config in straxen? Would this work? Will we confuse analysts with having a different config in the lineage with respect to the plugin code?

jmosbacher:
Right, but on the other hand this is nothing new right? The context already performs modifications on the config of the plugin prior to setting it, such as using the default value if it's not set or taking the value from a dict based on the run I'd if the defaults by run I'd is enabled. This just adds to that logic... 
The change to the url is rather minor since it just changes the order of the parameters at the end of the url. 
That being said I was not a strong advocate of making all equivalent urls resolve to the same hash, I just mentioned that is easy to implement using the preprocessor framework. But it did seem to be the preferred option by most in the team a meeting we had about it

WenzDaniel:
> Right, but on the other hand this is nothing new right? The context already performs modifications on the config of the plugin prior to setting it, such as using the default value if it's not set or taking the value from a dict based on the run I'd if the defaults by run I'd is enabled. This just adds to that logic...

I see, my impression is just that this is a bit more transparent, but also might be just personal bias.  

> That being said I was not a strong advocate of making all equivalent urls resolve to the same hash, I just mentioned that is easy to implement using the preprocessor framework. But it did seem to be the preferred option by most in the team a meeting we had about it

I also think that it makes a lot of sense and that the hash should depend on the data and not the order of the config. I am just wondering if this needs more communication with the users. E.g. force all default settings to be sorted, and warn the user in case a non-sorted setting is used, that it is going to be sorted in the hash. But might also be  over engineering.

jmosbacher:
A warning would be nice but it would probably have to be a single global warning otherwise we will have tens of warnings every time someone processes any data. This makes the implementation of the warning a bit more complex.

LuisSanchez25:
I think we can put a warning for maybe this version informing people about this change and to use older version of straxen if the want to access data they already processed with their URLs. We can then remove the warning in a version or two from now when people become familiar with this.

LuisSanchez25:
@jmosbacher @WenzDaniel I added a warning to the init file which should only be displayed when you first import the straxen file. Let me know if I should change the message in any way or if there are any concerns with the way I implemented it.

WenzDaniel:
Thanks @LuisSanchez25. Two little things I doubt that the next release will be 2.0.8, but more like 2.1.0. I think there were quite some changes to reason a major bump, but maybe @dachengx can comment. Maybe add a final sentence like, that nothing changes in the usage of URL configs or how one overwrites a default setting.

dachengx:
> Thanks @LuisSanchez25. Two little things I doubt that the next release will be 2.0.8, but more like 2.1.0. I think there were quite some changes to reason a major bump, but maybe @dachengx can comment. Maybe add a final sentence like, that nothing changes in the usage of URL configs or how one overwrites a default setting.

Thanks, @LuisSanchez25 . I agree that the next release should be at least 2.1.0

jmosbacher:
Can you only raise a warning if a URL was affected by the preprocessor? So can we move the warning logic to the preprocessor function and raise the error if both the URL was changed and warning has not yet been raised

LuisSanchez25:
From my understanding the warning function will only appear once when it is called, I have checked this by trying to invoke it multiple times so this setup should work. I believe this is ready to merge.

jmosbacher:
Can we change this to:
```
From straxen version 2.1.0 onward, URLConfig parameters will be sorted alphabetically before being passed to the plugins, this will change the lineage hash for non-sorted URLs. To load data processed with non-sorted URLs, you will need to use an older version.
```

jmosbacher (commented):
Thank Luis, looks good. The only thing i dont understand is why the test requires utilix database? Cant you just test it with a test context?

LuisSanchez25:
Ah yes you are correct, I will change it to using the testing context

tests/__init__.py
jmosbacher, line 1: this change should be reverted.

LuisSanchez25: done

dachengx:
Hey, @jmosbacher. Is this good enough? It looks good to me, especially the `URLConfig.preprocessor` and warning.

jmosbacher:
@dachengx there are still two open changes i requested:

- change the warning text to: "From straxen version 2.1.0 onward, URLConfig parameters will be sorted alphabetically before being passed to the plugins, this will change the lineage hash for non-sorted URLs. To load data processed with non-sorted URLs, you will need to use an older version."
- Remove the utilix condition from the test.

dachengx:
@LuisSanchez25 For the 2nd request, I guess you can try https://github.com/XENONnT/straxen/blob/8da4b6fe28c3bcbbee6a68b11fd411cb98e95ace/tests/plugins/_core.py#L68.

tests/test_url_config.py
```diff
@@ -344,6 +344,22 @@ def test_global_version_preprocessor(self):
...
+        """URLConfig preprocessor to rearange the order of arguments given buy a 
+        url to ensure the same url with a different hash order gives the same hash"""
+        url = "xedocs://electron_lifetimes?run_id=034678&version=v5&attr=value"
+        intended_url = "xedocs://electron_lifetimes?attr=value&run_id=034678&version=v5"
+        preprocessed_url = straxen.url_config.alphabetize_url_kwargs(url)
+        self.assertEqual(intended_url, preprocessed_url)
+
+    def test_xedocs_global_version_hash_coinsistency(self):
+        st1 = self.st
+        st2 = self.st
```
jmosbacher, line 358: This doesnt test anything, you are using the same context object for both st1 and st2 so of course the lineages will match no matter whether the sorting is working or not, you are comparing the context to itself.

tests/test_url_config.py
```diff
@@ -344,6 +344,22 @@ def test_global_version_preprocessor(self):
...
+        # Same URLs but the queries are in a different order
+        st1.set_config({"test_config":"fake://electron_lifetimes?run_id=25000&version=v5&attr=value"})
```
jmosbacher, line 360:
```suggestion
        st1 = self.st.new_context(config={"test_config":"fake://electron_lifetimes?run_id=25000&version=v5&attr=value"})
```

tests/test_url_config.py
```diff
@@ -344,6 +344,22 @@ def test_global_version_preprocessor(self):
...
+        st2.set_config({"test_config":"fake://electron_lifetimes?attr=value&run_id=25000&version=v5"})
```
jmosbacher, line 361:
```suggestion
        st2 = self.st.new_context(config={"test_config":"fake://electron_lifetimes?attr=value&run_id=25000&version=v5"})
```

tests/test_url_config.py
```diff
@@ -344,6 +344,22 @@ def test_global_version_preprocessor(self):
...
+    def test_alphabetize_url_kwargs(self):
+        """URLConfig preprocessor to rearange the order of arguments given buy a 
+        url to ensure the same url with a different hash order gives the same hash"""
+        url = "xedocs://electron_lifetimes?run_id=034678&version=v5&attr=value"
+        intended_url = "xedocs://electron_lifetimes?attr=value&run_id=034678&version=v5"
+        preprocessed_url = straxen.url_config.alphabetize_url_kwargs(url)
+        self.assertEqual(intended_url, preprocessed_url)
+
+    def test_xedocs_global_version_hash_coinsistency(self):
+        st1 = self.st
```
jmosbacher, line 357:
```suggestion
```

tests/test_url_config.py
jmosbacher, line 358:
```suggestion
```

jmosbacher: they dont, but you did not make a copy of the context. You simply set two names to point to the same python object. 
This has nothing to do with strax or contexts specifically. This is just how assignment works in python.

jmosbacher (approved):

PR #1208, opened 2023-06-23
Fix timing of peaks when ordering in `center_time`

dachengx:

`strax.touching_windows` requires that the `time` of `things` and `containers` are all sorted. 

When the `time` and `endtime` are defined based on `center_time` in some plugins like `PeakShadow` and `PeakAmbience`, the `touching_windows` will show an error, because the `center_time` is not sorted. 

Reference: https://github.com/AxFoundation/strax/pull/725

First, order the peaks by `center_time`, after the features are calculated, recover the original order. 

### _Notes on testing_

### DIFF
--- a/straxen/plugins/peaks/peak_ambience.py
+++ b/straxen/plugins/peaks/peak_ambience.py
@@ -16,7 +16,7 @@ class PeakAmbience(strax.OverlapWindowPlugin):
     References:
         * v0.0.7 reference: xenon:xenonnt:ac:prediction:shadow_ambience
     """
-    __version__ = '0.0.7'
+    __version__ = '0.0.8'
     depends_on = ('lone_hits', 'peak_basics', 'peak_positions')
     provides = 'peak_ambience'
     data_kind = 'peaks'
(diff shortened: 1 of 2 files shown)


FaroutYLq (approved):
Looks good to me. Just one quick sanity check: is it possible for two peaks to share same time? Seems no to me.

FaroutYLq:
Will it be possible that a merged_s2s has same time as the S1 inside?

dachengx:
> Looks good to me. Just one quick sanity check: is it possible for two peaks to share same time? Seems no to me.

We have no functions to prevent that. 

Similar case to `center_time`. But the `mergesort` algorithm of `argsort` is stable sorting so that even though the `center_time` of two peaks can be the same, they will still be reordered correctly in the final result, because the `argsort` of two same `center_time` is still in the order of indices when it is stable.

FaroutYLq:
> > Looks good to me. Just one quick sanity check: is it possible for two peaks to share same time? Seems no to me.
> 
> 
> 
> We have no functions to prevent that. 
> 
> 
> 
> Similar case to `center_time`. But the `mergesort` algorithm of `argsort` is stable sorting so that even though the `center_time` of two peaks can be the same, they will still be reordered correctly in the final result, because the `argsort` of two same `center_time` is still in the order of indices when it is stable. 

Then I have no concern. Please feel feee to merge it.

PR #1209, opened 2023-06-26
Move `get_window_size` factor of merged_s2s as untracked configuration

dachengx:

The original factor is 5. This PR moves it to an untracked context configuration, because in reprocessing we might want to change it, but better without changing the source code. 

It is trying to resolve: https://github.com/AxFoundation/strax/issues/490

This PR will not change the result of `MergedS2s`. It will not change the lineage either. 

### _Notes on testing_


WenzDaniel (approved):
Thanks, from a technical point of few it should be fine.

dachengx:
Thanks, both of you!

PR #1211, opened 2023-06-28
Only print out warning once

LuisSanchez25:

Make sure the warning regarding the change in URL configuration to restructure the arguments alphabetically only prints out once.

Create a singleton class that always returns the same instance when you try to create it. Then when the warning function is called it changes the warning_shown flag to True to ensure this code will not run again unless you restart your kernel.

### _Notes on testing_


LuisSanchez25:
Maybe I should put the warning class somewhere else since it could be used outside of just URL Config issues?

dachengx:
> Maybe I should put the warning class somewhere else since it could be used outside of just URL Config issues?

Yes, maybe to the utils.

jmosbacher (commented):
This looks way overly complex. The singleton pattern is useful for some cases, but in this case you just need a simple flag so im not sure why you need all of this complexity. 
I would just use a global flag. is there any specific reason you implemented a singleton?

jmosbacher:
Also there is a reason we use the warnings builtin instead of print. The warnings package allows the user to suppress specific categories of warnings etc via global configuration. print statements cannot be controlled by the user.

jmosbacher:
Please make this more pythonic. the current solution looks like a Java style solution.
Here are some more pythonic options:

### Option 1: Use warnings package
```python
import warnings

class URLSortWarning(warnings.UserWarning):
    pass

warnings.filterwarnings("once", category=URLSortWarning,)

``` 
Then raise this warning category in the protocol.

### Option 2: Use pythons `global` keyword

```python

WARNING_RAISED = False

def alphabetize_url_kwargs(url: str):
    global WARNING_RAISED
    ...
    if url != URLConfig.format_url_kwargs(url) and not WARNING_RAISED:
         warnings.warn("...")
         WARNING_RAISED = True

```

### Option 3: decorator with flag
```python
import warnings
from functools import wraps


def warn_if_changed(f):
    WARN = True
    @wraps(f)
    def wrapper(url, **kwargs):
        nonlocal WARN
        new_url = f(url, **kwargs)
        if new_url != url and WARN:
            warnings.warn("...")
            WARN = False
        return new_url
    return wrapper

```

jmosbacher:
@LuisSanchez25 please simplify the code so we can merge it.

LuisSanchez25:
Hey Yossi, I did not have a particular reason for doing things the way I did before other than the other ways I thought about doing this did not work. I will work on this again soon but I was told it was fine to leave the warning as they were for a while to make sure people where aware of the changes.

jmosbacher:
Please see above, i added three examples of more pythonic ways of doing this

LuisSanchez25:
Yes I did see your suggestions, thanks for showing me these options.

jmosbacher (approved):

PR #1212, opened 2023-07-04
Nv time sync sr1

jjakob03:
Permits negative delays between NV and TPC

Defines a certain error value and checks for exact equality instead only for negative numbers.


WenzDaniel:
This one is not needed anymore.

PR #1216, opened 2023-07-07
Bad url warnings

LuisSanchez25:

This code should help prevent using bad/incorrect urls when using xedocs URLConfigurations. It gives warning if things that are expected to be part of certain urls is not present.

I made a preprocessor that looks through many of the potential errors someone could making a url. If any issues is found, a warning will be given to the user. This should also help prevent mistakes when building new global version.

The following code will have a very wrong URL configuration that should trigger a good amount of the warnings.
st = straxen.contexts.xenonnt(xedocs_version='global_v11')
st.set_config({'elife': 'list-to-array://xedocs://posrec_models?verion=v5&run_id=034343&atrr=value'})
st.get_single_plugin('034343', 'corrected_areas').elife

### _Notes on testing_


LuisSanchez25:
I also swapped the position of the run-docs to keep all the preprocessing functions together towards the end of the document.

jmosbacher (commented):
@LuisSanchez25 thanks for these checks. I think we should try to make this a bit more extensible and maintainable instead of just hardcoding a list of functions to call.
I think one pattern that may be suitable here would be to have a regex dispatcher that you can register functions to be called on URLs that match the regex. 
For an example of a regex dispatcher see [xedocs](https://github.com/XENONnT/xedocs/blob/master/xedocs/regex_dispatcher.py), you will have to modify it to support multiple calls of course.
I think is would be nice to also support passing arguments to the `register` method like "warn" and "raise" which accept a warning or exception to be raised if the test doesnt pass. 

One more comment is that I think maybe its time to organize the urlconfig code since its growing. Perhaps we should take the opportunity to make urlconfigs a folder will multiple files, one for the descriptor code, one for the protocols and one for the preprocessors and maybe even one for URL tests (most of this PRs code would go there)

LuisSanchez25:
@dachengx any clue why the tests are not running in this PR?

straxen/url_config/dispatcher.py
```diff
@@ -0,0 +1,85 @@
...
+
+@check_urls.register(r".*(list-to-array|'object-to-dict').*")
+def data_as_list_check(url: str):
+    if not ("as_list=True" in url):
+        warnings.warn("When using the list-to-array or object-to-dict protocol, you must include an as_list=True in the URL arguments")
+    if not ("sort" in url):
+        warnings.warn("When using the list-to-array or object-to-dict protocol, you must include a sort argument in the URL")
+
+
+@check_urls.register(r".*")
```
jmosbacher, line 25: only relavent to xedocs URLs, regex should require xedocs://

straxen/url_config/dispatcher.py
```diff
@@ -0,0 +1,85 @@
...
+def confirm_xedocs_schema_exists_check(url: str):
+    import xedocs
+    # Selects the schema name in the URL and confirms it exists in the xedocs db
+    if not re.findall(r"(?<=\://)[^\://]*(?=\?)", url)[0] in xedocs.list_schemas():
+        warnings.warn("You tried to use a schema that is not part of the xedocs database")
+
+
+@check_urls.register(r".*(list-to-array|'object-to-dict').*")
```
jmosbacher, line 17: only relevant to the combination of xedocs protocol and list-to-array protocol. 
also, why does object-to-dict protocol require a list?

jmosbacher (dismissed):
Thanks Luis. Also have some general comments

- most of the regexs are too general for the tests you wrote, try to use the regex to filter only the URLs that actually are relevant to the check.
- file naming: rename "url_config" folder --> "config", "dispatcher.py" --> "config_tests.py"
- add a new file called "preprocessors.py" for the preprocessors
-  I think we should have the dispatcher raise the warnings and add the URL that caused the issue to the warning. also would be good to use a custom warning class so all these warnings can be disabled via the warnings package if needed.

LuisSanchez25:
In the Regex_Dispatcher itself, the call function only triggers if the urls contains 'xedocs' as I was not thinking about other kinds of urls, I also thought it would be cleaner to put that check there so I did not have to write a separate xedocs check for each function. If this is bad practice or something let me know, but then maybe in the call function I can register a separate set of functions that are more general? Or do you think its better to remove the 'xedocs' check in the regex_dispatcher itself and just add that as a condition to check the the dispatchers we register that are only for xedocs?

straxen/url_config/dispatcher.py
```diff
@@ -0,0 +1,85 @@
...
+
+@check_urls.register(r".*posrec_models.*")
+def tf_check(url: str):
+    if not ("tf://" in url):
+        warnings.warn("Warning, you are requesting a position reconstruction model with this URL however, the protocol tf:// was not requested as part of the URL. The tf:// protocol is requiered for position reconstruction corrections.")
+
+
+@check_urls.register(r".*")
+def url_attr_check(url: str):
+    if not ("attr" in url):
```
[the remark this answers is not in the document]
LuisSanchez25: I havent run into any of those cases in the corrections so far, I think all the urls in global_v11 had this, and since its just a warning I think anyone digging deep enough to change the attribute should be fine?

straxen/url_config/dispatcher.py
LuisSanchez25: object-to-list expects a list as an input

jmosbacher: so i guess its a typo and should be "objects-to-dict"?

jmosbacher:
I think its a useful to collect sanity checks for all URLs, not just xedocs related ones.

straxen/url_config/dispatcher.py
```diff
@@ -0,0 +1,85 @@
...
+def confirm_xedocs_schema_exists_check(url: str):
+    import xedocs
+    # Selects the schema name in the URL and confirms it exists in the xedocs db
+    if not re.findall(r"(?<=\://)[^\://]*(?=\?)", url)[0] in xedocs.list_schemas():
+        warnings.warn("You tried to use a schema that is not part of the xedocs database")
+
+
+@check_urls.register(r".*(list-to-array|'object-to-dict').*")
+def data_as_list_check(url: str):
+    if not ("as_list=True" in url):
```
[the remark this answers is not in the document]
LuisSanchez25: Hey Yossi, could you explain the logic behind this change a little more?

straxen/config/preprocessors.py
```diff
@@ -0,0 +1,33 @@
+from .config_tests import check_urls
+from .url_config import URLConfig
+import warnings
+
+WARN = True
+
+
+@URLConfig.preprocessor
+def alphabetize_url_kwargs(url: str):
```
jmosbacher, line 9: can we rename this to `sort_url_kwargs` ?

jmosbacher:
Maybe group all the xedocs related checks into a single dispatcher function and register it under the main dispatcher in the preprocessor with a `xedocs://` regex to avoid repeating yourself.

LuisSanchez25:
wouldnt putting all of the xedocs preprocessor in the same function make the code harder to read/ to modify in the future?

WenzDaniel:
Long time no new news converting into a draft.

LuisSanchez25:
There is one slight problem that I noticed while testing this. When we use URLConfigs here and a version is not spesified, it seems like the ONLINE version is the one assumed, this can however be a bit dangerous as it could lead it issues. There is a warning that flags when the "version" is missing but we might what the produce an error instead. Here is an example:

 straxen.URLConfig.evaluate_dry("xedocs://electron_lifetimes?verion=v12&run_id=044332")

Here "version" is deliberately missplead to trigger the warning however we want a real error in this case instead of the ONLINE version.

dachengx (changes requested):
Thanks @LuisSanchez25 .

I fixed a bug in the codes here 9bdd5212b08ea5711c1bcda5b11dcd2f5f07abd0.

When I run:
```
import cutax
st = cutax.xenonnt_offline()
run_id = '026000'
st.key_for(run_id, 'event_info')
```

I got:
```
/home/xudc/straxen/straxen/config/config_tests.py:106: URLWarning: A URL for fdc was given without a [scale_coordinates] argument. This could lead to issues when reprocessing data. Please include the scaling in the URL. url: itp_map://resource://format://xedocs://fdc_maps?algorithm=plugin.default_reconstruction_algorithm&fmt=json.gz&attr=value&run_id=plugin.run_id&version=v9
  warnings.warn(
/home/xudc/straxen/straxen/config/config_tests.py:50: URLWarning: A URL which requests the resource:// was given. However resource:// was not found within the URL, not data will be loaded. url: xedocs://s1_xyz_maps?version=v7&run_id=plugin.run_id&algorithm=plugin.default_reconstruction_algorithm&attr=map
  warnings.warn(
/home/xudc/straxen/straxen/config/config_tests.py:63: URLWarning: Warning, you are requesting a map file with this URL. However, the protocol itp_map:// was not requested as part of the URL. The itp_map:// protocol is requiered for map corrections. url: xedocs://s1_xyz_maps?version=v7&run_id=plugin.run_id&algorithm=plugin.default_reconstruction_algorithm&attr=map
  warnings.warn(
```

But when I run:
```
url = 'xedocs://s1_xyz_maps?version=v7&run_id=plugin.run_id&algorithm=plugin.default_reconstruction_algorithm&attr=map'
straxen.URLConfig.evaluate_dry(url, run_id=run_id, algorithm=DEFAULT_POSREC_ALGO)
url = 'itp_map://resource://format://xedocs://fdc_maps?algorithm=plugin.default_reconstruction_algorithm&fmt=json.gz&attr=value&run_id=plugin.run_id&version=v9'
straxen.URLConfig.evaluate_dry(url, run_id=run_id, algorithm=DEFAULT_POSREC_ALGO)
```

Everything runs well. So maybe those warnings are not necessary?

LuisSanchez25:
Hey Dacheng, those warnings are all necessary for the following reasons:

1.  Loading a map without the [scale_coordinates] URL will lead to incorrect loading of the data itself, this is why this protocol was added in the first place.
2.  This will only load the name of the file, not the file itself. Resources have the protocol to actually fetch the file, when you try to process data without this it will lead to errors.
3.  Itp_maps is a similar story to 1.

So these protocols are necessary, if you have any doubts as to what these protocols are doing maybe we can work to comment the protocols better.

LuisSanchez25:
@dachengx can you tell me if you are fine with the points I raised to see if we can merge this?

dachengx:
Overall, if we do need these warnings, how can our previous processing be correct?

Running

```
url = 'itp_map://resource://format://xedocs://fdc_maps?algorithm=plugin.default_reconstruction_algorithm&fmt=json.gz&attr=value&run_id=plugin.run_id&version=v6'
fdc_map = straxen.URLConfig.evaluate_dry(url, run_id=run_id, algorithm=DEFAULT_POSREC_ALGO)
```
then `fdc_map.coordinate_system[:, 2].max()` is O(1e6), whose exact number is hidden because this is a public repo.
but when change the `v6` to `v8`, the `fdc_map.coordinate_system[:, 2].max()` is O(1e2), which is the size of XENONnT TPC.

1. Since v8 `fdc_map`, no `scale_coordinates` need to be applied: https://github.com/XENONnT/corrections/blob/1b553d0f259b137fe2dcbaf971b46f24d15f4f23/XENONnT/global_versions/global_v12.json, so that check is not always needed. I will remove it.
2. v7 `s1_xyz_map` is just a json file: https://github.com/XENONnT/corrections/blob/1b553d0f259b137fe2dcbaf971b46f24d15f4f23/XENONnT/global_versions/global_v14.json#L95 https://github.com/XENONnT/corrections/blob/1b553d0f259b137fe2dcbaf971b46f24d15f4f23/XENONnT/s1_xyz/data/mlp_v7.json#L44. `straxen.InterpolatingMap` will load it automatically: https://github.com/XENONnT/straxen/blob/7e1b84b4a9ac317606d38f908f2753931e9a725e/straxen/itp_map.py#L135. So that check is not really needed when `attr=map`, because https://github.com/XENONnT/xedocs/blob/42ccece67cc568df827814e3f24c342bca039436/xedocs/schemas/corrections/base_references.py#L122. I will add another condition before output the warning.

dachengx (approved):

dachengx (approved):
Looks cool now. Thanks @LuisSanchez25 for these works!

PR #1217, opened 2023-07-09
Validate final type after URL eval

jmosbacher:
If type is given, validate the config type after URL evaluation.
Also added the run_id to the cache key.

### DIFF
--- a/straxen/url_config.py
+++ b/straxen/url_config.py
@@ -280,6 +280,22 @@ class URLConfig(strax.Config):
                     f"Invalid type for option {self.name}. "
                     f"Excepted a {self.final_type}, got a {type(cfg)}")
 
+    def validate_type(self, value):
+        """This method is called by the context on plugin initialization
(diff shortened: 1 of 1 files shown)


dachengx:
Thanks @jmosbacher . Are codes that can change the results only `def validate_type`?

jmosbacher:
@dachengx Oh right, sorry about that. I ran auto formatting so there are a bunch of formatting changes not really relevant to the actual change.
The `validate_type` and `fetch` methods indeed include the actual "logic" changes. 
I am a bit on the fence whether we should warn when the type is incorrect or raise an error. Any opinion?

dachengx:
No problem. Since `type`  does not influence lineage, maybe we can raise errors and also modify the current code if needed.

jmosbacher:
Ok, I will switch it to raise an error.

dachengx (approved):
Looks great. Thanks @jmosbacher .

PR #1218, opened 2023-07-14
scale coordinate properly

jingqiangye:

The FDC map is made based on (x, y, drift time), it has to be scaled by drift velocity before applying corrections based on (x, y, z). However, the current implementation of coordinate scaling is wrong.

Use 
```
self.map.scale_coordinates([1., 1., - self.electron_drift_velocity])
```
to properly scale the coordinate.

The illustration can be found [here](https://xenonnt.slack.com/archives/C016UJZ090B/p1689371326763939).


jingqiangye:
This is solved by https://github.com/XENONnT/corrections/pull/197.

PR #1230, opened 2023-08-07
Allow saving peaklets in main and alt peaks

shenyangshi:
Allow saving peaklets in main and alt S1 and S2
image


straxen/plugins/peaklets_events/peaklets_events.py
```diff
@@ -0,0 +1,75 @@
...
+        help='Number of TPC PMTs')
+
+    sum_waveform_top_array = straxen.URLConfig(
+        default=True,
+        type=bool,
+        help='Digitize the sum waveform of the top array separately'
+    )
+
+    def setup(self):
+        return
```
WenzDaniel, line 35: ?

straxen/plugins/peaklets_events/peaklets_events.py
```diff
@@ -0,0 +1,75 @@
...
+    def infer_dtype(self):
+        return strax.peak_dtype(
+            n_channels=self.n_tpc_pmts,
+            digitize_top=self.sum_waveform_top_array,
+        )
+
+    def fake_mask(self, df, start_field='time', end_field='end_time', pad=10e3):
+        """
+        Create a fake mask to touch the peaklets
+        :param df: events
```
WenzDaniel, line 46: I would call it `events` and not `df` for the sake of clean code and readability. If you call it events I know directly what the object is suppose to represent. Like it is know I will think that I have to provide a dataframe here which is most likely wrong as plugins handle data in from of structure arrays. 

In addition I think the entire function is not needed. See my comment below.

straxen/plugins/peaklets_events/peaklets_events.py
```diff
@@ -0,0 +1,75 @@
...
+        for tag in ['s1', 'alt_s1', 's2', 'alt_s2']:
+            result = strax.touching_windows(peaklets,
+                                            self.fake_mask(events, f'{tag}_time',
+                                                           f'{tag}_endtime'))  # touch the peaklets
+            if len(result) > 0:
+                result = np.concatenate([np.arange(result[i][0], result[i][1], dtype=int) for i in
+                                         range(len(result))])  # find the index of the touched peaklets
+                peaklets_events_id.append(np.array(result))
+        peaklets_events = peaklets[np.sort(np.concatenate(peaklets_events_id))]  # sort by time
+        peaklets_events = peaklets_events[np.argsort(peaklets_events['time'])]  # sort again just in case
```
WenzDaniel, line 73: I think this can be done easier. We provide an index field for peaks in a given event is not it? You can simply do:

```python
peaks_indicies_in_event = strax.touching_windows(peaks, events)

for indicies, event in zip(peaks_indicies_in_event , event):
    start_index, end_index = indicies
    peaks_in_event = peaks[start_index:end_index]
    s1_in_event = peaks_in_event[event['s1_index']] 
```

You can even numbafy this for loop if necessary. You know that you can have only n_events S1/S2 etc. So you fill first the corresponding buffer in the numbafied for-loop, drop all null results and concetante and sort afterwards,

WenzDaniel (commented):
Hej Shenyang, thanks but I would like to ask some changes. I think the code can be much simpler.

straxen/plugins/peaklets_events/peaklets_events.py
WenzDaniel: Ah sorry, I made a mistake to make this work via the index you need to use fully_contained_in or split_by_containment instead of touching_windows. See:
https://github.com/XENONnT/straxen/blob/5f232eb2c1ab39e11fb14d4e6ee2db369ed2c2ec/straxen/plugins/events/event_basics.py#L205
and
https://github.com/AxFoundation/strax/blob/9b508f7f8d441bf1fe441695115d292c59ce631a/strax/processing/general.py#L224

WenzDaniel:
Long time no new news, converting to a draft.

WenzDaniel:
Since no activity in a long time I will close now.

PR #1231, opened 2023-08-09
Use formatted float in `dataframe_to_wiki`

dachengx:

In the previous implementation, `straxen.dataframe_to_wiki` will give you '0.5' when `float_digits=2`, which gives the output different lengths. 

Previous:
```
import pandas as pd
import straxen

df = pd.DataFrame({'x': [0.51], 'y': [0.50]})
print(straxen.dataframe_to_wiki(df, float_digits=2))
```

gives output:
```
^ Awesome table ^^
^ x ^ y ^
| 0.51 | 0.5 |
```

Now the new code gives output:
```
^ Awesome table ^^
^ x ^ y ^
| 0.51 | 0.50 |
```

### _Notes on testing_

### DIFF
--- a/straxen/misc.py
+++ b/straxen/misc.py
@@ -32,7 +32,7 @@ export, __all__ = strax.exporter()
 
 @export
 def dataframe_to_wiki(
-    df, float_digits=5, title='Awesome table', force_int=()):
+    df, float_digits=5, title='Awesome table', force_int:ty.Tuple=()):
     """Convert a pandas dataframe to a dokuwiki table 
     (which you can copy-paste onto the XENON wiki)
     :param df: dataframe to convert
@@ -47,8 +47,8 @@ def dataframe_to_wiki(
         if isinstance(x, float):
             return f'{x:.{float_digits}f}'
         return x
-    force_int = np.where(np.in1d(
-        df.columns.values, strax.to_str_tuple(force_int)))[0]
+    force_int = np.where(np.in1d(df.columns.values,
+                                 strax.to_str_tuple(force_int)))[0]
 
     for _, row in df.iterrows():
         table += "| " + ' | '.join([
--- a/straxen/misc.py
+++ b/straxen/misc.py
@@ -32,7 +32,7 @@ export, __all__ = strax.exporter()
 
 @export
 def dataframe_to_wiki(
-    df, float_digits=5, title='Awesome table', force_int=tuple()):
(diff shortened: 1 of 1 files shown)


JYangQi00:
Hey Dacheng, I read through the commits and they look fine to me. Correct me if I'm missing something but, it seems like the failing tests are unrelated to your commits?

straxen/misc.py
```diff
@@ -31,27 +31,28 @@
 
 
 @export
-def dataframe_to_wiki(df, float_digits=5, title='Awesome table', 
-                      force_int=tuple()):
+def dataframe_to_wiki(
+    df, float_digits=5, title='Awesome table', force_int=()):
```
WenzDaniel, line 35: General question, should we do explicit typing whenever we change one of those functions? I guess you know what I mean, just for reference: https://docs.python.org/3/library/typing.html

straxen/misc.py
```diff
@@ -31,27 +31,28 @@
...
-    def do_round(x):
+    def format_float(x):
         if isinstance(x, float):
-            return round(x, float_digits)
+            return f'{x:.{float_digits}f}'
         return x
-    force_int = np.where(np.in1d(df.columns.values,
-                                 strax.to_str_tuple(force_int)))[0]
+    force_int = np.where(np.in1d(
+        df.columns.values, strax.to_str_tuple(force_int)))[0]
```
WenzDaniel, line 51: General comment for future code, no need to change here: this guy is in general a bit ugly many operations for a single line.

WenzDaniel (approved):
Looks good.

WenzDaniel:
The error seems to be related to the holoviews displays, but I do not know why.

straxen/misc.py
JYangQi00: Just wondering, numpy's documentation says that np.isin is preferable to np.in1d, but does it have any functional differences that we should worry about?

JYangQi00 (approved):
Everything seems to work for me! 👍

straxen/misc.py
dachengx: `in1d` is much faster as they said https://numpy.org/doc/stable/reference/generated/numpy.isin.html#numpy.isin

JYangQi00: Gotcha, thanks!

straxen/misc.py
dachengx: Updated.

PR #1244, opened 2023-08-27
Give `RunDB` an option to find files in storage

dachengx:
Give RunDB an option to find files in storage but not in database in find_several function.

Historically, the `RunDB` knows where are all the available files located on servers and assumes they are all available. But if we have multiple partitions, then RunDB will provide runs in the database but not available.

This happens a lot when running `strax.Context.select_runs`, in the `strax.StorageFrontend.find_several` function. So this PR gives RunDB an option to run `strax.StorageFrontend.find_several` directly.

### _Notes on testing_

### DIFF
--- a/straxen/contexts.py
+++ b/straxen/contexts.py
@@ -158,6 +158,7 @@ def xenonnt_online(output_folder: str = './strax_data',
                    _auto_append_rucio_local: bool = True,
                    _rucio_path: str = '/dali/lgrandi/rucio/',
                    _rucio_local_path: ty.Optional[str] = None,
+                   _find_in_storage: bool = False,
                    _raw_paths: ty.Optional[str] = ['/dali/lgrandi/xenonnt/raw'],
                    _processed_paths: ty.Optional[ty.List[str]] = ['/dali/lgrandi/xenonnt/processed',
                                                                   '/project2/lgrandi/xenonnt/processed',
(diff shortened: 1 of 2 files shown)


JYangQi00:
Thanks Dacheng, I tried to use `st.select_runs` with `_find_in_storage = False` and with `_find_in_storage = True`, however, I'm having a bit of trouble understanding the output. I'd like to find some runs that cannot be loaded when found using `st.select_runs` with _find_in_storage=False. Then I would expect that when _find_in_storage=True, these runs do not show up when `st.select_runs` is called. Is my interpretation of this fix correct?

However, instead I find that when using _find_in_storage=False, I do not get any extra runs compared to when _find_in_storage=True. I also specified that peak basics should be available when using `select_runs`. Is there an example of a run+data_type that cannot be loaded despite the output of `select_runs` saying that it can? I knew there were some in the past but I haven't found any this time around.

dachengx:
> However, instead I find that when using _find_in_storage=False, I do not get any extra runs compared to when _find_in_storage=True. I also specified that peak basics should be available when using `select_runs`. Is there an example of a run+data_type that cannot be loaded despite the output of `select_runs` saying that it can? I knew there were some in the past but I haven't found any this time around.

Like the `050034-peaklets-ui5hguaz2k` of offline context, but you need to be on a midway computing node like jupyter notebook.

PR #1254, opened 2023-09-04
FDC uses corrected position

shenyangshi:
See https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:analysis:signals_and_detector:20230815:fdc, now straxen uses original positions for the FDC correction but the map is developed for z_dv_corr.

Note that the SR0 result won't be affected as the z_bias_map in SR0 is a dummy map so z_corr=0.

I also delete some fields that don't make sense to me.

### DIFF
--- a/straxen/plugins/events/event_positions.py
+++ b/straxen/plugins/events/event_positions.py
@@ -19,7 +19,7 @@ class EventPositions(strax.Plugin):
 
     depends_on = ('event_basics',)
 
-    __version__ = '0.2.1'
+    __version__ = '0.3.0'
 
     default_reconstruction_algorithm = straxen.URLConfig(
         default=DEFAULT_POSREC_ALGO,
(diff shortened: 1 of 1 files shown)


dachengx:
> The change is minimal, but the logic will be consistent. image

What is the `data` here?

shenyangshi:
It's 043256 with the master branch of straxen and global_v12 correction https://github.com/XENONnT/corrections/pull/219

dachengx:
So can we explain the difference of `r` in your demo by the change from `z_naive` to `z_dv_corr`? Do we actually expect a change, or is the reason for the change only digital precision?

shenyangshi:
I don't know, something is not fully reproducible, but the change is negligible.
image

straxen/plugins/events/event_positions.py
```diff
@@ -116,34 +116,21 @@ def compute(self, events):
...
             r_obs = np.linalg.norm(orig_pos[:, :2], axis=1)
-            delta_r = self.map(orig_pos)
             z_obs = z_obs + self.electron_drift_velocity * self.electron_drift_time_gate
 
-            # apply radial correction
+            # apply z bias correction
+            z_dv_delta = self.z_bias_map(np.array([r_obs, z_obs]).T, map_name='z_bias_map')
+            corr_pos = np.vstack([events[f"{xy_pos}x"], events[f"{xy_pos}y"], z_obs - z_dv_delta]).T
+            # apply r bias correction
+            delta_r = self.map(corr_pos)
```
dachengx, line 129: @shenyangshi If I am correct, you changed the coordinate the map applied on from `z_obs ` to `z_obs - z_dv_delta`. So actually you will see a difference between results before and after this PR. So I do not understand why the difference is so small.

shenyangshi: Right, because the Z correction is only happening at the bottom outside corner, so other events won't be changed by a lot.

dachengx: So we actually do not expect a non-zero difference, no? And whatever the master branch gives us, it is wrong.

dachengx (approved):

straxen/plugins/events/event_positions.py
shenyangshi: No we do expect a non-zero difference for R, as the z_bias map will do some nearly negligible thing inside the bulk tpc too. What I don't understand is why the z has a change, but it's also negligible.

shenyangshi:
Do you want to do a test by yourself? Corrections and versions in my folder are changing rapidly, I cannot promise I didn't make any stupid mistake in checking.

dachengx:
Using environment:
```
^ Awesome table ^^^^
^ module ^ version ^ path ^ git ^
| python | 3.9.17 | /opt/XENONnT/anaconda/envs/XENONnT_2023.07.2/bin/python | None |
| strax | 1.5.3 | /home/xudc/strax/strax | branch:master | a466a94 |
| straxen | 2.1.2 | /home/xudc/straxen/straxen | branch:correct_fdc | f2f8825 |
| cutax | 1.15.3 | /home/xudc/cutax/cutax | branch:master | 4aa29c9 |
```

and code
```
run_id = '043256'

st = cutax.xenonnt_offline()

st.set_config(
    {'fdc_map': 'itp_map://resource://'
     'XnT_3D_FDC_xyz_MLP_v1.2_B2d75n_C2d74n_G0d3p_A5d0p_T0d9n_PMTs1d3n_FSR0d65p.json.gz'
     '?fmt=json.gz&scale_coordinates=plugin.coordinate_scales'}
)

EventPositions0 = st.get_single_plugin(run_id, 'event_positions')
```

I get
image

dachengx:
And I do not think you should use `&scale_coordinates=plugin.coordinate_scales`.

shenyangshi:
Yes indeed there is some change... Maybe due to interpolation errors. We don't have to use `&scale_coordinates=plugin.coordinate_scales` if the map itself is in z but not in drift time, but unfortunately in my reprocessings the url config has this, so I have to make the samething in the new plugin to compare.

jingqiangye (approved):
Hey I did some [checks](https://xe1t-wiki.lngs.infn.it/doku.php?id=jingqiang:straxen_review:1254) and it looks good to me.

dachengx:
> Hey I did some [checks](https://xe1t-wiki.lngs.infn.it/doku.php?id=jingqiang:straxen_review:1254) and it looks good to me.

Professional!

dachengx (approved):

PR #1256, opened 2023-09-06
Use the new S1 optical map for global_v12 reprocessing.

minzhong98:
Use the newer S1 pattern map for global_v12 reprocessing. Related note: [https://xe1t-wiki.lngs.infn.it/doku.php?id=min:s1opticalmap_comparison](url).


jingqiangye:
Hi, xedocs has added the [s1 pattern map](https://github.com/XENONnT/xedocs/blob/master/xedocs/schemas/corrections/s1_pattern_maps.py) schema now. I think it's better to add this to the corrections repo and close this PR.

PR #1266, opened 2023-09-14
Add pad-array protocol

jmosbacher:
Adds a URLConfig protocol wich will pad an array on the left or right or both with a configurable number of elements of the specified value

The protocol accepts the arguments pad_left, pad_right and pad_value to specify how many elements to add and what value to fill them with.

### DIFF
--- a/straxen/legacy/xenon1t_url_configs.py
+++ b/straxen/legacy/xenon1t_url_configs.py
@@ -95,3 +95,9 @@ def get_z_bias(offset: str):
         return np.zeros(len(rz)) * int(offset)
 
     return fake_z_bias
+
+
+@URLConfig.register('pad-array')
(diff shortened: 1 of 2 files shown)


WenzDaniel:
Looks good. Thanks for the quick implementation.

WenzDaniel (approved):

LuisSanchez25:
I was trying to test this protocol but I am a bit confused as to why it is in the xe1t_url_config and not in the normal url_config file, is there something I am missing here?

LuisSanchez25:
I successfully used the protocol to process some nv records so this is ready to be merged

jmosbacher:
@LuisSanchez25 If you tested this, can you add your review? thanks.

LuisSanchez25 (approved):

PR #1270, opened 2023-09-25
Add top channel at peak_basics and above

### LINKED ISSUE #1263: Add n_top_channel at the peak level
Add a field n_top_channel at the [peak_basics level](https://github.com/XENONnT/straxen/blob/master/straxen/plugins/peaks/peak_basics.py#L32-L33) at the event level, since 2-fold S1s mostly wouldn't make to the event level. Thus we can get rid of the dependency on area_per_channel in 2-fold analysis.

shenyangshi:
Add n_top_channel to let s1 bdt get rid of the peaks computational-expensive area_per_channel dependence.

Resolve #1263.


shenyangshi:
image
image

shenyangshi:
@dachengx Do you think this is good to merge?

dachengx:
> @dachengx Do you think this is good to merge?

I suggest putting the new fields in another plugin. Because the already processed data will be affected.

dachengx:
> We'll start the global_v13 reprocessing soon, how about only merging this PR before releasing the new container? @jingqiangye

Please discuss those in other place than public platform.

PR #1285, opened 2023-10-31
Add gps plugins

WenzDaniel:
Removed old syncing plugins and move GPS plugins from cutax to straxen.


WenzDaniel:
I can fix the reviewdoc things, but not sure if I should wait for the pre-commit thing.

WenzDaniel:
To make tests working we need to updated our test data to a run for which also GPS signals are available: https://github.com/XENONnT/straxen/blob/11cdb5c8ae7e27990a84c269c4c3e7778acae50b/straxen/test_utils.py#L16-L21
This should to be discussed in Team A.

WenzDaniel:
The failing test has something to do with a numerical issue in the event pattern fit which cannot be solved right now.

straxen/plugins/gps_syncing/gps_syncing.py
```diff
@@ -0,0 +1,306 @@
...
+    1. Finds the TTL GPS pulses coming into the AM from the gps
+    module and their pairs coming from the module for the
+    correspondent run.
+      2. Corrects the timestamp of all events by linearly interpolating
+    between the previous and next sync pulses.
+
+    """
+
+    __version__ = "0.2.1"
+    depends_on = ("aqmon_hits", "event_info")
```
jjakob03, line 24:
```suggestion
    depends_on = ("aqmon_hits", "event_basics")
```

jjakob03 (dismissed):
Looks good. I did not check mv but nv works :)

WenzDaniel:
Thanks Johanna, the MV plugins were just copied over from cutax. I will try to verify them again.

PR #1287, opened 2023-11-02
Check by default basics

WenzDaniel:
Trivial change, check by default availability of peak and event basics instead of raw records.


dachengx (approved):
Thanks @WenzDaniel

PR #1292, opened 2023-11-20
Pull out FakeDAQ to legacy plugins

WenzDaniel:

Move Fake1TDAQ into legacy plugins to not register in nT context.


WenzDaniel:
Should be merged together with: https://github.com/AxFoundation/strax/pull/776

chnlkx (commented):
Looks good to me. I can import and register this plugin. 
Is there other thing I should check in the reviewing process?

JelleAalbers (approved):
Sounds good; `raw_records_diagnostic` does not exist anymore so this clearly doesn't belong with the nT plugins.

PR #1240, opened 2023-08-24
Proposal to use pre-commit for continuous integration

dachengx:
## What does the code in this PR do / what does it improve?

Stop using the `wemake-python-styleguide` which gives too many errors. And switch to `pre-commit`'s continuous integration.

The advantage of `pre-commit` is that:

1. It runs automatically when you `git commit`.
2. You will not upload codes that are not aligned with some styles.

The codes are changed only in code style, the requirement of hooks.

Note: `black` will force you to use double quotes, it actually auto-format for you, so do not worry.

You will see no reviewdog complaining, and pre-commit will do almost all formatting for you, and it is very fast.

## Can you briefly describe how it works?

To setup pre-commit:

```
cd straxen
pip install pre-commit
pre-commit install
```

Besides basic whitespace and large files check,
1. `black` will **auto-format** your code, mainly about indent and line-length, and make auto-commit if necessary.
2. `docformatter` will **auto-format** your docstring, following google convention, and make auto-commit if necessary.
3. `flake8` will **check** the code-style, but in a more lite way than `wemake-python-styleguide`.

So there will be two pre-commit checks, one check of **changed files** when you `git commit`, and one check of **all files** by pre-commit CI.

## Can you give a minimal working example (or illustrate with a figure)?

When you `git commit`, it will automatically `pre-commit run` for the **changed** files, so actually you would not always run it manually.

If you want to check all the files, run `pre-commit run --all-files`. If you want to check the difference between the current branch and `main`, run `pre-commit run --from-ref main --to-ref .`. If you want to remove the color from the output, add `--color never`.

`black` and `docformatter` will change your code, you still need to make a commit by yourself on the changed part.

For more information, please go to https://pre-commit.com/ and https://pre-commit.ci/.

_Please include the following if applicable:_
  - [x] _Update the docstring(s)_
  - [ ] _Update the documentation_
  - [x] _Tests to check the (new) code is working as desired._
  - [x] _Does it solve one of the open issues on github?_

Inspired by and closes https://github.com/XENONnT/straxen/issues/1192

PR #1293, opened 2023-11-22
Use pre-commit for continuous integration also for scripts

dachengx:

Complete https://github.com/XENONnT/straxen/pull/1240

Change mode of files under `bin` to 775, so that they can be scanned by pre-commit.


dachengx:
Trivial one which is just auto-format and some small fix on the variables' type.

PR #1294, opened 2023-11-23
Report PR changes phase B

dachengx:
Organize codes into python script but there is no functionality changed.

```
bash .github/scripts/check_lineage_updates.sh track_lineage_changes test_plugin_changes
```


.github/scripts/check_lineage_updates.sh
```diff
@@ -1,176 +1,32 @@
...
+git checkout $old_branch
+./report_pr_changes --branch old --computation lineage_hash
+git checkout $new_branch
+./report_pr_changes --branch new --computation lineage_hash
+./report_pr_changes --branch new --computation hash_comparison
+./report_pr_changes --branch new --computation print_added_plugin
+./report_pr_changes --branch new --computation changed_affected_plugin
+git checkout $old_branch
+./report_pr_changes --branch old --computation changed_affected_plugin
+./report_pr_changes --branch old --computation report_changes
```
JYangQi00, line 32: Does the cleanup function automatically delete all of the temporary json files that we've made?

JYangQi00 (commented):
Thanks Dacheng, everything makes sense and looks good to me. For record-keeping, we will meet later today to discuss this PR and if we agree that this looks good and works well then I will approve.

.github/scripts/check_lineage_updates.sh
```diff
@@ -1,176 +1,32 @@
 #!/bin/bash
+set -e
 
-if [ "$#" -ne 4 ]; then
-    echo "Usage: $0 <first_branch_name> <second_branch_name> <output folder path> <run_id>"
+if [ "$#" -ne 2 ]; then
+    echo "Usage: $0 <first_branch_name> <second_branch_name>"
```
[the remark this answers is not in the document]
dachengx: Yes, currently yes.

.github/scripts/check_lineage_updates.sh
dachengx: No, I have not deleted them. Maybe should delete it by python scripts. Because they are generated by the python scripts.

PR #1300, opened 2023-11-30
Update som classifcation

WenzDaniel:
**This PR will change the lineage for > peaklets!**

In this PR we change the SOM plugin such that it only returns a single data-type. It is based on the changes made in https://github.com/XENONnT/straxen/pull/1299. Further, we propagate the SOM and straxen classification, including the SOM subclasses, up to event basics for a better analysis.

I test processed a short run 025423 with the normal offline and SOM context.


straxen/contexts.py
```diff
@@ -124,7 +124,14 @@ def xenonnt_som(cmt_version="global_ONLINE", xedocs_version=None, _from_cutax=Fa
...
     del st._plugin_class_registry["peaklet_classification"]
-    st.register(straxen.PeakletClassificationSOM)
+    st.register(
+        (
+            straxen.PeakletClassificationSOM,
+            straxen.PeaksSOM,
+            straxen.PeakBasicsSOM,
+            straxen.EventBasicsSOM,
+        )
+    )
```
JYangQi00, line 135: Just for clarification, the PeaksSOM, PeakBasicsSOM, and EventBasicsSOM is so that we do not interfere with the lineage of the non-SOM plugins?

straxen/plugins/events/event_basics_som.py
```diff
@@ -0,0 +1,31 @@
...
+        self.peak_properties += [
+            ("som_sub_type", np.int32, "SOM subtype of the peak(let)"),
+            ("straxen_type", np.int8, "Old straxen type of the peak(let)"),
+            ("loc_x_som", np.int16, "x location of the peak(let) in the SOM"),
+            ("loc_y_som", np.int16, "y location of the peak(let) in the SOM"),
+        ]
+        self.peak_properties = tuple(self.peak_properties)
+
+    def compute(self, events, peaks):
+        result = super().compute(events, peaks)
```
JYangQi00, line 30: I presume that the `super().compute()` will copy the peak info into EventBasics. But just for clarification, the reason we don't need to use `depends_on = (..., peaks_som)` or something like that is because `peaks` itself is being provided by the PeaksSOM plugin which is registered in the context?

straxen/plugins/merged_s2s/merged_s2s.py
```diff
@@ -16,7 +16,7 @@ class MergedS2s(strax.OverlapWindowPlugin):
     """Merge together peaklets if peak finding favours that they would form a single peak
     instead."""
 
-    __version__ = "1.0.2"
+    __version__ = "1.1.0"
```
JYangQi00, line 19: Presumably changes in this file are all from PR #1299?

straxen/plugins/peaklets/peaklet_classification.py
```diff
@@ -12,7 +12,7 @@
 class PeakletClassification(strax.Plugin):
     """Classify peaklets as unknown, S1, or S2."""
 
-    __version__ = "3.0.3"
+    __version__ = "3.0.4"
```
JYangQi00, line 15: I have no problems with the change here, but out of curiosity, why did we return peaklet_classification as a dict before? Also, @FaroutYLq , will this be an issue for reprocessing if the version changes but the output is functionally the same?

FaroutYLq: Thanks for the headup @JYangQi00 , very important catch! I would try to avoid touching the standard `peaklet_classification` in this SOM PR. It doesn't look to me that the performance of `peaklet_classification` is meaningfully changed, and if it is just about style we should also have a different PR for this. I would try to avoid bumping version which changes lineage so we have to reprocess again the standard pipeline.

straxen/plugins/peaklets/peaklet_classification.py
```diff
@@ -105,14 +105,11 @@ def compute(self, peaklets):
         is_s2[is_large_s1 | is_small_s1] = False
         ptype[is_s2] = 2
 
-        return dict(
```
FaroutYLq, line 108: Actually what is the motivation we want to change here?

straxen/plugins/peaks/peak_basics.py
```diff
@@ -15,40 +15,10 @@ class PeakBasics(strax.Plugin):
 
     """
 
-    __version__ = "0.1.4"
+    __version__ = "0.1.5"
```
JYangQi00, line 18: Presumably this is also just from PR #1299 ?

FaroutYLq (changes requested):
I agree with concern by @WenzDaniel and suggest we don't touch the regular `peaklet_classification` in this PR. The rest look good to me, but I have to admit I didn't run tests locally. 
Another general comments: I think for this level of complexity and significance in update we should add some tests. I would let strax* maintainers to comment more here.

straxen/plugins/peaks/peaks_som.py
```diff
@@ -0,0 +1,42 @@
...
+        merged_s2s_dtype = strax.merged_dtype((peaklet_classification_dtype, peaklets_dtype))
+        return merged_s2s_dtype
+
+    def setup(self):
+        self.copy_function_name = "_copy_requested_peak_fields_som"
+
+    def compute(self, peaklets, merged_s2s):
+        result = super().compute(peaklets, merged_s2s)
+
+        # For merged S2s SOM and straxen type are undefined:
```
JYangQi00, line 35: Perhaps this is a really basic thing that I'm missing, but why are types undefined for merged S2s?

JYangQi00 (commented):
Hi Daniel, thanks for the improvements! Everything looks fine to me I am just confused about a few things that I left in the comments. If you think they are fine then I will approve.

straxen/contexts.py
WenzDaniel: I had a few more thoughts about it during the weekend. This is the idea yes. However, this is only true for not using SOM at all. If we want to use SOM as a cut we also have to use the SOM plugins to do it properly.

straxen/plugins/events/event_basics_som.py
WenzDaniel: Yes, but the reason is PeakBasicsSOM provides a different version of `peak_basics` which EventBasics(SOM) depends on.

straxen/plugins/merged_s2s/merged_s2s.py
WenzDaniel: Yes, they are. Sorry that it is a bit messy, but I thought you guys would look first at #1299

straxen/plugins/peaklets/peaklet_classification.py
WenzDaniel: The change is coming from #1299 and indeed the version bump is purely optional. Since I do not modify the outcome of the plugin we can also merge without version bump if you guys prefer. If you agree I change #1299.

straxen/plugins/peaks/peak_basics.py
WenzDaniel: Yes, same as above purely optional.

straxen/plugins/peaklets/peaklet_classification.py
WenzDaniel: The motivation is that I use `super.compute()` in PeakletClassifcationSOM and it is easier to work with a returned structure array than a dictionary. This style change has no impact on the remaining straxen workflow.

straxen/plugins/peaks/peaks_som.py
WenzDaniel: Because SOM classifies only based on peaklets. Merged S2s is only called on S2s (`type==2`) and we assume that these merged peaks stay S2s (`type==2`). However, for each S2 there are multiple SOM sub-types and it is currently not defined which sub-type should be inherited to the merged S2s after merging. We could do something like the sub-type of the peak with the largest area, but not sure if this make sense. 
If one uses straxen types as classification for peaks then of course `type == straxen_type` and the `straxen_type` of MergedS2s would be defined again, however in that case the information is trivial to retrieve. So I preferred to not add an extra if case and make the code more complicated.

WenzDaniel:
Hmm @FaroutYLq which type of test do you have in mind? I can try to add some in #1299 because there we make actually all technical changes.

straxen/contexts.py
LuisSanchez25: The lineages of the two will always be separate since the peaklet clasification will have a different liniage as that propagates all the way down the processing pipeline, I thought the main Idea was just to propagate the SOM clusters down the chain no?

WenzDaniel: Well yes and no, when I made this PR I made different assumptions. Right now one cannot separate the two.

straxen/contexts.py
```diff
@@ -19,12 +19,15 @@
         straxen.PulseProcessing,
         straxen.Peaklets,
         straxen.PeakletClassification,
+        straxen.PeakletSOMClass,
```
WenzDaniel, line 22: These three plugins registered to the default nT context are additional plugins. They will allow us to keep the already processed data as is and provide the additional SOM fields at peak and event level as additional data-types. Thus, after processing on midway we have to use OSG to process `peak_som_classification` and `event_som_classification`.

straxen/contexts.py
```diff
@@ -124,7 +127,14 @@ def xenonnt_som(cmt_version="global_ONLINE", xedocs_version=None, _from_cutax=Fa
         cmt_version=cmt_version, xedocs_version=xedocs_version, _from_cutax=_from_cutax, **kwargs
     )
     del st._plugin_class_registry["peaklet_classification"]
-    st.register(straxen.PeakletClassificationSOM)
+    st.register(
+        (
+            straxen.PeakletClassificationSOM,
+            straxen.PeaksSOM,
+            straxen.PeakBasicsSOM,
+            straxen.EventBasicsSOM,
```
WenzDaniel, line 135: These are new plugins which extend the already existing classical versions. With this context we will have a new lineage for everything above peaklets. The same data-type names as for the classical case are used (peaklet_classification, peaks, peak_basics, event_basics), such that we can use the standard processing procedure. 

* `PeakletClassificationSOM` computes first the classical classification before adding the SOM information.
* `PeaksSOM` propagates the information to Peaks such that we can propagate it to peak basics in  `PeakBasicsSOM`. 
* `EventBasicsSOM` is the same as the classical EventBasics, but also extended by the additional SOM fields.

straxen/plugins/events/event_basics_som.py
```diff
@@ -0,0 +1,93 @@
+import strax
+import numpy as np
+
+from straxen.plugins.events.event_basics import EventBasics
+
+export, __all__ = strax.exporter()
+
+
+@export
+class EventBasicsSOM(EventBasics):
```
WenzDaniel, line 10: This plugin is inherited from the classic event-basics and only adds in addition the new SOM fields. It will also provide `event_basics` and is used in the SOM context.

straxen/plugins/events/event_basics_som.py
```diff
@@ -0,0 +1,93 @@
...
+        ]
+        self.peak_properties = tuple(self.peak_properties)
+
+    def compute(self, events, peaks):
+        result = super().compute(events, peaks)
+        return result
+
+
+@export
+class EventSOMClassification(strax.Plugin):
```
WenzDaniel, line 35: This plugin is the extra plugin for the default context case in which we provide in addition `event_som_clasification.` For peaks which do not provide any SOM information like S2s -1 is set as default value.

straxen/plugins/peaklets/peaklet_classification.py
```diff
@@ -105,14 +105,11 @@ def compute(self, peaklets):
...
-            time=peaklets["time"],
-            dt=peaklets["dt"],
-            # Channel is added so the field order of the merger of
-            # peaklet_classification and peaklets matches that
-            # of peaklets.
-            # This way S2 merging works on arrays of the same dtype.
-            channel=-1,
-            length=peaklets["length"],
-        )
+        peaklets_classification = np.zeros(len(peaklets), dtype=self.dtype)
```
WenzDaniel, line 108: Here I change the plugin return from a dict into an array. There is no impact whatsoever for our processing. Only more convenient to work with when inhering.

straxen/plugins/peaklets/peaklet_classification_som.py
```diff
@@ -27,31 +27,31 @@ class PeakletClassificationSOM(PeakletClassification):
...
-
-    parallel = True
+    __version__ = "0.2.0"
+    child_plugin = True
 
     som_files = straxen.URLConfig(
         default="resource://xedocs://som_classifiers?attr=value&version=v1&run_id=045000&fmt=npy"
     )
 
+    use_som_as_default = straxen.URLConfig(
```
WenzDaniel, line 37: Add option for the future to default back to classic classification if required.

straxen/plugins/peaklets/peaklet_classification_som.py
```diff
@@ -65,67 +65,38 @@ def setup(self):
...
-        )
-
-        is_small_s1 &= rise_time < self.upper_rise_time_aft_boundary(
-            area_fraction_top,
-            *self.s1_risetime_aft_parameters,
-            *self.s1_flatten_threshold_aft,
-        )
-
-        is_small_s1 &= peaklets["tight_coincidence"] >= self.s1_min_coincidence
+        peaklets_classifcation = super().compute(peaklets)
```
WenzDaniel, line 68: Inherit method to avoid copy & paste errors.

straxen/plugins/peaklets/peaklet_classification_som.py
```diff
@@ -65,67 +65,38 @@ def setup(self):
...
 
-        ptype[is_large_s1 | is_small_s1] = 1
-
-        is_s2 = n_channels >= self.s2_min_pmts
-        is_s2[is_large_s1 | is_small_s1] = False
-        ptype[is_s2] = 2
+        peaklet_with_som = np.zeros(len(peaklets_classifcation), dtype=self.dtype)
```
WenzDaniel, line 70: In the following are some changes required due to the inheritance and some more explicit naming conventions.

straxen/plugins/peaklets/peaklet_classification_som.py
```diff
@@ -298,3 +269,17 @@ def compute_wf_attributes(data, sample_length, n_samples: int):
         quantiles[i] = cumsum_steps[1:] - cumsum_steps[:-1]
 
     return quantiles
+
+
+@export
+class PeakletSOMClass(PeakletClassificationSOM):
```
WenzDaniel, line 275: This plugin is used in the default straxen context as the provided data-type needs to be different from `peaklet_classification`. The rest is inherited from above.

straxen/plugins/peaks/peak_basics.py
```diff
@@ -66,6 +36,45 @@ class PeakBasics(strax.Plugin):
         ),
     )
 
+    def infer_dtype(self):
```
WenzDaniel, line 39: Only moving the dtype in here so it can be better inherited.

straxen/plugins/peaks/peak_basics_som.py
```diff
@@ -0,0 +1,30 @@
+import numpy as np
+import strax
+from straxen.plugins.peaks.peak_basics import PeakBasics
+
+export, __all__ = strax.exporter()
+
+
+@export
+class PeakBasicsSOM(PeakBasics):
```
WenzDaniel, line 9: This plugin is used in the SOM context and replaces peak_basics with a new version including the SOM parameters.

straxen/plugins/peaks/peaks_som.py
```diff
@@ -0,0 +1,81 @@
+import strax
+import numpy as np
+from straxen.plugins.peaks.peaks import Peaks
+
+export, __all__ = strax.exporter()
+
+
+@export
+class PeaksSOM(Peaks):
```
WenzDaniel, line 9: This plugin is used in the SOM context to propagate the SOM parameters from `peaklet_classification` to `peaks`.

straxen/plugins/peaks/peaks_som.py
```diff
@@ -0,0 +1,81 @@
...
+            strax.endtime(result), strax.endtime(merged_s2s)
+        )
+        result["straxen_type"][_is_merged_s2] = -1
+        result["som_sub_type"][_is_merged_s2] = -1
+
+        return result
+
+
+@export
+class PeaksSOMClassification(strax.Plugin):
```
WenzDaniel, line 44: This is again a stand alone plugin to propagate the SOM settings to peaks for the default nT context using the default classification.

WenzDaniel (commented):
Hey guys I think I made all needed modifications. I also added some comments to the code for guidance. I conducted the following test to see if everything works as intended:

1. Process run 025423 up to event basics using the current master using the default offline context. Switch to this branch and see if I can still load peaklet_classifcation, merged_s2s, peak_basics and event_basics with the default nT offline context.
2. Processed for the same run the three new SOM data-types and checked if I can load them with meaningful values too.
3. I repeated the test of 1. with run `053494` using the XENONnT online context.

If you guys do not have any objections I think we can merge.

FaroutYLq (approved):

JYangQi00:
I checked that the That peaklet_classification (SOM context) has the same straxen_type as the type of peaklet_classification (nT online context), and the same som fields as peaklet_classification_som (also in nT context). I did similar things as well for the peak and event basics as well, but only looking at the S1 SOM information (as intended by this PR). Everything looks good on my end and I think we are good to merge!

PR #1302, opened 2023-12-05
Update NaN filtering in InterpolatingMap

JelleAalbers:

Fixes InterpolatingMap so we do not crash with scipy 1.11.

Scipy 1.11's kdtree no longer accepts nonfinite values (https://github.com/scipy/scipy/pull/18502). However, we sometimes use NaN to represent missing/undefined values, e.g. for drift time if an event has no S1.

Currently we pass all points to InterpolatingMap, then force the result to NaN for points that have nonfinite distance to the neighbours the kdtree finds. That's fine in our current scipy 1.10 environment, but once we update (https://github.com/XENONnT/ax_env/pull/353) there will be problems.

This PR instead checks for NaN before passing the points to InterpolatingMap to avoid crashing.

See the new unit test. 

I tried processing a small run 026195 locally with scipy 1.11.3, and it crashed here https://github.com/XENONnT/straxen/blob/933cc547dcd0931252f606d5c1692f36b367cc48/straxen/plugins/events/event_positions.py#L164 where events with NaN drift time are being passed to a map. Now it no longer crashes.

### DIFF
--- a/straxen/itp_map.py
+++ b/straxen/itp_map.py
@@ -41,19 +41,21 @@ class InterpolateAndExtrapolate:
             self.n_dim = self.values.shape[-1]
 
     def __call__(self, points):
-        distances, indices = self.kdtree.query(points, self.neighbours_to_use)
+        points = np.asarray(points)
 
+        # kdtree doesn't grok NaNs
+        # Start with all Nans, then overwrite for the finite points
         result = np.ones(len(points)) * float("nan")
         if self.array_valued:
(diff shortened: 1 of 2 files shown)


dachengx:
Thanks @JelleAalbers . Just for my education. How did you find this potential future crash?

JelleAalbers:
Thanks Dacheng -- I was processing a tiny run on my laptop, which meant all the package versions were different from the ones we usually use.

dachengx (approved):
Thanks @JelleAalbers .

PR #1303, opened 2023-12-06
change integration window

marianarajado:
The updated code changes the integration window from [78, 116] to [78, 132] ADC

This is the ADC window in which the waveform is expected to appear during an LED calibration. This change was necessary because the previous window did not contain the whole waveform (see [this note](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:giovo:the_led_window_crisis))

### DIFF
--- a/straxen/plugins/led_cal/led_calibration.py
+++ b/straxen/plugins/led_cal/led_calibration.py
@@ -34,7 +34,7 @@ class LEDCalibration(strax.Plugin):
           from the signal one.
     """
 
-    __version__ = "0.2.3"
+    __version__ = "0.2.4"
 
     depends_on = "raw_records"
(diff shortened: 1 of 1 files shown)


GiovanniVolta:
Maybe we should also [update the plugin version](https://github.com/XENONnT/straxen/blob/04a90993c589eda7acf9baaa3a918ff6e924f95d/straxen/plugins/led_cal/led_calibration.py#L37)?

cwittweg:
Can we merge this now or should we wait @dachengx @JYangQi00 ?

GiovanniVolta:
I think we can proceed since this affects only led_calibration data and nothing else. Namely, the lineage of what ever is above records won't be touched

PR #1307, opened 2023-12-20
add_ref_mon_nv_plugin

eangelino:

The code in this PR introduce a new plugin which computes parameters used for the Reflectivity Monitor for the Neutron Veto

From the raw records, the start time of each trigger window is extracted (by taking the smallest "time" of all the channels)    
Using the hitlets, the delta times wrt the beginning of the trigger are computed.
Cuts can be made in area (of all the hitlets), time (since the beginning of the trigger window) and number of contributing PMTs (in each window).
The returned values are the delta times (used for the reflectivity), PMT channels and areas.


straxen/plugins/ref_mon_nv/ref_mon_nv.py
```diff
@@ -0,0 +1,81 @@
...
+    def compute(self, raw_records_nv, hitlets_nv):
+        split_time = 5000
+
+        # find the trigger times
+        time_unique = np.unique(raw_records_nv["time"])
+        time_diff = np.diff(time_unique, prepend=time_unique[0])
+        time_unique_split = np.split(time_unique, np.where(time_diff >= split_time)[0])
+        trigger_times = np.array([i[0] for i in time_unique_split])
+
+        # split the hits in trigger windows
```
WenzDaniel, line 55: FYI: For splitting things by some container you can also use `strax.split_by_containment` or `strax.fully_contained_in`. Could be that those functions are faster if performance is an issue.

WenzDaniel (commented):
Looks fine I will make a small test process before approving and merging.

straxen/plugins/ref_mon_nv/ref_mon_nv.py
eangelino: It should be not a problem, but for future upgrades I could also test strax functions to see improvements. Thanks a lot!

WenzDaniel:
I fixed a small import error, and test processed run `057562` now it works like a charm.

WenzDaniel:
and removed some not needed imports

WenzDaniel (approved):

PR #1308, opened 2023-12-21
Move ref mon to online and add to bootstrax

WenzDaniel:
Adds ref mon plugin to bootstrax.


WenzDaniel:
@mflierm tested the PR on the ebs and they work. We can merge once test are done.

PR #1311, opened 2023-12-21
Bump to v2.1.6

dachengx:
Just bump the version.


WenzDaniel (approved):
For bumping versions I am fine merging without review. This is also how Joran and I did it.

PR #1317, opened 2024-01-14
Use `straxen.EventBasics.set_nan_defaults` to set default values

dachengx:

With the old code and `numpy==1.26.3`, we will see error like:

```
FAILED tests/plugins/test_plugins.py::PluginTest::test_event_ms_naive - ValueError: cannot convert float NaN to integer
```

This is due to https://github.com/XENONnT/straxen/blob/2c48b773a56f446d0f46bb42e664a918a5620b99/straxen/plugins/peaks/peak_per_event.py#L33 is not compatible anymore with the new numpy when there is `int` in `result`.

So this PR use `straxen.EventBasics.set_nan_defaults` to set the default value of result.

### DIFF
--- a/straxen/plugins/events/event_shadow.py
+++ b/straxen/plugins/events/event_shadow.py
@@ -1,6 +1,8 @@
 import numpy as np
 import strax
 
+import straxen
+
 export, __all__ = strax.exporter()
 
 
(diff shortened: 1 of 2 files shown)


WenzDaniel:
Thanks for the quick update.

Issue #1206, opened 2023-06-22
Update bokeh version to 3


dachengx:
`bokeh` has released their version 3 for a while https://pypi.org/project/bokeh/. It takes some time to fix the discrepancy between straxen code and the newest bokeh.

cfuselli:
Hey @dachengx @FaroutYLq,
I had a quick look at this and updating versions of `panel` and `holoviews` seems sufficient to solve the discrepancies. 

In particular:

```
bokeh==3.1.1 (previously 2.4.3)
panel==1.2.0 (previously 1.15.4)
holoviews==1.16.0 (previously 0.14.3)
```

I am running the straxen tests now and seems like they don't raise bokeh related errors anymore. 
I believe that bumping panels and holoviews will not create additional problems.. 

Let me know what you think.

Best,
Carlo

cfuselli:
Ok I failed, there is one issue with .opts and .options in holoviews that I can't fix atm. Will have a look another day maybe.

dachengx:
> Ok I failed, there is one issue with .opts and .options in holoviews that I can't fix atm. Will have a look another day maybe.

Thanks, @cfuselli . That is exactly I am concerned about. No hurry, this is not urgent.

JYangQi00:
I have a minor update on this issue. Basically, I was able to get the `event_display_interactive` function to work with plot_record_matrix=False using bokeh version 3.2.1 along with some modifications to the event display. However, I run into what seems to be some backwards compatibility issues when plot_record_matrix=True. Here is a summary:

- Using bokeh version 3.2.1, panel version 1.2.1, and holoviews version 1.17.0 (i.e. the most updated versions of each package up until 8/8/2023), I get the error:
 ```ValueError: Calling the .opts method with options broken down by options group (i.e. separate plot, style and norm groups) has been removed. Use the .options method converting to the simplified format instead or use hv.opts.apply_groups for backward compatibility.``` which is due to the internal function calls within holoviews
 - Using holoviews version 1.15.4 runs into the issue where `hv.extension('bokeh')` does not properly load the `arrow_heads` model from bokeh, because this has been changed to `arrowheads`

WenzDaniel:
[...]

PR #1318, opened 2024-01-15
Update to bokeh v3 and holoviews v1, drop py3.8 support

dachengx:

Close https://github.com/XENONnT/straxen/issues/1206

Depends on https://github.com/XENONnT/ax_env/pull/362 and https://github.com/XENONnT/ax_env/pull/363

Drop python 3.8 because bokeh>=3.2.0 stops support py3.8.

The test for Python 3.11 is experimental, not expect can pass. nestpy==2.0.0 does not support py3.11 now.

Also, minor update of SCADA interface: `if np.all((run_id, self.context)):` to `if run_id is not None and self.context is not None:` for numpy compatibility.

1. change `plot_width` to `width` and `plot_height` to `height` for `bokeh.plotting.figure`
2. change `style` to `styles` for `bokeh.models.Div`
3. change `names=list` to `name=str` for `bokeh.models.HoverTool`
4. set all `sizing_mode` to `"scale_width"` for better visualization in the jupyter notebook
5. use `hv.opts.apply_groups` to set grouped options, to be compatible with https://github.com/holoviz/holoviews/blob/62b7f1ea47461173253a123765a7d555a0c239ea/holoviews/core/accessors.py#L573

### DIFF
--- a/.github/workflows/pypi_install.yml
+++ b/.github/workflows/pypi_install.yml
@@ -15,7 +15,7 @@ jobs:
       - name: Setup python
         uses: actions/setup-python@v5.0.0
         with:
-          python-version: '3.8'
+          python-version: '3.9'
 
       - name: Checkout repo
         uses: actions/checkout@v4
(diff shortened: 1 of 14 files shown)


WenzDaniel (approved):
We will drop 3.8  python support. We will also drop nestpy from ax_env to prevent failing tests.

PR #1319, opened 2024-01-15
Add nopython

WenzDaniel:
Update decorator for future numba

Reference: https://numba.readthedocs.io/en/stable/reference/deprecation.html#deprecation-of-object-mode-fall-back-behaviour-when-using-jit


dachengx:
The error is because we have not dropped py3.8. We should wait for https://github.com/XENONnT/straxen/pull/1318.

dachengx:
I see the failing test because we drop wfsim from the requirement. https://github.com/XENONnT/straxen/actions/runs/7530878003/job/20498214152?pr=1319#step:9:289
Shall we make an issue or PR to move these tests or even the code of context to wfsim?

dachengx:
So I implemented https://github.com/XENONnT/straxen/pull/1320 and https://github.com/XENONnT/WFSim/pull/430.

dachengx (dismissed):

dachengx (approved):

PR #1327, opened 2024-02-06
Loosen `save_when` of `Events`

dachengx:

Allow users to save `events` for debugging. Related to https://github.com/XENONnT/outsource/pull/127.


FaroutYLq (approved):
Straightforward enough and I don't see why we shouldn't do it. LGTM

yuema137:
It looks good to me. I believe it can benefit our analysis. Thanks!

PR #1328, opened 2024-02-08
Fixing hitlets boundary out of chunk

WenzDaniel:

Limits hilt(lets) bounds to be within chunk. Fixes n_hits != tight_coincidence error.

### DIFF
--- a/straxen/plugins/peaklets/peaklets.py
+++ b/straxen/plugins/peaklets/peaklets.py
@@ -248,8 +248,21 @@ class Peaklets(strax.Plugin):
         hitlets = hits
         del hits
 
+        # Extend hits into hitlets and clip at chunk boundaries:
         hitlets["time"] -= (hitlets["left"] - hitlets["left_integration"]) * hitlets["dt"]
+        hitlets["time"] = np.clip(hitlets["time"], start, straxen.INFINITY_64BIT_SIGNED)
+
         hitlets["length"] = hitlets["right_integration"] - hitlets["left_integration"]
+        hitlet_endtime = strax.endtime(hitlets)
+        _hitlet_beyond_chunk = hitlet_endtime > end
+        if np.any(_hitlet_beyond_chunk):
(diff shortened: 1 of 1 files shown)


straxen/plugins/peaklets/peaklets.py
```diff
@@ -248,8 +248,21 @@ def compute(self, records, start, end):
         hitlets = hits
         del hits
 
+        # Extend hits into hitlets and clip at chunk boundaries:
         hitlets["time"] -= (hitlets["left"] - hitlets["left_integration"]) * hitlets["dt"]
+        hitlets["time"] = np.clip(hitlets["time"], start, straxen.INFINITY_64BIT_SIGNED)
```
dachengx, line 253: Thanks @WenzDaniel for the quick fix. Two points:
1. we could use https://github.com/XENONnT/straxen/blob/476bf1afd9dcf5015f6931c8dd1b27811ea6f1b9/straxen/plugins/peaklets/peaklets.py#L377 to clip the time and endtime.
2. this line https://github.com/XENONnT/straxen/blob/476bf1afd9dcf5015f6931c8dd1b27811ea6f1b9/straxen/plugins/peaklets/peaklets.py#L306 reuse the `left_integration` to recover the `hitlets["time"]` before https://github.com/XENONnT/straxen/blob/476bf1afd9dcf5015f6931c8dd1b27811ea6f1b9/straxen/plugins/peaklets/peaklets.py#L251. So maybe we can save the original `hitlets["time"]` into a variable so that we do not need to recover it, or we can clip on `left_integration`.

dachengx (changes requested):

straxen/plugins/peaklets/peaklets.py
WenzDaniel: Thanks, done.

WenzDaniel: Or not..

WenzDaniel:
Added back time shift this works now.

dachengx:
I made a PR to fix point 2: https://github.com/XENONnT/straxen/pull/1338

PR #1340, opened 2024-02-29
Fix restrax small bug in warning module

cfuselli:
AttributeError: 'ReStrax' object has no attribute 'warning'


yuema137 (approved):

cfuselli:
Was already fixed in https://github.com/XENONnT/straxen/pull/1402/files

PR #1349, opened 2024-03-21
Rotated map interpolation

l-althueser:
This PR adds functionality for pattern maps that are saved in rotated parameter space such that the input coordinates need to be translated. In addition, a new (slow) interpolation algorithm is added which only takes the inverse distant weights in 1D for each row in y-direction.
We will need further confirmation from @andrieub and @odadoun if this is exactly what is needed from their side - lets discuss offline but keep this draft PR for communication.

Thanks!


andrieub:
First test of rotated map interpolation using 1D algorithm in fuse successful, see
https://xenonnt.slack.com/archives/C07GXK1KH36/p1726237489137709
More tests needed, especially on timing, but it looks promising!
A small correction @l-althueser comment: 1D algorithm should work in x-direction, not y

cfuselli:
Hey @andrieub @l-althueser do you think we can merge this one?

cfuselli:
I tested this again now with the new map and `Linear1Din2D`. Seems to work all good.

sebvetter (approved):

Issue #1354, opened 2024-04-06
Corner case of `test_inverse_pdf` failure


dachengx:
**Describe the bug**

In the ordinary pytest of https://github.com/XENONnT/straxen/pull/1353, an error of `test_inverse_pdf` is observed: https://github.com/XENONnT/straxen/actions/runs/8576757387/job/23508277771?pr=1353.

**To Reproduce**
Insert the MWE of how to reproduce the error

1. add `@reproduce_failure('6.98.15', b'AXicY2CAAUYgPqABYrGugwsyMAEAGbMBnw==')` as the decorator of `test_inverse_pdf`.
2. run `pytest -xv ./tests -k test_inverse_pdf` in `straxen` folder

like:

```
@settings(deadline=None)
@given(strategies.floats(0.0, 1.0), strategies.floats(0.01, 0.99), strategies.floats(2.0, 1000))
@reproduce_failure('6.98.15', b'AXicY2CAAUYgPqABYrGugwsyMAEAGbMBnw==')
def test_inverse_pdf(aftobs, aft, s1tot):
    s1top = aftobs * s1tot
    ...
```

error observed:
```
FAILED tests/test_patternfit.py::test_inverse_pdf - AssertionError:
```

In that corner case: `aftobs` is 1.0, `aft` is 0.18243408203125003, `s1tot` is 2.0.

**Expected behavior**
No errors.

**Screenshots**

**Versions**
Please add the output of:
```
straxen.print_versions()
```

PR #1355, opened 2024-04-06
Return the edge closer to the target in `_numeric_derivative`

dachengx:

Try to fix https://github.com/XENONnT/straxen/issues/1354

The change is minor so I will not update the version of plugins.

Like the MWE in https://github.com/XENONnT/straxen/issues/1354 listed, when
```
aftobs, aft, s1tot = 1.0, 0.18243408203125003, 2.0
```

The `j_min` and `j_max` are too close (`j_min` is 0, and `j_max` is 2.8076670357142592e-05) and the difference between their corresponding `y0` and `y1` is smaller than `err=1e-7`. Then before this PR, this line https://github.com/XENONnT/straxen/blob/92bcf54cc236d3de1c86544fa0fd720cb8daf088/straxen/plugins/events/event_pattern_fit.py#L572 will directly return `j_max`. But in principle, it should return the `x` which will give the `y` closer to the `target`. And in the MWE of https://github.com/XENONnT/straxen/issues/1354, it should return `j_min`.

image
The vertical lines are `j_min` and `j_max` which are very close.

### _Notes on testing_

### DIFF
--- a/straxen/plugins/events/event_pattern_fit.py
+++ b/straxen/plugins/events/event_pattern_fit.py
@@ -569,7 +569,11 @@ def _numeric_derivative(y0, y1, err, target, x_min, x_max, x0, x1):
     """Get close to <target> by doing a numeric derivative."""
     if abs(y1 - y0) < err:
         # break by passing dx == 0
-        return 0.0, x1, x1
+        if abs(y0 - target) < abs(y1 - target):
+            x = x0
+        else:
+            x = x1
+        return 0.0, x, x
 
     x = (target - y0) / (y1 - y0) * (x1 - x0) + x0
     x = min(x, x_max)


yuema137 (approved):
Thanks @dachengx. I looked into the code and found that the problem is caused by this edge case (see attached drawing). It happens when there is actually no root for pmf(j) =pmf(k). In the old version, the program should return `x0` which is 0 and will be converted to `np.nan` by the following functions, but it returned non-zero`x1` which made the following functions think there is a root. 
While the current modification can solve this problem, I think it's better to also highlight the case when there is actually no root:
```
if abs(y1 - y0) < err:
     if abs(target-y0) > err * 2 (any safeguard will be good):
           return 0.0, 0.0, 0.0
```

image
It doesn't really change the performance considering there is only one edge case when we cannot find the root. So feel free to add the above line or not.

PR #1371, opened 2024-04-24
Let xedocs to handle avg seg and seg partitioning

GiovanniVolta:

Changed how to handle the TPC partition for AB and CD SEG and avg SEG.

Relevant for [xedocs PR#120](https://github.com/XENONnT/xedocs/pull/120) and [correction PR#273](https://github.com/XENONnT/corrections/pull/273)

The partition AB is distinguished from CD based on a linear and a radial cut on the (xy) plane. For the two partitions, we have different average SE gains.

We have add two schemas, one for the definition of AB/CD and one for the avg_se_gain. Here they are implemeneted.

### DIFF
--- a/straxen/plugins/events/corrected_areas.py
+++ b/straxen/plugins/events/corrected_areas.py
@@ -85,12 +85,11 @@ class CorrectedAreas(strax.Plugin):
     # Single electron gain partition
     # AB and CD partitons distiguished based on
     # linear and circular regions
+    # SR0 values set as default
     # https://xe1t-wiki.lngs.infn.it/doku.php?id=jlong:sr0_2_region_se_correction
     # https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:noahhood:corrections:se_gain_ee_final
     single_electron_gain_partition = straxen.URLConfig(
-        default="objects-to-dict://xedocs://avg_se_gains?run_id=plugin.run_id&version=v2&"
-        "field=region_circular&field=region_linear&"
-        "as_list=True&sort=field&key_attr=field&value_attr=value",
+        default={"linear": 28, "circular": 60},
         help=(
             "Two distinct patterns of evolution of single electron corrections between AB and CD. "
             "Distinguish thanks to linear and circular regions"
@@ -150,9 +149,9 @@ class CorrectedAreas(strax.Plugin):
 
     def ab_region(self, x, y):
         new_x, new_y = rotate_perp_wires(x, y)
-        cond = new_x < self.single_electron_gain_partition["region_circular"]
-        cond &= new_x > -self.single_electron_gain_partition["region_circular"]
-        cond &= new_x**2 + new_y**2 < self.single_electron_gain_partition["region_circular"] ** 2
+        cond = new_x < self.single_electron_gain_partition["linear"]
+        cond &= new_x > -self.single_electron_gain_partition["linear"]
+        cond &= new_x**2 + new_y**2 < self.single_electron_gain_partition["circular"] ** 2
         return cond
 
     def cd_region(self, x, y):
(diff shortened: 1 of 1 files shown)


GiovanniVolta:
@dachengx, do you have a suggestion on verifying that nothing changed for SR0? From the test performed in correction [PR#274](https://github.com/XENONnT/corrections/pull/274) the SR0 values of avg SEG and SEg partitioning are good!

yuema137:
@GiovanniVolta Thanks for the PR! It looks fine for me as the dates and values are correct. But I'm wondering should we also replace the avg SEG? For now only the partitions are replaced.

yuema137 (commented):
Another comment: I understand that the tests may fail because of not updated xedocs and other relevant repos. Could you provide the dependencies between them and point out which PRs we should use to test together? Thanks a lot

GiovanniVolta:
For the following test you have to be on correction [seg_partition](https://github.com/XENONnT/corrections/tree/seg_partition) branch and on xedocs [se_gain_partition_schema](https://github.com/XENONnT/xedocs/tree/se_gain_partition_schema) branch

immagine
immagine
immagine
immagine

Note that for these test I have specified where to fetch the data. So basically, this is what the new global version will look like.

I have also tested the online context, and it works well for the partitioning definition. For AB and CD values, we don't have to worry since it is handled by the plugin itself [here](https://github.com/XENONnT/straxen/blob/194a44e32aa2e6a0638ef6463e6f6c48afb66201/straxen/plugins/events/corrected_areas.py#L181) (this should reply to your first question @yuema137)

immagine

GiovanniVolta:
> Another comment: I understand that the tests may fail because of not updated xedocs and other relevant repos. Could you provide the dependencies between them and point out which PRs we should use to test together? Thanks a lot

Thank you, @yuema137, for the comment. So these PR should be tested with correction [seg_partition](https://github.com/XENONnT/corrections/tree/seg_partition) branch and on xedocs [se_gain_partition_schema](https://github.com/XENONnT/xedocs/tree/se_gain_partition_schema)

LuisSanchez25:
I thought we discussed yesterday that github does not have access to our mongo database so the URLs could not be set to xedocs?

LuisSanchez25 (approved):

yuema137:
@GiovanniVolta Thanks for the explanation. So, in my understanding, the `single_electron_gain_partition' config already contains the SEG values. 
https://github.com/XENONnT/straxen/blob/194a44e32aa2e6a0638ef6463e6f6c48afb66201/straxen/plugins/events/corrected_areas.py#L90
In that case, to avoid confusion, should we remove those two configs? It seems that they are set by CMT

https://github.com/XENONnT/straxen/blob/194a44e32aa2e6a0638ef6463e6f6c48afb66201/straxen/plugins/events/corrected_areas.py#L58

https://github.com/XENONnT/straxen/blob/194a44e32aa2e6a0638ef6463e6f6c48afb66201/straxen/plugins/events/corrected_areas.py#L67

The current practice looks like a mixture of CMT and xedocs which is not ideal. So could you look into this? Thanks a lot

GiovanniVolta:
Hey @yuema137,
No, no, the `single_electron_gain_partition` now only takes care of linear and circular selection for defining the AB and CD partition. Similarly, the avg_se_gain and se_gain take care only of their respective variables. There is no nested indexing anymore. Sorry for the confusion. Please have a look at the correction [PR#273](https://github.com/XENONnT/corrections/pull/273) to further understand the two new schemas.
I have also updated the PR description. My bad; I should have done this before.

GiovanniVolta:
@flammhead asked me if these changes and the release of a new version of xedocs will affect the data lineage. My naive answer is only above the event level. The Strax/staxen version has not been changed, and the lineage does not care about the xedocs version.
@yuema137 , @dachengx is this correct?

GiovanniVolta:
The relevant PRs in xedocs and correction have been merged! We are good to go imo

yuema137 (approved):
Thanks @GiovanniVolta ! Now it's clear to me and I think we can go ahead to merge.

Issue #1362, opened 2024-04-13
Remove `resource_cache` on dali


dachengx:
At https://github.com/XENONnT/straxen/blob/e7d95878b091ebe0a6d80732b256ba677c7a6ba1/straxen/common.py#L264 and https://github.com/XENONnT/straxen/blob/e7d95878b091ebe0a6d80732b256ba677c7a6ba1/straxen/storage/mongo_storage.py#L264 we stored `resource_cache` files on `/dali`. But sometimes `/dali` is not accessible because `/dali` and `/project(2)` are decoupled. I am wondering if we can remove the `resource_cache` on dali.

yuema137:
I think letting users store the resource_cache under their local folders doesn't harm, as the size of those caches is not terrible. So, should we just remove the dali path from the list?

PR #1372, opened 2024-04-25
remove resource_cache from dali

yuema137:
See issue #1362 
The `resource_cache` on dali is just a backup folder to store the caches from url/mongodb. Most of the time, they are stored in users' local folders unless there is a permission issue. To avoid possible mounting issues between Dali and other servers, we removed this backup location from Straxen.


dachengx (approved):
Thanks @yuema137

PR #1382, opened 2024-05-16
Add plugin to get data of multiple peaks at event level

HenningSE:

This PR adds a plugin that propagates multiple peak properties to event level. 


HenningSE:
Need to wait for https://github.com/AxFoundation/strax/pull/867 and https://github.com/XENONnT/straxen/pull/1406

yuema137:
Hi, @HenningSE, thanks for this high-level PR. On top of the small comments that I made for the input parameters and docstring, I have a more general one - if we really need such a plugin here?

 This plugin reads the relative time, position, and AFT of every peak within events. For a plugin named `multiple_peak_data`, other users may expect to get more information about those peaks but not just a few selected features. So I would suggest moving it to the `analyses` module if it's something needed for certain analyses. Basically, we just need to put a function to read from `event` data.

WenzDaniel:
Maybe @HenningSE can pick this up again, or let me know if we can close.

PR #1387, opened 2024-06-07
Allow local blinding files

WenzDaniel:
Allows to use blinding files from a local repository to be used for testing.


dachengx (approved):
Thanks @WenzDaniel

PR #1397, opened 2024-07-02
Initialize `MongoDownloader` only once to limit unnecessary connection to DB

dachengx:
_Before you submit this PR: make sure to put all operations-related information in a wiki-note, a PR should be about code and is publicly accessible_

## What does the code in this PR do / what does it improve?

Started from [slack thread](https://xenonnt.slack.com/archives/C0172947DFW/p1717974709915159?thread_ts=1717597229.473489&cid=C0172947DFW), we need to find a way to limit connections to DB.

This line controls the connection to DB: https://github.com/XENONnT/straxen/blob/b8fe520deca55b8c5ac0cb53249b38639f0ae5de/straxen/common.py#L225. Since in fuse we used a lot of resource protocol to download maps: https://github.com/XENONnT/straxen/blob/b8fe520deca55b8c5ac0cb53249b38639f0ae5de/straxen/config/protocols.py#L47, each calling of resource protocol will trigger a connection.

This PR checks if the cached file exist BEFORE build the connection to DB: caching `downloader.list_files()` in line https://github.com/XENONnT/straxen/blob/b8fe520deca55b8c5ac0cb53249b38639f0ae5de/straxen/common.py#L226C17-L226C40 for future reference so that you will not need to initialize `straxen.MongoDownloader` each time you call resource protocol.

## Can you briefly describe how it works?

Put `mongo_downloader = mongo_storage.MongoDownloader()` and `mongo_downloader_files = mongo_downloader.list_files()` in `straxen/storage/__init__.py` and replace `downloader = straxen.MongoDownloader()` with `mongo_downloader`.

## Can you give a minimal working example (or illustrate with a figure)?

_Please include the following if applicable:_
  - [ ] _Update the docstring(s)_
  - [ ] _Update the documentation_
  - [ ] _Tests to check the (new) code is working as desired._
  - [ ] _Does it solve one of the open issues on github?_

### _Notes on testing_
 - _Until the automated tests pass, please mark the PR as a draft._
 - _On the XENONnT fork we test with database access, on private forks there is no database access for security considerations._

All _italic_ comments can be removed from this template.

PR #1398, opened 2024-07-02
Generate only one instance for `MongoDownloader`

dachengx:

Similar to https://github.com/XENONnT/straxen/pull/1397 but in another approach. When trying `downloader = straxen.MongoDownloader()`, unless `*args` or `**kwargs` changed for  `MongoDownloader`, return the already existed instance of `MongoDownloader`.

### _Notes on testing_

### DIFF
--- a/straxen/storage/mongo_storage.py
+++ b/straxen/storage/mongo_storage.py
@@ -3,6 +3,7 @@ import tempfile
 from datetime import datetime
 from warnings import warn
 import pytz
+from typing import Tuple, Dict
 from strax import exporter, to_str_tuple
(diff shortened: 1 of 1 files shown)


SaelMo (approved):
DachengReviewwDachengBranch10-07
DachengReviewwoDachengBranch10-07
Here is a before and after (with and without the patch) of running the example_runconfig in OSG with my resource_cache (full coverage). While it is difficult to properly quantify, Dacheng's patch seems to reduce the number of redundant connections being made, and I can't find a difference in the fuse.out files from the two tests as expected.

PR #1400, opened 2024-07-10
fix peak per event plugin

RoiFrankel:

Fix the Bug of the mismatch between peaks['event_number'] and event['event_number'] given by peak_per_event plugin.
More details on the bug and his solution are at the following wiki note:
https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:analysis:analysts_overview_page:roi_frankel:Event_number_bug

 We start by sorting the indices of fully contained peaks. Then, we create a dictionary called 'mapping',
 where keys represent event numbers in peak level data, and values represent event numbers in event level data.

The 'mapping' dictionary ensures that each event number at the peak level corresponds to the correct event
 number at the event level.
 We then use 'corrected_split_peaks_ind' to update the event numbers of
 peaks based on the 'mapping' dictionary. 
This process ensures that event numbers match between peaks and events.
 The 'result['event_number']' array is updated with 'corrected_split_peaks_ind', ensuring that event numbers
 in peaks are consistent with those in events, maintaining the original length and preserving -1 values for peaks
 where event numbers are not applicable.
    
An example could be found at the wiki note : 
https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:analysis:analysts_overview_page:roi_frankel:Event_number_bug

Test with detailed example for bug and solution could be found at this  notebook:
https://github.com/XENONnT/analysiscode/blob/master/peak_per_event_corrected/new_peak_per_event_plugin.ipynb


HenningSE:
Hi @RoiFrankel thanks for the PR. Right now you have a long explanation of the proposed changes as comments in the code. It is of course nice to explain changes, but I would propose to move these comments from the code to the description of the PR otherwise the code will get difficult to track as more and more comments would pile up at some point.

RoiFrankel:
> Hi @RoiFrankel thanks for the PR. Right now you have a long explanation of the proposed changes as comments in the code. It is of course nice to explain changes, but I would propose to move these comments from the code to the description of the PR otherwise the code will get difficult to track as more and more comments would pile up at some point.

yuema137 (changes requested):
Hi @RoiFrankel good catch and nice implementation! I went through your note and module in analysis code, which are quite clear. As you already have those checks in hand, can I ask you for another favor to add an unit test for this plugin, i.e. create a `peak_per_event.py` under `straxen/tests/plugins`? 
Don't worry if you don't want to do it as it's not necessary.
Thanks a lot :)

The commit by me is just to remove the redundant `#` in the comment to resolve the error from pre-commit checking. Nothing changed functionally.

yuema137 (approved):

RoiFrankel:
Hi @yuema137  thanks for your respond ! I will add the unit test for the plugin :)

yuema137:
@RoiFrankel How's it going there? Don't worry if you don't have time to add the test. We can merge this PR without it.

yuema137:
@dachengx we can go ahead to merge this PR. I talked to Roi offline and we decided to add the test in the future.

PR #1401, opened 2024-07-30
Dynamic led window

tflehmke:
This PR implements a dnyamic window  for the LED pulse integration in the `afterpulse_processing` and `led_calibration` plugins.

It uses the straxen hit finder to find the first hit inside a fixed, wider window in the record. In the afterpulse plugin, all hits around this hit inside the `LED_window_width` are then integrated to give the full LED hit. In led calibration plugin, the samples around the first hit given by `led_hit_extension` are integrated and the maximum is taken from this area.


straxen/plugins/led_cal/led_calibration.py
```diff
@@ -122,48 +178,130 @@ def get_records(raw_records, baseline_window):
...
+
+        left = hit["left"] + led_hit_extension[0]
+        # Limit the position of the window so it stays inside the record.
+        if left < default_windows[0, 0]:
+            left = default_windows[0, 0]
+        elif left > 150 - (led_hit_extension[1] - led_hit_extension[0]):
+            left = 150 - (led_hit_extension[1] - led_hit_extension[0])
+
+        right = hit["left"] + led_hit_extension[1]
+        if right < default_windows[0, 1]:
```
yuema137, line 233: I would suggest to avoid hard-coded numbers in the plugin. Could you name it and add it as a URLConfig?

straxen/plugins/led_cal/led_calibration.py
```diff
@@ -122,48 +178,130 @@ def get_records(raw_records, baseline_window):
...
+    return _get_led_windows(hits, default_windows, led_hit_extension)
+
+
+@numba.jit(nopython=True)
+def _get_led_windows(hits, default_windows, led_hit_extension):
+    windows = default_windows
+    last = -1
+    for hit in hits:
+        if hit["record_i"] == last:
+            continue  # If there are multiple hits in one record, ignore after the first
```
yuema137, line 223: My guess for these lines is that we only want to keep the first hit (by time). If that's the case, please note that the hits from `find_hits` are not sorted yet: https://github.com/AxFoundation/strax/blob/a58cd5d68e5e5e6623d2ac29122537c57d04bb8f/strax/processing/pulse_processing.py#L163-L173
and consider using something like:
`sorted_hits = sorted(hits, key=lambda hit: hit['record_i'])`

straxen/plugins/led_cal/led_calibration.py
```diff
@@ -122,48 +178,130 @@ def get_records(raw_records, baseline_window):
...
+    :param ndarray led_windows : 2d array of shape (len(records), 2) with the window to use as the
+        integration boundaries.
+    :return ndarray area: 1d array of length len(records) with the averaged integrated areas for
+        each record.
 
     """
-    left = led_window[0]
-    end_pos = [led_window[1] + 2 * i for i in range(6)]
+    area = np.zeros(len(records), dtype=_area_dtype)
+    end_pos = [2 * i for i in range(6)]
```
yuema137, line 295: Same here - probably it's better to make a URLConfig for it (something like `max_integration_window`)

straxen/plugins/led_cal/led_calibration.py
```diff
@@ -122,48 +178,130 @@ def get_records(raw_records, baseline_window):
     mask = np.where((records["record_i"] == 0) & (records["length"] == 160))[0]
```
yuema137, line 178: Probably add an URL config for this parameter instead of hard coding?

yuema137, line 178: Although it was like that in the previous version, it will be better if we could add a URLConfig for it instead of using the hard coded number

straxen/plugins/afterpulses/afterpulse_processing.py
```diff
@@ -153,6 +209,7 @@ def _find_ap(
 
     is_LED = False
     t_LED = None
+    t_LED_hit = None
```
yuema137, line 212: My understanding for those two times is:
- `t_LED_hit`: the weighted time of the first hit found, used for calculating the dynamic window -> only updated once for the first hit
- `t_LED`: the weighted time of all of the hits found in the window -> updated after reading every hit
If this is correct, could you consider adding some comments for that? Currently the names are a little bit confusing

yuema137:
Hi @tflehmke Many thanks to this very low-level PR! I went through the two updated files and think they are mostly good. The only possible functional problem that I found is that:

- In `led_calibration.py`, `hits` are calculated by the `find_hits` function in `strax`, which are not sorted yet (in the afterpulse plugin they are sorted).  If we want to only keep the first hit by time, we need to add the sorting here.

I also have some other minor suggestions for naming/comments/changing hard-coded numbers to URLConfigs, please see the single comments for details. 

For the speed and memory I'm not really worried, as the calculation of dynamic windows is still O(N), and the size of the dynamic window is much smaller than the hits themselves. But it will still be great if you could provide me an example dataset so that I can test the performance.

straxen/plugins/led_cal/led_calibration.py
```diff
@@ -33,7 +37,7 @@ class LEDCalibration(strax.Plugin):
           from the signal one.
     """
 
-    __version__ = "0.2.4"
+    __version__ = "0.3.0"
```
GiovanniVolta, line 40: I would suggest to go to 1.0.0

straxen/plugins/led_cal/led_calibration.py
```diff
@@ -122,48 +178,130 @@ def get_records(raw_records, baseline_window):
...
+    :param default_window: Default window to use if no LED hit is found given as a tuple (start,
+        end)
+    :param led_hit_extension: The integration window around the first hit found to use. A tuple of
+        form (samples_before, samples_after) the first LED hit.
+    :param hit_min_amplitude: Minimum amplitude of the signal to be considered a hit.
+    :param hit_min_height_over_noise: Minimum height of the signal over noise to be considered a
+        hit. :return (len(records), 2) array: Integration window for each record
+
+    """
+    hits = strax.find_hits(
```
GiovanniVolta, line 208: What is the lowest hits detectable? 15 ADC? Can this bias the result?

GiovanniVolta, line 208: What is the lowest hits detectable? 15 ADC? Can this bias the result?

straxen/plugins/led_cal/led_calibration.py
tflehmke: Yes, you are right. I fixed the hard coded numbers.

straxen/plugins/led_cal/led_calibration.py
tflehmke: fixed

straxen/plugins/led_cal/led_calibration.py
tflehmke: fixed

straxen/plugins/afterpulses/afterpulse_processing.py
tflehmke: I renamed one variable and added comments explaining both.

straxen/plugins/led_cal/led_calibration.py
tflehmke: I didn't know how to test the plugin in place in straxen earlier, so I copied the function and run it outside. However doing that I did not have access to the URL config resources. I tested it again in place and it turns out with the hard cutoff of 15 ADC I see a hard cut in the amplitude spectra at 15 ADC. I removed the hard cutoff and always use the `hit_min_height_over_noise` which removes the hard cut in the spectra.

straxen/plugins/led_cal/led_calibration.py
yuema137: seems that the thrid argument of `get_records` should be `led_cal_record_length`, but a same variable is defined inside the function itself. Please check my commit to see if it's the right fix: https://github.com/XENONnT/straxen/pull/1401/commits/c08d6c00827881c9afb7704c9863c1a22d8e378e

straxen/plugins/led_cal/led_calibration.py
yuema137: Sounds good. For LED calibration only the sequence within the same channel matters. So we are good here. Thanks for the clarification!

straxen/plugins/led_cal/led_calibration.py
yuema137: @tflehmke any update here?

tflehmke: Hey, it seems I forgot to commit push that change. I had done the same thing as you did, but also renamed the `record_length` for clarity.

yuema137 (approved):
It looks all good to me now. Thank you @tflehmke !

straxen/plugins/afterpulses/afterpulse_processing.py
```diff
@@ -117,8 +130,49 @@ def compute(self, raw_records):
 
 @export
 def find_ap(
-    hits, records, LED_window_left, LED_window_right, hit_left_extension, hit_right_extension
+    hits,
+    records,
+    LED_hit_left_boundary,
+    LED_hit_right_boundary,
+    LED_window_width,
```
dachengx, line 137: These arguments should be in the snake case.

PR #1402, opened 2024-07-30
Small restrax fix - DAQ

cfuselli:
[DAQ only] restrax did not fail a run in case of problems with the data docs themselves, getting into an infinite loop of try - fail.


bin/restrax
```diff
@@ -367,16 +365,25 @@ class ReStrax(daq_core.DataBases):
...
         }
+
         if self.production:
             self.run_coll.update_one(
                 {"_id": run_doc["_id"]},
                 {"$set": update},
             )
+
+            # Perform the $inc update to increment n_tries
+            result_inc = self.run_coll.update_one(
```
mflierm, line 382: Random question but how was the n_tries incremented before? Was it even?

stevensa22 (approved):
Looks good to me at a glance! Let us see if this then fixes the issues we have been seeing.

PR #1404, opened 2024-08-02
Plugins for position reconstruction with conditional normalizing flow

juehang:

In this PR we introduce a set of new plugins to perform probabilistic position reconstruction using a conditional normalising flow model. This position reconstruction method provides uncertainties on the reconstructed position, and the best-fit point improves on the resolution of the MLP. Details can be found in this [wiki note](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:svetter:conditional_norm_flow_posrec_principle).

We introduce the use of a normalizing flow model for position reconstruction. This model provides both a best-fit position and uncertainties.

### _Notes on testing_


LuisSanchez25:
I wonder if it would make sense to integrate PeakPositionsFlow by inheriting from PeakPositionsBaseNT since it looks like there is a lot of overlap between what the two are computing and instead just having any additional datatypes you need get called in PeakPositionsFlow? Same with the events, in an ideal world we want to have as little duplicate code as possible but I am not 100% sure that this all can work like that.

Other than that, if the plugin is working and once it passes all the tests it will be fine by me

juehang:
> I wonder if it would make sense to integrate PeakPositionsFlow by inheriting from PeakPositionsBaseNT since it looks like there is a lot of overlap between what the two are computing and instead just having any additional datatypes you need get called in PeakPositionsFlow? Same with the events, in an ideal world we want to have as little duplicate code as possible but I am not 100% sure that this all can work like that.
> 
> Other than that, if the plugin is working and once it passes all the tests it will be fine by me

I found it quite difficult to inherit from `PeakPositionsBaseNT` because that one cannot currently deal with the extra fields for uncertainty, and if I inherited it in a naive way I would end up overloading almost everything.

LuisSanchez25:
What do you mean by overloading everything?

juehang:
I just meant re-defining the methods to accommodate the new data fields

straxen/plugins/events/event_s2_positions_flow.py
```diff
@@ -0,0 +1,314 @@
...
+                        )
+                        result["x_flow_fdc"][event_i] = sp["x_flow"][type_index] * scale
+                        result["y_flow_fdc"][event_i] = sp["y_flow"][type_index] * scale
+
+        return result
+
+
+@export
+class EventPositionUncertainty(strax.Plugin):
+    """Plugin to calculate position uncertainties for events.
```
yuema137, line 204: Could you also add the relevant note in the docstring here? Same suggestion applied for other plugins too

straxen/plugins/events/event_s2_positions_flow.py
```diff
@@ -0,0 +1,314 @@
...
+
+    depends_on = ("events", "event_basics", "event_info", "peak_positions_flow", "peaks")
+
+    provides = "event_position_contour"
+    compressor = "zstd"
+    data_kind = "events"
+    loop_over = "events"
+
+    # Configuration options
+    default_reconstruction_algorithm = straxen.URLConfig(
```
yuema137, line 46: To me it looks that the `default_reconstruction_algorithm` is not used/called at all in this plugin. Is it a redundant config variable or something is missing in the compute function?

yuema137 (changes requested):

yuema137:
Hi @juehang and @sebvetter thanks for this great work! The flow-based position reconstruction undoubtfully improves the interpretability of our ML efforts a lot! The PR is generally good and I put some minor suggestions in the review above. 

One thing that I'm a little bit concerned about is the speed. The flow-based model kindly gives us a position contour, unlike other algorithms which only offer deterministic outputs. This nice feature, on the other hand, usually also lead to much more computation cost. So it will be great if you could provide some kind of estimation on the speed (like compare the speed of mlp, cnn, gcn and flow model for the same dataset) so that we have a sense for it.

sebvetter:
Hi @yuema137 ,
In this note: https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:juehang:flow_posrec_proposal_sr2 you can find the attached plot
cnf_vs_current_timing_per_sample

The Flow model is faster than the CNN

straxen/plugins/peaks/peak_positions.py
```diff
@@ -22,7 +22,12 @@ class PeakPositionsNT(strax.MergeOnlyPlugin):
     """
 
     provides = "peak_positions"
-    depends_on = ("peak_positions_cnn", "peak_positions_mlp", "peak_positions_gcn")
+    depends_on = (
+        "peak_positions_cnn",
+        "peak_positions_mlp",
+        "peak_positions_gcn",
+        "peak_positions_flow",
```
dachengx, line 29: "peak_positions_flow" or "peak_positions_cnf"? :) I think the latter is better.

dachengx (approved):

PR #1406, opened 2024-08-15
Save first samples of peak(lets) waveform

HenningSE:

This PR uses the code changes in https://github.com/AxFoundation/strax/pull/867 to save the first 110 samples of not-downsampled waveforms.

### DIFF
--- a/straxen/plugins/merged_s2s/merged_s2s.py
+++ b/straxen/plugins/merged_s2s/merged_s2s.py
@@ -62,6 +62,10 @@ class MergedS2s(strax.OverlapWindowPlugin):
         default=True, type=bool, help="Digitize the sum waveform of the top array separately"
     )
 
+    save_waveform_start = straxen.URLConfig(
+        default=True, type=bool, help="Save the start time of the waveform with 10 ns dt"
+    )
+
     merged_s2s_get_window_size_factor = straxen.URLConfig(
         default=5, type=int, track=False, help="Factor of the window size for the merged_s2s plugin"
     )
@@ -109,6 +113,14 @@ class MergedS2s(strax.OverlapWindowPlugin):
             del peaklets
             peaklets = peaklets_w_field
 
+        if "data_start" not in peaklets.dtype.names:
+            peaklets_w_field = np.zeros(
+                len(peaklets), dtype=strax.peak_dtype(n_channels=self.n_tpc_pmts, digitize_top=True, save_waveform_start=True)
+            )
+            strax.copy_to_buffer(peaklets, peaklets_w_field, "_add_data_start_field")
+            del peaklets
+            peaklets = peaklets_w_field
+
         # Max gap and area should be set by the gap thresholds
         # to avoid contradictions
         start_merge_at, end_merge_at = self.get_merge_instructions(
@@ -146,8 +158,15 @@ class MergedS2s(strax.OverlapWindowPlugin):
 
         strax.compute_widths(merged_s2s)
 
-        if n_top_pmts_if_digitize_top <= 0:
-            merged_s2s = drop_data_top_field(merged_s2s, self.dtype, "_drop_top_merged_s2s")
+        #if n_top_pmts_if_digitize_top <= 0:
+        #    merged_s2s = drop_data_top_field(merged_s2s, self.dtype, "_drop_top_merged_s2s")
+        
+        merged_s2s = drop_data_top_field(merged_s2s,
+                                       self.dtype,
+                                       "_drop_top_and_start_merged_s2s"
+                                        )
+
+        
         return merged_s2s
 
     @staticmethod
--- a/straxen/plugins/peaklets/peaklets.py
+++ b/straxen/plugins/peaklets/peaklets.py
@@ -120,6 +120,10 @@ class Peaklets(strax.Plugin):
         default=True, type=bool, help="Digitize the sum waveform of the top array separately"
     )
 
+    save_waveform_start = straxen.URLConfig(
+        default=True, type=bool, help="Save the start time of the waveform with 10 ns dt"
+    )
+
     saturation_correction_on = straxen.URLConfig(
         default=True, infer_type=False, help="On off switch for saturation correction"
     )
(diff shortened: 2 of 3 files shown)


straxen/plugins/merged_s2s/merged_s2s.py
```diff
@@ -58,6 +58,20 @@ class MergedS2s(strax.OverlapWindowPlugin):
 
     n_tpc_pmts = straxen.URLConfig(type=int, help="Number of TPC PMTs")
 
+    sum_waveform_top_array = straxen.URLConfig(
```
dachengx, line 61: Why `sum_waveform_top_array` is needed?

straxen/plugins/merged_s2s/merged_s2s.py
```diff
@@ -141,8 +168,11 @@ def compute(self, peaklets, lone_hits):
 
         strax.compute_widths(merged_s2s)
 
-        if n_top_pmts_if_digitize_top <= 0:
-            merged_s2s = drop_data_top_field(merged_s2s, self.dtype, "_drop_top_merged_s2s")
+        # if n_top_pmts_if_digitize_top <= 0:
```
dachengx, line 171: Why these two lines are commented out?

straxen/plugins/peaklets/peaklets.py
```diff
@@ -348,9 +372,15 @@ def compute(self, records, start, end):
         counts = np.diff(counts, axis=1).flatten()
         peaklets["n_hits"] = counts
 
-        # Drop the data_top field
-        if n_top_pmts_if_digitize_top <= 0:
-            peaklets = drop_data_top_field(peaklets, self.dtype_for("peaklets"))
+        # Drop the data_top field and waveform start field if needed
+        peaklets = drop_data_top_field(
+            peaklets,
+            strax.peak_dtype(
```
dachengx, line 378: Why can not use `self.dtype_for("peaklets")`?

straxen/plugins/merged_s2s/merged_s2s.py
dachengx: If we set `sum_waveform_top_array` as `False`, there might be a bug. I have not tested, though.

dachengx (changes requested):
Do we also need to add lone_hits to the start of waveform like https://github.com/XENONnT/straxen/blob/f514d5ef9b48efa05d15da196b8a4a0ed5f75584/straxen/plugins/merged_s2s/merged_s2s.py#L140?

HenningSE:
> Do we also need to add lone_hits to the start of waveform like
> 
> https://github.com/XENONnT/straxen/blob/f514d5ef9b48efa05d15da196b8a4a0ed5f75584/straxen/plugins/merged_s2s/merged_s2s.py#L140
> 
> ?

Thanks @dachengx, yes we should add it! When checking the corresponding code I discovered a bug in strax that we should also fix: https://github.com/AxFoundation/strax/issues/906

straxen/plugins/merged_s2s/merged_s2s.py
HenningSE: You are right, I removed it.

straxen/plugins/merged_s2s/merged_s2s.py
HenningSE: I removed the lines.

straxen/plugins/peaklets/peaklets.py
HenningSE: Done!

HenningSE:
Needs https://github.com/AxFoundation/strax/pull/908 to merge lone hits into the start of the waveforms

straxen/plugins/merged_s2s/merged_s2s.py
```diff
@@ -137,12 +161,12 @@ def compute(self, peaklets, lone_hits):
         lh = strax.sort_by_time(lh)
 
         n_top_pmts_if_digitize_top = self.n_top_pmts if "data_top" in self.dtype.names else -1
-        strax.add_lone_hits(merged_s2s, lh, self.to_pe, n_top_channels=n_top_pmts_if_digitize_top)
+        strax.add_lone_hits(merged_s2s, lh, self.to_pe, n_top_channels=n_top_pmts_if_digitize_top, waveform_start=self.save_waveform_start)
 
         strax.compute_widths(merged_s2s)
 
-        if n_top_pmts_if_digitize_top <= 0:
```
dachengx, line 144: I think you should keep logic `if n_top_pmts_if_digitize_top <= 0:` because you can not stop people form setting `sum_waveform_top_array` URLConfig to `False`.

dachengx (changes requested):

yuema137:
Thanks @HenningSE ! On top of @dachengx's suggestions, I have a few comments
- The autotest would only work with https://github.com/AxFoundation/strax/pull/908 (that one looks good to me, and I already approved.), so please rerun the GitHub action after that PR is merged.
- Sorry that I may miss the context of this PR - is it widely needed by general analysis? I'm wondering if it's necessary to set the default value of `save_waveform_start` to be True

dachengx:
> Thanks @HenningSE ! On top of @dachengx's suggestions, I have a few comments
> 
> * The autotest would only work with [Add option to merge LHs into data_start AxFoundation/strax#908](https://github.com/AxFoundation/strax/pull/908) (that one looks good to me, and I already approved.), so please rerun the GitHub action after that PR is merged.
> * Sorry that I may miss the context of this PR - is it widely needed by general analysis? I'm wondering if it's necessary to set the default value of `save_waveform_start` to be True

We can not process `peaklets` twice. So we'd better same everything during that only round of processing.

straxen/plugins/merged_s2s/merged_s2s.py
HenningSE: Is this strictly needed? I think we already cover this case with the dtype of the plugins. The `drop_data_top_field` function copies data from the input to an array with the desired dtype. If the desired dtype (the one from the plugin) does not contain the top waveform it should do it automatically without the need for the if case.

dachengx: Yes, you are right in that sense. But that will make the computing longer if we do not want to drop.

dachengx: I pushed to recover to the original implementation.

dachengx (dismissed):

PR #1417, opened 2024-08-27
Debug for `EventwBayesClass` because peaks overlapping

dachengx:

Debug `EventwBayesClass`. In very rare cases the peaks' `"time"` can be the same, so simply use the matching between time https://github.com/XENONnT/straxen/blob/c61848018f2fd49504e6d3b50de1343ecfb056b0/straxen/plugins/events/event_w_bayes_class.py#L29-L30 can cause error like:
```
CRITICAL:ThreadedMailboxProcessor:Target Mailbox (event_w_bayes_class_paired) killed, exception <class 'strax.mailbox.MailboxKilled'>, message (<class 'ValueError'>, ValueError('NumPy boolean array indexing assignment cannot assign 48003 input values to the 47986 output values where the mask is true'), <traceback object at 0x7f62537a4e00>)
```

Mimicking what `EventAmbience` does, split the peaks for each event and assign the fields based on indices: https://github.com/XENONnT/straxen/blob/c61848018f2fd49504e6d3b50de1343ecfb056b0/straxen/plugins/events/event_ambience.py#L42-L57.

The results are not changed so I will not update the `__version__`.

### _Notes on testing_

### DIFF
--- a/straxen/plugins/events/event_w_bayes_class.py
+++ b/straxen/plugins/events/event_w_bayes_class.py
@@ -20,17 +20,20 @@ class EventwBayesClass(strax.Plugin):
         return dtype
 
     def compute(self, peaks, events):
-        result = np.empty(len(events), dtype=self.dtype)
+        split_peaks = strax.split_by_containment(peaks, events)
 
+        # 1. Initialization
+        result = np.empty(len(events), dtype=self.dtype)
+        # 2. Set time and endtime for events
         for name in ["s1", "s2", "alt_s1", "alt_s2"]:
             result[f"{name}_ln_prob_s1"] = np.nan
             result[f"{name}_ln_prob_s2"] = np.nan
-            # Select peaks based on their start time
-            mask = np.in1d(peaks["time"], events[f"{name}_time"])
-            mask_ev = np.in1d(events[f"{name}_time"], peaks["time"])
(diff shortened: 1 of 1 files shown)


dachengx:
However, this plugin is not widely used.

WenzDaniel:
I am wondering whether we should not remove this plugin. Is it still being maintained? I do not think so.

dachengx:
> I am wondering whether we should not remove this plugin. Is it still being maintained? I do not think so.

For the reproducibility of old results, we still need it. I think we can keep it for a while.

WenzDaniel:
Old results use old container no? Otherwise it is not reproducible anyhow?

dachengx:
> Old results use old container no? Otherwise it is not reproducible anyhow?

Yes. You are right. I mean the reproducibility of the codes in other repo which still uses this plugin.

MerzJohannes (approved):
Looks good to me. If we want to keep it for reproducibility we can merge it.

yuema137 (approved):
In the previous code, `mask` may have a larger length than `mask_ev` (and `events`) when there are duplicated 'time' in `peaks`. The PR fixes this problem by looping over split peaks together with events, and only assigning the probability when a valid peak index is found.  It looks good to me.

PR #1421, opened 2024-09-11
Update definition of the SE density

chnlkx:

Update the se_density calculated for the hotspot veto cut to use a score based on the distance and position reconstruction resolution.

### _Notes on testing_

### DIFF
--- a/straxen/plugins/peaks/peak_se_sensity.py
+++ b/straxen/plugins/peaks/peak_se_sensity.py
@@ -6,9 +6,8 @@ import straxen
 
 
 class PeakSEDensity(strax.OverlapWindowPlugin):
-    """
-    This plugin is designed to calculate a score 
-    based on the position relations to single electrons.
+    """This plugin is designed to calculate a score based on the position relations to single
+    electrons.
 
     References:
         * v0.2.0: xenon:xenonnt:ac:sr1:hotspot_veto_cut:wimp_roi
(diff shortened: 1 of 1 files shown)


chnlkx:
Hi @dachengx and @yuema137, I have a question: this plugin depends on the ''peak_position'' in v2.2.5. So how should we proceed properly?

yuema137:
@chnlkx Thanks for this PR. `peak_positions` is a merge-only plugin, and what it does is just to select the positions computed with different algorithms. So if you want to make sure that your plugin depends on the `mlp` one you can just use that directly. But in principle there should be be substantial difference when we use different position reconstruction algorithms and the result should be consistent.

dachengx:
> @chnlkx Thanks for this PR. `peak_positions` is a merge-only plugin, and what it does is just to select the positions computed with different algorithms. So if you want to make sure that your plugin depends on the `mlp` one you can just use that directly. But in principle there should not be substantial difference when we use different position reconstruction algorithms and the result should be consistent.

https://xenonnt.slack.com/archives/C016UJZ090B/p1726239673345349

yuema137:
@dachengx Thanks dacheng. This is for sure a much better practice. We should separate the developments as some low level plugins are frozen and will not be reprocessed. Should we update the readme and somehow make an annoucement?

dachengx:
> @dachengx Thanks dacheng. This is for sure a much better practice. We should separate the developments as some low level plugins are frozen and will not be reprocessed. Should we update the readme and somehow make an annoucement?

You are right. I think this decision is for people in XENONnT collaboration so modification of `README.md` is not really needed.

yuema137:
Submission to the correct branch is in #1427

PR #1427, opened 2024-09-14
Update definition of the SE density for SR1 WIMP

chnlkx:

Update the se_density calculated for the hotspot veto cut to use a score based on the distance and position reconstruction resolution.

### _Notes on testing_

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 2.2.4
+current_version = 2.2.5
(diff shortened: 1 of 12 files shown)


yuema137:
Hi @chnlkx thanks for this nice PR, but please let the PR remain as "draft" and only mark it as ready when the automatic tests are passed.

Also please make sure that only the files that you intend to update are changed. At first glance, there are some change, for example, in veto_interval, which are not relevant. Please resolve that so that we can proceed.

PR #1429, opened 2024-09-16
Fixed default window position

tflehmke:

The fixed default window leads to a time depended bias in how much untriggered PEs are seen.

See here the note: https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:dsg:pmt:gains:dynamic_led_window

This is fixed by positioning the default window on the mode of the position of all hits found in one chunk.

### _Notes on testing_

### DIFF
--- a/straxen/plugins/led_cal/led_calibration.py
+++ b/straxen/plugins/led_cal/led_calibration.py
@@ -60,7 +60,7 @@ class LEDCalibration(strax.Plugin):
         default=60,
         infer_type=False,
         help=(
-            "The minimum sample index to consider for LED hits. Hits before this sample are"
+            "The minimum sample index to consider for LED hits. Hits before this sample are "
             "ignored."
         ),
     )
@@ -75,9 +75,9 @@ class LEDCalibration(strax.Plugin):
         default=7,
         infer_type=False,
         help=(
-            "The total length of the averaging window for the area calculation."
-            "To mitigate a possiple bias from noise, the area is integrated multiple times with"
-            "sligntly different window lengths and then averaged. integration_averaging_length"
+            "The total length of the averaging window for the area calculation. "
+            "To mitigate a possiple bias from noise, the area is integrated multiple times with "
+            "sligntly different window lengths and then averaged. integration_averaging_length "
             "should be divisible by step."
         ),
     )
@@ -86,9 +86,9 @@ class LEDCalibration(strax.Plugin):
         default=1,
         infer_type=False,
         help=(
-            "The step size used for the different windows, averaged for the area calculation."
-            "To mitigate a possiple bias from noise, the area is integrated multiple times with"
-            "sligntly different window lengths and then averaged. integration_averaging_length"
+            "The step size used for the different windows, averaged for the area calculation. "
+            "To mitigate a possiple bias from noise, the area is integrated multiple times with "
+            "sligntly different window lengths and then averaged. integration_averaging_length "
             "should be divisible by step."
         ),
     )
@@ -107,7 +107,7 @@ class LEDCalibration(strax.Plugin):
         default=6,
         infer_type=False,
         help=(
-            "Minimum hit amplitude in numbers of baseline_rms above baseline."
+            "Minimum hit amplitude in numbers of baseline_rms above baseline. "
             "Actual threshold used is max(hit_min_amplitude, hit_min_"
             "height_over_noise * baseline_rms)."
         ),
--- a/straxen/plugins/led_cal/led_calibration.py
+++ b/straxen/plugins/led_cal/led_calibration.py
@@ -62,7 +62,7 @@ class LEDCalibration(strax.Plugin):
         help=(
             "The minimum sample index to consider for LED hits. Hits before this sample are"
             "ignored."
-        )
+        ),
     )
 
     led_hit_extension = straxen.URLConfig(
@@ -224,8 +224,8 @@ def get_led_windows(
     led_hit_extension. If no hit is found in the record, return the default window.
 
     :param records: Array of the records to search for LED hits.
-    :param minimum_led_position: The minimum simple index of the LED hits. Hits before this sample are
-        ignored.
+    :param minimum_led_position: The minimum simple index of the LED hits. Hits before this sample
+        are ignored.
     :param led_hit_extension: The integration window around the first hit found to use. A tuple of
         form (samples_before, samples_after) the first LED hit.
     :param hit_min_amplitude: Minimum amplitude of the signal to be considered a hit.
--- a/straxen/plugins/led_cal/led_calibration.py
+++ b/straxen/plugins/led_cal/led_calibration.py
@@ -276,9 +276,7 @@ def get_led_windows(
 
 
 @numba.jit(nopython=True)
-def _get_led_windows(
-    hits, default_windows, led_hit_extension, maximum_led_position, triggered
-):
+def _get_led_windows(hits, default_windows, led_hit_extension, maximum_led_position, triggered):
     windows = default_windows
     last = -1
 
--- a/straxen/plugins/led_cal/led_calibration.py
+++ b/straxen/plugins/led_cal/led_calibration.py
@@ -70,7 +70,7 @@ class LEDCalibration(strax.Plugin):
     )
 
     area_averaging_length = straxen.URLConfig(
-        default=8,
+        default=7,
         infer_type=False,
         help=(
             "The total length of the averaging window for the area calculation."
@@ -81,7 +81,7 @@ class LEDCalibration(strax.Plugin):
     )
 
     area_averaging_step = straxen.URLConfig(
-        default=2,
+        default=1,
         infer_type=False,
         help=(
             "The step size used for the different windows, averaged for the area calculation."
--- a/straxen/plugins/led_cal/led_calibration.py
+++ b/straxen/plugins/led_cal/led_calibration.py
@@ -234,7 +234,7 @@ def get_led_windows(
 
     """
     if len(records) == 0:  # If input is empty, return empty arrays of correct shape
-        return np.empty((0, 2), dtype=np.int64), np.empty((0, 1), dtype=bool)
+        return np.empty((0, 2), dtype=np.int64), np.empty(0, dtype=bool)
 
     hits = strax.find_hits(
         records,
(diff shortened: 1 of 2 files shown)


straxen/plugins/led_cal/led_calibration.py
```diff
@@ -204,7 +212,7 @@ def get_records(raw_records, baseline_window, LED_cal_record_length):
 
 def get_led_windows(
     records,
-    default_window,
+    minimum_led_position,
     led_hit_extension,
     hit_min_height_over_noise,
```
GiovanniVolta, line 217: I would use the same name that appears in L184 (when you call the function)

GiovanniVolta (changes requested):

straxen/plugins/led_cal/led_calibration.py
```diff
@@ -214,8 +222,8 @@ def get_led_windows(
     led_hit_extension. If no hit is found in the record, return the default window.
 
     :param records: Array of the records to search for LED hits.
-    :param default_window: Default window to use if no LED hit is found given as a tuple (start,
-        end)
+    :param minimum_led_position: The minimum position of the LED hits. Hits before this sample are
```
dachengx, line 225: What is the meaning of "position" here?

dachengx (changes requested):

straxen/plugins/led_cal/led_calibration.py
tflehmke: Yes, that name also better describes what the parameter does. I forgot to change that.

straxen/plugins/led_cal/led_calibration.py
tflehmke: The sample index of the record.

straxen/plugins/led_cal/led_calibration.py
```diff
@@ -111,13 +112,16 @@ class LEDCalibration(strax.Plugin):
...
 
     dtype = [
-        ("area", np.float32, "Area averaged in integration windows"),
-        ("amplitude_led", np.float32, "Amplitude in LED window"),
-        ("amplitude_noise", np.float32, "Amplitude in off LED window"),
-        ("channel", np.int16, "Channel"),
-        ("time", np.int64, "Start time of the interval (ns since unix epoch)"),
-        ("dt", np.int16, "Time resolution in ns"),
-        ("length", np.int32, "Length of the interval in samples"),
+        (("Area averaged in integration windows", "area"), np.float32),
```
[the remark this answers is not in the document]
tflehmke: In the first format, you cannot specify a field which contains a tuple. Like I used for the window.

straxen/plugins/led_cal/led_calibration.py
dachengx: Then shall it be "index" instead? :)

tflehmke: If you think that makes it clearer, I am fine with that.

dachengx: Please change it. Thank you very much!

dachengx: Sorry for being tedious. Maybe it is better to change all "position" to "index" in `straxen/plugins/led_cal/led_calibration.py` file.

straxen/plugins/led_cal/led_calibration.py
```diff
@@ -55,21 +56,23 @@ class LEDCalibration(strax.Plugin):
         help="Window (samples) for baseline calculation.",
     )
 
-    default_led_window = straxen.URLConfig(
-        default=(78, 107),
+    minimum_led_position = straxen.URLConfig(
```
yuema137, line 59: Is there a reference for this change? It seems that we change the double-sided range to a single lower limit for the triggering time. Isn't the original structure more robust?

yuema137 (commented):
I'm in general happy with the PR as the changes are mostly renaming and adjusting the default parameters. I just have one small question about the change from the double-sided `default_led_window` to single-sided `mininum_led_window`. 

To me it seems that it makes the plugin more flexible by keeping the double-sided range and set it as `(60,None)` here, in sense of that we will only need to update the default value but not the structure of the plugin if we need double-sided window again in the future. But if for some reasons there will never be such a need, I think it's fine.

straxen/plugins/led_cal/led_calibration.py
tflehmke: Thanks for the comment, this is actually the major change: In the previous version, I used a fixed default window, given by the old setting. However that means we see a different amount of underamplified PE, which we do not trigger on, depending on where they happen to be relative to the window. Now there is no default window anymore, instead if there is no hit found in a record, I use the mode of the distribution of all hits in one chunk.
Additionally the previous was redundant. If you wanted to change the window length, you had to remember to change both the `led_hit_extension` and the `default_led_window`. Now, `led_hit_extension` is always used.

yuema137: Thanks @tflehmke. It's clear to me now

straxen/plugins/led_cal/led_calibration.py
dachengx: @tflehmke do you want to make further changes? :)

yuema137 (approved):

straxen/plugins/led_cal/led_calibration.py
```diff
@@ -55,21 +56,23 @@ class LEDCalibration(strax.Plugin):
...
+        default=60,
         infer_type=False,
-        help="Default window (samples) to integrate and get the maximum amplitude \
-            if no hit was found in the record.",
+        help=(
+            "The minimum sample index to consider for LED hits. Hits before this sample are"
```
dachengx, line 63: `"The minimum sample index to consider for LED hits. Hits before this sample are"` -> `"The minimum sample index to consider for LED hits. Hits before this sample are "`

dachengx (changes requested):

straxen/plugins/led_cal/led_calibration.py
tflehmke: fixed

straxen/plugins/led_cal/led_calibration.py
dachengx: OK, since in `LEDAfterpulseProcessing` we used "position", we do not need to change this further.

dachengx (changes requested):

straxen/plugins/led_cal/led_calibration.py
```diff
@@ -55,43 +56,45 @@ class LEDCalibration(strax.Plugin):
...
-        default=10,
+        default=7,
         infer_type=False,
         help=(
-            "The total length of the averaging window for the area calculation."
-            "To mitigate a possiple bias from noise, the area is integrated multiple times with"
-            "sligntly different window lengths and then averaged. integration_averaging_length"
+            "The total length of the averaging window for the area calculation. "
+            "To mitigate a possiple bias from noise, the area is integrated multiple times with "
+            "sligntly different window lengths and then averaged. integration_averaging_length "
```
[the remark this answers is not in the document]
dachengx: We should delete this if it is not relevant. @tflehmke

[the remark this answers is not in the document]
tflehmke: We take the integrate the area a couple of times, with slightly adding to the integration window, and then take the average. This was already done before I made my changes to the plugin. Previously it was hardcoded. I added this and the area averaging step, to determine the length of the longest window, and the step size.
I am not quite sure what exactly you want to delete.

straxen/plugins/led_cal/led_calibration.py
```diff
@@ -151,10 +157,14 @@ def compute(self, raw_records):
...
+        temp["hit_position"] = led_windows[:, 0] - self.led_hit_extension[0]
+        temp["integration_window"] = led_windows
         return temp
 
 
-def get_records(raw_records, baseline_window, LED_cal_record_length):
+def get_records(raw_records, baseline_window, led_cal_record_length):
     """Determine baseline as the average of the first baseline_samples of each pulse.
 
     Subtract the pulse float(data) from baseline.
```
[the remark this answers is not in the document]
tflehmke: This was also like this before, I think the reasoning was, that in case there have to be changes made to this internal datatype. But maybe @GiovanniVolta can comment on that.

straxen/plugins/led_cal/led_calibration.py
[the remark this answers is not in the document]
dachengx: There is no variable called `integration_averaging_length`. So I suggest changing it to `Integration averaging length` or something meaningful you like.

straxen/plugins/led_cal/led_calibration.py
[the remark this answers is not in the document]
dachengx: The fact that it was there before does not mean that it is optimal.

straxen/plugins/led_cal/led_calibration.py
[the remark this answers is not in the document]
tflehmke: Ah, I see what you mean. I messed up with the names of the variables. Should be fixed now.

straxen/plugins/led_cal/led_calibration.py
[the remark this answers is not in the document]
tflehmke: I clarified with Giovanni, it is actually only for historical reasons, because this plugin existed before the other code. I will look if we can change this easily.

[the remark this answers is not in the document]
tflehmke: Actually it is a bit more involved to change this. Due to the different handling of the baseline and the fact that the `data` field is an integer in the strax records and a float here. 
We are also here in a different, externally triggered mode. Giovanni and I think it's justifiable to keep it as is.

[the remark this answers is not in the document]
GiovanniVolta: Yes, I understand that it is not pleasant that led_calibration has his get_records, but since the DAQ setting and the purpose are different WRT to the default strax get_records (which, BTW, it is basically copied and pasted here) I think it is okay to leave it as it is

dachengx (dismissed):

GiovanniVolta (approved):

yuema137 (dismissed):

yuema137 (approved):

GiovanniVolta:
Please @yuema137 or @dachengx proceed with the merging

PR #1430, opened 2024-09-18
Update definition of the SE Score (previously the SE density) for SR1 WIMP

chnlkx:

Update the se_score calculated for the hotspot veto cut to use a score based on the distance and position reconstruction resolution.

This PR will provide input for the hotspot veto cut in https://github.com/XENONnT/cutax/pull/479

### _Notes on testing_

### DIFF
--- a/straxen/plugins/events/event_se_score.py
+++ b/straxen/plugins/events/event_se_score.py
@@ -1,19 +1,17 @@
 import numpy as np
 import strax
 
-
-class EventSEDensity(strax.Plugin):
+class EventSEScore(strax.Plugin):
     """This plugin is designed to calculate the single electron rate density for main and alt S2 in
     events.
 
     References:
         * v0.1.0: xenon:xenonnt:ac:sr1:hotspot_veto_cut:wimp_roi
-
     """
 
-    __version__ = "0.0.0"
-    depends_on = ("event_basics", "peak_se_density")
-    provides = "event_se_density"
+    __version__ = "0.1.0"
+    depends_on = ("event_basics", "peak_se_score")
+    provides = "event_se_score"
 
     def infer_dtype(self):
         dtype = []
@@ -22,8 +20,8 @@ class EventSEDensity(strax.Plugin):
                 dtype += [
                     (
                         (
-                            f"Neiboring single-electron score of {description} S{s_i} [Hz/cm^2]",
-                            f"{main_or_alt}s{s_i}_se_nearby_probability",
+                            f"Neiboring single-electron score of {description} S{s_i} ",
+                            f"{main_or_alt}s{s_i}_se_score",
                         ),
                         np.float32,
                     ),
@@ -40,7 +38,7 @@ class EventSEDensity(strax.Plugin):
             for type_ in ["s1", "alt_s1", "s2", "alt_s2"]:
                 type_index = event[f"{type_}_index"]
                 if type_index != -1:
-                    result[f"{type_}_se_nearby_probability"][event_i] = sp["se_nearby_probability"][
+                    result[f"{type_}_se_score"][event_i] = sp["se_score"][
                         type_index
                     ]
 
--- a/straxen/plugins/events/event_se_score.py
+++ b/straxen/plugins/events/event_se_score.py
@@ -1,12 +1,15 @@
 import numpy as np
 import strax
 
+
 class EventSEScore(strax.Plugin):
     """This plugin is designed to calculate the single electron rate density for main and alt S2 in
     events.
 
     References:
+        * v0.0.0: xenon:xenonnt:analysis:hot_spot_cut_summary
         * v0.1.0: xenon:xenonnt:ac:sr1:hotspot_veto_cut:wimp_roi
+
     """
 
     __version__ = "0.1.0"
@@ -38,9 +41,7 @@ class EventSEScore(strax.Plugin):
             for type_ in ["s1", "alt_s1", "s2", "alt_s2"]:
                 type_index = event[f"{type_}_index"]
                 if type_index != -1:
-                    result[f"{type_}_se_score"][event_i] = sp["se_score"][
-                        type_index
-                    ]
+                    result[f"{type_}_se_score"][event_i] = sp["se_score"][type_index]
 
         # 2. Set time and endtime for events
         result["time"] = events["time"]
--- a/straxen/plugins/peaks/peak_se_score.py
+++ b/straxen/plugins/peaks/peak_se_score.py
@@ -118,6 +118,7 @@ class PeakSEScore(strax.OverlapWindowPlugin):
                 se_position_resolution**2 + peak_position_resolution**2
             )
             d_squre = (se_in_time["x"] - peak_i["x"]) ** 2 + (se_in_time["y"] - peak_i["y"]) ** 2
+            # 2D Gaussian with zero off-diagonal covairance matirx elements
             constant = 1 / (2 * np.pi * combined_position_resolution**2)
             exponent = np.exp(-1 / 2 * (d_squre / combined_position_resolution**2))
             _se_nearby_prob = constant * exponent
--- a/straxen/plugins/peaks/peak_se_score.py
+++ b/straxen/plugins/peaks/peak_se_score.py
@@ -50,12 +50,12 @@ class PeakSEScore(strax.OverlapWindowPlugin):
         help="Area range for single electron selection.[PE]",
     )
 
-    para_a_place_holder = straxen.URLConfig(
+    para_a = straxen.URLConfig(
         default=1,
         help="Place holder for Parameter A in the position resolution function.",
     )
 
-    para_b_place_holder = straxen.URLConfig(
+    para_b = straxen.URLConfig(
         default=1,
         help="Place holder for Parameter B in the position resolution function.",
     )
@@ -65,8 +65,8 @@ class PeakSEScore(strax.OverlapWindowPlugin):
         return 2 * (self.se_time_search_window_left + self.se_time_search_window_right)
 
     def setup(self):
-        self._para_a = self.para_a_place_holder
-        self._para_b = self.para_b_place_holder
+        self._para_a = self.para_a
+        self._para_b = self.para_b
 
     def select_se(self, peaks):
         """Function which select single electrons from peaks.
--- a/straxen/plugins/peaks/peak_se_score.py
+++ b/straxen/plugins/peaks/peak_se_score.py
@@ -87,13 +87,31 @@ class PeakSEScore(strax.OverlapWindowPlugin):
 
     @staticmethod
     @njit
-    def get_se_count_and_pdf_sum(peaks, se_peaks, se_indices, para_a, para_b, eps):
+    def get_se_count_and_pdf_sum(peaks, se_peaks, se_indices, para_a, para_b):
+        """Function to calculate the single electron (SE) score for each peak.
+
+        This function calculates the probability of nearby SE events for each peak
+        based on their spatial and temporal proximity. The probability is calculated
+        using a Gaussian distribution, where the standard deviation is the combined
+        position resolution of the peak and the SE event.
+
+        Parameters:
+        :peaks (np.ndarray): Array of peaks.
+        :se_peaks (np.ndarray): Array of single electron peaks.
+        :se_indices (np.ndarray): Array of indices for SE that are
+        in temporal proximity to each peak.
+        :para_a (float): Parameter 'a' used in the calculation of position resolution.
+        :para_b (float): Parameter 'b' used in the calculation of position resolution.
+
+        :return: np.ndarray: Array of probabilities of nearby SE events for each peak.
+
+        """
         se_nearby_probability = np.zeros(len(peaks))
         for index, peak_i in enumerate(peaks):
             indices = se_indices[index]
             se_in_time = se_peaks[indices[0] : indices[1]]
-            peak_area_top = peak_i["area"] * (peak_i["area_fraction_top"] + eps)
-            se_area_top = se_in_time["area"] * (se_in_time["area_fraction_top"] + eps)
+            peak_area_top = peak_i["area"] * peak_i["area_fraction_top"]
+            se_area_top = se_in_time["area"] * se_in_time["area_fraction_top"]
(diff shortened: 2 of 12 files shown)


straxen/plugins/events/event_se_score.py
```diff
@@ -2,18 +2,18 @@
...
 
 
-class EventSEDensity(strax.Plugin):
+class EventSEScore(strax.Plugin):
     """This plugin is designed to calculate the single electron rate density for main and alt S2 in
     events.
 
     References:
-        * v0.0.0: xenon:xenonnt:analysis:hot_spot_cut_summary
+        * v0.1.0: xenon:xenonnt:ac:sr1:hotspot_veto_cut:wimp_roi
```
dachengx, line 10: Reference to v0.0.0 should also be kept.

straxen/plugins/peaks/peak_se_score.py
```diff
@@ -0,0 +1,119 @@
...
+import strax
+import straxen
+
+
+class PeakSEScore(strax.OverlapWindowPlugin):
+    """This plugin is designed to calculate a score based on the position relations to single
+    electrons.
+
+    References:
+        * v0.2.0: xenon:xenonnt:ac:sr1:hotspot_veto_cut:wimp_roi
```
dachengx, line 13: Reference to v0.1.0 should also be kept.

straxen/plugins/peaks/peak_se_score.py
```diff
@@ -0,0 +1,119 @@
...
+        type=int,
+        help="Time searching window right side extension, one max drift time [ns]",
+    )
+
+    def get_window_size(self):
+        # This method is required by the OverlapWindowPlugin class
+        return 2 * (self.se_time_search_window_left + self.se_time_search_window_right)
+
+    def setup(self):
+        self._para_a = 1
```
dachengx, line 45: You should better set `self._para_a = self.para_a` and, make `para_a` a `straxen.URLConfig`.

straxen/plugins/peaks/peak_se_score.py
```diff
@@ -0,0 +1,119 @@
...
+        self._para_b = 1
+        self._sr_phase = 1
```
dachengx, line 47: `self._sr_phase` is not needed based on your code.

straxen/plugins/peaks/peak_se_score.py
```diff
@@ -0,0 +1,119 @@
...
+    def select_se(self, peaks):
+        """Function which select single electrons from peaks.
+
+        :param peaks: peaks data contains single electrons.
+        :return: single electron peaks data
+
+        """
+        mask = peaks["type"] == 2
+        mask &= (peaks["area"] > 10) & (peaks["area"] < 80)
+        mask &= (peaks["range_50p_area"] > 80) & (peaks["range_50p_area"] < 700)
```
dachengx, line 58: Sorry, this comment should be raised when developing v0.1.0. But here `10`, `80`, and `80`, `700` should be set as `straxen.URLConfig`.

straxen/plugins/peaks/peak_se_score.py
```diff
@@ -0,0 +1,119 @@
...
+
+    @staticmethod
+    @njit
+    def get_se_count_and_pdf_sum(peaks, se_peaks, se_indices, para_a, para_b, eps):
+        se_nearby_probability = np.zeros(len(peaks))
+        for index, peak_i in enumerate(peaks):
+            indices = se_indices[index]
+            se_in_time = se_peaks[indices[0] : indices[1]]
+            peak_area_top = peak_i["area"] * (peak_i["area_fraction_top"] + eps)
+            se_area_top = se_in_time["area"] * (se_in_time["area_fraction_top"] + eps)
```
dachengx, line 69: What is the purpose of `eps` here?

dachengx (changes requested):

straxen/plugins/events/event_se_score.py
chnlkx: Ok

straxen/plugins/peaks/peak_se_score.py
chnlkx: To avoid having same name with the `para_a` in cutax, I added 'para_a_place_holder' here

straxen/plugins/peaks/peak_se_score.py
chnlkx: True, removed

straxen/plugins/peaks/peak_se_score.py
chnlkx: Agree. Fixed

straxen/plugins/peaks/peak_se_score.py
chnlkx: Some peaks have `aft == 0` which will cause numba to error. So this way to avoid divided by zero.

dachengx: Will the pos-rec of the peaks with `area_fraction_top  == 0` be `np.nan`?

chnlkx: Yes, the pos-rec gives these peaks `np.nan`. These peaks mainly `type==0`

chnlkx: Yes, the pos-rec gives these peaks `np.nan`. These peaks mainly `type==0`

dachengx: Then is it better to replace the `eps` to mask about whether pos-rec is `np.nan`?

straxen/plugins/peaks/peak_se_score.py
```diff
@@ -0,0 +1,145 @@
...
+        help="Area range for single electron selection.[PE]",
+    )
+
+    se_selection_width_roi = straxen.URLConfig(
+        default=[80, 700],
+        type=int,
+        help="Area range for single electron selection.[PE]",
+    )
+
+    para_a_place_holder = straxen.URLConfig(
```
dachengx, line 53: I think we still should need to use `para_a` rather than `para_a_place_holder` here, and overwrite them in cutax.

dachengx (changes requested):

straxen/plugins/peaks/peak_se_score.py
chnlkx: Ok! Done

straxen/plugins/peaks/peak_se_score.py
dachengx: No `eps` is not easily understood by people.

chnlkx: Ok, let's do it by adding a comment to help the readability.

dachengx: I am pretty sure you should identify the `np.nan` and use a mask to make it clear. If you need help with how the mask works, @yuema137 and I can help.

straxen/plugins/peaks/peak_se_score.py
```diff
@@ -0,0 +1,145 @@
...
+    @staticmethod
+    @njit
+    def get_se_count_and_pdf_sum(peaks, se_peaks, se_indices, para_a, para_b, eps):
+        se_nearby_probability = np.zeros(len(peaks))
+        for index, peak_i in enumerate(peaks):
+            indices = se_indices[index]
+            se_in_time = se_peaks[indices[0] : indices[1]]
+            peak_area_top = peak_i["area"] * (peak_i["area_fraction_top"] + eps)
+            se_area_top = se_in_time["area"] * (se_in_time["area_fraction_top"] + eps)
+            peak_position_resolution = para_a * 1 / np.sqrt(peak_area_top) + para_b
```
dachengx, line 96: `* 1` here is not needed.

straxen/plugins/peaks/peak_se_score.py
chnlkx: Could you be more explicit about the reason: )

straxen/plugins/peaks/peak_se_score.py
```diff
@@ -0,0 +1,146 @@
...
+            se_position_resolution = para_a * 1 / np.sqrt(se_area_top) + para_b
+            combined_position_resolution = np.sqrt(
+                se_position_resolution**2 + peak_position_resolution**2
+            )
+            d_squre = (se_in_time["x"] - peak_i["x"]) ** 2 + (se_in_time["y"] - peak_i["y"]) ** 2
+            constant = 1 / (2 * np.pi * combined_position_resolution**2)
+            exponent = np.exp(-1 / 2 * (d_squre / combined_position_resolution**2))
```
dachengx, line 103: The calculation here is not Gaussian PDF. Is this deliberately designed or a typo?

straxen/plugins/peaks/peak_se_score.py
dachengx: 1. If a peak has position `np.nan`, the `se_score` of it should be `np.nan`, but in your calculation, it is a relatively small number but not `np.nan`.
2. It is not hard to just set the score of peaks with position `np.nan` to `np.nan`.
3. Though people can read the comments, they will also wonder why the developer does not set the score directly to `np.nan`.
4. The value of `eps` depends on a specific IEEE standard which is hard to track.

straxen/plugins/peaks/peak_se_score.py
chnlkx: It is a standard 2D gaussian with \pho = 0.

straxen/plugins/peaks/peak_se_score.py
chnlkx: Well as you insist, I'll modify it accordingly to proceed.

straxen/plugins/peaks/peak_se_score.py
dachengx: Sorry for the previous naive question. Could you use [scipy.stats.multivariate_normal](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.multivariate_normal.html) to make the code concise and allow future updates like 2D Gaussian with non-zero off-diagonal covariance matrix elements?

straxen/plugins/peaks/peak_se_score.py
```diff
@@ -0,0 +1,149 @@
...
+    )
+
+    def get_window_size(self):
+        # This method is required by the OverlapWindowPlugin class
+        return 2 * (self.se_time_search_window_left + self.se_time_search_window_right)
+
+    def setup(self):
+        self._para_a = self.para_a_naive
+        self._para_b = self.para_b_naive
+
```
GiovanniVolta, line 70: Hey @chnlkx, sorry for the dummy question, but is this necessary?

straxen/plugins/peaks/peak_se_score.py
```diff
@@ -0,0 +1,149 @@
...
+        )
+        mask &= (peaks["range_50p_area"] > self.se_selection_width_roi[0]) & (
+            peaks["range_50p_area"] < self.se_selection_width_roi[1]
+        )
+        mask &= peaks["area_fraction_top"] != 0
+        return mask
+
+    @staticmethod
+    @njit
+    def get_se_count_and_pdf_sum(peaks, se_peaks, se_indices, para_a, para_b, eps):
```
GiovanniVolta, line 90: If possible, I would like to ask you to add a short description to this function :)

straxen/plugins/peaks/peak_se_score.py
```diff
@@ -0,0 +1,149 @@
...
+        # get se score
+        # eps: smallest positive float, used to prevent division by zero.
+        eps = np.finfo(float).eps
+        _se_nearby_probability = self.get_se_count_and_pdf_sum(
+            _peaks,
+            se_peaks,
+            split_result,
+            self._para_a,
+            self._para_b,
+            eps,
```
GiovanniVolta, line 128: Now that you removed the peaks with S1 aft = 0, is the eps still necessary?

GiovanniVolta (changes requested):

straxen/plugins/peaks/peak_se_score.py
```diff
@@ -0,0 +1,149 @@
...
+        return _se_nearby_probability
+
+    def compute(self, peaks):
+        # sort peaks by center_time
+        argsort = np.argsort(peaks["center_time"], kind="mergesort")
+        _peaks = np.sort(peaks, order="center_time")
+        mask_nan = np.isnan(_peaks["x"]) | np.isnan(_peaks["y"])
+        # prepare output
+        se_nearby_probability = np.zeros(len(peaks))
+        _se_nearby_probability = np.zeros(len(peaks))
```
dachengx, line 139: Minor one, if you could, `np.zeros(len(peaks))` -> `np.full(len(peaks), np.nan)`. Then you do not need `_se_nearby_probability[mask_nan] = np.nan`.

straxen/plugins/peaks/peak_se_score.py
chnlkx: Hi @GiovanniVolta, the parameters a and b change between phase_0 and phase_1, but this information could only be implemented in cutax. So we put the phase-dependent parameters in https://github.com/XENONnT/cutax/pull/479, which will override the para_a and para_b here. So here they are more like a naive placeholder.

straxen/plugins/peaks/peak_se_score.py
chnlkx: The single electrons with aft==0 are removed, but in more cases, the 'current peaks' have zero aft. Now I give up using eps but to directly set the score of these zero-aft peaks to be nan.

straxen/plugins/peaks/peak_se_score.py
chnlkx: Ok, done!

straxen/plugins/peaks/peak_se_score.py
chnlkx: The problem is [scipy.stats.multivariate_normal](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.multivariate_normal.html) doesn't work with numba, and that's way we use this form. I add a docstring to make it more readable.

dachengx (approved):
Thanks a lot! @chnlkx

GiovanniVolta (approved):

PR #1440, opened 2024-10-08
Changed save type and bootstrax

WenzDaniel:

Update bootstrax to ensure to not make software triggered data for calibrations.

### DIFF
--- a/bin/bootstrax
+++ b/bin/bootstrax
@@ -601,6 +601,11 @@ def infer_target(rd: dict) -> dict:
     nv_ref_mon = [
         "nVeto_LASER_calibration",
     ]
+    nv_calibration_modes = [
+        "nVeto_LASER_calibration_LONG_window",
+        "nVeto_LED_calibration",
+        "nVeto_LED_calibration_2",
+    ]
     mode = str(rd.get("mode"))
     detectors = list(rd.get("detectors"))  # type: ignore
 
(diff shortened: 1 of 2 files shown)


yuema137 (dismissed):
Before this change, the `target` and `post_process` for nv calibration runs were both set to `raw_records` as the default. Now they are both set to `raw_records_nv`, which is without the software coincidence trigger. This PR looks good to me.

yuema137:
The tests fail as the NV data is not available - we only downloaded one commissioning run for the test: https://github.com/XENONnT/straxen/blob/9813b300086eb1eb27c90688370a88b2bb86a968/straxen/test_utils.py#L16-L27
And currently all tests are using this run, which is hardcoded in the `/tests/plugins` module
Should we write a particular test for NV to solve the problem? @WenzDaniel

yuema137 (changes requested):
The tests fail as the NV data is not available - we only downloaded one commissioning run for the test: https://github.com/XENONnT/straxen/blob/9813b300086eb1eb27c90688370a88b2bb86a968/straxen/test_utils.py#L16-L27
And currently all tests are using this run, which is hardcoded in the `/tests/plugins` module
Should we write a particular test for NV to solve the problem? @WenzDaniel

yuema137:
@WenzDaniel any updates here?

PR #1442, opened 2024-10-10
Put the RunDB API interface and MongoDB interface together

yuema137:
As most of XENON packages (https://github.com/XENONnT/utilix/issues/136) are using the DB-related functions implemented in straxen, we propose to remove the redundant module in `utilix` (https://github.com/XENONnT/utilix/pull/138) and keep them within `straxen`. In general, `utilix` should only contain fundamental tools that are widely used while `straxen` should handle the data management and processing tools which are specifically related to xenonnt data.  However, as a historical problem, the `APIUploader` being used by `private_nt_aux_files` is currently only in `utilix`.

Therefore, here we migrate the `utilix` code to `straxen,` with improved type hint and docstrings. And we switch to the new implementation here in `private_nt_aux_file`: https://github.com/XENONnT/private_nt_aux_files/pull/387

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -2,7 +2,7 @@
 # See https://pre-commit.com/hooks.html for more hooks
 repos:
 -   repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v4.6.0
+    rev: v5.0.0
     hooks:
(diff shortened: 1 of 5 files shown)


dachengx (approved):
Thanks @yuema137 . Please move these later to utilix.

PR #1445, opened 2024-10-11
Move the whole mongo_storage module to utilix

yuema137:
https://github.com/XENONnT/utilix/pull/141


docs/source/reference/straxen.storage.rst
```diff
@@ -7,7 +7,7 @@ Submodules
 straxen.storage.mongo\_storage module
 -------------------------------------
 
-.. automodule:: straxen.storage.mongo_storage
+.. automodule:: utilix.mongo_storage
```
dachengx, line 10: If `mongo_storage.py` is removed, you do not need this section.

pyproject.toml
```diff
@@ -60,6 +60,9 @@ Jinja2 = { version = "3.1.4", optional = true }
 urllib3 = { version = "2.2.2", optional = true }
 lxml_html_clean = { version = "*", optional = true }
 
+[tool.poetry.overrides]
+utilix = "*"
```
dachengx, line 64: I do not think you will need this.

dachengx (changes requested):
Thanks @yuema137 . This is generally good.

docs/source/reference/straxen.storage.rst
dachengx: If possible, please update the docs with https://github.com/XENONnT/straxen/blob/c8b2e2800264ffeae22974b5f9190382d6c65b4f/docs/make_docs.sh

yuema137: That script modifies too many docs that are not relevant to this PR (I think that is worth another PR). So, here I just deleted this section manually.

pyproject.toml
yuema137: Right this is redundant as the overriding actually happens in `pytest.yml`

pyproject.toml
```diff
@@ -47,10 +47,10 @@ docutils = "==0.18.1"
 mistune = "==0.8.4"
 pymongo = "*"
 requests = "*"
-utilix = ">=0.5.3"
 xedocs = "*"
 sharedarray = { url = "https://xenon.isi.edu/python/SharedArray-3.2.3.tar.gz", optional = true }
 base_environment = { git = "https://github.com/XENONnT/base_environment.git", optional = true }
+utilix = {version = "*", allow-prereleases = false}
```
dachengx, line 53: What is `allow-prereleases` here?

docs/source/config_storage.rst
```diff
@@ -9,13 +9,13 @@ what is done behind the scenes.
 
 Downloading XENONnT files from the database
 -------------------------------------------
-Most generically one downloads files using the :py:class:`straxen.MongoDownloader`
+Most generically one downloads files using the :py:class:`utilix.mongo_storage.MongoDownloader`
```
dachengx, line 12: Have you deleted this? @yuema137

pyproject.toml
dachengx: Why this is needed? @yuema137

Issue #1444, opened 2024-10-10
Trim CNN and GCN from the dependency tree for SR2 


yuema137:
For SR2, we will only use CNF and MLP for the position reconstruction, considering their better performance. 
CNN and GCN should, therefore, be trimmed from the dependency tree to avoid unnecessary massive processing. However, the plugins should be kept for potential cross-checking.

dachengx:
Related [slack thread](https://xenonnt.slack.com/archives/C016UJZ090B/p1728575173324429)

yuema137:
Fixing PR: #1449

juehang:
Looks good, I assume the tests failing is due to #1452 ? Also, is the plan to push this for now and add another PR after reprocessing and after the FDC map has been constructed?

PR #1449, opened 2024-10-15
Remove redundant pos recon

yuema137:
#1444
The only point that may need to be discussed is that the `cnf` based corrections are not available yet. So, I had to remove them from `corrections_services` for now. At some point we need to add this back


straxen/analyses/posrec_comparison.py
```diff
@@ -13,7 +13,7 @@ def load_corrected_positions(
     alt_s1=False,
     alt_s2=False,
     cmt_version=None,
-    posrec_algos=("mlp", "gcn", "cnn"),
```
dachengx, line 16: Why no `"cnf"` here?

straxen/corrections_services.py
```diff
@@ -19,15 +19,11 @@
     "cnn_model",
     "gcn_model",
     "s2_xy_map_mlp",
-    "s2_xy_map_cnn",
```
dachengx, line 22: I think you still need to keep these. Because the remaining MLP and CNN plugins need them.

straxen/plugins/events/__init__.py
```diff
@@ -52,15 +52,9 @@
 from . import event_s2_positions_mlp
 from .event_s2_positions_mlp import *
 
-from . import event_s2_positions_cnn
```
dachengx, line 55: These should be kept also?

straxen/plugins/peaks/__init__.py
```diff
@@ -16,15 +16,9 @@
 from . import peak_positions
 from .peak_positions import *
 
-from . import peak_positions_cnn
```
dachengx, line 19: These should be kept also?

dachengx (changes requested):

straxen/analyses/posrec_comparison.py
yuema137: It compares the FDC map, and the `cnf` one is currently unavailable

dachengx: Please make an issue about this.

straxen/corrections_services.py
yuema137: You are right as we probably still need to calculate those plugins at some point. Including them there doesn't harm as those maps are not necessarily called during the processing. I added them back.

juehang:
Looks good, I assume the tests failing is due to #1452 ? Also, is the plan to push this for now and add another PR after reprocessing and after the FDC map has been constructed?

yuema137:
@juehang Yes see this issue: #1454
The `analyses` module is actually not super concerned, as there are no further dependencies. When the map is ready, we can add it back.

dachengx:
@yuema137 add a TODO then?

straxen/corrections_services.py
```diff
@@ -26,8 +26,6 @@
     "s1_xyz_map_cnn",
     "s1_xyz_map_gcn",
     "fdc_map_mlp",
-    "fdc_map_cnn",
```
dachengx, line 29: This removed? @yuema137

straxen/plugins/events/__init__.py
```diff
@@ -55,12 +55,12 @@
 from . import event_s2_positions_cnn
 from .event_s2_positions_cnn import *
 
-from . import event_s2_positions_cnf
```
dachengx, line 58: Why change the order of importing?

dachengx (changes requested):

dachengx (approved):

PR #1451, opened 2024-10-15
Adjust saving preference

yuema137:
The general logic of the saving strategy could be summarized as:
1. The plugins that involve heavy computation should be saved at least once (there could be duplications in subsequent plugins)
2. The plugins that could be easily calculated from parents should not be saved

The `veto_regions` and `event_pattern_fit` fall into the first categories. As they are not computed again for other child plugins, we should always save them during their own computations.


straxen/plugins/events/event_pattern_fit.py
```diff
@@ -13,6 +13,7 @@ class EventPatternFit(strax.Plugin):
     """Plugin that provides patter information for events."""
 
     depends_on = ("event_area_per_channel", "event_basics", "event_positions")
+    save_when = strax.SaveWhen.ALWAYS
```
dachengx, line 16: This line is not needed because the default `save_when` is `ALWAYS`.

dachengx: Meanwhile, you should make `event_area_per_channel` always saved, but in another plugin.

dachengx (changes requested):

Issue #1450, opened 2024-10-15
Pre-commit removed the support for venv


yuema137:
See this post: **https://stackoverflow.com/questions/79057817/invalidmanifesterror-at-key-language-expected-one-of-but-got-python-venv**
And the pre-commit config in `straxen`: https://github.com/XENONnT/straxen/blob/df0d4887fd26a2350dcbec1567bca9aedff8b361/.pre-commit-config.yaml#L23C11-L23C48

Basically, `pre-commit` doesn't support the keyword `venv` anymore, but the [docformatter](https://github.com/pycqa/docformatter) repo is still using the `venv` hook in its [.pre-commit-hooks.yaml](https://github.com/PyCQA/docformatter/blob/v1.7.5/.pre-commit-hooks.yaml). This fails all of the `pre-commit` tests. That hook is already removed in the master branch, but unfortunately not in the latest release.

yuema137:
This is a fresh issue fixed only several days ago in `docformatter`: https://github.com/PyCQA/docformatter/issues/289
I submitted another issue to request for a new release. If it could not happen very soon, we could consider switching to the master branch

PR #1453, opened 2024-10-16
Switch to master for docformatter

yuema137:
It seems that the maintainers of [docformatter](https://github.com/pycqa/docformatter) don't make new releases frequently (the last one was on July 2023...) 
So we should better use the master branch.

Close: https://github.com/XENONnT/straxen/issues/1450


dachengx (approved):
You are the hero!

PR #1457, opened 2024-10-17
Add fix integration window for the noise runs

GiovanniVolta:

The decreasing occupancy shown in [Peak finding a plugin for LED calibration processing](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:dsg:pmt:gains:dynamic_led_window#impact_on_gain_and_spe) is because the integration window of the noise run is moving too much inside the waveforms.

See here below the case for PMT 0:
integration_windows

Given that the pulser can/is causing some noise in the waveforms in coincidence with the LED (therefore it is fully caught by the LED on-runs integration window), it is better to place the integration window of the LED off-runs (aka noise runs) in proximity to where we would expect the LED.

The pulse is delayed by ~990 ns WRT the DAQ (see [here](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:dsg:pmt:calibration:pulser_upgrade#iii_timing_adjustments)). ~~Therefore the `expected_led_position` is placed at **98 ADC sample**~~.  

Although the delay between the pulser and the DAQ, we suggest putting the fixed integration window at 98 ADC. This led to some instability in the occupancy for certain calibration campaigns. Empirically, a fixed integration window defined starting from the **88 ADC** sample gives more stable results. The finding are summarized in the [note](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:dsg:pmt:gains:dynamic_led_window#impact_on_gain_and_spe).

~~Note that the check I performed was done with the fixed position at 86 ADC sample, but we have determined that the position of the integration window for the noise has an impact of 1% on occupancy and gain estimation.~~

To discriminate between the noise and the led runs, we used the comments from the run_doc. The noise runs have always (thanks to the automatic script for the LED calibration) the 'SPE_calibration_step0', 'Gain_calibration_step3'.

I check this using the LED-on run '054597' (Gain_calibration_step2) and the LED-off run '065876' (Gain_calibration_step3) with some ad-hoc print:

Screenshot from 2024-10-23 15-55-57

As you can see the noise run has the integration window defined starting from the fixed positon.

The gains' stability and dependence on the position of the fixed integration window for the nose runs are discussed [here](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:dsg:pmt:gains:dynamic_led_window).

### _Notes on testing_

### DIFF
--- a/straxen/plugins/led_cal/led_calibration.py
+++ b/straxen/plugins/led_cal/led_calibration.py
@@ -161,7 +161,6 @@ class LEDCalibration(strax.Plugin):
         """
 
         self.is_led_on = is_the_led_on(self.run_doc, self.noise_run_comments)
-        print("is_led_on: ", self.is_led_on)
 
         mask = np.where(np.in1d(raw_records["channel"], self.channel_list))[0]
         raw_records_active_channels = raw_records[mask]
@@ -332,8 +331,6 @@ def get_led_windows(
     # hard-coded ADC sample
     if (not is_led_on) or (len(hits) == 0):
         default_hit_position = fixed_position
-        print("is_led_on: ", is_led_on)
-        print("len(hits): ", len(hits))
     else:
         default_hit_position = sps.mode(hits["left"])[0]
 
@@ -342,7 +339,7 @@ def get_led_windows(
 
         if default_hit_position > maximum_led_position:
             default_hit_position = maximum_led_position
-    print("default_hit_position: ", default_hit_position)
+
     triggered = np.zeros(len(records), dtype=bool)
 
     default_windows = np.tile(default_hit_position + np.array(led_hit_extension), (len(records), 1))
--- a/straxen/plugins/led_cal/led_calibration.py
+++ b/straxen/plugins/led_cal/led_calibration.py
@@ -66,7 +66,7 @@ class LEDCalibration(strax.Plugin):
     )
 
     expected_led_position = straxen.URLConfig(
-        default=98,
+        default=88,
         infer_type=False,
         help=(
             "Expected LED position given the delayed set on the LED pulse. "
--- a/straxen/plugins/led_cal/led_calibration.py
+++ b/straxen/plugins/led_cal/led_calibration.py
@@ -38,7 +38,7 @@ class LEDCalibration(strax.Plugin):
           from the signal one.
     """
 
-    __version__ = "0.3.0"
+    __version__ = "0.3.1"
 
     depends_on = "raw_records"
     data_kind = "led_cal"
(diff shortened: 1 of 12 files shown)


FaroutYLq (commented):
Hi quick comment. Can you link a notes next to the default `expected_led_position` and `minimum_hits_requirement` in annotation?

GiovanniVolta:
This is the draft PR for fixing the issues causing the downward trend in occupancy. I need to make the plots and update the notes.

@tflehmke, please have a look and check it independently.

GiovanniVolta:
I have further checked the impact the fixed integration window for the noise on the occupancy and gain.
I have summarized [here](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:dsg:pmt:gains:dynamic_led_window#impact_on_gain_and_spe)
In a few words:

- the position of the window as a <2% impact on the estimation of the occupancy and gain
- the more stable condition (stability of occupancy and gain for different lengths of the integration window) is achieved by defining ``expected_led_position`` to 88 ADC

It is unclear why we have this behavior, and I would like @tflehmke , @sanaouahada, and @MaxiA04  to keep looking into it. Given the impact is within the 2.5% value that we use to model the time trend of the gain, I would suggest moving on.

straxen/plugins/led_cal/led_calibration.py
```diff
@@ -65,6 +65,25 @@ class LEDCalibration(strax.Plugin):
...
+        infer_type=False,
+        help=(
+            "Expected LED position given the delayed set on the LED pulse. "
+            "This is used as default when no hits or less than a certain amount "
+            "are identified."
+        ),
+    )
+
+    minimum_hits_requirement = straxen.URLConfig(
+        default=500,
```
tflehmke, line 79: For the bottom array I get around 300-500 hits per PMT and chunk. The noise run has O(1) hits. I think we should decrease the value to around 100.

tflehmke (dismissed):

dachengx (changes requested):
We should replace the `minimum_hits_requirement` by the tagging, like https://github.com/XENONnT/cutax/blob/87f4d25a2bd36cbb96fefab26c00bc3779b47396/cutax/utils.py#L190. The tag can tell us which run is a noise run.

straxen/plugins/led_cal/led_calibration.py
```diff
@@ -228,6 +283,11 @@ def get_led_windows(
     :param records: Array of the records to search for LED hits.
     :param minimum_led_position: The minimum simple index of the LED hits. Hits before this sample
         are ignored.
+    : is_led_on: Fetch from the run database. It is used for discriminate between LED on
+        and LED off runs.
+    :param fixed_position: Fixed ADC sample upon which the integration window is defined. Used
+        if no hits are identified :minimum_hits_requirement: Minimum hits requirement for the
```
dachengx, line 289: `minimum_hits_requirement` is not needed anymore

straxen/plugins/led_cal/led_calibration.py
```diff
@@ -38,14 +38,30 @@ class LEDCalibration(strax.Plugin):
...
+        default="run_doc://comments?run_id=plugin.run_id",
+        infer_type=False,
+        help=(
+            "Comments dictionary from the run metadata. "
+            "It is used for discriminate between LED on and LED "
+            "off runs."
+        ),
+    )
+
+    noise_run_comments = straxen.URLConfig(
```
dachengx, line 59: This line is a very good implementation! @GiovanniVolta

dachengx (approved):

PR #1459, opened 2024-10-19
Use `max_time` when calculating peaklets properties

dachengx:

Benefitted from https://github.com/AxFoundation/strax/pull/913, we do not need to calculate `max_time` again.

### _Notes on testing_

### DIFF
--- a/straxen/plugins/peaklets/peaklets.py
+++ b/straxen/plugins/peaklets/peaklets.py
@@ -262,7 +262,6 @@ class Peaklets(strax.Plugin):
         hitlets["length"] = hitlets["right_integration"] - hitlets["left_integration"]
 
         hitlets = strax.sort_by_time(hitlets)
-        hitlets_time = np.copy(hitlets["time"])
         self.clip_peaklet_times(hitlets, start, end)
         rlinks = strax.record_links(records)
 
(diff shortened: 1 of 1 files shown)


straxen/plugins/peaklets/peaklets.py
```diff
@@ -302,15 +299,10 @@ def compute(self, records, start, end):
         # Saturation correction using non-saturated channels
         # similar method used in pax
         # see https://github.com/XENON1T/pax/pull/712
-        # Cases when records is not writeable for unclear reason
-        # only see this when loading 1T test data
-        # more details on https://numpy.org/doc/stable/reference/generated/numpy.ndarray.flags.html
-        if not r["data"].flags.writeable:
```
dachengx, line 308: If `r["data"].flags.writeable == False`, you will see error like `ValueError: assignment destination is read-only`. So I would not worry about this PR changes the result. Let's see how the reprocessing goes.

straxen/plugins/peaklets/peaklets.py
```diff
@@ -190,9 +190,7 @@ def setup(self):
         self.channel_range = self.channel_map["tpc"]
 
     def compute(self, records, start, end):
-        r = records
```
dachengx, line 193: This is a handy renaming, but confusing in debugging.

PR #1462, opened 2024-10-28
Fix run_doc for led plugin

GiovanniVolta:

I realized that there are run_ids with more than one comment, which will cause an error as it is right now. The first comment is always from the automatic script.

I have fixed that and I have added a check for verifying that the first comments has the right format.

### _Notes on testing_

### DIFF
--- a/straxen/plugins/led_cal/led_calibration.py
+++ b/straxen/plugins/led_cal/led_calibration.py
@@ -57,11 +57,11 @@ class LEDCalibration(strax.Plugin):
     )
 
     defualt_run_comments = straxen.URLConfig(
-        default=['auto, SC user: ', 'pulserScript: ', 'userComment: Weekly LED calibration'],
+        default=["auto, SC user: ", "pulserScript: ", "userComment: Weekly LED calibration"],
         infer_type=False,
         help=("List of default comments the automatic script for runs in PMT calibration. "),
     )
-    
+
     noise_run_comments = straxen.URLConfig(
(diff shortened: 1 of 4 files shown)


dachengx (dismissed):

GiovanniVolta:
[Here](https://xenonnt.slack.com/archives/C016UJZ090B/p1730120125020709), @dachengx suggested changing the config in the test instead of adding a specific run_id for the led_calibration. Unfortunately, I don't know where to change it. If you prefer me to change the config, I need some help

GiovanniVolta:
Ah, of course, the data are not stored so the test are still failing

straxen/plugins/led_cal/led_calibration.py
```diff
@@ -56,6 +56,12 @@ class LEDCalibration(strax.Plugin):
         ),
     )
 
+    defualt_run_comments = straxen.URLConfig(
```
dachengx, line 59: Please assign its type to be `list`.

dachengx (changes requested):

straxen/plugins/led_cal/led_calibration.py
GiovanniVolta: Done

GiovanniVolta:
@dachengx @minzhong98 @MerzJohannes @yuema137 shall we proceed here?

dachengx (approved):

GiovanniVolta:
Thank you Dacheng <3

PR #1468, opened 2024-11-14
Enforce stable sorting in `np.sort` and `np.argsort`

dachengx:
Depends on https://github.com/AxFoundation/strax/pull/918.

Because no one cared about whether the sorting was stable before so I did not change the version of the affected plugins in this PR.


yuema137 (approved):
Thankd @dachengx

PR #1492, opened 2024-12-03
[DAQ] Bootstrax dynamic adapt targets for AmBe high rate

cfuselli:
We introduce this PR to allow a dynamic modification of targets processed on the DAQ during difficult times like AmBe calibrations. 

We add a specific case for which if "ambe" is in the run mode name, the targets and post-targets are updated based on some configuration that is defined in the DAQ database, separately for each of the event builders. This PR is just to allow for a dynamic change of targets. The nice game of deciding who will process what at a given time is implemented in the scripts used on the DAQ and in daqnt (See related daqnt PR https://github.com/XENONnT/daqnt/pull/110). 

This method allowed us to survive and successfully process the last two heavy rate AmBe campaigns. 

On top of that, if "ambe" is in the mode name, the data rate used to calculate how many resources to use ( cores and workers ) is set to the max of 550MB/s, such that a low amount of resources will be chosen. 

- New argument fix_resources: it is just for testing, it just ignores the n_failures in deciding how many resources to use.


cfuselli:
Goes together with https://github.com/XENONnT/daqnt/pull/110

yuema137 (approved):
This PR allows the online processing manager to read the `"bootstrax_config"` from `db_daq` so that the targets and post_process in the online processing could be changed dynamically by updating the config in `db_daq`. 

Although we currently don't have auto tests for `bootstrax`, I believe @cfuselli has tested locally and already makes sure the hard-coded patches here work well for our previous data. Considering that this PR is straight forward and doesn't have any influence on other part of straxen, I think we should merge it soon for the upcoming calibrations.

PR #1493, opened 2024-12-03
Accelerate Euclidean distance by numba

dachengx:

Nothing is changed numerically.

```
import os
from memory_profiler import memory_usage
import numpy as np
import numba
import matplotlib.pyplot as plt
import strax
import cutax


home = os.path.expanduser('~')
st = cutax.xenonnt_offline(xedocs_version='global_v16')
run_id = '050775'
st.storage = [strax.DataDirectory(os.path.join(home, f'{run_id}/input'), readonly=True)]

PeakletClassificationSOM0 = st.get_single_plugin(run_id, 'peaklet_classification')
peaklets = st.get_array(run_id, 'peaklets', chunk_number={'peaklets': 0}, progress_bar=False)

def compute():
    PeakletClassificationSOM0.compute(peaklets)

mem = memory_usage(proc=compute, interval=0.1, timestamps=True)
mem = np.array(mem)

fig, ax = plt.subplots(1, 1, figsize=(4, 3))
ax.plot(mem[:, 1] - mem[0, 1], mem[:, 0] / 1e3, color='black', lw=0.5)
ax.set_xlabel('Time [s]')
ax.set_ylabel('Memory usage [GB]')

fig.savefig('peaklet_classification.svg', transparent=True, dpi=800)
plt.close(fig)
```

The improvement is visible:
image

### _Notes on testing_

### DIFF
--- a/straxen/plugins/peaklets/peaklet_classification_som.py
+++ b/straxen/plugins/peaklets/peaklet_classification_som.py
@@ -1,7 +1,7 @@
 import numpy as np
 import numpy.lib.recfunctions as rfn
-from straxen.plugins.peaklets.peaklet_classification_vanilla import PeakletClassificationVanilla
 import numba
+from straxen.plugins.peaklets.peaklet_classification_vanilla import PeakletClassificationVanilla
 
 import strax
 import straxen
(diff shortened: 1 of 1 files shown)


LuisSanchez25:
This looks really promising! Do you know what the speed impact would be here? Is this implementation faster than cdist?

LuisSanchez25:
Looks good to me, let me just do some quick tests to make sure we get the same numbers for both methods

LuisSanchez25:
I have a quick question since I don't seem to be able to get the function to run locally since numba seems to reject using the assert function, did you not run into this error? The error goes away when using @numba.jit(nopython=True, cache=True) instead

dachengx:
> I have a quick question since I don't seem to be able to get the function to run locally since numba seems to reject using the assert function, did you not run into this error? The error goes away when using @numba.jit(nopython=True, cache=True) instead

My gut feeling is that the discrepancy is from different numba versions. But anyway, I moved the asssert out.

LuisSanchez25:
Ok then other than that I think everything looks good!

PR #1507, opened 2024-12-23
Clean unnecessary codes

dachengx:
No result will be changed after the PR.

### DIFF
--- a/straxen/plugins/peaks/peak_positions.py
+++ b/straxen/plugins/peaks/peak_positions.py
@@ -34,7 +34,7 @@ class PeakPositions(strax.MergeOnlyPlugin):
     )
 
     def infer_dtype(self):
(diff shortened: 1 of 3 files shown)


straxen/plugins/peaks/peak_positions.py
```diff
@@ -34,7 +34,7 @@ class PeakPositions(strax.MergeOnlyPlugin):
     )
 
     def infer_dtype(self):
-        dtype = strax.merged_dtype([self.deps[d].dtype_for(d) for d in self.depends_on])
+        dtype = strax.merged_dtype([self.deps[d].dtype_for(d) for d in sorted(self.depends_on)])
```
dachengx, line 37: Following https://github.com/AxFoundation/strax/blob/2ed84f9de82c2f084f9480ab25b0d4764db3dfc4/strax/plugins/merge_only_plugin.py#L33

PR #1514, opened 2025-01-08
Add new protocol and give SOM capabilities to use less deciles

LuisSanchez25:

This PR makes a new URL protocol, this is mainly used for testing purposes. Additionally it also lets us the new SOM model in which the first decile was excluded.

### _Notes on testing_


straxen/config/protocols.py
```diff
@@ -309,3 +309,11 @@ def open_jax_model(model_path: str, **kwargs):
         serialized_jax_object = file_obj.read()
     # Deserialize the JAX object and return its callable function
     return export.deserialize(serialized_jax_object).call
+
+
+@URLConfig.register("load_numpy")
```
dachengx, line 314: Why this protocol is needed? And I would call it just "numpy" if really needed.

dachengx (commented):
Thanks, @LuisSanchez25

straxen/config/protocols.py
LuisSanchez25: Ah so I was running into a small error when trying to run my local SOM map and just made this as it was a quicker solution but I just checked what the deeper problem is.
https://github.com/XENONnT/straxen/blob/69265ff85a3d7fc8d9e57b32ad02f71605f5d501/straxen/common.py#L136
This line will run into issues if the assigned object is not an array, I can do a quick change in my code and change the line a care about to an array but it might be better to just modify this code so it can handle some of the values not being arrays. If you think this makes sense I can make a PR and change the code

dachengx: Why your local SOM map is not compatible with `result_slurped = {k: v[:] for k, v in result.items()}`?

LuisSanchez25:
I believe the remianing error is a larger issue right? Not something to do with this PR

PR #1535, opened 2025-01-22
Change dark rate monitoring of the software trigger to seconds

WenzDaniel:
In this PR we change the exception for the software trigger from chunks to seconds. This means that always in the very first N seconds of a run no software trigger is applied. 

The plot below shows the number of raw records fragments per 0.1 s as function of time of a small test run. As it can be seen after 30 s the rate of fragments drops thanks to the software trigger. 

example

### DIFF
--- a/straxen/config/protocols.py
+++ b/straxen/config/protocols.py
@@ -311,3 +311,21 @@ def open_jax_model(model_path: str, **kwargs):
         serialized_jax_object = file_obj.read()
     # Deserialize the JAX object and return its callable function
     return export.deserialize(serialized_jax_object).call
+
+
+@URLConfig.register("runstart")
+def get_run_start(run_id):
+    """Protocol which returns start time of a given run as unix time in ns."""
+    import pytz
+
+    rundb = utilix.xent_collection()
+    doc = rundb.find_one(
+        {"number": int(run_id)},
+        projection={
+            "start": 1,
+        },
+    )
+    start_time = doc["start"]
+    start_time_unix = start_time.replace(tzinfo=pytz.utc).timestamp()
+    start_time_unix = np.int64(start_time_unix) * 10**9
+    return start_time_unix
--- a/straxen/plugins/raw_records_coin_nv/nveto_recorder.py
+++ b/straxen/plugins/raw_records_coin_nv/nveto_recorder.py
@@ -80,7 +80,7 @@ class nVETORecorder(strax.Plugin):
     )
 
     keep_n_seconds_for_monitoring = straxen.URLConfig(
-        default=30,
+        default=10,
         track=False,
         infer_type=False,
         help=(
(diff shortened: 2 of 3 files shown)


straxen/config/protocols.py
```diff
@@ -311,3 +311,21 @@ def open_jax_model(model_path: str, **kwargs):
...
+    rundb = utilix.xent_collection()
+    doc = rundb.find_one(
+        {"number": int(run_id)},
+        projection={
+            "start": 1,
+        },
+    )
+    start_time = doc["start"]
+    start_time_unix = start_time.replace(tzinfo=pytz.utc).timestamp()
+    start_time_unix = np.int64(start_time_unix) * 10**9
```
dachengx, line 330: `10**9` -> `straxen.units.s`

straxen/plugins/raw_records_coin_nv/nveto_recorder.py
```diff
@@ -128,6 +131,30 @@ def compute(self, raw_records_nv, start, end, chunk_i):
                 "lone_raw_record_statistics_nv": lrs,
             }
 
+        # Keep all raw data for the very first n seconds of a run.
+        # This case is tricky since it can also be just a
+        # subset of a chunk and in this case we need to make sure to also keep
+        # all fragments of a pulse beyond the first n seconds boundary.
+        # For performance check very first fragment if in applicable time range:
+        _is_first_record_at_run_start = raw_records_nv[0]["time"] < (
```
dachengx, line 139: `_is_first_record_at_run_start` -> `_need_save_for_monitoring` Maybe this is a better name?

dachengx (commented):
Thanks @WenzDaniel

straxen/plugins/raw_records_coin_nv/nveto_recorder.py
```diff
@@ -79,16 +79,23 @@ class nVETORecorder(strax.Plugin):
...
+            "Number of seconds which should be stored without applying the software trigger."
+            "Use -1 if all records should be kept without applying the software trigger."
         ),
     )
 
+    run_start = straxen.URLConfig(
+        default="runstart://plugin.run_id?",
+        track=False,
+        infer_type=False,
+        help=("Returns run start in utc unix time in ns."),
```
dachengx, line 96: Being picky: `help=("Returns run start in utc unix time in ns."),` -> `help="Returns run start in utc unix time in ns.",`

straxen/plugins/raw_records_coin_nv/nveto_recorder.py
WenzDaniel: Done

straxen/plugins/raw_records_coin_nv/nveto_recorder.py
WenzDaniel: Done

mrcselvi (approved):
I cannot comment on the coding, but I agree on the general purpose of this script. As I suggested via email (NV group), let's use 10s at the beginning of the run *without* applying the software trigger

dachengx (approved):
The errors in pytest are not related to this PR, so I would merge this.

Issue #1475, opened 2024-11-15
Remove CMT


dachengx:
https://github.com/XENONnT/straxen/blob/6a166b97501d249234d1975a5253f1aaf974af96/straxen/corrections_services.py#L1

We need to remove CMT because the ONLINE corrections will be handled by xedocs.
When removing codes, please keep the functions that are still needed by xedocs.

dachengx:
This issue will be partially solved by https://github.com/XENONnT/straxen/pull/1235.

PR #1235, opened 2023-08-11
Remove CMT URLs

jmosbacher:
- Replace all CMT URLs with xedoc URLs
- unify "cmt_version" and "xedocs_version" arguments in the context functions, replace with single "global_version" argument which loads configs from xedocs.

PR #1539, opened 2025-01-23
Remove CMT

dachengx:
Close https://github.com/XENONnT/straxen/issues/1475


dachengx:
The tests can not pass because https://github.com/XENONnT/straxen/pull/1235#issuecomment-2609448584. After the DB is synced correctly, the test should (partially) pass.

tests/test_url_config.py
```diff
@@ -117,12 +117,6 @@ def test_leading_zero_int(self):
         self.assertEqual(p.test_config, "0666")
         self.assertIsInstance(p.test_config, str)
 
-    @unittest.skipIf(not straxen.utilix_is_configured(), "No db access, cannot test CMT.")
```
GiovanniVolta, line 120: Shall we add a test for xedocs protocol like what @LuisSanchez25 did [here](https://github.com/XENONnT/straxen/pull/1124/files)?

dachengx: xedocs protocol is almost tested everywhere in straxen after https://github.com/XENONnT/straxen/pull/1235

PR #1576, opened 2025-03-06
adapt to new channel occupancy on aqmon

RoBGlaBe:
It puts the correct labels on the veto and trigger start and stop hits. The labels are not static anymore since the recent Caen V1495 firmware upgrade.
When e.g. on channel 802 there was always the "high energy veto (HEV) stop" signal, now it can be one of the following: HEV, HEV_tag, Neutron-Generator-Anti-Veto or Fractional Lifetime Veto stop signal.
Now the channels are still named after their primary usage, but the occupancy in a run is used for the naming of the interval.

The class used gets the run-doc, finds what parameters the V1495 was run with, and labels the intervals accordingly.

The usage does not change. You just load the data with the plugin, as you did before. There are just more "veto"-types now. Backwards compatibility is ensured and also automatic. If it doesn't find the new configs it knows that the V1495 was run with the old firmware.
In cutax, there is a branch with the same name, that adapts further processing to be compatible with the changes in this PR.

  - [ ] Update the docstring(s)
  - Documentation will be updated next
  -  Code was tested, but no new tests in repo


RoBGlaBe:
The issue seems to be DB access for creation of a context. This context is not used for the files I changed.

cfuselli (dismissed):
Hey @RoBGlaBe !
I had some time to look at your PR. 
Thanks a lot, it looks very good to me. The logic makes sense and the implementation seems correct. I have to say I did not really test the code myself, but given you tested it extensively and the tests are passing, for me this would be okay. 
I only have a copule of comments/requests. 
It seems just a bit strange to me to use the utilix xent_collection stuff inside a plugin, it's just not really in the strax style I think. What do you think about using the URLConfig to read the rundoc, as we already do in other plugins (like the LEDCalibration plugin). 
You could for example just define in the VetoIntervals plugin something like: 

```
    v1495_config = straxen.URLConfig(
        default="run_doc://daq_config.V1495.tpc?run_id=plugin.run_id",
        infer_type=True,
        cache=True,
        help="DAQ config for the V1495 board (firmware v10+).",
    )
```

and then you just have access to that part of the rundoc inside the plugin. If this is something that makes sense to you, the follow-up question would be if we still need the additional py file with the class, or if at this point we cannot just add the implementation in the VetoIntervals plugin, also given the fact that it is used only there. 
So this would basically just mean moving the functions of your new class directly to the plugin. It could get a bit messy so I am not sure if it's the best idea, but it might also cleanup a bit. If we think this is not a good option, I would still go for using the URLConfig and changing the class to accept a dictionary (the config) instead of the runid, leaving the RunDB interface to the URLConfig method. 

Let me know what you think and thanks again for the nice PR!

RoBGlaBe:
Hi Carlo,
thanks a lot for your review!
We can definitely use the URL-Config instead. This seems to be the intended way. I will take a look into it this week.
About the class: I think for a better overview and readability, it might make sense to have it still in a separate file, especially with the checks. I'll think about it and try it out, since the class can probably be a little slimmer with the URL-Config change.
I'll let you know when I can present some changes :)
Thanks again!

cfuselli:
Need to look at https://github.com/XENONnT/cutax/pull/525 as well

RoBGlaBe:
Hi @cfuselli,
thanks a lot for taking a look at this.
It's been a while, however I remember some of the tests failed due to some connectivity issue or similar. Was this a temporary thing, or did you fix that with the pytest.yml changes?

I see you also changed the test cases for version 9 runs by explicitly stating the firmware version. While this is a nice idea to state it explicitly, the reality of the run_docs is that old runs don't have this version tag (it wasn't foreseen at the time that we need to make this distinction). So straxen should assume the run uses version 9 without the tag. Only when version 10 is tagged (which is enforced for runs using this firmware), we want the new straxen behavior. For backwards compatibility we leave the implicit version 9 check in.
I'll make a suggestion to make this explicit to whoever reads the code while not putting the explicit version 9 tag in the test.

Your other changes are solely better variable namings. Always a plus, thanks a lot!

tests/test_daq_reader.py
```diff
@@ -86,6 +86,7 @@ def test_insert_deadtime(self):
             {
                 "safe_break_in_pulses": int(0.5e9),
                 "dummy_version": "test_insert_deadtime",
+                "v1495_config": {"firmware_version": 9},
```
RoBGlaBe:
```suggestion
                # No firmware_version tag is given for old runs.
                # Need v9 behavior. Silently assuming:
                # "v1495_config": {"firmware_version": 9},
```

cfuselli: @RoBGlaBe the thing is that because of https://github.com/XENONnT/straxen/blob/279e6bbe2d9476d5712636867ea745d0c3a82911/straxen/plugins/veto_intervals/veto_intervals.py#L48 we are not backwards compatible. I think its fine to not be backwards compatible? 
Also, why would we want to test on the old runs instead of the new ones?

cfuselli:
Thanks @RoBGlaBe. 

The change from "straxen_deadtime_veto" to "straxen_deadtime" in veto_intervals.py was not a simple rename of variable but a bug, because we would just not see this type of veto otherwise. 

So.. if everything looks good to you here and in https://github.com/XENONnT/cutax/pull/525, we can merge?

DonNabla:
I tested (with LLM help) this branch against released straxen on 48 runs (firmware v9 background, firmware v10 background, NG, fractional-lifetime, LED): the fw9 backwards compatibility is exact and the fracLT reclassification works as intended (intervals previously mislabeled `hev_veto` now correctly come out as `fractional_lifetime`). While testing I ran into three issues in `veto_intervals.py` that I could reproduce concretely:

### 1. Runs whose rundoc has no `daq_config.V1495.tpc` crash instead of falling back to fw9

```
>>> read_rundoc("daq_config.V1495.tpc", run_id="083498")   # nVeto LED calibration run
ValueError: No path daq_config.V1495.tpc found in rundoc for run 083498
```

This makes `veto_intervals` (and everything downstream) unprocessable for such runs, although `AqMonChannelOccupancy` itself handles a minimal config fine. The `run_doc` protocol already supports a `default=`, so the fallback can be fully automatic:

```python
v1495_config = straxen.URLConfig(
    default=(
        "run_doc://daq_config.V1495.tpc"
        "?run_id=plugin.run_id&default=plugin.pre_v10_v1495_config"
    ),
    type=dict,
    help="DAQ config for the V1495 board (firmware v10+).",
)

# Fallback for rundocs without daq_config.V1495.tpc (pre-v10 DAQ, MC, special runs)
pre_v10_v1495_config: dict = {"firmware_version": 9}
```

### 2. Reconstructed anti-veto intervals can end after the run → strax refuses the chunk

The synthesized stop (NG trigger + delay + duration) is not a real hit, so nothing keeps it inside the run. Feeding one NG trigger 0.1 s before the chunk end through the plugin (anti-veto on, 600 s duration):

```
anti_veto  time=900000000  endtime=600900000000
ValueError: Attempt to create chunk (...veto_intervals: 0sec - 1sec, 2 items)
            whose data ends late at 600900000000
```

`handle_starts_and_stops_outside_of_run` doesn't help here because the synthesized starts/stops are always paired. Suggested fix in the reconstruction branch:

```python
    veto_hits_stop = copy(veto_hits_start)
    veto_hits_stop["time"] += duration_ns

    # Synthesized times are not real hits, so nothing guarantees they stay
    # inside the run; clip to the chunk or strax raises "data ends late".
    np.clip(veto_hits_start["time"], start, end, out=veto_hits_start["time"])
    np.clip(veto_hits_stop["time"], start, end, out=veto_hits_stop["time"])
```

### 3. Artificial-deadtime slice bug (pre-existing, but this PR touches the block)

```python
result[vetos_seen:n_artificial]["time"] = artificial_deadtime["time"]
```

The slice length is `n_artificial - vetos_seen`, not `n_artificial`. With one artificial-deadtime hit plus any normal veto in the chunk this **silently corrupts the output** (numpy broadcasts the size-1 array into the empty slice: the real `straxen_deadtime` row is lost and an all-zero row is emitted); with two or more hits it raises `ValueError: could not broadcast input array from shape (2,) into shape (1,)`. Fix:

```python
if n_artificial:
    sel = slice(vetos_seen, vetos_seen + n_artificial)
    result[sel]["time"] = artificial_deadtime["time"]
    result[sel]["endtime"] = strax.endtime(artificial_deadtime)
    result[sel]["veto_type"] = "straxen_deadtime"
    vetos_seen += n_artificial
```

Minor while you're at it: `_settings_check_plausilble` → `_check_settings_plausible`, and the `anti_veto_delay_µs`/`anti_veto_duration_µs` attributes use a non-ASCII `µ` (the config keys they're read from are `_us`) -> worth renaming.

PR #1577, opened 2025-03-07
Specify strax mailbox processor in bootstrax

cfuselli:
This change is needed after https://github.com/AxFoundation/strax/pull/904 that sets default processor to single thread. At the DAQ we want to keep mailbox, so we need to specify it now in bootstrax.


cfuselli:
Close in favor of https://github.com/XENONnT/straxen/pull/1643

PR #1578, opened 2025-03-07
Accelerate `peak_se_score`, results are not changed

dachengx:


dachengx:
It is impossible to accelerate more. The 5 second window is too long.

PR #1581, opened 2025-03-11
Added more colors+types to waveform_plot

MerzJohannes:
Due to recent low level changes (i.e. classification, merging, etc.) the ["type"] field of peaks has more than just [0,1,2]. Instead we can find [-1,0,1,2,3,20,22]. When using the plot_waveform function this results in a key error, because of this line:
`plot_peak(p, t0=t_reference, color={0: "gray", 1: "b", 2: "g"}[p["type"]])`
This is fixed by extending the dictionary, which assigns a color to the other types (image added below):
`plot_peak(p, t0=t_reference, color={-1:"black", 0:"gray",1:"b",2:"g",3:"purple",20:"orange",22:"teal"}[p["type"]])`
Here the image with the different colors
ColorsWaveform

### DIFF
--- a/straxen/analyses/bokeh_waveform_plot.py
+++ b/straxen/analyses/bokeh_waveform_plot.py
@@ -486,7 +486,7 @@ def plot_peaks(peaks, time_scalar=1, fig=None, colors=("gray", "blue", "green"),
     if not fig:
         fig = straxen.bokeh_utils.default_fig(width=1600, height=400, y_axis_type=yscale)
 
-    for i in range(0, 3):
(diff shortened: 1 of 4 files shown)


dachengx:
> Hey @MerzJohannes thank you for this PR! Just a question from my side, are we sure there are no other possible values coming out from the peak classification? Maybe we could just use the gray color for all numbers that are not 1 or 2?

Thank you, @MerzJohannes. I agree with @cfuselli . Now we can set everything not S1 or S2 as gray.

PR #1585, opened 2025-03-26
Update print_versions to include all local packages

cfuselli:
With this simple PR I would like to propose an enhancement for print_versions(). 
With the proposed change, print_versions will by default print all the versions of the loaded python modules that are not sourced from an /opt/XENONnT path (the container path). 
I think this is very useful and educative as it keeps all of us always aware of which packages we are using. Because we are used to work with containers, we think we always use the packages in the containers, but this does not hold if we have locally installed packages. In this way, it's more in your eyes :)

### DIFF
--- a/straxen/misc.py
+++ b/straxen/misc.py
@@ -92,6 +92,7 @@ def print_versions(
     include_python=True,
     return_string=False,
     include_git=True,
+    include_all_local=True
 ):
     """Print versions of modules installed.
 
@@ -99,6 +100,7 @@ def print_versions(
         print_versions(modules=('numpy', 'dddm',))
     :param return_string: optional. Instead of printing the message, return a string
     :param include_git: Include the current branch and latest commit hash
+    :param include_all_local: Include all local modules (not in /opt/XENONnT/).
     :return: optional, the message that would have been printed
 
     """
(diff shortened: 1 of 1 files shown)


straxen/misc.py
```diff
@@ -108,6 +110,24 @@ def print_versions(
         versions["version"] = [python_version()]
         versions["path"] = [sys.executable]
         versions["git"] = [None]
+
+    local_modules = []
+    if include_all_local:
+        # Only activate local-module detection if container paths are present
+        module_values = sys.modules.values()
+        in_container = any(
+            mod and (getattr(mod, "__file__", "") or "").startswith("/opt/XENONnT/")
```
dachengx, line 119: The usage of `"/opt/XENONnT/"` is too XENON-specified. Can https://github.com/XENONnT/utilix/blob/d416e2e8c506ab26f9c9b7da0b6bbba01b2da071/utilix/tarball.py#L84 work?

dachengx: Another point to improve is that `"/opt/XENONnT/"` can not work in cvmfs.

dachengx (changes requested):
Thank you @cfuselli . I suggested some improvements.

straxen/misc.py
cfuselli: Hey @dachengx , thanks for looking it and for the nice suggestions, and sorry for the delay. 
I had a second look at it but I still think the scope of the utilix function is a bit different, here it is more general. It could not work with packages installed in other locations than .local (for example if you just add the path to the pythonpath). So I would still go for checking the startswith.. 
I changed it into just /opt and now check also /cvmfs, so it is less xenon. 
What do you think?

PR #1594, opened 2025-04-23
Make sure `version=ONLINE` is not used when global version is not `global_ONLINE`

dachengx:
Similar to https://github.com/XENONnT/straxen/pull/1546

Sometimes you might forget to overwrite some config when initializing the offline context. This happens when the offline global config does not provide enough configuration.

To bypass the check, run

```
st.set_context_config({'check_global_version_configs': False})
```

### DIFF
--- a/straxen/config/__init__.py
+++ b/straxen/config/__init__.py
@@ -2,3 +2,4 @@ from .url_config import *
 from .protocols import *
 from .preprocessors import *
 from .check_superruns import *
+from .check_global_version import *
--- a/straxen/config/check_global_version.py
+++ b/straxen/config/check_global_version.py
@@ -9,7 +9,7 @@ def check_global_version_wrapper(func):
         if (
             not self.context_config["check_global_version_configs"]
             or "xedocs_version" not in self.context_config
-            or self.context_config["xedocs_version"] is not None
+            or self.context_config["xedocs_version"] is None
             or self.context_config["xedocs_version"] == "global_ONLINE"
         ):
             return func(self, run_id=run_id, targets=targets, **kwargs)
--- a/straxen/config/check_global_version.py
+++ b/straxen/config/check_global_version.py
@@ -1,0 +1,67 @@
+import strax
+import straxen
+
+
+def check_global_version_wrapper(func):
+    """Check whether all configs for superruns-allowed plugins are the same for subruns."""
+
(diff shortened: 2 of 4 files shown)


cfuselli (changes requested):
Hey @dachengx thanks for this cool PR. 
I don't fully understand how all of this works, but I tested it.. 
I added one line to check if xedocs version is None, as that is set by default in the context if none is passed, otherwise all tests were failing with "xedocs version is None and you have ONLINE". 
It seems to work but... I think it works only when you request the components (or try to make) a data type that uses the config that is set to ONLINE. So I think this is why the test fails (I think!). 
Anyway, with the current status of the PR it is not possible to produce data types that have online configs, but it's still possible to define the context and make the rest of data types. Maybe this is fine and we can just change the logic of the test, I leave the decision up to you :) 
Thanks again,
Carlo

dachengx:
> Hey @dachengx thanks for this cool PR. I don't fully understand how all of this works, but I tested it.. I added one line to check if xedocs version is None, as that is set by default in the context if none is passed, otherwise all tests were failing with "xedocs version is None and you have ONLINE". It seems to work but... I think it works only when you request the components (or try to make) a data type that uses the config that is set to ONLINE. So I think this is why the test fails (I think!). Anyway, with the current status of the PR it is not possible to produce data types that have online configs, but it's still possible to define the context and make the rest of data types. Maybe this is fine and we can just change the logic of the test, I leave the decision up to you :) Thanks again, Carlo

Hi Carlo. Thank you for the check.

> but it's still possible to define the context and make the rest of data types

I think this is fine.

cfuselli (approved):
Great!

cfuselli:
It gives me this with v16: 

`ValueError: The global version is set to be global_v16. But PeakletPositionsCNF is still using ONLINE version config cnf_pred_function, which is [jax://resource://xedocs://posrec_models?attr=value&fmt=abs_path&kind=cnf&n_poly=plugin.n_poly&run_id=plugin.run_id&sig=plugin.sig&version=ONLINE](jax://resource//xedocs://posrec_models?attr=value&fmt=abs_path&kind=cnf&n_poly=plugin.n_poly&run_id=plugin.run_id&sig=plugin.sig&version=ONLINE).`

and it just works with v17. 

So all as excpected.

PR #1592, opened 2025-04-08
Add peak bias correction

cfuselli:
For now just a draft. 
Testing adding a correction for peak reconstruction bias. 
Is it only for corrected areas or should we add it at peak level?

PR #1598, opened 2025-05-03
IMplement N-1 correction variables

DonNabla:

Add types to CorrectedAreas plugin to facilitate the study of correction impact.

### _Notes on testing_

### DIFF
--- a/straxen/plugins/events/corrected_areas.py
+++ b/straxen/plugins/events/corrected_areas.py
@@ -185,18 +185,6 @@ class CorrectedAreas(strax.Plugin):
                 ),
             ]
             # 3. All corrections except photon ionization correction
-            dtype += [
-                (
-                    f"{peak_type}cs2_wo_picorr",
-                    np.float32,
-                    f"Corrected area of {peak_name} S2 (without photon ionization correction) [PE]",
-                ),
-                (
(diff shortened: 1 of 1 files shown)


DonNabla:
Need to implement the peak bias corr based on #1592

straxen/plugins/events/corrected_areas.py
```diff
@@ -138,49 +151,75 @@ def infer_dtype(self):
...
+                ]
+
+            # N-1 corrections for S2
+            n1_versions = {
+                "xycorr": "xy position",
+                "segee": "SEG/EE",
+                "picorr": "photoionization",
+                "elifecorr": "electron lifetime",
+                "peakbiascorr": "peak bias",
+                "timecorr": "all time dependent",
```
dachengx, line 210: In principle, PI correction is also time-dependent. But in the codes, only SEG/EE + elife are not applied.

dachengx (commented):
Thanks @DonNabla . I made a PR to simplify (?) a bit the logic https://github.com/XENONnT/straxen/pull/1617.

straxen/plugins/events/corrected_areas.py
DonNabla: You are right, I should also remove the PI correction

PR #1600, opened 2025-05-10
Add relative z- t- dependent LY correction

GianMarco98:
Adding a relative z- t- dependent LY correction, as described in [this note ](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:lucchetti:xenonnt:sr2:relativelycorrection).

As discussed in the [slack thread](https://xenonnt.slack.com/archives/C049JJZ2U3A/p1746798697119419), we apply the z- dependent correction only until the low PI period, as a conservative approach.

### _Notes on testing_

### DIFF
--- a/straxen/plugins/events/corrected_areas.py
+++ b/straxen/plugins/events/corrected_areas.py
@@ -214,7 +214,7 @@ class CorrectedAreas(strax.Plugin):
         b = self.b_rel_light_yield
 
         # Compute full z- and t-dependent correction
-        rel_ly_zt_corr = self.rel_light_yield * (a * (events["z"]**2 + b * events["z"]) + 1)
+        rel_ly_zt_corr = self.rel_light_yield * (a * (events["z"] ** 2 + b * events["z"]) + 1)
 
         return rel_ly_zt_corr
 
--- a/straxen/plugins/events/corrected_areas.py
+++ b/straxen/plugins/events/corrected_areas.py
@@ -76,11 +76,6 @@ class CorrectedAreas(strax.Plugin):
         help="Relative extraction efficiency for this run (allows for time dependence)",
     )
 
-    sr = straxen.URLConfig(
-        default="science_run://plugin.run_id?&phase=False",
-        help="Science run to be used for the corrections. It can affect the parameters of the corrections.",
-    )
-
     # relative light yield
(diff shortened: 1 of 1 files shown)


straxen/plugins/events/corrected_areas.py
```diff
@@ -76,13 +76,36 @@ class CorrectedAreas(strax.Plugin):
...
+
+    # slope parameter for z-dependent relative light yield correction
+    slope_rel_ly = straxen.URLConfig(
+        default={"sr0": 0, "sr1": 0, "sr2": -6.323e-5},
+        help="Slope parameter for z-dependent relative light yield correction",
+    )
+
+    # date before which the z-dependent correction is applied
+    lowPI_date = straxen.URLConfig(
+        default=1733428781045000000,
```
dachengx, line 105: What exactly is the splitting run_id?

GianMarco98: That’s the start timestamp of the 25th Kr calibration, at which we decided to not apply the z-dependent correction, and to use instead only the time dependent correction as in SR1

dachengx (changes requested):
Hey, Gian Marco. Nice implementation! Would you please be more detailed in the PR description, even details are discussed in SLACK. For example, would you link the summary note of the study?

dachengx (approved):

dachengx:
We do not have `z` for peaks (though we have `z_obs_ms`), but now the relative light yield depends on `z` in https://github.com/XENONnT/straxen/blob/53135ee4a27e4423fc3e66c8a68c5b16ecbe2f1d/straxen/plugins/peaks/peak_corrections.py#L94. Do you think this is a problem? @cfuselli

cfuselli:
Hey @dachengx , I had a quick look and I understand the problem.. 
Why not just use the `z_obs_ms` that you mentioned? Afterall this is exactly the same situation as for the s1_xyz correction map used in the line above, no? 
In the entire codebase of xenonnt I only found a reference for the  `s1_rel_light_yield_correction_factor ` in [here](https://github.com/XENONnT/straxen/blob/53135ee4a27e4423fc3e66c8a68c5b16ecbe2f1d/straxen/plugins/events/multi_scatter.py#L123)

PR #1617, opened 2025-07-01
Simplify https://github.com/XENONnT/straxen/pull/1598

dachengx:
Simplify https://github.com/XENONnT/straxen/pull/1598, though the number of lines of code increases, the logic is (hopefully) clearer.

Confine the S2 corrections application in `apply_s2_corrections`. When calculating N-1 corrections, set the corresponding dropped correction as 1.

### DIFF
--- a/straxen/plugins/events/corrected_areas.py
+++ b/straxen/plugins/events/corrected_areas.py
@@ -200,14 +200,14 @@ class CorrectedAreas(strax.Plugin):
                     (f"{peak_type}cs2_area_fraction_top{name}", np.float32, aft_comment),
                 ]
 
-            # N-1 corrections for S2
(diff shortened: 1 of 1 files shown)


straxen/plugins/events/corrected_areas.py
```diff
@@ -200,14 +200,14 @@ def infer_dtype(self):
                     (f"{peak_type}cs2_area_fraction_top{name}", np.float32, aft_comment),
                 ]
 
-            # N-1 corrections for S2
+            # N-1 corrections for S2, list them in the order of application
             n1_versions = {
+                "peakbiascorr": "peak bias",
```
dachengx, line 205: The bias corrections will be applied first, so here I drop it first.

PR #1626, opened 2025-08-20
Remove cAFT bottom corr form cS2

sebvetter:
In [straxen 2.2.7](https://github.com/XENONnT/straxen/blob/v2.2.7/straxen/plugins/events/corrected_areas.py#L270) the corrected areas plugin specifically excluded the bottom top ratio correction that is caused by PI from the total area returned as cS2. Instead, this correction was only applied to the AFT correction.

In [straxen 3.2.3](https://github.com/XENONnT/straxen/blob/v3.2.3/straxen/plugins/events/corrected_areas.py#L324) the bottom PI correction is applied to the bottom area, and therefore the correction is included in the final returned cS2.

A discussion of this can be found in [this note](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:svetter:sr2_cy_pi_correlation_plot_collection#plugin_rework_proposal).

The `cs2` variable is now computed as follows:
<code>
cs2_top_before_pi = s2_area_top / s2_bias_correction / s2_xy_correction_top / seg_ee_corr
cs2_bottom_before_pi = (
    s2_area_bottom / s2_bias_correction / s2_xy_correction_bottom / seg_ee_corr
)
cs2 = (cs2_top_before_pi + cs2_bottom_before_pi) * elife_correction
</code>

The `cs2_area_fraction_top` variable is computed as follows:
<code>
cs2_area_fraction_top = cs2_top_before_pi / (cs2_top_before_pi + cs2_bottom_before_pi * pi_corr_bottom)
</code>

The variable distributions of the resulting and intermediate variables for selected Kr runs look like this:

sr2_cy_pi_corr_plugin_rework_distributions

### DIFF
--- a/straxen/plugins/events/corrected_areas.py
+++ b/straxen/plugins/events/corrected_areas.py
@@ -37,7 +37,7 @@ class CorrectedAreas(strax.Plugin):
 
     """
 
-    __version__ = "0.5.6"
+    __version__ = "0.5.7"
 
     depends_on: Tuple[str, ...] = ("event_basics", "event_positions")
 
@@ -297,6 +297,15 @@ class CorrectedAreas(strax.Plugin):
         elife_correction,
     ):
         """Apply S2 corrections and return various corrected areas.
+        The variables include the following corrections:
+            cs2: peak bias, S2xy, SEG/EE, elife
+            cs2_area_fraction_top: peak bias, S2xy, SEG/EE, cAFT, elife
+            cs2_before_pi: peak bias, S2xy, SEG/EE
+            cs2_area_fraction_top_before_pi: peak bias, S2xy, SEG/EE
+            cs2_before_elife: peak bias, S2xy, SEG/EE, cAFT
+            cs2_area_fraction_top_before_elife: peak bias, S2xy, SEG/EE, cAFT
+            cs2_wo_segee_picorr: peak bias, S2xy, elife
+            cs2_area_fraction_top_wo_segee_picorr: peak bias, S2xy, elife
 
         Returns:
             cs2,
@@ -324,8 +333,10 @@ class CorrectedAreas(strax.Plugin):
         cs2_bottom_before_elife = cs2_bottom_before_pi * pi_corr_bottom
         cs2_before_elife = cs2_top_before_elife + cs2_bottom_before_elife
 
-        cs2 = cs2_before_elife * elife_correction
-        cs2_top_final = cs2_top_before_elife * elife_correction
+        # Do not correct total area for cAFT
+        cs2 = (cs2_top_before_pi + cs2_bottom_before_pi) * elife_correction
+        # Correct AFT for cAFT
+        cs2_area_fraction_top = cs2_top_before_pi / (cs2_top_before_pi + cs2_bottom_before_pi * pi_corr_bottom)
 
         cs2_top_wo_segee_picorr = s2_area_top / s2_bias_correction / s2_xy_correction_top
         cs2_bottom_wo_segee_picorr = s2_area_bottom / s2_bias_correction / s2_xy_correction_bottom
@@ -335,7 +346,7 @@ class CorrectedAreas(strax.Plugin):
         cs2_top_wo_segee_picorr = cs2_top_wo_segee_picorr * elife_correction
         return (
             cs2,
-            cs2_top_final / cs2,
+            cs2_area_fraction_top,
             cs2_before_pi,
             cs2_top_before_pi / cs2_before_pi,
             cs2_before_elife,


GiovanniVolta:
It looks good to me! It is indeed what we discussed.
I failed now to compare it with the [v2.7.2](https://github.com/XENONnT/straxen/blob/bdaa023e4385ac88ecd31eec6d6d28b33a364012/straxen/plugins/events/corrected_areas.py#L272). Those line code a bit difficult to digest.

I am also wondering if we really want to give out:
```
           cs2_before_elife,
           cs2_top_before_elife / cs2_before_elife
```
            
This are:
```
cs2_top_before_elife = cs2_top_before_pi
cs2_bottom_before_elife = cs2_bottom_before_pi * pi_corr_bottom
cs2_before_elife = cs2_top_before_elife + cs2_bottom_before_elife
```

And the PI correction is eventually used only to correct the AFT. So the `before_elife` should be `cs2_before_pi`. In any case, it is explained in the docstring.

Are we also interested in cS2 w/ only peak_bias and/or cS2 w/ only S2xy?

straxen/plugins/events/corrected_areas.py
```diff
@@ -297,6 +297,15 @@ def apply_s2_corrections(
         elife_correction,
     ):
         """Apply S2 corrections and return various corrected areas.
+        The variables include the following corrections:
+            cs2: peak bias, S2xy, SEG/EE, elife
+            cs2_area_fraction_top: peak bias, S2xy, SEG/EE, cAFT, elife
```
dachengx, line 302: `elife` should not change `cs2_area_fraction_top`. Only the docstring should be changed; the code is fine.

sebvetter: Same applies for all values that do not separate between top and bottom, namely SEG/EE + peak bias + elife. They get cancelled mathematically in the AFT computation. So far at least in theory. But we compute e.g. the `cs2_area_fraction_top` on the variables where elife and so on are applied. In case something goes wrong and we see weird numbers, I want the docstring to include the actual applied corrections from the code. Even if they get cancelled out. 

What do you think? I can also add this justification to the docstring. Or if you say the docstring should only include what is the mathematical truth, then I remove SEG/EE + peak bias  + elife from the AFT docstrings.

GiovanniVolta: For me, more information in the docstring, better it is (to some extent). To avoid confusion, I would add the justification.

straxen/plugins/events/corrected_areas.py
```diff
@@ -297,51 +313,106 @@ def apply_s2_corrections(
         elife_correction,
     ):
         """Apply S2 corrections and return various corrected areas.
+        The variables include the following corrections:
+            cs2: peak bias, S2xy, SEG/EE, elife
+            cs2_area_fraction_top: peak bias, S2xy, SEG/EE, elife, PI
+            cs2_w_bias: peak bias
+            cs2_area_fraction_top_w_bias: peak bias
+            cs2_w_bias_xy: peak bias, S2xy
+            cs2_area_fraction_top_w_bias_xy: peak bias
```
dachengx, line 322: `cs2_area_fraction_top_w_bias_xy: peak bias` -> `cs2_area_fraction_top_w_bias_xy: peak bias, S2xy`

straxen/plugins/events/corrected_areas.py
```diff
@@ -297,51 +313,106 @@ def apply_s2_corrections(
...
+        # SEG/EE
+        cs2_top_w_bias_xy_segee = cs2_top_w_bias_xy / seg_ee_corr
+        cs2_bottom_w_bias_xy_segee = cs2_bottom_w_bias_xy / seg_ee_corr
+        cs2_w_bias_xy_segee = cs2_top_w_bias_xy_segee + cs2_bottom_w_bias_xy_segee
+        # elife
+        cs2_top = cs2_top_w_bias_xy_segee * elife_correction
+        cs2_bottom = cs2_bottom_w_bias_xy_segee * elife_correction
+        cs2 = cs2_top + cs2_bottom
+        # PI AFT
+        cs2_area_fraction_top = cs2_top / (cs2_top + cs2_bottom * pi_corr_bottom)
```
dachengx, line 380: There might be a very common bug. For S2-Only events, the `elife_correction` will be `np.nan`, so `cs2_area_fraction_top` will be `np.nan` if `cs2_top` is affected by `elife_correction`.

dachengx: Please make sure also in the docstring that the elife does not affect aft.

dachengx (changes requested):

straxen/plugins/events/corrected_areas.py
dachengx: Sorry I was wrong, this line https://github.com/XENONnT/straxen/blob/e9fe5dd7fb61fb537d35732730b41639ebc39980/straxen/plugins/events/corrected_areas.py#L532 guaranteed that the S2-Only events' AFTs are not np.nan

dachengx (commented):
In general, can we use code like https://github.com/XENONnT/straxen/blob/e9fe5dd7fb61fb537d35732730b41639ebc39980/straxen/plugins/events/corrected_areas.py#L427-L443 to calculate the cS2 and cS2AFT with or without some particular corrections?

straxen/plugins/events/corrected_areas.py
dachengx: It should be `cs2_area_fraction_top_wo_elifecorr`, but later we assigned `cs2_area_fraction_top` as `cs2_area_fraction_top_wo_elifecorr` so in cutax code it is `cs2_area_fraction_top` https://github.com/XENONnT/cutax/blob/1e2d98cfde1683c52c873c5eaad63f260b93003a/cutax/cuts/s2_aft.py#L93

PR #1638, opened 2026-01-16
fixed the order of the corrections in apply_s2_correction()

caioishikawa:

The order of two arguments: "rel_cy_correction" and "elife_correction" in the function apply_s2_correction() were inverted when it was defined and when it was called, making our cS2 look much smaller than it should, since we were applying the correction in the wrong way. This PR changes the order in the definition of the function in a way to leave the last correction to be "elife_correction".

### _Notes on testing_

### DIFF
--- a/straxen/plugins/events/corrected_areas.py
+++ b/straxen/plugins/events/corrected_areas.py
@@ -38,7 +38,7 @@ class CorrectedAreas(strax.Plugin):
 
     """
 
-    __version__ = "0.5.11"
+    __version__ = "0.6.0"
 
     depends_on: Tuple[str, ...] = ("event_basics", "event_positions")
 


caioishikawa:
> Let me summarize my understanding quickly: We introduced the rel CY correction because even after applying all other corrections, we still saw some time dependence of cS2. Therefore the rel CY is derived on data with all other corrections applied. This requires the rel CY to be applied as the last correction, specifically on elife corrected data. However, the elife correction will artificially assign NaNs to data without S1s, which messes with the S2-only analysis. Thats why we shifted the elife correction to the back of the queue. Now those two things are conflicting.
> 
> Is this understanding correct?

Partially yes, the motivation is as you said because of the leftover time dependence. However the problem here is much simpler, it is simply that we changed the order of the arguments when calling the function apply_s2_correction(), as you said, to make the elife correction to be the last one. But we forgot to change the order of the arguments when defining the function. So we were applying elife_correction as if it is rel_cy_correction and applying rel_cy_correction as if it is elife_corr and since one of them multiply and the other divide, we were applying both in a wrong way

GiovanniVolta:
Hey, the fix looks fine! I have a doubt on the other still

GiovanniVolta:
Right now we are applying the rel_cy to top + bottom instead of applying it separately. 
So basically:
```
cs2_top_wo_elife_w_rel_cy = cs2_top_wo_elife / rel_cy_correction
cs2_bottom_wo_elife_w_rel_cy = cs2_bottom_wo_elife / rel_cy_correction

cs2 = (cs2_top_wo_elife_w_rel_cy + cs2_bottom_wo_elife_w_rel_cy) * elife_correction

cs2_area_fraction_top = cs2_top_wo_elife_w_rel_cy / (
            cs2_top_wo_elife_w_rel_cy + cs2_bottom_wo_elife_w_rel_cy * pi_corr_bottom
        )
```

But this is IDENTICAL to the current implementation. All good!

GiovanniVolta:
When @sebvetter has done with his check on Kr we can go ahead for me

GiovanniVolta:
Let's go

PR #1643, opened 2026-02-03
Straxen vanilla

cfuselli:
This pull request introduces new "vanilla" versions of several key plugins and updates the context configuration to support them, primarily for the XENONnT experiment data processing. The changes include the addition of new plugins for merged S2 signals and peak position reconstruction, updates to context options to select between standard and vanilla processing chains, and various improvements and fixes in plugin dependencies and dtype handling.

**New vanilla plugins and context options:**

* Added `MergedS2sVanilla` and `MergedS2sHighEnergyVanilla` plugins, along with their imports and registration, to provide alternative algorithms for merging S2 signals. [[1]](diffhunk://#diff-0ae82305375d57aa91e828e161218b318dfe7ca2351ab75cee384ebd7052387fR1-R270) [[2]](diffhunk://#diff-004197a824632407fc50381db863995468a7ca62cde37eb56d8d12be237c8b0eR1-R34) [[3]](diffhunk://#diff-eef6149b3a4c3c7de7b00c6f72b19e7557653adf037d74fc39b7b0b1bb9667b5R4-R7) [[4]](diffhunk://#diff-638ab3c44f1a1482c83ea92ca53b46e1ad63fa811742ad75a34f63d98e596035R3-R5)
* Introduced `PeakPositionsMLPVanilla` and `PeakPositionsCNFVanilla` plugins for peak position reconstruction using MLP and CNF algorithms, respectively. [[1]](diffhunk://#diff-a0d3f12ee1799d3623ef8629e4964d58575b2f20da479045bbaf56a1ca3c2328R20-R33) [[2]](diffhunk://#diff-35d626d3a75fb2595af026738a40ee24c091a18682bea9e36cf678585e2259b5R20-R33)
* Updated context configuration in `straxen/contexts.py` to add `common_opts_vanilla` and logic in `xenonnt_online` for selecting vanilla plugins, with additional debug print statements. [[1]](diffhunk://#diff-ebd5e6c1e333dd3011a01c5a7be9333983da6d07722650ac8745ca560fadb723R37-R62) [[2]](diffhunk://#diff-ebd5e6c1e333dd3011a01c5a7be9333983da6d07722650ac8745ca560fadb723L301-R338) [[3]](diffhunk://#diff-ebd5e6c1e333dd3011a01c5a7be9333983da6d07722650ac8745ca560fadb723R193-R195)

**Plugin dependency and dtype handling improvements:**

* Updated `PeaksVanilla` to depend on `peaklet_classification` instead of `enhanced_peaklet_classification`, and improved dtype merging and alignment for compatibility with numba. [[1]](diffhunk://#diff-76e256c0f2cf834bc27df350ba29fef2ed38902854f81aaf2481aba985fe64e3L24-R24) [[2]](diffhunk://#diff-76e256c0f2cf834bc27df350ba29fef2ed38902854f81aaf2481aba985fe64e3L49-L61)
* Improved dtype alignment and merging logic in `PeaksSOM` plugin to prevent numba errors.
* Commented out the `merged` field in vanilla peak and event basics plugins to match the new vanilla processing chain. [[1]](diffhunk://#diff-d8c89bad729519104f11621873316097e1e10c4a00c2f70e36e01705fafec026L57-R57) [[2]](diffhunk://#diff-d8c89bad729519104f11621873316097e1e10c4a00c2f70e36e01705fafec026L84-R85) [[3]](diffhunk://#diff-92307fc79e5c77d8453f8256945b0cbdd9320789e1785af9bbaef73214ff95e5L169-R169) (!!!!) This field we would actually like to add, but I did not manage to do it well!

**Bug fixes and minor changes:**

* Fixed a bug in `is_the_led_on` to use the last document in a list instead of the first.
* Minor formatting and import improvements in various files. [[1]](diffhunk://#diff-b8042fdb4a1b310312c8b777d87feb51fc58fa64557a821eb9870a03186be844R5) [[2]](diffhunk://#diff-26b437972e666142ed832154317618abbbcb251a69ed54df689e68b6da4e173bR232)

These changes collectively enable an alternative "vanilla" data processing path in the XENONnT analysis framework, with improved plugin modularity and maintainability.

---

**References:**
- [[1]](diffhunk://#diff-0ae82305375d57aa91e828e161218b318dfe7ca2351ab75cee384ebd7052387fR1-R270) [[2]](diffhunk://#diff-004197a824632407fc50381db863995468a7ca62cde37eb56d8d12be237c8b0eR1-R34) [[3]](diffhunk://#diff-eef6149b3a4c3c7de7b00c6f72b19e7557653adf037d74fc39b7b0b1bb9667b5R4-R7) [[4]](diffhunk://#diff-638ab3c44f1a1482c83ea92ca53b46e1ad63fa811742ad75a34f63d98e596035R3-R5) [[5]](diffhunk://#diff-a0d3f12ee1799d3623ef8629e4964d58575b2f20da479045bbaf56a1ca3c2328R20-R33) [[6]](diffhunk://#diff-35d626d3a75fb2595af026738a40ee24c091a18682bea9e36cf678585e2259b5R20-R33) [[7]](diffhunk://#diff-ebd5e6c1e333dd3011a01c5a7be9333983da6d07722650ac8745ca560fadb723R37-R62) [[8]](diffhunk://#diff-ebd5e6c1e333dd3011a01c5a7be9333983da6d07722650ac8745ca560fadb723L301-R338) [[9]](diffhunk://#diff-ebd5e6c1e333dd3011a01c5a7be9333983da6d07722650ac8745ca560fadb723R193-R195) [[10]](diffhunk://#diff-76e256c0f2cf834bc27df350ba29fef2ed38902854f81aaf2481aba985fe64e3L24-R24) [[11]](diffhunk://#diff-76e256c0f2cf834bc27df350ba29fef2ed38902854f81aaf2481aba985fe64e3L49-L61) [[12]](diffhunk://#diff-b8042fdb4a1b310312c8b777d87feb51fc58fa64557a821eb9870a03186be844R21-R38) [[13]](diffhunk://#diff-d8c89bad729519104f11621873316097e1e10c4a00c2f70e36e01705fafec026L57-R57) [[14]](diffhunk://#diff-d8c89bad729519104f11621873316097e1e10c4a00c2f70e36e01705fafec026L84-R85) [[15]](diffhunk://#diff-92307fc79e5c77d8453f8256945b0cbdd9320789e1785af9bbaef73214ff95e5L169-R169) [[16]](diffhunk://#diff-5b73fb2b1583eaaed7d4293063a7e41ca2d4fefb2da2a739fa8e9648878e376eL249-R249) [[17]](diffhunk://#diff-b8042fdb4a1b310312c8b777d87feb51fc58fa64557a821eb9870a03186be844R5) [[18]](diffhunk://#diff-26b437972e666142ed832154317618abbbcb251a69ed54df689e68b6da4e173bR232)


cfuselli:
@WenzDaniel @dachengx I think if you would like to have a look at the changes most things make sense to me, and I see no issues or conflicts with the default pipeline. Just adding Vanilla/Online related stuff (plus scripts changes that are just DAQ stuff). Only thing still to fix is this merged field I removed for now.

straxen/plugins/events/event_basics_vanilla.py
```diff
@@ -169,7 +169,7 @@ def _set_dtype_requirements(self):
             ("area_fraction_top", np.float32, "fraction of area seen by the top PMT array"),
             ("tight_coincidence", np.int16, "channel within tight range of mean"),
             ("n_saturated_channels", np.int16, "total number of saturated channels"),
-            ("merged", bool, "is merged from peaklets"),
+            # ("merged", bool, "is merged from peaklets"),
```
WenzDaniel: Can you elaborate what the difficulty is? Is it because it is only in S2 and not S1?

straxen/plugins/led_cal/led_calibration.py
```diff
@@ -246,7 +246,7 @@ def is_the_led_on(run_doc, default_run_comments, noise_run_comments):
     # Check if run_doc is a list with a dictionary
     if isinstance(run_doc, list) and isinstance(run_doc[0], dict):
         # Extract the dictionary
-        doc = run_doc[0]
+        doc = run_doc[-1]
```
WenzDaniel, line 249: Why this change?

straxen/plugins/merged_s2s/merged_s2s_vanilla.py
```diff
@@ -0,0 +1,270 @@
+from typing import Tuple
```
WenzDaniel, line 3: Can you maybe add as reference from which straxen version you took the plugin? That we at lest to some degree are able to track its history.

straxen/plugins/led_cal/led_calibration.py
WenzDaniel: There can be more than one document?

straxen/plugins/merged_s2s_he/merged_s2s_he_vanilla.py
```diff
@@ -0,0 +1,34 @@
+import strax
```
WenzDaniel, line 1: I assume you used here the same straxen version?

straxen/plugins/peaks/peak_basics_vanilla.py
```diff
@@ -54,7 +54,7 @@ def infer_dtype(self):
                 np.int16,
             ),
             (("Classification of the peak(let)", "type"), np.int8),
```
WenzDaniel, line 56: Even if it does not work at event basics we can still leave it here is not it?

straxen/plugins/merged_s2s/merged_s2s_vanilla.py
```diff
@@ -0,0 +1,270 @@
...
+        peaklets_dtype = self.deps["peaklets"].dtype_for("peaklets")
+        # The merged dtype is argument position dependent!
+        # It must be first classification then peaklet
+        # Otherwise strax will raise an error
+        # when checking for the returned dtype!
+        merged_s2s_dtype = strax.merged_dtype(
+            (
+                peaklet_classification_dtype,
+                peaklets_dtype,
+                # self.indicator_dtype
```
WenzDaniel: Can we keep it at least here?

straxen/plugins/peaks/peaks_som.py
```diff
@@ -17,6 +18,24 @@ class PeaksSOM(PeaksVanilla):
...
+        # we need to merge them
+        peaklets_dtype = self.deps["peaklets"].dtype_for("peaklets")
+        peaklet_classification_dtype = self.deps["enhanced_peaklet_classification"].dtype_for(
+            "enhanced_peaklet_classification"
+        )
+        merged_dtype = strax.merged_dtype((peaklets_dtype, peaklet_classification_dtype))
+        # Numba is very picky about alignment for structured dtypes. Ensure an aligned dtype
+        # so assignments inside strax.replace_merged don't see "unaligned array(...)".
+        return np.dtype(merged_dtype, align=True)
+
```
WenzDaniel, line 38: Is this a AI addition? Can you elaborate?

straxen/plugins/peaks/peaks_vanilla.py
```diff
@@ -46,19 +46,32 @@ def infer_dtype(self):
...
+        # # produce unaligned dtypes, which triggers TypingError in numba.
+        # common_dtype = self.dtype
+
+        # _peaklets = np.zeros(len(peaklets), dtype=common_dtype)
+        # strax.copy_to_buffer(peaklets, _peaklets, f"_cast_{self.provides[0]}_peaklets")
+
+        # _merged_s2s = np.zeros(len(merged_s2s), dtype=common_dtype)
+        # strax.copy_to_buffer(merged_s2s, _merged_s2s, f"_cast_{self.provides[0]}_merged_s2s")
+
+        # _peaks = self.replace_merged(_peaklets, _merged_s2s, merge_s0=self.merge_s0)
```
WenzDaniel: Can this go? I assume that the copying is not needed anymore because of this unaligned/aligned thing. I would be still interested to understand what the problem was.

straxen/scripts/bootstrax.py
```diff
@@ -44,11 +44,41 @@
...
+# Register plugins
+PLUGINS_TO_REGISTER = [
+    straxen.PeaksVanilla,
+    straxen.PeakBasicsVanilla,
+    straxen.PeakletClassificationVanilla,
+]
+# Save when to target
+straxen.MergedS2s.save_when = SaveWhen.TARGET
+straxen.PeakletClassificationVanilla.save_when = SaveWhen.TARGET
+straxen.PeakletPositionsCNF.save_when = SaveWhen.TARGET
```
WenzDaniel: Is it not possible to adjust the plugins themselves? Or does this collide with OSG reprocessing?

straxen/scripts/bootstrax.py
```diff
@@ -44,11 +44,41 @@
...
+# Don't do the rechunk on load for raw_records and peaklets
+# It's something we do for offline reprocessing
+straxen.DAQReader.rechunk_on_load = False
+straxen.Peaklets.rechunk_on_load = False
+# Register plugins
+PLUGINS_TO_REGISTER = [
+    straxen.PeaksVanilla,
+    straxen.PeakBasicsVanilla,
+    straxen.PeakletClassificationVanilla,
+]
```
WenzDaniel: Why in bootstrax and not a context?

straxen/contexts.py
```diff
@@ -300,7 +326,12 @@ def xenonnt_online(xedocs_version="global_ONLINE", _from_cutax=False, **kwargs):
     if not _from_cutax and xedocs_version != "global_ONLINE":
         warnings.warn("Don't load a context directly from straxen, use cutax instead!")
 
-    st = straxen.contexts.xenonnt(**kwargs)
+    if _from_cutax and xedocs_version != "global_ONLINE":
+        context_options = {**straxen.contexts.common_opts, **kwargs}
+    else:
+        context_options = {**straxen.contexts.common_opts_vanilla, **kwargs}
+
+    st = straxen.contexts.xenonnt(**context_options)
```
WenzDaniel: This implies that we will never move back to the other plugins for online processing? I think it make sense, but was this discussed with a wider audience? However, if this is the case we also should not make this if/else statement for the "online" context, but always register the vanilla plugins.

WenzDaniel (commented):
Thanks Carlo I think it looks good. I have not checked if you copied the code correctly, but I assume you did. Neither did I test process some data, but here I assume you did as well. I cannot really judge the ajax and bootstrax changes though. Below you can find a few smaller questions from my side. I think generally the base is good to iterate from here.

straxen/plugins/events/event_basics_vanilla.py
cfuselli: Simply I didn't have the time to manage to make it work with the MergedS2sVanilla. Just some issues with dtypes and stuff, and I didnt know how to assign its value. It should not be easy to add again. Maybe @dachengx do you want to give it a quick try?

straxen/plugins/led_cal/led_calibration.py
cfuselli: The code in this plugin is actually misleading, what is called run_doc in the code is actually the comment of the rundoc (see the URLConfig, it calls comments on run_doc). So because sometimes the comment is wrong, you can updated it from the DAQ website, and this way we take the last one. So you can fix a comment for a run that would not work.

straxen/plugins/merged_s2s/merged_s2s_vanilla.py
cfuselli: Added as a comment at the top

straxen/plugins/peaks/peak_basics_vanilla.py
cfuselli: you mean the merged field? if we dont propagate it better to remove it..

straxen/plugins/merged_s2s/merged_s2s_vanilla.py
cfuselli: this is where I was having issues actually I think

straxen/plugins/peaks/peaks_som.py
cfuselli: its from the peaks vanilla plugin... but I had to change the name of the deps. 
this part: 
```
#numba is very picky about alignment for structured dtypes. Ensure an aligned dtype
# so assignments inside strax.replace_merged don't see "unaligned array(...)".
return np.dtype(merged_dtype, align=True)
```
I dont remember if the unaligned array was a temporary error I had or if still needed..

straxen/plugins/peaks/peaks_vanilla.py
cfuselli: yes this can go. Dont remember about the unaligned issue..

cfuselli: Numba failing inside strax.replace_merged because it’s being fed an unaligned structured array dtype:
	•	Key clue: setitem(unaligned array(Record(...)), ...)
	•	Numba’s jitted _replace_merged does assignments like result[result_i] = merge[window_i], and it refuses when the structured dtype is unaligned (or when orig and merge are not exactly the same aligned dtype).

this was it

straxen/scripts/bootstrax.py
cfuselli: Yeah, I am not sure why they were set like this, maybe needed for reprocessing? 
This needs to be coordinated well with DAQ, Transfers, Reprocessing..

straxen/scripts/bootstrax.py
cfuselli: this is changes I made in bootstrax for testing before making the new online context with the vanilla plugins registered. can go.

straxen/contexts.py
cfuselli: mmm "always" means in this straxen version, if this will change it would be anyway with a new straxen version where we can tune again the online context. 
at the moment, i think its good to keep online from straxen using vanilla, and leave the option to use the common_opts if specifying online from cutax? idk.. this we can decide

cfuselli:
Thank you @WenzDaniel for the careful review. 
I answered all the comments and implemented the changes where needed.
On a few points we might need to discuss a bit more..

cfuselli:
Since the last review, I've added CI testing for vanilla plugins..

**What changed:**
- Added `_vanilla` parameter to `xenonnt()` context to switch between SOM/vanilla plugins
- Added `pytest_vanilla` CI job that runs all plugin tests with vanilla
- Simple implementation - just one line to choose the plugin set

**Test results:**
The skipped/excluded ones:
- SE score tests (2): these plugins don't exist in vanilla
- test_sum_wf (1): tests the store_data_top feature which has an issue in vanilla where data_top contains zeros instead of actual waveform data. This is an edge case that needs deeper investigation.

Overall the vanilla plugins work well including the new merged field (not sure if its implementation makes sense though)

cfuselli:
Ok I figured out what was going on with the area_fraction_top test failing!

**The Problem:**
The test_sum_wf was failing because `area_fraction_top` was being set to **zero** for all merged S2 peaks in vanilla. The pattern was interesting - every other peak had zero (unmerged peaks were fine, merged S2s had zeros).

**Root Cause:**
Found two bugs in `merged_s2s_vanilla.py`:

1. **Line 115 typo**: Used `digitize_top=True` which isn't even a valid parameter (should be `store_data_top=True`)

2. **Missing recomputation**: After `strax.merge_peaks()` creates new peak arrays with `np.zeros()`, the old peak data gets copied over BUT fields like `area_fraction_top` are initialized to zero and never recomputed. The SOM version calls `strax.compute_properties()` to fix this, vanilla didn't.

**Impact:**
This is actually a **critical data quality bug**! All merged S2 peaks in vanilla had `area_fraction_top=0`, which breaks:
- S1/S2 discrimination 
- Position reconstruction
- Energy reconstruction

**The Fix:**
1. Fixed the parameter: `digitize_top=True` → `store_data_top=True`
2. Added `strax.compute_properties(merged_s2s, n_top_channels=self.n_top_pmts)` after merging to properly recompute all peak properties
3. Removed the test skip - now test_sum_wf passes for vanilla!

Good thing we added these tests, they caught a real bug! 🐛

cfuselli:
Here the context comparison between "offline" and "online" contexts, meaning "default" vs "vanilla" plugins. 

straxen_vanilla_offline
straxen_vanilla_online

cfuselli:
I changed the way we start the mongo instance in the tests because it was giving: 
```
  docker: Error response from daemon: client version 1.40 is too old. Minimum supported API version is 1.44, please upgrade your client to a newer version.
```

PR #1644, opened 2026-02-11
Add 'merged' field to vanilla plugins

cfuselli:
This pull request reintroduces and standardizes the `merged` field across several event and peak data structures in the codebase. The main goal is to ensure that merged peaks and S2 signals are consistently marked and their data types are correctly defined throughout the relevant plugins.

**Data structure and field standardization:**

* Added the `merged` boolean field back to the dtype definitions in `event_basics_vanilla.py` and `peak_basics_vanilla.py`, ensuring that merged peaks/events are explicitly tracked in the data schema. [[1]](diffhunk://#diff-92307fc79e5c77d8453f8256945b0cbdd9320789e1785af9bbaef73214ff95e5L172-R172) [[2]](diffhunk://#diff-d8c89bad729519104f11621873316097e1e10c4a00c2f70e36e01705fafec026L57-R57)
* Updated the list of fields copied in the `compute` method of `peak_basics_vanilla.py` to include `merged`, so this property is preserved in output arrays.

**Merged S2 plugin improvements:**

* Restored the `indicator_dtype` to the dtype inference in `merged_s2s_vanilla.py`, ensuring all relevant fields are included in the merged S2s data type.
* Explicitly set the `merged` field to `True` for all merged S2s in the `compute` method, providing a clear marker for downstream processing.

### DIFF
--- a/straxen/plugins/events/event_basics_vanilla.py
+++ b/straxen/plugins/events/event_basics_vanilla.py
@@ -169,7 +169,7 @@ class EventBasicsVanilla(strax.Plugin):
             ("area_fraction_top", np.float32, "fraction of area seen by the top PMT array"),
             ("tight_coincidence", np.int16, "channel within tight range of mean"),
             ("n_saturated_channels", np.int16, "total number of saturated channels"),
-            # ("merged", bool, "is merged from peaklets"),
+            ("merged", bool, "is merged from peaklets"),
         )
 
     def setup(self):
--- a/straxen/plugins/merged_s2s/merged_s2s_vanilla.py
+++ b/straxen/plugins/merged_s2s/merged_s2s_vanilla.py
@@ -79,12 +79,14 @@ class MergedS2sVanilla(strax.OverlapWindowPlugin):
             "peaklet_classification"
         )
         peaklets_dtype = self.deps["peaklets"].dtype_for("peaklets")
+        # DON'T include indicator_dtype here! It causes dtype mismatch in strax.replace_merged()
+        # The merged field will be added later in PeaksVanilla
         # The merged dtype is argument position dependent!
         # It must be first classification then peaklet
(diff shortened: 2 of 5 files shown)


cfuselli:
## Implementation Summary: Option A - Match Non-Vanilla Architecture

After several iterations to solve the dtype mismatch error, we implemented **Option A** which fully matches the non-vanilla architecture.

### Final Solution (Commit d38e155e)

The merged field now exists throughout the vanilla pipeline:

1. **PeakletClassificationVanilla** (v3.0.5): Added `merged` field (always `False` for peaklets)
2. **MergedS2sVanilla** (v1.0.3): Re-enabled `merged` field (set to `True` for all output)
3. **PeaksVanilla** (v0.1.3): Gets `merged` from peaklet_classification, converts merged_s2s dtype before replace_merged()

### Why This Architecture?

This matches exactly what non-vanilla does:
- `enhanced_peaklet_classification` has merged field → `peaklet_classification` has merged field ✓
- `MergedS2s` outputs merged field → `MergedS2sVanilla` outputs merged field ✓
- `PeaksSOM` converts dtype before replace_merged → `PeaksVanilla` converts dtype before replace_merged ✓

### Key Insight

The dtype mismatch error occurred because `strax.replace_merged(peaklets, merged_s2s)` requires both inputs to have the same dtype. Since `peaklets` never has the merged field, we:
1. Convert `merged_s2s` to `peaklets.dtype` (like PeaksSOM does)
2. Get the merged field in final peaks from `peaklet_classification` instead
3. Set correct values using `fully_contained_in()`

### Testing

✅ All local dtype tests pass:
- `peaklet_classification`: has merged field (v3.0.5)
- `merged_s2s`: has merged field (v1.0.3)
- `peaks`: has merged field (v0.1.3)

Ready for DAQ testing!

PR #1646, opened 2026-02-13
Daq vanilla no mlp?

cfuselli:
This pull request introduces a new plugin to the peak position reconstruction pipeline, streamlining how peak positions are selected for downstream processing. The main change is the addition of the `PeakPositionsVanilla` plugin, which simplifies the selection process by using only the CNF algorithm for peak position reconstruction.

New plugin for peak position reconstruction:

* Added the `PeakPositionsVanilla` plugin in `straxen/plugins/peaks/peak_positions.py`, which merges reconstructed peak positions using only the CNF algorithm, making the computation trivial and not saving its output.

Pipeline configuration update:

* Registered the new `PeakPositionsVanilla` plugin in the list of plugins used in `straxen/contexts.py`, enabling its use in the reconstruction pipeline._Before you submit this PR: make sure to put all operations-related information in a wiki-note, a PR should be about code and is publicly accessible_


yabbyhome:
Thanks, Carlo! To move forward on this, I want to check my understanding:
1. Data comes into the DAQ and should have the minimum processing to be useful for data quality monitoring/fast response.
2. You have written this vanilla straxen for the DAQ machines (bootstrax). To streamline, you suggest removing mlp positions. I think there is support for this after Tuesday, so this no_mlp branch can be merged to the vanilla straxen branch.
3. You have checked the performance. @WenzDaniel has made other checks on your vanilla straxen
4. We can merge to master, and then git pull for the machines and create a container (tag/singularity image/etc) such that the online context is now this one, which we will likely adjust rapidly parameters-wise over the next few months, but we may never need to change structure-wise ever again.

PR #1647, opened 2026-02-13
Peak only posrec

cfuselli:
This pull request introduces a new, memory-efficient position reconstruction plugin for XENONnT online processing and updates the context logic to support its use. The main goal is to reduce RAM usage during online data acquisition by enabling "peak-only" position reconstruction, which skips peaklet-level processing. The changes also improve configurability and documentation for online contexts.

**Major new (old) feature: Peak-only position reconstruction**

- Added new plugin `PeakPositionsCNFPeakOnly` in `straxen/plugins/peaks/peak_positions_cnf_peakonly.py`, which reconstructs S2 positions directly from peak-level data, reducing RAM usage by ~27 GB at the cost of slightly lower accuracy and less effective S2 merging. This is intended for online DAQ environments with limited memory.
- Registered the new plugin in `straxen/plugins/peaks/__init__.py` so it is available for use in contexts.

**Context and configuration updates**

- Updated the `xenonnt` context in `straxen/contexts.py` to accept a new `_peak_only_positions` boolean option. When enabled (and in vanilla mode), the context registers the new peak-only position plugin instead of the standard one. [[1]](diffhunk://#diff-ebd5e6c1e333dd3011a01c5a7be9333983da6d07722650ac8745ca560fadb723R162) [[2]](diffhunk://#diff-ebd5e6c1e333dd3011a01c5a7be9333983da6d07722650ac8745ca560fadb723R206-R212)
- Improved documentation for the new `_peak_only_positions` parameter, explaining its benefits, drawbacks, and recommended use cases.
- Refactored the `xenonnt_online` context logic to automatically select appropriate defaults for `_vanilla` and `_peak_only_positions` based on whether the context is called from `cutax` and which `xedocs_version` is used. User overrides are respected, and documentation is expanded for clarity.


cfuselli:
Also removing data start and data top from here.

PR #1650, opened 2026-03-02
Fix NumPy 2.4+ / pandas 3 compatibility

cfuselli:
## Why
Recent CI started failing after dependency refreshes (NumPy 2.4+, pandas 3 / pyarrow strings), with errors like:
- `AttributeError: module 'numpy' has no attribute 'in1d'`
- `TypeError: Expected string or tuple of strings, got <class 'pandas.arrays.ArrowStringArray'>`
- event display tests failing due to scalar vs 1-element array time handling.

## What changed
- Replaced `np.in1d` with `np.isin` in straxen code and tests where applicable.
- Fixed `_event_display` to consistently use scalar `event[...]` fields (instead of `events[...]` array fields) for time ranges and titles.
- Added runtime compatibility layer in `straxen/_compat.py` and invoked it from `straxen/__init__.py`:
  - alias `np.in1d` to `np.isin` when missing (NumPy >= 2.4)
  - patch `strax.to_str_tuple` / `strax.utils.to_str_tuple` to accept pandas `ExtensionArray` (e.g. `ArrowStringArray`)
- Updated SCADA tests for newer pandas behavior:
  - timezone assertion compares string representation (`zoneinfo` vs `pytz`)
  - positional indexing now uses `.iloc`.

## Notes
- Compatibility patches were moved out of inline `__init__` logic into `straxen/_compat.py` to keep imports clean and satisfy pre-commit (`flake8 E402`).


caioishikawa:
Hi Carlo
Sorry about that, I was trying to copy your branch in another branch to remove the numpy requirement (<2.4.0) to test in base_environment, but I did something wrong and ended up making a commit here.

The test is failing because of some changes in the base environment, I will push at least one more time later to fix them

WenzDaniel:
Close since the same as #1655

PR #1653, opened 2026-04-01
Make Peaklets chunking settings lineage-tracked for online/offline contexts

cfuselli:
## Summary
- keep Peaklets chunking behavior context-configurable by adding plugin config options
- intentionally keep these chunking options **untracked** in lineage (`track=False`)
- set `xenonnt_online` defaults to DAQ-compatible Peaklets chunking (`rechunk_on_load=False`, `chunk_target_size_mb=strax.DEFAULT_CHUNK_SIZE_MB`)
- keep cutax-offline behavior on offline-friendly defaults
- remove bootstrax Peaklets monkey-patching now that chunking is configured in context
- add a context test asserting that changing Peaklets chunking config does **not** change `peaklets` lineage hash

## Motivation
Recent commissioning runs showed mismatches when DAQ and local processing used different Peaklets chunking behavior. We want explicit online/offline defaults without making chunking part of lineage, so re-reprocessing can still change chunking freely.

## Validation
- local `py_compile` on modified files
- local runtime check:
  - default online config is `peaklets_rechunk_on_load=False`, `peaklets_chunk_target_size_mb=200`
  - toggling these options keeps the same `peaklets` key/hash (`mqxa33b2om`)
- note: full `pytest` suite not run in this environment


WenzDaniel:
Would not it be easier to just change this single line in strax; https://github.com/AxFoundation/strax/blob/bdc4728db79a68df2544f1c122c2d2f603b6ce95/strax/context.py#L981 Then we make all plugins aware of their chunk size to prevent this issue in the future.

cfuselli:
Thank you @WenzDaniel for having a look!!!
Indeed, I think we could also just change that. 
I still think its porbably not enough, because changing just that still does not really give us a good way to change the parameter for online and offline context easily?

WenzDaniel:
This is another side to it, true. I am also not sure how fuse's downchunking might be affected.

cfuselli:
Following @dachengx advice, we keep the chunking untracked. It could cause issues with reprocessing oflline. So now we have the possibility to tune it via config but without lineage change. 
Maybe @Naderschah can have a look and test that it works?

WenzDaniel:
Converting to a draft for now to not forget about it, but must be rebased to new master.

PR #1657, opened 2026-04-24
pre-commit test fix

caioishikawa:

This PR was made to fix pre-commit test after the update on base_environment.
The problem was that docformatter 1.7.6 was having problems, but instead of updating it, it is easier to just fix the python version to 3.12.
Also updated some python versions in pytest workflow files.

### _Notes on testing_

### DIFF
--- a/.github/workflows/pypi_install.yml
+++ b/.github/workflows/pypi_install.yml
@@ -15,7 +15,7 @@ jobs:
       - name: Setup python
         uses: actions/setup-python@v5.3.0
         with:
-          python-version: "3.10"
+          python-version: "3.12"
 
       - name: Checkout repo
         uses: actions/checkout@v4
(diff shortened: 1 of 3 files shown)


caioishikawa:
If I simply update docformatter to 1.7.8 to be compatible with python3.14, the pre-commit.ci bot pushes a commit that changes the comment styles and then it does not pass the test by flake8. Therefore I reverted some commits and I am now investigating the docformatter config to see what I can do

caioishikawa:
back to zero again. I didn't want to use this, but since the problem is the compatibility between docformatter 1.7.6 and python3.14 (why is it using python3.14 though), we can just fix the python version in docformatter hook settings

WenzDaniel:
Rebased to a different branch, see https://github.com/XENONnT/straxen/pull/1663.

PR #1658, opened 2026-05-05
New S2 Merging parameters based for Comissioning based on fuse simula…

MerzJohannes:
Change of the S2 Merging Parameters based on fuse simulations for the commissioning. The results were discussed in a Team B&C Meeting. 
**Also Big Sorry:** 
I accidentally pushed previously to the daq-vanilla branch without a PR and reverted the changes to open this one.

MergingParameters


MerzJohannes:
Also do we want to update the s2_max_duration? At least from simulation it looks like we could probably reduce it by quite a bit?
image

WenzDaniel:
> Also do we want to update the s2_max_duration? At least from simulation it looks like we could probably reduce it by quite a bit?

Yes please. Maybe round up to x1.5 of the max size you found. For the moment this should be sufficient.

MerzJohannes:
Max value I found was 19050.0 ns, so I reduced now to 30000.
Some tests keep feeling, probably due to the changed values but I couldn't figure out, what values to change yet.

WenzDaniel:
I picky backed an additional check for the scada interface which tests if a user query is older than 5 years. This is not possible with the web API and made one of the tests fail. I updated the test with a newer query and added an additional raise condition + test if a query is older than 5 years.

WenzDaniel:
See https://github.com/XENONnT/straxen/pull/1667#issue-4845492505

PR #1660, opened 2026-05-27
Temporarily ignore channel 799 overlap checks

cfuselli:
## Summary

This is an intentionally minimal emergency DAQ patch.

- Keep the existing raw-record overlap check enabled.
- Exclude channel `799` from that check before calling `check_overlaps`.
- This lets bootstrax continue past overlapping artificial-deadtime markers written by redax on channel `799`.

## Context

Recent DAQ tests with graceful corruption handling can write multiple artificial-deadtime records to channel `799`. These records may overlap each other in raw-record space. Bootstrax currently fails in `PulseProcessing.compute` before downstream AQMon/veto-interval logic has a chance to interpret or merge the artificial-deadtime records.

This patch does not claim that the channel-`799` deadtime bookkeeping is fully correct. It only prevents the TPC raw-record overlap sanity check from aborting on this dedicated artificial-deadtime channel while we debug the proper redax/straxen deadtime handling.

## Validation

- Ran `python3 -m py_compile straxen/plugins/records/records.py`.

## Caveat

This is a temporary operational workaround. A proper fix should either merge/de-duplicate channel-`799` intervals before writing them, or represent corruption deadtime on non-overlapping AQMon channels / downstream logic that can safely merge intervals.


WenzDaniel:
I think we should not merge. Channel 799 should not be among the TPC channels

PR #1662, opened 2026-07-07
Implementation of Natural Break GOFs check before S2 merging

caioishikawa:
This PR introduces and hardens the vanilla merged-S2 workflow with a more selective merge decision based on natural-break goodness-of-fit (GoF). This will allow the merges2 algorithm to separate single scatter from multi scatters.

Main improvements:

- Improves merge quality by evaluating natural-break GoF at candidate gaps before merging. This allows the discrimination of multi-scatters and single scatters
- Ensures merged outputs are consistently typed and explicitly marked as merged.
- Keeps lone-hit integration and recomputes peak properties/widths after merging.

The overview of the flow to merge s2s is described below:
1. Builds merge candidates from adjacent peaklet gaps, ordered from smallest gap upward.
2. Applies constraints on gap size, merged area, and merged duration.
3. Computes natural-break GoF at each candidate split using weighted waveform time-variance from both sides of the gap.
4. Skips merging when the split quality is high (good split), and merges otherwise.
5. The rest works the same as before

Some important things:
- The position check was also added back to this version, but I opted to leave it commented out for the time being until we have a reliable position reconstruction.
- The tuning of the threshold will be done by other analysts I suppose. But by design we can use the same threshold as the natural break splitter, so no need to tune this independently

image
This figure shows two peaklet waveforms that would be merged if natural break gof was not tested. It means that it passes the gap_size, max_duration and other checks, but its merge is blocked by the natural break gof.
<br><br \>

image
This other figure is showing simulations of double scattered events' S2s and their GOFs. The black line shows the current natural break split threshold and we can see that many events above the black line that should be split are being merged, this includes multi-scatters. The points below the threshold come from the 700 ns gap threshold that should be merged. So if this is true, then it is not a problem in the algorithm. I checked some of their waveforms and will put them in the note.
<br><br \>


caioishikawa:
scada test fix taken from another PR where Daniel said we can not query data from more than 5 years ago or something like this

WenzDaniel:
Please rebase your changes form the current master

PR #1663, opened 2026-07-08
Fix python versions of tests

WenzDaniel:

A copy of @caioishikawa #1657 referenced to daq-vanilla


WenzDaniel:
I may have to add additional test fixes, let see.

caioishikawa (approved):
Looks good.
The only thing is that I changed the pytest.yml following what was already written:

>         exclude:
>           - python-version: "3.11"
>           - python-version: "3.12"
>             test: coveralls
>           - python-version: "3.11"
>           - python-version: "3.12"
>             test: pytest_no_database
>           - python-version: "3.11"
>           - python-version: "3.12"
>             test: pytest_vanilla

changing what was 3.11 to 3.12, but I don't know why we would skip these tests for the newest python version and keep the test for old versions. Maybe it makes more sense to skip the tests for 3.10 and 3.11 and run the test for 3.12.
Of course if this is not relevant then we just go ahead as it is

WenzDaniel:
Honestly I do not know why we added this skipping. I will fix first some other issues, then I want to check if also 11 and 12 can pass.

WenzDaniel:
I added now everything from https://github.com/XENONnT/straxen/pull/1655 except for these strange alias for numpy and pandas. This looked rather AI hacky to me, better to also bump in strax, e.g., merging https://github.com/AxFoundation/strax/pull/1007.

PR #1664, opened 2026-07-08
Add an additional channel for Backscatter PMT

MerzJohannes:

An additional channel range (in the 3000s) is openend for the Backscatter PMT and it's raw_records are added to the DAQ Reader Plugin. 
Adding the new channels to the channel map + add it's raw_records so it can get processed.


WenzDaniel:
Please rebase to current master

PR #1667, opened 2026-07-09
S2 merging gaps

WenzDaniel:
Rebased #1658

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 3.2.8
+current_version = 4.0.0
 files = straxen/__init__.py
 commit = True
 tag = True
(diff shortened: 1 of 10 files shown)


WenzDaniel:
Locally I could not reproduce the failing test of `test_sum_wf`. Lets rerun and see what happens.

WenzDaniel:
Okay now I was able to reproduce as well. The issue arises from the fact that this test is run on the full and top-only summed waveform after down sampling. This alters the total area of the summed waveform. For the very same reason the field "area" and "area_per_channel", which eventually determines the "area_fraction_top", are computed via the area of the individual hits before down sampling. Given that the difference in area of the downsampled summed waveform is still on the O(0.1%), I think it is fine to raise the threshold of the test to 3 digits. An O(0.1 %) difference in area does not significantly impact the peak shape, while the correct areas of the peak are retained via the summation before down sampling.

PR #1668, opened 2026-07-09
Add additional channel range for Backscatter PMT of neutron Gun

MerzJohannes:
Reopened #1664 (because original branch was closed)

### DIFF
--- a/straxen/contexts.py
+++ b/straxen/contexts.py
@@ -94,8 +94,8 @@ common_config = dict(
         mv_blank=(1999, 1999),
         nveto=(2000, 2119),
         nveto_blank=(2999, 2999),
-        sc = (3000, 3000),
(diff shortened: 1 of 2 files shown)


WenzDaniel:
@MerzJohannes did you test process a short run to check if everything works as anticipated?

MerzJohannes:
@WenzDaniel Not yet, I think a test run was delayed, I would need to create some fake data, to test it.

WenzDaniel:
Okay then maybe we should wait with the merging until we tested it. Should be quick,

WenzDaniel (approved):
Code looks fine, we can merge once we tested with some data.

stevensa22:
@WenzDaniel @MerzJohannes does this work without the Scintillator cell? I.e. is it backwards compatible?

